PR #1, opened 2026-06-21
Match 327 functions across ov002, ov006, arm9, ov034

RyanCopley:
Adds **327 byte-matched functions** (233 C + 94 C++), taking the project from
**39.2% → 42.0%** function coverage. All hand-written from scratch and verified
relocation-aware against the EU retail ROM with `tools/match.py` — each ends in
`MATCHING VERSIONS: 1.2/sp2p3`.

### Breakdown by module
- **ov002 — 211**: Player/Actor state handlers (guard → `SetAnim` → set fields →
  `ChangeState`), state transitions, deleting destructor chains, `new`/ctor stubs,
  object-table & level loaders, collision/raycast setup, matrix/collider transforms,
  fixed-point math, particle/sound/music glue, static initializers, and a large
  set of C++ virtual-call / pointer-to-member dispatch stubs.
- **ov006 — 92**: constant returns, field getters/setters, struct/`.data` copies,
  bounded init/scan loops, guards, linked-list unlink/search, fixed-point clamps,
  and PMF/virtual dispatch.
- **arm9 (main) — 22**: pointer stores, masked field reads, global getters/setters,
  clamps, `AngleDiff`, small accessors.
- **ov034 — 2**: `Actor::ClosestPlayer` cache, a 7-table `SharedFilePtr` release loop.

### Verification
- Compiler: **mwccarm 1.2/sp2p3**
- Flags: `-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`
  (C++ files carry a `//cpp` first line and compile with `-lang c++`)
- Every function was independently re-verified byte-for-byte under the pinned
  `-O4,p` flags before landing.

Import knowledge, write code: only community symbol names / struct field offsets
were used; all C/C++ is original, written against my own ROM dump. No ROM or
extracted data is included.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2, opened 2026-06-24
Match 9 functions across arm9, ov002, ov006, ov007

andrewboudreau:
Matches 9 functions across arm9 and overlays ov002, ov006, and ov007 (488 bytes of code).

All were written from scratch in C and verified byte-for-byte (relocation-aware) against the EU retail ROM with `tools/match.py`; each file ends in `MATCHING VERSIONS: 1.2/sp2p3`. These are reloc tail-call / guard-return / call-forwarding shapes that the automated template sweep deliberately skips (reloc false-positive risk), so they were left for hand matching.

## Functions

**arm9 (main)**
- `LoadFile` @ 0x0201816c (0x10)
- `LoadFileAt` @ 0x020182ac (0x10)
- `func_02012120` @ 0x02012120 (0x34) — `Sound::PlayLong` wrapper

**ov002**
- `func_ov002_020bd20c` @ 0x020bd20c (0x44)
- `func_ov002_020e04a4` @ 0x020e04a4 (0x48)

**ov006**
- `func_ov006_0211d75c` @ 0x0211d75c (0x54)
- `func_ov006_02120a64` @ 0x02120a64 (0x54)
- `func_ov006_0212318c` @ 0x0212318c (0x20)

**ov007**
- `func_ov007_020bde70` @ 0x020bde70 (0x40)

## Verification

Compiled with mwccarm 1.2/sp2p3 using the project's pinned flags
(`-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`)
and confirmed identical to the ROM via `tools/match.py` (relocation slots treated as wildcards).

These were produced by importing knowledge of the target's behavior and writing the C by hand — no decompiler output was pasted, and no inline `asm {}` was used.

Progress after this PR: 6,316 / 11,390 functions (55.45%).

## A note on process

Full disclosure: this is mostly Claude-driven, with me as an active decomp learner / relative newbie steering and verifying the output. I know PR #1 ran into overlap with unpushed work — so if any of this collides with what you've already got locally, or the approach is off-base, please just tell me to go away. No hard feelings at all; I'd much rather coordinate than make cleanup work for you. Happy to claim a range in CLAIMS.md or check in on Discord first if that's preferred for future passes.

### DIFF
--- a/src/LoadFile.c
+++ b/src/LoadFile.c
@@ -1,0 +1,10 @@
+/* LoadFile at 0x0201816c
+ *
+ * Matched byte-for-byte with mwccarm 1.2/sp2p3 (arm9 main).
+ */
+
+extern int func_0201818c(int handle, int mode);
+
+int LoadFile(int handle) {
+    return func_0201818c(handle, 1);
+}
--- a/src/LoadFileAt.c
+++ b/src/LoadFileAt.c
@@ -1,0 +1,10 @@
+/* LoadFileAt at 0x020182ac
+ *
+ * Matched byte-for-byte with mwccarm 1.2/sp2p3 (arm9 main).
+ */
+
+extern int func_02018270(int handle, int dest, int size);
+
+int LoadFileAt(int handle, int dest) {
+    return func_02018270(handle, dest, -1);
+}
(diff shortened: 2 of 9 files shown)


andrewboudreau:
I'm kind of interested in making a "tiny" coordination server that simply hosts claims and is essentially a hosted lock for work units our agents can use to coordinate in real-time without humans in the loop for that part.

PR #15, opened 2026-06-27
Match func_ov018_021117e8

Moundistz:
Byte-exact match for `func_ov018_021117e8` verified with mwccarm 1.2/sp2p3.

**Oracle verification (tools/match.py):**
```
TARGET func_ov018_021117e8 @ 0x021117e8 size 0x1c  bytes: bc20d1e1bf0052e30120a0030020a013000052e3201380151eff2fe1
  1.2/sp2p3: MATCH

========================================
MATCHING VERSIONS: 1.2/sp2p3
```

- Module: ov018
- Address: 0x021117e8
- Size: 0x1c

PR #44, opened 2026-06-27
Match func_ov100_02143ae0

Moundistz:
Byte-exact match for `func_ov100_02143ae0` verified with mwccarm 1.2/sp2p3.

**Oracle verification (tools/match.py):**
```
TARGET func_ov100_02143ae0 @ 0x02143ae0 size 0x38  bytes: 00402de904d04de20020a0e320109fe500208de5041091e5c30f80e2013aa0e3104bfbeb0100a0e304d08de20040bde81eff2fe1ac861402
  1.2/sp2p3: MATCH

========================================
MATCHING VERSIONS: 1.2/sp2p3
```

- Module: ov100
- Address: 0x02143ae0
- Size: 0x38


andrewboudreau:
# Heads-up: matched-function PRs #15–#44 are flattened to one line + unverified

👋 Noticed this while browsing the open PRs — flagging in one place in case it's useful to @bmanus2 and @Moundistz, rather than repeating it on every PR. (I'm just another contributor here, not speaking for the project — take or leave as you see fit.)

There's a systematic issue across the whole batch (#15–#44):
1. The files are flattened to a single line, which comments the function out. Each file is one physical line like:
```
//cpp extern "C" int func_ov100_02143ae0(char *c); … return 1; }
```

The repo's //cpp header needs a newline after it, with the code below. As written, // comments out the entire body → the symbol is never defined → the overlay won't link. Looks like a newline-stripping step somewhere in the generation tooling; un-flattening fixes it. Reference shape: src/func_ov001_020aa6e4.cpp.

2. No match verification. A matched-function PR really needs the tools/match.py output proving the bytes match the retail ROM — none of these include it, so there's no way to confirm correctness.
3. Symbol declarations. Prefer the project's extern decls over raw mangled names (_ZN…) where applicable, for consistency with the rest of src/.

tangosdev:
Thanks for the contribution. This function was already matched on main, so I am closing this individual PR to avoid a duplicate source file. Your batch landed 3 matches plus 8 nonmatching entries and a round of callee renames (27c8ee6), and you are credited in the README. Thanks again.

PR #48, opened 2026-06-27
Match 81 functions across ov002/ov004/ov006/ov007/ov015/ov060/ov102

andrewboudreau:
Adds 81 byte-matched functions across seven overlays, all compiled with mwccarm 1.2 (`-O4,p -enum int -lang c99|c++ -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`).

**Per-overlay breakdown (81 total):**
- ov002 × 20 (incl. `HUD::Render`, `Platform::Kill`, `Player::St_Throw_Init`, `Player::St_Throw_Main`, `LoadDoorObjects`)
- ov004 × 11
- ov006 × 29 (incl. `__sinit_ov006_0212fc7c`)
- ov007 × 16
- ov015 × 1
- ov060 × 3
- ov102 × 1

**Verification — every function passes three gates:**
1. **Byte-identical** (relocation-aware) to the retail ROM.
2. **Trio-clean** — reproduces under all three of mwccarm `1.2/base`, `1.2/sp2`, `1.2/sp2p3` (codegen-identical).
3. **Reloc-destination clean** — every emitted relocation resolves to its ground-truth `config/**/relocs.txt` entry (no wrong-callee / wrong-global hidden behind the wildcard; see the audit in #47). Callees are declared by their exact (mangled) symbol so the destinations are verifiable, not left as unresolvable invented names.

src-only; no tooling or bookkeeping changes.

### DIFF
--- a/src/_Z15LoadDoorObjectsRN11LVL_Overlay11ObjSubTableEij.cpp
+++ b/src/_Z15LoadDoorObjectsRN11LVL_Overlay11ObjSubTableEij.cpp
@@ -1,0 +1,55 @@
+//cpp
+// _Z15LoadDoorObjectsRN11LVL_Overlay11ObjSubTableEij at 0x020fe4f0
+// Matched byte-for-byte with mwccarm 1.2/sp2p3 (ov002).
+struct Vector3 { int x, y, z; };
+struct Vector3_16 { short x, y, z; };
+
+struct DoorEntry {
+    short x;
+    short y;
+    short z;
+    short angle;
+    unsigned short field8;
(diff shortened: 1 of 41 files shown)


andrewboudreau:
Rebased onto current main and **trimmed to the 41 functions main doesn't have yet**. The other 40 from the original batch already landed on main (some matched independently, a few as `.cpp` where this PR had `.c`), so I dropped them — this is now a clean, conflict-free set of 41 still-missing byte-matches. Should merge without conflicts.

PR #50, opened 2026-06-29
Match 107 functions across arm9-main + 19 overlays (verified, rebased on 63.9%)

andrewboudreau:
Matches **107 functions** across arm9-main and 19 overlays, produced by the parallel fan-out matching harness and independently re-verified against the current `main`.

## Per-module

| module | funcs |
|---|---|
| arm9-main | 53 |
| ov006 | 14 |
| ov002 | 10 |
| ov098 | 6 |
| ov081 | 4 |
| ov007 / ov018 / ov080 / ov091 / ov100 | 2 each |
| ov003 / ov036 / ov062 / ov063 / ov064 / ov066 / ov072 / ov073 / ov074 / ov102 | 1 each |

## Verification

- **Byte-match:** all 107 recompile byte-identical (relocation-aware) to the ROM under `mwccarm 1.2/sp2p3` with the canonical flags, confirmed against this PR's merge base.
- **Reloc destinations:** link-checked — **103 VERIFIED + 4 BLIND, 0 WRONG, 0 NO-REPRO** (every resolved relocation points at the correct symbol; BLIND = relocs to symbols not yet in the delinker config, no wrong destination).

## Refreshed onto `main` @ 63.9%

Merged the latest upstream `main` (`c51439d`, 63.9%). Five functions that the upstream idiom-enriched fan-out matched in the same window were resolved in favor of upstream's copies and dropped from this PR, so it proposes **only functions `main` does not already have**: `__sinit_ov063_0211e29c`, `func_020485f0`, `func_02051918`, `func_ov004_020b6430`, `func_ov098_0213b6e0`.

src-only — no tooling, config, symbol, or notes changes.

### DIFF
--- a/src/DotVec3.c
+++ b/src/DotVec3.c
@@ -1,0 +1,14 @@
+/* DotVec3 at 0x0205380c
+ *
+ * Matched byte-for-byte with mwccarm 1.2/sp2p3 (arm9 main).
+ */
+typedef int Fix12i;
+typedef struct Vector3 { Fix12i x, y, z; } Vector3;
+
+Fix12i DotVec3(const Vector3 *a, const Vector3 *b)
+{
+    long long sum = (long long)a->x * b->x
+                  + (long long)a->y * b->y
+                  + (long long)a->z * b->z;
+    return (Fix12i)((sum + 0x800) >> 12);
+}
--- a/src/_ZN21ExtendingMeshCollider9GetNormalEsR7Vector3.c
+++ b/src/_ZN21ExtendingMeshCollider9GetNormalEsR7Vector3.c
@@ -1,0 +1,27 @@
+/* _ZN21ExtendingMeshCollider9GetNormalEsR7Vector3 at 0x0203a4dc
+ *
+ * Matched byte-for-byte with mwccarm 1.2/sp2p3 (arm9 main).
+ */
+typedef int Fix12i;
+struct Vector3 { int x, y, z; };
+extern void NormalizeVec3(struct Vector3 *src, struct Vector3 *dst);
+extern void func_02039db8(void *thiz, struct Vector3 *vec, struct Vector3 *res);
+
+void _ZN21ExtendingMeshCollider9GetNormalEsR7Vector3(char *c, int idx, struct Vector3 *res) {
+    char *p = *(char **)(c + 0x20);
+    char *ip = *(char **)(p + 8);
+    char *verts = *(char **)(p + 4);
+    unsigned short n = *(unsigned short *)(ip + idx*16 + 6);
+    short *v = (short *)(verts + n*6);
+    struct Vector3 tmp;
+    Fix12i sy;
+    tmp.x = v[0] << 2;
+    tmp.y = v[1] << 2;
+    tmp.z = v[2] << 2;
+    sy = *(Fix12i*)(c + 0x1c8);
+    tmp.x = (Fix12i)(((long long)tmp.x * sy + 0x800) >> 12);
+    sy = *(Fix12i*)(c + 0x1c8);
+    tmp.z = (Fix12i)(((long long)tmp.z * sy + 0x800) >> 12);
+    NormalizeVec3(&tmp, &tmp);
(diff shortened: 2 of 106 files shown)


andrewboudreau:
## Link-destination audit pass (51 → 12 WRONG)

Ran `tools/linkcheck.py` against the full matched corpus and fixed 39 wrong relocation destinations. The match gate wildcards every relocated word, so a function can compile to identical bytes while pointing at the wrong callee or global — `linkcheck` catches this by actually resolving each symbol and byte-comparing the linked result against the ROM.

### Corpus result

| verdict | before | after |
|---|---:|---:|
| VERIFIED | 4,550 | 4,587 |
| BENIGN | 16 | 16 |
| BLIND | 1,699 | 1,701 |
| WRONG | **51** | **12** |
| NO-REPRO | 1 | 1 |

### Fix shapes

**Wrong overlay prefix on `data_` symbols** (~30 files): functions in one overlay (`ov032`, `ov070`, `ov085`, `ov090`, `ov091`, `ov100`, etc.) referenced data symbols with the wrong overlay prefix — they compiled and matched because the relocation slot was wildcarded. Fixed by renaming externs to point at the correct overlay address.

**`__sinit` initializers** (8 files): `__sinit_ov002`, `__sinit_ov055`, `__sinit_ov064`, `__sinit_ov079`, `__sinit_ov081` all had the wrong three data symbol names (copy-pasted from a template for a different overlay). `__sinit_ov006_021303d0` and `__sinit_ov006_02132894` had 9 wrong `data_ov002_*` symbols each. `__sinit_ov070_02123030` had 5.

**Swapped ctor/dtor args to `func_020733a8`** (4 files): `_ZN6PlayerC1Ev`, `func_ov006_020fa6ac`, `func_ov006_021101bc`, `func_ov006_021255f8` — the 4th and 5th arguments (constructor/destructor function pointers) were in the wrong order. MWCC's pool layout puts the 5th-arg push *before* the 4th-arg r3 load, so the fix is to swap the arg pair in the call site.

**`func_02030aa4`**: two zeroing assignments in wrong order (same pool-ordering effect).

**`_ZN8CapEnemyD2Ev`**: wrong destructor symbol for the Enemy base dtor — used `_ZN5EnemyD2Ev` (0x020aed74) but ROM branches to `func_020aed18`.

**`func_ov077_02126a50`**: callee and data both had wrong overlay prefix (ov090 → ov077).

### Remaining 12 WRONG

- **7 animation thunks** (`_ZThn80_N9Animation{D0,D1}Ev` ×3 sets) — call `_ZN9AnimationD1Ev` but ROM calls a `ModelAnim2` destructor; fixing requires renaming the class in `nearmiss/db.jsonl`
- **3 BL→BLX false positives** (`func_0206a318`, `func_0206a3a4`, `func_0206a5c0`) — `WaitByLoop`/`CpuSet` are Thumb entry points; the linker rewrites BL→BLX but `link_function()` preserves the source `0xEB` upper byte; these are not genuine wrong matches
- **`func_ov005_020c0250`** — ROM uses byte-offset symbols `data_020a0de9/ea/eb` but source uses struct-array access from `data_020a0de8`
- **`func_ov098_02137eec`** — `data_ov098_0213c380[idx2].b/.c` pool constants should be `base+4`/`base+8`; struct access pattern differs from ROM's layout

Details in `notes/link-verification.md`.

andrewboudreau:
@bmanus2-dotcom — refreshed this onto current `main` (`c51439d`, 63.9%) so it's mergeable again.

The merge surfaced **5 functions your idiom-enriched fan-out matched in the same window**. I took your copies and dropped mine, so this PR now proposes only functions `main` doesn't already have:

`__sinit_ov063_0211e29c`, `func_020485f0`, `func_02051918`, `func_ov004_020b6430`, `func_ov098_0213b6e0`

The remaining **107** all recompile byte-identical under `mwccarm 1.2/sp2p3` and link-check clean (103 VERIFIED + 4 BLIND, **0 WRONG**). src-only — no tooling/config/symbol changes.

If any of these overlap work you have unpushed, happy to drop that subset — just point me at the names.

PR #63, opened 2026-07-01
Match __sinit_ov081_02128154

andrewboudreau:
One function, src-only: `__sinit_ov081_02128154` (ov081 static initializer, 0x94 bytes).

From a claims-coordinated coddog batch in the 0x280-0x400 band (locked/released via the live API, fully disjoint from your concurrent batch). Byte-MATCH verified independently at bank time, relocation destinations checked against config/**/relocs.txt (0 WRONG-DEST).

Tooling/workflow changes from tonight are coming separately on another branch.


tangosdev:
This function landed on main earlier tonight while salvaging closed #62 (commit 23f96b7, credited there) - closing as redundant. Verified identical bytes both ways.

PR #64, opened 2026-07-02
Reloc-destination gates: bank-time (harvest/bank-matches), source-level (clone/paramclone), whole-run link sweep

andrewboudreau:
Restructured to the (1)+(3) decision — the ledger.bank version that briefly replaced this was pushed before I saw the decision comment (13 minutes apart, my miss) and is gone. Two commits off current main (abad87b), 6 files, +137/-24:

## Commit 1 — the original bank-time gate, rebased (`bank_harvest`, `bank-matches`)

Unchanged in substance from the draft: `swarm.oracle_check` returns the compiled object alongside the verdict (`oracle_ok` delegates, all callers untouched), `reloc_audit.gate_wrong_dests` wraps `check_destinations` with cached config indexes, and both bank paths reject on WRONG-DEST rows before `ledger.bank`. `--no-strict` restores bytes-only banking. The 88712ce dup-filter commit is dropped per your note (main already banks against `matched_set`).

## Commit 2 — source-level gates for clone/paramclone (no recompile) + whole-run link sweep

- **clone.py**: the structural key already carries every reloc slot offset, so the gate is a pure config-map comparison — refuse when `config/**/relocs.txt` records a different destination for candidate vs template at any slot. Zero compiles spent on refused candidates.
- **paramclone.py**: same comparison over the slots the two functions share, applied per template before substitute/compile; other templates in the skeleton bucket still get their chance.
- **crackloop land**: snapshots `matched_set()` at entry, then linkchecks **everything any tier banked during the run** and shouts WRONG banks with unbank instructions — the backstop for any banker the source gates miss. Free tiers run with `--apply` again now that they're gated.

## Evidence (all rerun on this branch)

- Wrong-callee `func_02064a94` byte-passes the oracle, rejected with 5 WRONG-DEST relocs; genuine source passes with 0.
- `clone.py --min 0x40 --max 0x100` dry-run rejects **exactly the 10 ov006 PMF siblings from the incident** (`func_ov006_0211a4b0`…`0211b590`), each pinned to its distinct table pool word at `+0x4c` — before the oracle ever compiles them.
- `paramclone.py --min 0x40 --max 0x80` dest-gates all 40 PMF template pairings (10 siblings × 4 templates), candidates tried: 0.
- Sweep plumbing: `linkcheck()` on a known-banked function returns VERIFIED through the exact call path `land` uses.
- py_compile clean on all six files.

Thanks for the sec 6e credit and the 6g pointer — will re-run the zero-arg PMF twins (div=11) against the u64-mask laundering idiom.

https://claude.ai/code/session_01HgKJraYsbg7Gg967B8dCpc

### DIFF
--- a/tools/bank_harvest.py
+++ b/tools/bank_harvest.py
@@ -1,6 +1,9 @@
 """Central harvest banker. Collects {name,c_source} JSONL produced by fan-out agents,
-INDEPENDENTLY re-verifies each via the oracle (never trusts agent say-so), and banks the
-real byte-matches to src/<name>.c(pp) + progress/matched.jsonl (tag 'harvest'). Resilient
+INDEPENDENTLY re-verifies each via the oracle (never trusts agent say-so), checks that
+every relocation the object emits points where config/**/relocs.txt says (the byte
+compare wildcards reloc slots, so a wrong same-shaped callee would otherwise pass --
+--no-strict skips this), and banks the real byte-matches to src/<name>.c(pp) +
+progress/matched.jsonl (tag 'harvest'). Resilient
 to agents that died mid-run -- it banks whatever verified lines they saved.
 
 Dry-run by default; pass --apply to bank.
@@ -14,6 +17,7 @@ sys.path.insert(0, str(pathlib.Path(__file__).resolve().parent))
 import swarm as S
 import nearmiss_db as NM
 import ledger as L
+import reloc_audit as RA
 
 
 def main():
@@ -22,6 +26,8 @@ def main():
     ap.add_argument("--apply", action="store_true", help="bank the verified matches")
     ap.add_argument("--dry-run", action="store_true",
                     help="no-op; dry-run is the default (kept for compatibility)")
+    ap.add_argument("--no-strict", action="store_true",
+                    help="skip the reloc-destination gate (bytes-only banking)")
     args = ap.parse_args()
     do_apply = args.apply and not args.dry_run
 
@@ -51,11 +57,19 @@ def main():
         if L.make_key(module, addr) in done:
             dup += 1; continue
         try:
-            ok = S.oracle_ok(src, name, bytes.fromhex(thex))
+            ok, obj = S.oracle_check(src, name, bytes.fromhex(thex))
         except Exception as e:
             rejected.append((name, f"exc {e}")); continue
         if not ok:
             rejected.append((name, "oracle FALSE")); continue
+        if not args.no_strict:
+            # the byte oracle wildcards reloc slots; refuse a candidate whose
+            # relocations point somewhere other than config/**/relocs.txt records
+            bad = RA.gate_wrong_dests(obj, name, L.norm_addr(addr), size, module)
+            if bad:
+                rejected.append((name, f"WRONG-DEST reloc: {bad[0]['cand']} "
+                                       f"({bad[0]['cand_addr']}) != config {bad[0]['cfg']}"))
+                continue
         if do_apply:
             status = L.bank({"addr": addr, "name": name, "size": size,
                              "module": module, "versions": ["harvest"]}, src)
--- a/tools/clone.py
+++ b/tools/clone.py
@@ -7,6 +7,14 @@ so the call/data targets may differ). When that holds, the verified source of th
 matched function, retargeted to the clone's name, matches the clone too. No model
 is involved: we retarget and re-run the oracle, banking only what passes.
 
+One more check, because the oracle wildcards reloc slots: the retargeted source
+still names the TEMPLATE's callees and globals, so the clone only genuinely
+matches when config/**/relocs.txt records the SAME destination at every slot for
+both functions. Byte-identical PMF/table dispatchers fail exactly there (each
+sibling reads its own table) -- 24 of them byte-passed and banked wrong across
+two machines on 2026-07-02. The gate compares the two config reloc maps directly
+(no compile) and rejects mismatches before the oracle ever runs.
+
 This compounds: every function the LLM loop lands becomes a template for its
 byte-clones, so re-running clone.py after each bank harvests the new clones for
 free. Run it as the cheap first pass before spending tokens on a fan-out.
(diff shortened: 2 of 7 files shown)


tangosdev:
Heads up before you take this out of draft: the ledger consolidation landed on main tonight (51bce0a and follow-ups) - all bank paths now go through tools/ledger.py (locked atomic appends, an adoption path for merged PR files, and verified matches now override stale nonmatching parks). The blocking reloc gate you are adding belongs inside ledger.bank() now, so every caller inherits it. Also note bank_harvest.py is dry-run by default with --apply since the refactor. Happy to merge a rebased version - the gate itself is exactly right (your #62 evidence convinced us; we wired the non-blocking linkcheck pass as a stopgap at bank time in bank_run).

andrewboudreau:
Live evidence from tonight's batch that strengthens the case here — and exposes a coverage gap in the current post-bank linkcheck:

After banking 6 verified PMF-dispatch matches, `clone.py --apply` auto-banked **10** byte-identical ov006 siblings (`func_ov006_0211a4b0` … `0211b590`). `linkcheck --name` on each: **10/10 WRONG** — every clone dispatches through a different PMF table symbol, so the bytes match but every data reference points at the wrong table. We unbanked all 10 (src + ledger rows pruned).

The gap: the runbook says every banked function passes linkcheck, but `bank_run` only linkchecks *its own* `sources` — the free tiers (`clone.py`, `paramclone.py`, `nearmiss_db bank-matches`, `sweep`, `swarm`) bank through `ledger.bank` with no destination check at all. Honest caveat: this PR's gate covers `bank_harvest` + `bank-matches` only, so it wouldn't have stopped this either.

Options, in order of how much they centralize the invariant:
1. Extend this PR's gate to `clone.py`/`paramclone.py` (clone already knows both functions' reloc maps, so it can compare destination equivalence directly — cheaper than a recompile);
2. Move the check into `ledger.bank` itself (needs the object or a reloc summary passed in — bigger signature change, but then nothing can bank unchecked);
3. Keep gates at callers but add a `linkcheck` sweep to `crackloop land` covering everything banked in the run, as a backstop.

Happy to implement whichever shape you prefer — (1)+(3) is my suggestion: cheap at the source, belt-and-suspenders at land time.

andrewboudreau:
Follow-up from the rest of tonight's batch — the evidence got stronger, and the PR grew one commit:

**paramclone is wrong-dest too.** After we unbanked clone.py's 10 WRONG ov006 PMF dispatchers, `paramclone.py --apply` re-banked the *same 10* ("parameterized clones VERIFIED 10") — and linkcheck says **10/10 WRONG** again. It parameterizes the code but not the PMF-table symbol destination. Tonight's tally for ungated free tiers: **20/20 WRONG banks** on this function family. Until the tiers check destinations, byte-identical PMF/table-dispatch shapes are poison for them (option 1 or 2 from the previous comment; happy to implement on your word).

**New commit on this branch (88712ce):** `bank_harvest`'s dup pre-filter used `load_done()` (matched **+ parked**), which permanently blocked a verified match for any function a previous run had parked as a close near-miss — `ledger.bank`'s hatch-upgrade path never got the chance. Hit live: 3 of 4 verified matches in an Opus retry were dup-skipped because the preceding Sonnet run parked them an hour earlier. Fixed to `matched_set()`; `ledger.bank` still guards the actual overwrite cases.

Happier news, same batch: the "regperm wall" (`r1/r2/r3` vs `r2/r3/ip`) on PMF dispatchers is **not a wall** — give the PMF an `int` parameter and forward the index so it stays live in r1, and the scratch coloring matches the ROM. Proven 7x tonight across ov002/ov006 (e.g. `func_ov002_020f3740`, `func_ov006_0211a2c4`); might be worth a line in `notes/mwccarm-codegen.md` sec. 2, since several agents mis-filed these as the documented floor. The zero-arg PMF variants (4 ov002 twins, div=11) are in the near-miss DB.

tangosdev:
Your clone/paramclone warning fired verbatim on our side within the hour: landing a batch after merging #66 auto-banked 14 byte-identical PMF siblings with wrong-table symbols - linkcheck WRONG on all 14, reverted (commit history has the details). Free tiers now run dry in crackloop land until this gate is in. Consider that a second, independent motivation for landing #64 - rebase and we merge.

tangosdev:
Decision: go with your (1)+(3) - gate clone.py/paramclone.py at the source via reloc-map destination comparison (no recompile), keep the bank_harvest/bank-matches gate from this branch, and add the whole-run linkcheck backstop to crackloop land so anything banked by any tier in a run gets swept. Skip option 2 for now; if a future banker appears that the backstop misses, we revisit moving it into ledger.bank.

One rebase note: your 88712ce (load_done -> matched_set in bank_harvest) is already on our main since c38fac1, together with ledger.bank dropping stale parks on a verified match - drop that commit or expect a conflict.

Also for your notes: the PMF int-parameter lever from your #66 is documented in notes/mwccarm-codegen.md sec 6e with credit, and one more thing landed tonight that changes the floor math substantially - the u64-mask laundering idiom (((int)base + off) & 0xFFFFFFFFFFFFFFFF) forces base materialization that was documented as unreachable; see sec 6g and commit abad87b. Your zero-arg PMF twins at div=11 might be worth re-checking against it plus the levers in 6e.

Take it out of draft when the clone gates are in and we will verify and merge same-day.

andrewboudreau:
Rebased onto abad87b and reworked per your note: the gate is inside `ledger.bank()` now — one blocking check before the lock, all bankers inherit it, zero caller changes. The draft's per-caller gates and the swarm split are gone; 88712ce is dropped (main already carries the matched_set fix). `crackloop land` re-enables clone/paramclone with `--apply`, since the gate now blocks the class they were dried out for.

Proof rerun on the rebase: the wrong-callee `func_02064a94` variant byte-passes the oracle and is REFUSED with 5 WRONG-DEST relocs before any write; the genuine source banks clean; a repeat bank returns "dup" via the unlocked pre-check; `strict=False` banks bytes-only. Out of draft.

andrewboudreau:
Restructured to your (1)+(3) — and an apology for the noise: your decision comment and my previous push crossed (04:14 vs 04:27; I'd read the thread before your decision landed and didn't re-poll before pushing). The ledger.bank version is gone; the branch is now exactly the shape you specified, two commits off abad87b.

The live proof wrote itself: `clone.py` dry-run on this branch rejects **exactly the 10 ov006 PMF siblings from the incident** (`func_ov006_0211a4b0`…`0211b590`), each with its distinct table word at `+0x4c`, before any compile — and `paramclone.py` dest-gates all 40 of their template pairings (candidates tried: 0). Wrong-callee `func_02064a94` still refuses with 5 WRONG-DEST at bank time; the land sweep's linkcheck path returns VERIFIED on a known-good bank. 88712ce dropped per your note. Already out of draft — ready for your verify.

Noted on 6e/6g — I'll re-check the zero-arg PMF twins (div=11) against the u64-mask laundering idiom.

tangosdev:
Verified and merged. Local checks: py_compile on all seven touched files clean; clone.py dry-run rejects all 11 wrong-table PMF siblings from both incidents pre-compile with exact per-slot diagnoses; a known-good re-bank passes the strict gate as dup; the land sweep integrates cleanly with the --tag parallel-batch plumbing you rebased onto. Free tiers are re-enabled in land. Thanks for turning this around same-night - the loop is now gated end to end.

PR #66, opened 2026-07-02
Match 10 PMF-dispatch functions across ov002/ov006 (two-stage Sonnet→Opus batch)

andrewboudreau:
Ten functions, src-only, +132/-0, all in the PMF-table-dispatch family from the 0x40-0x100 high-sim band. All byte-MATCH (independent oracle re-verify) and linkcheck-VERIFIED, 0 wrong destinations. Claims-coordinated: all 15 ranges in the batch locked per-function before dispatch, released after landing.

ov006 (9): `020d65c8`, `020d7c00`, `020e1214`, `020e4744`, `020e8a44`, `020f0ba0`, `02119eec`, `0211a0d8`, `0211a2c4` — ov002 (1): `020f3740`.

## The idiom that cracked this family (worth a codegen-notes line)
Several agents had filed these as the documented regperm floor — instruction stream byte-identical, registers consistently shifted (`r1/r2/r3` vs the ROM's `r2/r3/ip`). It's not a floor here: **declare the PMF as `void (C::*)(int)` and forward the index — `(c->*table[idx].pmf)(i)`**. Keeping `i` live in r1 through the `blx` forces the address temps into `r2/r3/ip`, matching the ROM exactly. Proven 7 times in this batch across both overlays. The zero-arg PMF variants (4 ov002 twins at div=11, where there's no argument to pin r1) are ingested in the near-miss DB.

## Two-stage batch economics
Sonnet 5 first pass: 6/15 at 67K tokens/landed. Then, **after banking those 6**, an Opus retry on the 9 failures: 4/9 at 69K/landed — the retry agents found the freshly-banked twins in `src/` and copied the int-param idiom (one agent cited `func_ov002_020f3740.cpp` by name). Retrying failures *after* the round's solutions land is a cheap second harvest; retrying before would have re-proven "walls."

Heads-up (details on PR #64): during this batch, `clone.py` and then `paramclone.py` each auto-banked the same 10 byte-identical ov006 dispatchers with wrong PMF-table symbols — 20/20 linkcheck-WRONG, all reverted. None of those are in this PR; the 10 here are the verified agent matches only.

https://claude.ai/code/session_01DL5gbYkvGdYWYJHiBdv4De

### DIFF
--- a/src/func_ov002_020f3740.cpp
+++ b/src/func_ov002_020f3740.cpp
@@ -1,0 +1,9 @@
+//cpp
+struct C;
+typedef void (C::*PMF)(int);
+extern PMF data_ov002_02110e9c[];
+
+extern "C" void func_ov002_020f3740(C* self, int idx) {
+    unsigned char sel = *((unsigned char*)((char*)self + idx * 0x30 + 0x15e));
+    (self->*data_ov002_02110e9c[sel])(idx);
+}
--- a/src/func_ov006_020d65c8.cpp
+++ b/src/func_ov006_020d65c8.cpp
@@ -1,0 +1,19 @@
+//cpp
+typedef unsigned char u8;
+class C;
+typedef void (C::*PMF)(int);
+class C { public: int dummy; };
+struct Row { u8 d[0x10]; };
+
+extern "C" PMF data_ov006_02141680[];
+
+extern "C" void func_ov006_020d65c8(C *self)
+{
+    Row *rows = (Row *)self;
+    for (int i = 0; i < 2; i++) {
+        if (rows[i].d[0x626d]) {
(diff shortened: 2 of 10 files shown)


tangosdev:
Merged after independent re-verification here: 10/10 oracle MATCH, 10/10 linkcheck VERIFIED, adopted into the local ledger. The PMF-argument-liveness lever is now in notes/mwccarm-codegen.md sec 6e and the two-stage retry pattern is in the runbook, both credited. On the clone/paramclone auto-bank incident: acknowledged as a real gap on our side too - our land step link-gates only the fan-out sources, not the clone tiers. Your #64 blocking gate inside ledger.bank() is the right fix since every banker inherits it; rebase onto current main and we will merge it.

PR #67, opened 2026-07-02
Automate the README progress bar with a GitHub Action

andrewboudreau:
## Summary
- Adds `.github/workflows/update-progress.yml`: on every push to `main` touching `config/**/symbols.txt` or `src/**`, it runs `tools/progress.py --write-readme` to refresh the `## Progress` bar in README.md and commits the change if the numbers moved. Also runnable manually via `workflow_dispatch`.
- `tools/progress.py` gains `synced_from_src()`, used by `--bar` and `--write-readme`: it computes the matched set from `config/**/symbols.txt` cross-referenced against which functions have a `src/<name>.c[pp]` file, so it works from a plain checkout with nothing else needed. The full local report (`python tools/progress.py`, no flags) is unchanged.

## Test plan
- [x] Ran `tools/progress.py --write-readme` in a clean worktree checked out from `origin/main` with no `progress/` directory present (simulating the hosted runner) - produced correct numbers and rewrote the README block.
- [x] Confirmed idempotency: running it a second time reports "already up to date" with no further diff.
- [ ] Maintainer confirms the workflow permissions (`contents: write` via the default `GITHUB_TOKEN`) are acceptable for auto-committing to `main`.

### DIFF
--- a/.github/workflows/update-progress.yml
+++ b/.github/workflows/update-progress.yml
@@ -1,0 +1,53 @@
+name: Update progress bar
+
+# Keeps the README "## Progress" bar in sync with the committed function
+# corpus. progress/matched.jsonl is git-ignored (local-only, per-contributor,
+# and known to drift stale), so it never reaches this runner - instead
+# tools/progress.py --write-readme derives the matched set purely from
+# committed data: config/**/symbols.txt (every declared function) cross-
+# referenced against which of those names have a src/<name>.c[pp] file.
+#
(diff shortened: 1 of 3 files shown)


tangosdev:
Merged on main (commit 20f8ae2). One fix applied on the way in: synced_from_src now excludes the 25 // NONMATCHING hatch files (they have a src file but are not byte-matches, so they were inflating the count by ~25), and I added the progress:start/end markers it writes between. The Action is live. Thanks - this kills the manual bar updates.

PR #74, opened 2026-07-02
Match 17: wide-band rescan (12) + first 0x800+ band batch (5) — 68.56% funcs / 40.34% bytes

andrewboudreau:
Batch 3 of the Opus->Fable two-stage loop: a wide-band (0x40-0x800) coddog rescan right after the big upstream merge, which resurfaced first-tier sibling scaffolds (sims 0.65-1.00). Clean sweep: **12/12 attempted functions matched.**

## Stage results
- **Opus fan-out: 7/12 @ ~42K tok/landed** - the cheapest Opus pass we have measured (post-merge rescan picks are that good)
- **Fable retry on all 5 misses: 5/5** (promotion triage said none were walled; lifetime promotion record now 14/14)

## Functions (all verified by the link-gate sweep)
arm9: ARMMathLoadState. ov002: func_ov002_020d7030. ov004: func_ov004_020b3cb8. ov007: func_ov007_020af84c. ov064: func_ov064_02116ec0. ov065: func_ov065_021177e4. ov066: func_ov066_02119ce8. ov078: func_ov078_021250f8. ov085: func_ov085_0212d8ec. Plus func_02008550, _ZN6Player24St_MetalWaterGround_MainEv, func_ov002_020d7030's siblings - full list in the three commits.

## New compiler levers mined into notes/mwccarm-codegen.md (+ the agent prompt)
- **Empty inline dtor blocks DSE** on a write-only stack object in //cpp (`~T() {}` = zero code, keeps the batched ldr/str of a dead Vector3; every plain-C spelling failed at 12-29 div)
- **Phantom-frame retention**: `volatile T x; (void)&x;` keeps dead-local stack slots at -O4 when the body matches but the frame is too small
- **//cpp front end materializes guard bools that C99 folds** (movne/moveq/cmp shape) - front-end switch alone can be the whole residual
- **NitroSDK CP math-context MMIO** (divider/sqrt regs at 0x4000280+) matches as plain volatile u64/u16 raw-pointer stores - no asm needed
- WRONG-DEST reloc gate caught two byte-identical zero-stores with swapped data symbols in func_ov066_02119ce8 (oracle-blind, link-gate caught it; fixed at source)

Claims for all ranges were held during the run and released on landing. Near-miss DB updated alongside.

https://claude.ai/code/session_01HgKJraYsbg7Gg967B8dCpc


andrewboudreau:
Pushed 3 more commits extending this PR with the first batch in the untouched **0x800–0x4000 band** (plus a merge from main, so it's current):

**5 new matches, +14.6KB — code bytes crossed 40%** (68.56% funcs / 40.34% bytes):
- `__sinit_ov060_021195dc` (0x814) and `__sinit_ov002_02107370` (0xc18) — the big-sinit vein holds at this size
- `Player::St_Wait_Main` (0x9cc)
- `func_02069994` (0x89c) and `func_0201f32c` (0x820) — Fable retries from stored near-miss drafts (div 10 and 62 → MATCH)

All 5 link-gate verified (one benign BLIND-1 on `func_0201f32c` — unresolvable slot, zero diffs).

Band economics for the runbook (committed in this PR): Sonnet lands the __sinits but 0x800+ *logic* is ~5% for Sonnet even with m2c drafts — its real product is the near-miss seed (19 compiling drafts at div 9–1054, most at exact target size, now in `nearmiss/db.jsonl`). The Fable-on-stored-draft retry then converts at 449K/landed and crushed the rest to div 4–37. New levers in `notes/mwccarm-codegen.md` sec 6h: call-in-pointer-initializer kills the post-`bl` mov, unnamed temps color before named vars, per-site laundering *spellings* defeat cross-site CSE, set-then-conditionally-clear guard flags, unsigned-char-for-plain-ldrb, jump-threading vs materialized bools, explicit `case 0:` forces jump-table. One new floor class: spilled-first-param stack-slot homing order on spill-heavy 0x800+ functions.

Also FYI: a new handle "Cursor/Grok" started claiming today (arm9 0x2009e70, ov006 0x21173c8) — the claims dance has a third dancer.

https://claude.ai/code/session_01HgKJraYsbg7Gg967B8dCpc

andrewboudreau:
Split the runbook/codegen-notes/sched_run.js changes out into #75 to keep this PR scoped to matched functions. Pushed an update (ccc2f05) that drops those three files.

PR #75, opened 2026-07-03
Runbook + codegen notes: 0x800+ band economics, sec 6h levers, new homing-order floor

andrewboudreau:
## Summary
- Runbook update: 0x800+ band economics (Sonnet vs Fable-retry-on-drafts performance)
- New mwccarm-codegen levers (sec 6h): call-in-pointer initializer, unnamed-temp coloring order, per-site laundering spellings, guard-flag set/clear pattern, unsigned char dead-check survival, jump-threading avoidance, explicit case 0, (void*) vs (int) function-address materialization
- sched_run.js prompt: new spilled-first-param stack-slot homing-order floor class

Split out of #74 to keep that PR scoped to matched functions only (no docs/tooling churn).


andrewboudreau:
Split out of #74 (Match 17: wide-band rescan (12) + first 0x800+ band batch (5) — 68.56% funcs / 40.34% bytes) — that PR now contains only the matched-function/near-miss-DB changes; this one carries the runbook + codegen-notes + sched_run.js prompt updates that came out of the same batch.

andrewboudreau:
my agents kind of getting carried away tracking notes in the various markdowns

PR #78, opened 2026-07-03
Update near-miss DB: +76 local-session entries (merged with upstream)

Moundistz:
Merges 76 near-miss records from local matching sessions into the upstream DB (1685 → 1761 entries). All 76 are new keys not present upstream. Lowest divergences: div=1 (func_02030e4c, func_ov006_020de440, func_ov062_02118cdc), div=2 (func_020078c4, func_0203fec4, ov002/ov007/ ov010/ov020/ov026 functions). Covers arm9 and overlays ov002–ov098.


tangosdev:
Thanks for these! Every single one was matched today using your drafts so it truly helped a ton

PR #87, opened 2026-07-05
Match 12 high-divergence structural functions (Fable refine, div 19-25)

andrewboudreau:
Twelve functions matched by pointing the refine specialist at the **high-divergence** end of the near-miss backlog — the "worst to crack" band that the standard `--max-div 6` refine tier never touches. All are genuine structural misses (not codegen floor), driven to byte-match with mwccarm 1.2/sp2p3.

### Functions (12)

| module | function | lang | note |
|---|---|---|---|
| arm9 | func_02070b98 | C | increment-before-store loop order + laundered RMW addrs; strength-reduction off |
| arm9 | func_0205c9b8 | C | split-const-fold laundering + store-order swap (missing-logic) |
| ov013 | func_ov013_021114cc | C | call hoisted to shared exit + `&&` range conjuncts + laundered base |
| ov016 | func_ov016_0211276c | C | u64-mask laundering forces the `+=` RMW full-address materialization |
| ov025 | func_ov025_02111fa0 | C++ | `case 0: break;` flag-reuse zero-test + laundered case-body RMW sites |
| ov030 | func_ov030_02111bc4 | C | CSE'd null-check load pins the arg register + merged `||` guard |
| ov036 | func_ov036_02111854 | C | unconditional +0x110 store block + three laundered RMW base adds |
| ov072 | func_ov072_02120e50 | C++ | merged `||` guard chain, fall-through kills if-conversion |
| ov079 | func_ov079_02123804 | C | volatile-qualified reload defeats guard-register CSE before adjacent store |
| ov081 | func_ov081_02127240 | C++ | laundered pooled-offset materialization + callee moved into case body |
| ov085 | func_ov085_0212a220 | C | laundering forces `addne`+ldrb/strb address materialization |
| ov091 | func_ov091_02132108 | C++ | laundered counter RMW + un-nested unconditional tail (real-vtable PMF) |

Four compile as C++ (real-vtable PMF dispatch). Each replaces a `// NONMATCHING` draft stub except `func_ov079_02123804`, which is new.

### Why high-divergence

A draft at divergence 22 is almost never 22 independent problems — it's usually **one** missing or CSE-elided instruction near the first diff whose absence shifts every downstream branch offset. Fix the earliest real divergence and the rest collapse. The dominant lever across this batch was u64-mask laundering (`*(int*)(((int)base + off) & 0xFFFFFFFFFFFFFFFF)`) to force address materialization the offset-fold otherwise hides.

### Near-miss DB (second commit)

Removes the 12 now-matched functions from the shared `nearmiss/db.jsonl` backlog so they no longer surface as refine candidates.

### Verification

All 12 independently oracle re-verified and link-checked (reconstructed linked bytes): 10 VERIFIED, 2 BLIND (unresolvable invented-extern reloc slots), **0 wrong-dest**.

Commits kept separate: src-only first, DB prune second. The `--high-div` tooling and codegen/runbook notes that produced this batch are in a separate PR.

### DIFF
--- a/src/func_0205c9b8.c
+++ b/src/func_0205c9b8.c
@@ -1,6 +1,3 @@
-// NONMATCHING: missing logic (ROM does more) (div=22). Logic verified correct vs ROM; not
-// byte-matchable from C at mwccarm 1.2/sp2p3 (see notes/matching-style.md).
-// Counts as decompiled, not matched.
 extern void func_0205d874(int *s);
 extern void func_0205d5e8(int *s, void *this, int a, int b, int e);
 extern void func_0205d3d4(int *s, int x, int n);
@@ -13,7 +10,7 @@ unsigned int func_0205c9b8(char *thiz, int arg1, unsigned int arg2) {
     int r4;
     int sz;
     sz = *(int*)(thiz + 0x24) + *(int*)(thiz + 0x2c) + 0x20;
-    r5 = (sz + 0x1f) & ~0x1f;
+    r5 = ((unsigned int)(sz & 0xFFFFFFFFFFFFFFFF) + 0x1f) & ~0x1f;
     if (r5 > arg2) goto done;
     r4 = (arg1 + 0x1f) & ~0x1f;
     func_0205d874(s);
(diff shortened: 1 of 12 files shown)


andrewboudreau:
Rebased onto current `main` and dropped the `nearmiss/db.jsonl` prune commit so this PR is now **src-only** (the DB snapshot was the sole source of conflict as the shared file churned). The 12 matches are unchanged; can prune the corresponding near-miss entries separately if you'd like. Now CLEAN/MERGEABLE.

PR #97, opened 2026-07-06
NONMATCHING: full decomp of Stage::PS_Update (arm9, 12.4KB, div 3032 -> 1170)

ruspecial:
From-scratch decompilation of `_ZN5Stage9PS_UpdateEv` (arm9, `0x0202635c`, size `0x30ac` — 3,115 instructions), the pause-screen update: six timer blocks plus a 20-case state machine. The previous stored draft was a div=3032 stub; this compiles to **div=1170 aligned**, and per-region verification (case bodies extracted via both jump tables and aligned independently) shows:

- **Byte-identical**: prologue + all six timer blocks, the jump table, **16 of 20 case bodies** (0, 2, 3, 4, 5, 6, 9, 0xb, 0xc, 0xd, 0xe, 0xf, 0x10, 0x12, 0x13) and the shared tail.
- **Residuals** (all documented in the file header for the refine tier):
  - case 1: the star-scroll arrow loop is 11 instructions short — the ROM spills five named constant locals to `[sp..sp+0x10]` and keeps `0xf/1/2` in r7/r8/r6; the first arm's dead x-coordinate re-check (`and`+`cmp`) resists every merged-arms spelling tried.
  - cases 8/0xa: touch-region handlers, structure correct but the t-flag chains are -8/-13 instructions and the whole block is register-rotated by one.
  - case 7: `d45c`/`d454` interleaved RMW pair colors r2/r3 swapped; case 0x11: size-exact, coloring rotation only.

Useful discoveries recorded in the header: the dead compare tails come from `if (...) return; return;` tail-merging (bare expressions get DCE'd); `data_020a0e40/dea/deb` behave volatile while `de8/de9` do not; message ID for case 2 is `0x277` (pooled, non-encodable).

Compiler mwccarm 1.2/sp2p3. CLAIMS.md updated — note this takes over the `_ZN5Stage9PS_UpdateEv` row paused by Cursor/Grok on 2026-07-02 (the belongto.us claims key had expired — 401 — so the lock is via CLAIMS.md only).

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -32,3 +32,4 @@ it is fair to take over: ping the claimant first.
 | ov002: func_ov002_020f3310, 020f3d98, 020f562c, 020f5848 (0x4c each) | Cursor/Grok | 2026-07-02 | done - all four verified byte-identical |
 | ov006: func_ov006_0211a4b0 (andrew, PR #74) + 0211a648/0211a7ac/0211aa44/0211abdc/0211ad44/0211af60/0211b17c (Cursor/Grok) | mixed | 2026-07-03 | done - all eight verified byte-identical |
 | arm9: func_02056674 (0x02056674, 0x68) | (already matched earlier) | 2026-07-02 | done previously; Grok edit was a no-op and was reverted |
+| arm9 _ZN5Stage9PS_UpdateEv (0x0202635c, 0x30ac) | ruspecial (Claude-assisted) | 2026-07-06 | NONMATCHING div=1170 from-scratch decomp (prior stub div=3032); 16/20 cases byte-identical; claims API was 401 so locked via this row only — taking over the paused Cursor/Grok row above |
--- a/src/_ZN5Stage9PS_UpdateEv.cpp
+++ b/src/_ZN5Stage9PS_UpdateEv.cpp
@@ -1,0 +1,1426 @@
+//cpp
+// NONMATCHING: 16/20 switch cases byte-identical; residuals: case1 star-scroll loop shape (-11 insn), case8/0xa controller+options touch handlers (-8/-13 insn, plus whole-block register rotations), case7 d45c/d454 RMW pair coloring (8), case0x11 block coloring (41). (div=1170 aligned)
+/* Stage::PS_Update at 0x0202635c (arm9), size 0x30ac (12,460 bytes, 3115 insns)
+ * Compiler mwccarm 1.2/sp2p3, flags:
+ * -O4,p -enum int -lang c++ -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc
+ *
+ * Full decompilation of the pause-screen state machine: 6 timer blocks +
+ * a 20-case switch on data_0209f248. Verified relocation-aware against the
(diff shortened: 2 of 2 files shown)


ruspecial:
Pushed a follow-up session's convergence work to this branch:

- **Case-8 controller-mode blocks decoded** — the change/keep control flow now matches the ROM exactly (`a==0 → change`; default-1/override-0 bool; `t==0` or `f2dc != MODE` → keep). Case-8 size gap −8 → −5 instructions.
- Region-anchored divergence (case bodies extracted via both jump tables, aligned independently): **704 → 690**, still 16/20 case bodies byte-identical.
- Note on metrics: the flat `abverify` aligned div reads 1239 vs 1170 before — that number is anchor-sensitive while the total size skew (−0x9c) persists; the region measure is the meaningful one and is in the header.
- The header now pins down the two hard walls for whoever picks this up: (1) the ROM's language arms are 6 instructions with `slot*4` materialized once at the join and *rematerialized* inside each mode block after call clobbers — every C spelling tried instead hoists one `lsl` into each arm and keeps a long-lived index register; (2) case 0xa's mid-case 47-word literal pool can't align until every preceding byte is exact.

Claims API still returns 401 with the key I have, so coordination remains via the CLAIMS.md row.

ruspecial:
Third session: no net code change (candidate stays at region div 690, 16/20 cases byte-identical), but three decisive **negative results** on the case-1 loop wall, recorded here so the refine tier doesn't re-tread them:

1. **The ROM's five spilled loop locals cannot be reproduced from C constants.** The ROM preheader stores `sl` (the zero in `done`) into three stack slots and `0x51/0x52` into two more. In C, constant-valued locals (`int f1z = 0;` and every variant tried: distinct values like `0x100` that truncate identically, copies `= done`, `register`) get constant-folded/rematerialized by mwcc and never spill — copy-propagation sees through all of them. Only `volatile int` pins the slots, **but** volatile also (a) forces the preheader stores into declaration order while the ROM's are scheduler-reordered (`[sp,4],[sp,8],[sp],[sp,0x10],[sp,0xc]`), proving the ROM locals are *plain*, and (b) shifts downstream allocation (net worse: 698–772).
2. **The spills are pressure-driven and the pressure is shape-coupled.** Rewriting the two touched-checks to the ROM's exact temp structure (`a` folded then `idx`; `de8+idx+1` in check 1 vs re-derived `de8+slot*4+1` in check 2; ternary-shaped `touched2`) *lowers* my build's pressure enough that the frame drops from 0x14 to 0xc — the opposite of the ROM. The missing pressure source is elsewhere in my case-1 codegen.
3. **Tick-pool-pointer hoisting is part of the coupling**: under some spellings my build hoists `&data_0208ee44` into `fp` as loop-invariant (the ROM reloads it from the pool inside each arm), consuming exactly the callee-saved register whose absence changes the spill decisions.

So the case-1 wall is a coupled system: exact temp shapes ↔ register pressure ↔ five spills ↔ whole-loop coloring. A permuter-style scripted sweep over the loop's temp-structure space is probably the right tool from here, not hand search.

Claims API: still 401 with the supplied key.

PR #99, opened 2026-07-06
Match 33 functions: 12 high-div Fable refine + 21 Opus coddog fan-out

andrewboudreau:
Twelve functions matched by pointing the Fable refine tier at the highest-divergence winnable-structural near-misses (`refine_wl.py --high-div`), across two batches.

**Per-module (12 src files):**
| module | functions |
|---|---|
| arm9 | func_020401e8, func_0206e06c |
| ov002 | func_ov002_020bf800, func_ov002_020de968 |
| ov006 | func_ov006_0211a500, func_ov006_0211aa94 |
| ov007 | func_ov007_020c9a2c, func_ov007_020b44ec |
| ov030 | func_ov030_02111384 |
| ov043 | func_ov043_021112a8 |
| ov063 | func_ov063_0211c82c |
| ov064 | func_ov064_0211a39c |

All byte-MATCH (div 0) + link-gate VERIFIED. Also prunes the 12 now-matched drafts from `nearmiss/db.jsonl`.

**Levers that cracked the div 24-25 shapes:** full `this->arr[idx].field` chains restoring mwcc's per-field combined-base pooling (also brings back the lr push/frame); empty inline dtor pinning a stack vector against scalarization; re-reading a field (`*(int*)(c+0x60)+0x50000`) instead of `+=` to reproduce a fresh load; passing `self` to the first callee to keep incoming r0 live; explicit enumeration of all switch cases to force the jump table; `#pragma opt_strength_reduction off` to keep indexed addressing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #104, opened 2026-07-06
NONMATCHING: full decomp of func_ov006_02115b0c (6.3KB) + chaos-data atlas bug report

ruspecial:
From-scratch decompilation of `func_ov006_02115b0c` (ov006, `0x02115b0c`, size `0x18bc` — 1,583 instructions), the minigame-board teardown/rebuild. **Note: the nearmiss DB entry under this name is mislabeled** — its `c_source` is actually `func_ov006_02118b70`; worth fixing so the refine tier doesn't trip on it.

The draft compiles with complete, structurally-aligned logic: the seven inlined-destructor loops are instruction-identical modulo a register permutation, the twelve tilemap scan loops reproduce the signed div/mod-by-32 shift idioms, and the spawn/virtual-init/zeroing sections are in place. `#pragma opt_strength_reduction off` is required (every loop in the ROM is un-reduced; the pragma does not affect div-by-const lowering). Residuals: +140 bytes from per-loop base-pointer hoisting, a frame pointer my build emits that the ROM lacks (suspected interaction with the by-value `V2` argument machinery), and the register skew downstream of that — details in the header.

**Separate finding worth attention: the chaos-data published disassembly text is lossy for some pool words.** Pool words whose garbage decode is `ldrd/strd`-family (`bits[7:4] ∈ {0xB,0xD,0xF}` with a register Rm) print WITHOUT their `imm4H` bits (bits [11:8]) — e.g. `0x3ff` and `0xff` both print as `strdeq r0, r1, [r0], -pc`. Reconstructing target bytes from the listing round-trips those words wrong. Two constants in this function are affected (the tilemap scan mask at `+0x59c`, written as `0x3ff` on domain grounds, and a pooled tile ID at `+0x5ac`, written as `0x2d2` — could be `0x1d2/0x3d2`). Anyone with a ROM can confirm in seconds. Suggested atlas fix: emit pool words as `.word 0x...` lines instead of instruction text — that also removes the reconstruction ambiguity for every other published function.

Compiler mwccarm 1.2/sp2p3. CLAIMS.md updated (claims API still 401 with my key; this also takes over the corresponding entry in the paused Cursor/Grok batch row).

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -43,3 +43,4 @@ it is fair to take over: ping the claimant first.
 | ov002 _ZN6Player18St_YoshiPower_MainEv (0x020d7504, 0x9cc) | ruspecial (Claude-assisted) | 2026-07-05 | done - matched byte-identical, PR open |
 | ov002/006/065/080 batch: func_ov006_020ef794, _ZN6Player19St_CrazedCrate_InitEv, func_ov080_02123fcc, func_ov006_020c87d0, func_ov002_020cec2c, func_ov065_0211a6d0 | ruspecial (Claude-assisted) | 2026-07-06 | done - 6 verified byte-identical |
 | ov063 func_ov063_0211640c (0x0211640c, 0x2a0) | ruspecial (Claude-assisted) | 2026-07-06 | NONMATCHING div=1 (ang*-1 rsb vs ROM smulbb); near-miss draft submitted |
+| ov006 func_ov006_02115b0c (0x02115b0c, 0x18bc) | ruspecial (Claude-assisted) | 2026-07-06 | NONMATCHING full decomp (no prior draft; db entry under this name was mislabeled func_ov006_02118b70); claims API 401 - locked via this row; takes over the paused Cursor/Grok batch row |
--- a/src/func_ov006_02115b0c.c
+++ b/src/func_ov006_02115b0c.c
@@ -1,0 +1,805 @@
+// NONMATCHING: complete from-scratch decomp; structure aligned (dtor loops instruction-identical mod regperm, scan loops match the div/mod-by-32 idioms); residuals: +140 bytes from per-loop base-pointer hoisting differences, an fp frame pointer my build emits (ROM has none; suspected by-value V2 arg machinery), and whole-function register skew downstream of it.
+/* func_ov006_02115b0c at 0x02115b0c (ov006), size 0x18bc (6,332 bytes, 1583 insns)
+ * Compiler mwccarm 1.2/sp2p3, flags:
+ * -O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc
+ * NOTE: #pragma opt_strength_reduction off is required - every loop in the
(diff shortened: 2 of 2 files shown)


andrewboudreau:
**Matched.** ✅ Kicked this draft to Fable (max effort) and it byte-matches the retail ROM now — banked on `matching-fresh` (PR #99), independently link-gate VERIFIED (0 reloc diffs, 0 blind). Puts the project at **69.45% funcs / 42.01% bytes**.

**The residual was one systemic defect, not a register permutation.** Local `abverify` against the ROM showed **1,649 divergences / +140 bytes**, all downstream of a frame pointer your build emits (`mov fp, sp`) that the ROM doesn't have. Root cause: the spawn callees took `V2` **by value**, so mwccarm splits it across `r3/[sp]` with a dynamic sp adjust at each call site, which pins the fp and pulls it into the push set. Fix: pass a pointer to a caller-local copy —
```c
V2 tmp = pos[j]; f(p, c, j, &tmp);
```
Killing the fp collapsed the 1,649 to a two-word load reorder, then to MATCH. (Written up as a general lever now: *by-value small-struct arg → phantom frame pointer*.)

**Your two flagged pool words are resolved** — the match pins the real ROM bytes:
- `+0x59c` scan mask = **0x3ff** ✔
- `+0x5ac` tile ID = **0x3d2** (not `0x2d2`/`0x1d2`)

Confirms your atlas bug report exactly: the `target_hex` I diffed against is ROM-derived and was never lossy — only the *published disassembly text* drops the `imm4H` bits on ldrd/strd-decoded pool words. The `.word 0x...` atlas fix you suggested is the right call.

Also noted your DB-mislabel report (this name's entry actually holding `func_ov006_02118b70`'s source upstream) — worth fixing so the refine tier doesn't trip on it.

Since the matched version supersedes this draft, feel free to close #104 in favor of #99 (or I can fold it in however you prefer).

tangosdev:
Superseded by #105, which byte-matched `func_ov006_02115b0c` by cracking this from-scratch draft with Fable. Thank you @ruspecial — the structural groundwork here (the seven inlined-destructor loops, the signed div/mod-by-32 shift idioms, and the `opt_strength_reduction off` requirement) is what made the match reachable, and #105 credits you for it.

Capturing your two findings so they are not lost:

1. **Atlas pool-word bug** (real, worth fixing repo-wide): the chaos-data disassembly prints ldrd/strd-decoded pool words without their imm4H bits [11:8], so e.g. 0x3ff and 0xff both render as `strdeq r0, r1, [r0], -pc`. Reconstructing target bytes from the listing round-trips those words wrong. Suggested fix: emit pool words as `.word 0x...` lines. #105's matched bytes pin the two affected constants in this function (+0x59c mask = 0x3ff, +0x5ac tile ID = 0x3d2).

2. **Mislabeled nearmiss entry**: the db entry under this name had its `c_source` pointing at `func_ov006_02118b70`. #105 removes the 02115b0c line; the 02118b70 mislabel is worth a separate check so the refine tier does not trip on it.

Closing as superseded by #105.

PR #113, opened 2026-07-07
Match 12 functions: flag-setters, pooled increments, thunks, struct-copy

ruspecial:
Matches 12 ARM9/overlay functions (all byte-exact at mwccarm 1.2/sp2p3):

- **func_ov006_02114fd0 / 02114fec / 02115008 / 02115024 / 02115060** — pooled-offset int increments; u64-mask laundering forces the base to materialize.
- **func_ov002_020aefa4** — base+offset flag setter (`|= 0x8000`), laundered to defeat the first-access fold.
- **func_02035860** — Vec3 double-copy to obj+0x5c and +0x68; laundered bases give `add` + pre-indexed writeback `str [r2,#0x68]!`.
- **func_ov071_0211f6f8** — state handler, adapted from sibling ov081_02126c8c.
- **_ZThn80_N9AnimationD0Ev/D1Ev, func_ov002_020f8838/020f8848** — -0x50 this-adjusting destructor thunks (//cpp dummy-inheritance idiom).

### DIFF
--- a/src/_ZThn80_N9AnimationD0Ev.cpp
+++ b/src/_ZThn80_N9AnimationD0Ev.cpp
@@ -1,0 +1,5 @@
+//cpp
+struct Base1 { int pad[19]; virtual ~Base1(); virtual void f1(); };
+struct Base2 { int b; virtual ~Base2(); virtual void g1(); };
+struct Animation : Base1, Base2 { virtual ~Animation(); };
+Animation::~Animation() {}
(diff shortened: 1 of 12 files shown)


tangosdev:
Superseded by #115, which lands the 8 CI-verified matches from this PR (byte-reproduce the ROM). Dropped the 2 wrong-dest _ZThn80 Animation thunks (reloc -> _ZN9AnimationD1Ev, wrong destructor) and 2 blind ov002 thunks whose relocs could not resolve. Thanks @ruspecial - those 4 can come back once the relocations resolve.

PR #135, opened 2026-07-08
Import 549 symbol names from SplattyDS/DynamicAllocationDecomp

andrewboudreau:
Adopts symbol names from the community [SplattyDS/DynamicAllocationDecomp](https://github.com/SplattyDS/DynamicAllocationDecomp) project (an independent SM64DS reverse-engineering effort) for symbols we still carried as `data_XXXXXXXX` placeholders. Cross-referenced its `symbols.x` against our `config/arm9/**/symbols.txt` by address.

## What's in here

**Commit 1 — symbol import (src + config):**
- **549 renames** across 6 `symbols.txt` files (arm9 main + ov000/001/002/089/100) and all referencing sources — 7,085 references in 1,407 `src/` files.
- Data symbols: level/lookup tables, vtables (`_ZTV*`), cutscene blobs, embedded assets, global state. E.g. `data_02082214 → SINE_TABLE`, `data_0208e3a4 → _ZTV5Actor`, `data_02075264 → LEVEL_PART_TABLE`.
- Aligns `symbols.txt` with the semantic names decompilers had already hand-written as externs in `src/` (the `@ 0xADDR` comments), unifying the two naming styles.

**Commit 2 — nearmiss sweep (companion):**
- Same 549 renames swept through `nearmiss/db.jsonl` stored `c_source` (2,796 replacements, 573/1,376 records). Kept as a separate commit so it can be split into a companion PR if preferred.

## Vetting (address-only joins pruned to avoid false matches)
- Excluded 11 targets whose name already existed elsewhere in our tables.
- Excluded 45 addresses mapping to >1 overlay (ov000/ov001 overlap) — an address-only map can't disambiguate.
- **Dropped the entire `func_` bucket** — those addresses collided across overlapping overlays (e.g. ov002 vs ov004 both at `0x020aea30`); join artifacts, not real matches.
- Left the 10 name-conflict cases (both projects named, differ) for manual review; not applied.

## Verification
No bytes change (identifier-only). Confirmed 0 leftover placeholders, no new duplicate symbols, line integrity preserved, and a sample of renamed functions across arm9 and ov002 still compile and match byte-exact via `tools/match.py` (1.2 base/sp2/sp2p3 trio). Near-miss DB re-validated as JSON with unchanged record count.


andrewboudreau:
yeah, i'm not sure when is the best time to do something like this, as soon as possible or at 100%.  Feel like there is so much in flight right now and how things are changing it might be better to do later.

PR #142, opened 2026-07-08
Match 21 functions + 5 near-misses (flag RMWs, state inits, dispatch loops, vector math)

ruspecial:
Adds **21 byte-exact matches** and **5 near-misses** (all `// NONMATCHING`, size-correct byte diffs) from an Opus matcher batch. Every match was independently re-verified against the ROM with `mwccarm 1.2/sp2p3`.

### Matches (21)
- **Flag ops / pooled RMWs** (u64-mask launder to force base materialization): `func_ov006_020f4ad4`, `func_ov006_020dce3c`, `func_ov002_020d6084`, `func_ov006_021082c4`, `func_ov006_02107d20`, `func_ov002_020c9e18`, `func_ov006_020fe2bc`
- **State inits** (SetAnim + laundered increment): `_ZN6Player19St_SwingPlayer_InitEv`, `func_ov002_020c9c4c`, `func_ov084_0212fc10`, `func_ov030_02112ff8`, `func_ov070_0211f62c`, `func_ov006_020df540`, `func_ov006_020dabec`
- **Loops**: `func_ov002_020e6b74` (per-element copy loop), `func_ov006_02106bc0` (per-entry `blx` dispatch loop)
- **Vector math**: `_ZN8Particle8Converge4FuncERNS_10EffectDataEPcR7Vector3` (per-axis Fix12 update)
- **Two-word struct-copy ordering** (no-op ternary idiom to force both loads before both stores): `func_ov006_02123b24`, `func_ov006_02121778`, `func_ov006_020c7ba4`
- `func_ov007_020aebac`

### Near-misses (5, `// NONMATCHING`)
`Vec3_AsrInPlace`, `Vec3_LslInPlace`, `func_0205c410`, `func_0205c448`, `func_0205d568` — all size-correct with logic verified against the ROM; the only divergence is pure register-coloring / instruction-scheduling that source reordering does not move (candidates for the permuter).

### Notes
- Based directly on `main`; the diff is exactly these 26 files (no unrelated commits).
- Three functions from the original batch were dropped because they had already landed upstream (`_ZN6Player18St_DizzyStars_InitEv`, `func_ov006_020e82c8`, `func_02009414`).
- Three additional near-misses were **excluded** because they are size-mismatched (structurally incomplete), not just byte-diffs, so they don't fit the `NONMATCHING` convention.


tangosdev:
Landed the 18 byte-verified matches. Dropped 8 files that failed `validate` (per MERGE.md section 2 — never commit a file that doesn't reproduce the ROM):
- **Near-miss (don't reproduce):** Vec3_AsrInPlace, Vec3_LslInPlace, func_0205c410, func_0205c448, func_0205d568
- **Blind (unresolvable reloc slot):** func_ov002_020c9c4c, func_ov006_020c7ba4, func_ov006_02106bc0

Thanks for the batch. The 5 near-misses are good candidates for the near-miss DB if you want to resubmit them there.

PR #170, opened 2026-07-09
tangos/glm: matched functions (20260709-1702)

tangosdev:
Automated per-agent PR from tangOS Console - matched functions from **glm** this session. CI validation + your review gate the merge.


tangosdev:
Both files are wrong-dest: they compile byte-identical but a relocation points at the wrong symbol, so they're not real matches (CI correctly fails them). The local link gate detected this but only warned instead of unbanking, so they got auto-pushed. Fixed in tools/crackloop.py (cd08a8d9): the LINK GATE now auto-unbanks wrong-dest banks (removes the src file + ledger row) so they never reach a PR.

PR #204, opened 2026-07-10
Fix duplicate _ZN5SceneD0Ev symbol (0x020235d4 is BootScene)

tangosdev:
Second of the duplicate-symbol landmines (after _Znwj). Two functions were both named `_ZN5SceneD0Ev`. The vtable each installs is decisive: 0x020235d4 installs `_ZTV9BootScene` -> it's `_ZN9BootSceneD0Ev`; 0x0202e170 installs `_ZTV5Scene` -> the real `Scene::~Scene`. Renames the BootScene symbol+src and drops the `_0x0202e170` filename workaround. Byte-neutral (both verified locally); validate re-checks the two renamed files.


tangosdev:
Superseded by a version that also switches the dtors' vtable refs to the config data_ names so they link-verify (they were committed blind on _ZTV* names).

PR #205, opened 2026-07-10
Fix duplicate _ZN5SceneD0Ev (0x020235d4 = BootScene) + link-verify both dtors

tangosdev:
Dedup + link-fix. 0x020235d4 installs _ZTV9BootScene -> _ZN9BootSceneD0Ev; 0x0202e170 installs _ZTV5Scene -> real Scene dtor. Both dtor files referenced vtables by _ZTV* names not in config (committed blind); now use the data_<addr> config symbols the ~40 other vtable users already reference, so linkcheck resolves. Byte-neutral.


tangosdev:
Closing: fixing the duplicate surfaced that both Scene-D0 files are pre-existing near-misses (0x0202e170 is NO-REPRO on 2 slots regardless of symbol names - a real codegen divergence, not resolution). They were committed as matches but never link-verify. Needs a decision on how to handle the false matches, not a blind rename.

PR #208, opened 2026-07-10
tangos/claude: matched functions (20260710-0118)

ruspecial:
Automated per-agent PR from tangOS Console (from fork `ruspecial/sm64ds-decomp`) - matched functions from **claude** this session. CI validation + your review gate the merge.

### DIFF
--- a/src/func_ov001_020ab228.c
+++ b/src/func_ov001_020ab228.c
@@ -1,27 +1,43 @@
-// NONMATCHING: different op / idiom (div=18). Logic verified correct vs ROM; not
-// byte-matchable from C at mwccarm 1.2/sp2p3 (see notes/matching-style.md).
-// Counts as decompiled, not matched.
-extern void* data_ov001_020ad634[];
-extern unsigned char data_ov001_020ad630[];
-extern unsigned char data_ov001_020ad628[];
 extern void func_ov001_020aa6cc(int r0);
 
-void func_ov001_020ab228(char* c, char* a1, int idx, int a3, unsigned char a5){
-    *(unsigned char*)(c+0x18) = (unsigned char)idx;
-    *(int*)(c+4) = (int)a1;
-    *(int*)(c+8) = *(int*)(a1+4);
-    *(unsigned char*)(c+0x19) = (unsigned char)a3;
-    *(int*)(c+0x10) = (int)data_ov001_020ad634[idx];
-    char* prev = (char*)data_ov001_020ad634[idx];
-    if(prev) *(int*)(prev+0xc) = (int)c;
-    data_ov001_020ad634[idx] = c;
-    data_ov001_020ad630[idx]++;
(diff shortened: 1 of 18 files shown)


tangosdev:
**`validate` is red on 2 of the 19 files -- both `operator new` veneers.**

`src/_ZN6Memory13operator_new2Ej.cpp` and `src/_Znwj.cpp` are 12-byte tail-call veneers whose `.word` relocation should target the real allocator `func_0203cc0c` (0x0203cc0c). Right now both target `_Znwj` (0x0203cbe4) -- which is itself the veneer, so `_Znwj.cpp` ends up tail-calling itself. Point both the `extern` decl and the call at `func_0203cc0c` and they reproduce the ROM (WRONG -> VERIFIED under `tools/linkcheck.py`).

The other 17 files verify clean (one BLIND-1 on `func_ov007_020ccad0` is a benign unresolved-symbol slot with no byte diff, not a failure).

Heads-up: the branch force-pushed over a fix commit I had pushed here, so this edit needs to happen in your generator to stick.

tangosdev:
Update: the veneer issue from my earlier comment is resolved (those two files are gone from the current push), and all 18 files now reproduce the ROM bytes. But `validate` is still red because of a naming mismatch introduced by #211 (the actor-symbol rename).

Three files use stale `func_ov*` names for addresses that #211 renamed. On this branch, config already carries the new names, so the build box cannot map these files to their symbols:

- `src/func_ov064_0211aa58.c` -> config names 0x0211aa58 `_ZN4Clam8BehaviorEv`
- `src/func_ov085_0212d39c.c` -> config names 0x0212d39c `_ZN9RabbitKey6RenderEv`
- `src/func_ov102_0214d2e8.cpp` -> config names 0x0214d2e8 `_ZN10KoopaShell8BehaviorEv`

Rename each file and its function definition to the config symbol name (one matched function per `src/<symbol>.c|.cpp`) and validate should go green -- the bytes themselves already match.

Root cause is upstream: the generator is producing this branch against config from before #211 landed. Pulling latest config (with the actor names) before regenerating will fix this for good; your newer PR #215 already used the correct `_ZN` names and passed green. Note a fix commit pushed directly to this branch gets reverted by the next force-push, so this needs to happen in the generator.

tangosdev:
Landed the matching content myself in e983c817. Breakdown of the 18 files here:

- 8 functions still needed landing -> matched byte-identical and pushed to main, with the one stale name fixed (`func_ov055_02111508` -> `_ZN11MirrorLuigi8BehaviorEv` per config after #211). All linkcheck VERIFIED.
- `func_ov064/085/102` (Clam/RabbitKey/KoopaShell) already landed under their real `_ZN` names via #217 + the batch merges.
- The remaining ov007/ov025 functions were already on main.

Credit for the 8 is attributed to you. Closing as integrated.

PR #234, opened 2026-07-10
NONMATCHING: func_ov070_0211f48c near-miss (div=4) for refine

lunavyqo:
## Summary

Near-miss draft for `func_ov070_0211f48c` @ `0x211f48c` (size `0x164`, ov070).

**div=4** — only remaining mismatch is r0/r1 coloring on:

```
ldrh r1, [r5, #0x8c]   ; ROM
ldrh r0, [r5, #0x8e]
```

Candidate swaps those two registers (values still correct). Everything else byte-matches including the dual Vector3 copy for HorzAngle, ApproachAngle, SpawnFireball, and Finished→state transition.

## Compiler

mwccarm `1.2/sp2p3` with the standard flags.

## Why land this

Highest-yield refine input: 4-word pure-coloring floor after the Crate-style Vector3 materialization and u64-mask base for `player+0x5c` already land.

```
// NONMATCHING: r0/r1 coloring on dual ldrh for vel.x/vel.y (div=4)
```


andrewboudreau:
Closing as already landed. This PR's source contribution is already on `main`, byte-for-byte identical, via direct commit 0ab90bc ("NONMATCHING: func_ov070_0211f48c near-miss for refine (div=4)"). GitHub kept this PR open because the content reached `main` through a separate direct commit rather than a merge of this branch, so no merge event was ever recorded. Verified: every src file in this PR matches `main` exactly. The branch's only remaining diff is a stale CLAIMS.md edit based on an older main (it would regress newer rows), so it should not be merged. No action needed — the work is preserved on main.

PR #244, opened 2026-07-10
NONMATCHING near-misses: three ov020 functions (refine tier)

lunavyqo:
## Summary
Near-miss compiling drafts for refine (not byte-identical). Highest-yield input for the refine tier.

| Function | Address | Size | Status |
|---|---|---|---|
| `func_ov020_02111340` | `0x02111340` | `0xd8` | NONMATCHING div=13 (mul/add scheduling after FindWithID) |
| `func_ov020_02113148` | `0x02113148` | `0xb0` | NONMATCHING div=22 (regperm: f not in ip; smulbb coloring) — **size-matched** |
| `func_ov020_021115ac` | `0x021115ac` | `0x1a0` | NONMATCHING (stack frame `#4` vs `#0xc`; velocity zeroing store shape) |

Logic verified against ROM disassembly. Compiler: **mwccarm 1.2/sp2p3**.

Related byte-match PR: https://github.com/tangosdev/sm64ds-decomp/pull/242

## Test plan
- [ ] Review as near-miss / refine input (validate expected red for NONMATCHING)
- [ ] Permuter / refine tier can pick up from these drafts


andrewboudreau:
Closing as already landed. This PR's source contribution is already on `main`, byte-for-byte identical, via direct commit 220256f ("NONMATCHING near-miss drafts for three ov020 functions"). GitHub kept this PR open because the content reached `main` through a separate direct commit rather than a merge of this branch, so no merge event was ever recorded. Verified: every src file in this PR matches `main` exactly. The branch's only remaining diff is a stale CLAIMS.md edit based on an older main (it would regress newer rows), so it should not be merged. No action needed — the work is preserved on main.

PR #248, opened 2026-07-10
Near-miss RollingRock Behavior + helper (ov021)

lunavyqo:
## Summary
Near-miss drafts for refine tier.

| Function | Addr | Size | Status |
|---|---|---|---|
| `_ZN11RollingRock8BehaviorEv` | `0x02112854` | `0x2d8` | size cand `0x2d4` (missing ~1 pool word, likely `0x3bf`) |
| `func_ov021_02112544` | `0x02112544` | `0x270` | **div=93** — LDM/STM struct pos vs interleaved field loads |

Logic is sound; both compile. Marked `// NONMATCHING`.

## Compiler
mwccarm **1.2/sp2p3**


andrewboudreau:
Closing as already landed. This PR's source contribution is already on `main`, byte-for-byte identical, via direct commit d42c7db ("Near-miss RollingRock Behavior + ground/wall helper (ov021)"). GitHub kept this PR open because the content reached `main` through a separate direct commit rather than a merge of this branch, so no merge event was ever recorded. Verified: every src file in this PR matches `main` exactly. The branch's only remaining diff is a stale CLAIMS.md edit based on an older main (it would regress newer rows), so it should not be merged. No action needed — the work is preserved on main.

PR #267, opened 2026-07-11
Bank 52 near-misses from today's backlog (nearmiss/db.jsonl)

ruspecial:
Coordination data only — no `src` changes, no `validate` gate.

Ingested 72 close-but-not-matching attempts from the day's working tree via `tools/nearmiss_db.py ingest` (each re-compiled, reloc-aware divergence measured, closest-per-function kept, already-matched entries auto-dropped): **+3 new, 49 improved** over the current DB. Net 1372 → 1375 entries with 49 existing rows brought closer.

These are the non-matching half of today's backlog; the byte-verified half is in #265.


tangosdev:
Banked into main via union merge with #266's near-miss entries (both preserved, 0 lost). Closing as integrated.

PR #291, opened 2026-07-12
Match 6 functions byte-identical (mwccarm 1.2/sp2p3)

ruspecial:
6 new byte-exact matches. Each compiles with mwccarm 1.2/sp2p3 and reproduces the ROM bytes (verified locally with `match`), and every relocation was audited to resolve to the config-recorded callee/global (`reloc_audit` CLEAN — no WRONG-DEST).

| function | module | size | shape |
|---|---|---|---|
| `func_ov002_020af0c0` | ov002 | 0x11c | player-tracking (atan2 / ApproachLinear / sine) |
| `func_ov006_020fc9b0` | ov006 | 0x6c | per-element range check |
| `func_ov006_02105c1c` | ov006 | 0x6c | predicated compare loop |
| `func_ov006_02119a18` | ov006 | 0x70 | counter-advance state fn (data-table reloc) |
| `func_ov006_0211e184` | ov006 | 0x9c | counter/threshold loop |
| `func_ov006_0211e4e0` | ov006 | 0x7c | walking-pointer decrement loop |

src-only, one function per file. `validate` is the gate.

### DIFF
--- a/src/func_ov002_020af0c0.c
+++ b/src/func_ov002_020af0c0.c
@@ -1,0 +1,37 @@
+typedef struct Vector3 { int x, y, z; } Vector3;
+#define AT(p,off) ((void*)(int)(((long long)(int)((char*)(p)+(off)))&0xFFFFFFFFFFFFFFFFLL))
+extern short data_02082214[];
+#define SINE_TABLE data_02082214
+extern char* _ZN5Actor13ClosestPlayerEv(void*);
+extern int func_ov002_020d0d2c(void*);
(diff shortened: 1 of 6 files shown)


tangosdev:
Superseded: all 6 functions in this PR are already on `main`, byte-identical to this branch (0 diff lines per file). Spot-checked func_ov002_020af0c0 and func_ov006_0211e4e0 with linkcheck against the ROM: both VERIFIED, 0 blind. Nothing to merge here since the matches already landed. Thanks ruspecial — closing as already-landed, not as a rejection of the work.

PR #293, opened 2026-07-12
tangos: matched work (tangos/andrew)

andrewboudreau:
Matched functions pushed from tangOS Console. Review + CI gate the merge.

### DIFF
--- a/src/_ZN6Player17St_Headstand_MainEv.cpp
+++ b/src/_ZN6Player17St_Headstand_MainEv.cpp
@@ -1,0 +1,74 @@
+//cpp
+typedef int s32;
+typedef short s16;
+typedef unsigned int u32;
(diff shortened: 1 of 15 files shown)


tangosdev:
Verified against main before acting on this: 14 of the 15 files are blob-identical to copies already on main, all authored by you and landed via earlier merges (647d539b, c768227f, 864e13e0, 6c67fa01), so this branch is a leftover slice of work that already shipped and your attribution is unaffected. The 15th, src/func_020469e8.c, was independently matched and landed on main by ruspecial (4b51fc3a) with a typed-struct source for the same bytes, which is why the PR shows as conflicting. Nothing here is missing from main, so closing as superseded rather than merging. Same situation as #271/#273. Thanks as always.

PR #295, opened 2026-07-12
Match func_020469e8 byte-identical (mwccarm 1.2/sp2p3)

andrewboudreau:
## Summary
- `func_020469e8` now matches byte-identical to the ROM under mwccarm 1.2/sp2p3.

## Context
This was part of a 15-function match session on `tangos/work`; 14 of the 15 had already landed on `main` via other merges (identical content confirmed by blob hash), so this PR carries just the one remaining function.

## Test plan
- [x] Verified byte-identical match locally during the matching session

### DIFF
--- a/src/func_020469e8.c
+++ b/src/func_020469e8.c
@@ -1,68 +1,62 @@
-typedef unsigned short u16;
-
-struct Entry {
-    u16 id;     /* +0x00 */
-    u16 p2;
-    u16 p4;
-    u16 p6;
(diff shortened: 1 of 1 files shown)


andrewboudreau:
False alarm — main's func_020469e8.c is already the matched+cleaned-up version (docs/index.html confirms matched:yes, and it's out of the active worklist). CI confirms my version also compiles byte-identical, but main's is the cleaner, already-polished form — no reason to replace it. Closing with no change needed.

PR #347, opened 2026-07-15
Match 15 functions byte-identical (mwccarm 1.2/sp2p3)

natbree:
Matched 15 functions using automatic template matching (swarm.py) and manual verification.

All functions verified byte-identical with mwccarm 1.2/sp2p3.

Generated with Qwen Code

### DIFF
--- a/src/SetNumPlayers.c
+++ b/src/SetNumPlayers.c
@@ -1,8 +1,7 @@
-extern unsigned char data_0208a0e0;
 extern void func_020308d0(void);
+extern unsigned char data_0208a0e0;
 
-void SetNumPlayers(int a)
-{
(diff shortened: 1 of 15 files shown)


tangosdev:
Thanks for the first time contribution! 9 of these merge perfectly! One small heads-up for next time: six of the files in the PR were edits to functions that were already matched on main. Since the originals already reproduce the ROM, changing them just adds churn on landed work, so I dropped those six and kept your 9 new matches. For future PRs, the smoothest path is new matches only if a src/ file already exists, it's already done.

If you're up for more, come hang out in the project Discord! Its where we colab and talk: https://discord.gg/JXckvdARc

Really nice work would love to have you around. Thanks again!

PR #361, opened 2026-07-16
Match 3 functions byte-identical (mwccarm 1.2/sp2p3)

ai-tdd-labs:
Three matches, each verified locally with `tools/match.py --strict-relocs` against an EU retail dump:

| Function | Address | Size |
|---|---|---|
| `_ZNK9Animation12WillHitFrameEi` | 0x02015a98 | 0x134 |
| `DMASyncFillTransfer` | 0x0205a1b4 | 0x68 |
| `func_02058568` | 0x02058568 | 0x64 |

Notes:
- `DMASyncFillTransfer` uses the u64-laundering idiom from `notes/mwccarm-codegen.md` §6m to get the pre-indexed writeback store; its `bl` resolves to `DMAStartTransferFB` (ITCM).
- `func_02058568` is an OS_InitContext-style hand-asm SDK primitive (sibling of the already-banked `ARMSaveContext`, same original TU). ~200 C formulations floor at a 5-word register-coloring residual, so it is banked as a byte-faithful `asm` block per the asm-hatch policy in `notes/mwccarm-codegen.md` §8. If maintainers prefer to hold out for C here, happy to drop it from the PR.

### DIFF
--- a/src/DMASyncFillTransfer.c
+++ b/src/DMASyncFillTransfer.c
@@ -1,0 +1,43 @@
+/* DMASyncFillTransfer at 0x0205a1b4
+ * Synchronous 32-bit DMA fill: waits for the DMA channel to be free, writes
+ * the fill value into the channel's DMA fill register (0x040000E0 + ch*4),
+ * then starts a source-fixed word DMA from that register (ctrl 0x85000000 =
+ * enable | 32-bit | src-fixed | immediate) and waits for completion.
+ *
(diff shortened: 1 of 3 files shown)


tangosdev:
Thanks Chris - three clean matches, all byte-verified against the ROM and merged. func_02058568 as an asm-hatch SDK primitive is fine, that's exactly what notes/mwccarm-codegen.md section 8 is for, no need to drop it.

If you want to claim spans and compare notes with the other contributors, here's the Discord: https://discord.gg/JXckvdARc

PR #387, opened 2026-07-16
asm: func_02068398 & func_02069918 (arm9) - null-check guard functions

natbree:
Two asm-hatched functions with early-return paths using conditional ldm/bx instructions.

**func_02068398** (arm9, 0x78): 28/30 entries match byte-for-byte.
**func_02069918** (arm9, 0x7c): 26/28 entries match byte-for-byte.

The remaining mismatches are an inherent mwccarm assembler limitation — the target uses `ldm{cond}`/`bx{cond}` for early-return paths, but mwccarm's inline assembler rejects those mnemonics entirely. The functionally equivalent `ldreq r12, [sp], #4` + `moveq pc, r12` substitute assembles to different ARM encodings (`popeq {ip}` + `moveq pc, ip`).


tangosdev:
Thanks for the submission. Closing this one, but here's exactly what happened so the next PR lands clean.

**Why this couldn't merge as-is:**

1. **Stale base.** The branch was cut from an old `main`, so the diff re-includes ~16 src files that other contributors have already landed (the `ov006` batch, the `Actor`/`Model`/`Sound` functions, etc.). Compared against current `main`, those 16 files are byte-identical to what's already there, so there's nothing to add from them.

2. **Generated files got rewritten.** `README.md`, `contributions.json`, `docs/index.html`, and `docs/progress-treemap.svg` are all tool-generated. They're regenerated automatically after merges, so hand-edited versions just create conflicts. Please leave them out of PRs.

3. **Tooling changes.** The PR also modifies `tools/match.py`, `tools/fdiff.py`, `tools/glm_refine.py`, and `tangos.json`. Tooling changes need to go in their own separate, reviewed PR, not bundled into a function-match PR.

4. **The two headline functions aren't matches.** `func_02068398` (28/30) and `func_02069918` (26/28) aren't byte-identical, and you've correctly identified why: mwccarm's inline assembler won't emit `ldm{cond}`/`bx{cond}`, and the `ldreq`/`moveq pc` substitute assembles to different encodings. That makes them near-misses, not matches. (`func_02069918` is also already matched on `main` via #388.) A near-miss shouldn't land as a committed `src/` file, since it reddens the `validate` gate; it belongs in `nearmiss/db.jsonl` instead.

**For next time:**

- Branch from the latest `main` (`git fetch origin main` then rebase/branch from `origin/main`) right before you start, so you don't pick up already-landed work.
- Keep a match PR to just the `src/` files you actually matched byte-for-byte. Nothing else.
- Never touch generated files (`README.md`, `contributions.json`, `docs/`, the treemap) or tooling (`tools/`, `tangos.json`) in a match PR.
- If a function can't reach byte-identical (like these two `ldm{cond}` cases), bank it as a near-miss in `nearmiss/db.jsonl` rather than committing a non-matching `src/` file.

Appreciate the analysis on the two guard functions, that `ldm{cond}` limitation writeup is genuinely useful. Please resubmit anything you've actually matched from a fresh `main` and it'll go right in.

PR #417, opened 2026-07-17
tangos/grok: matched functions (20260717-0200)

tangosdev:
Automated per-agent PR from tangOS Console - matched functions from **grok** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.2.9** (each push re-stamps the commit message with the build that made it).

### DIFF
--- a/src/_ZN7BooCage8BehaviorEv.cpp
+++ b/src/_ZN7BooCage8BehaviorEv.cpp
@@ -6,10 +6,11 @@ struct WithMeshClsn {
     bool JustHitGround() const;
     bool IsOnGround() const;
 };
+struct CylinderClsn;
 struct Actor {
     static Actor* FindWithID(unsigned int id);
     void LandingDust(bool b);
-    void UpdatePos(void* cyl);
+    void UpdatePos(CylinderClsn* cyl);
 };
 struct Player {
     void JumpIntoBooCage(Vector3& v);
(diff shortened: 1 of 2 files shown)


tangosdev:
Holding this one — `func_ov006_020d6170.c` verifies clean, but `_ZN7BooCage8BehaviorEv.cpp` byte-matches only because the match gate wildcards relocation slots. Local link-verification (the CI `validate` build box is temporarily down, so I'm verifying PRs locally) shows all three of its `bl` relocations point at symbols that **don't exist** — the simplified type signatures mangle to names the game doesn't have, so a real link leaves them undefined:

| call site | this source emits | correct symbol (in config) |
|---|---|---|
| `+0x24` PlaySecretSound | `_Z36_ZN5Sound15PlaySecretSoundEP5ActorPtPvPt` | `_ZN5Sound15PlaySecretSoundEP5ActorPt` @ 0x0200f8f8 |
| `+0xf4` Actor::UpdatePos | `_ZN5Actor9UpdatePosEPv` | `_ZN5Actor9UpdatePosEP12CylinderClsn` @ 0x02010c30 |
| `+0xe4` Particle::System::New | `_ZN8Particle6System3NewEjjiiiPvPNS_8CallbackE` | `_ZN8Particle6System3NewEjj5Fix12IiES2_S2_PK11Vector3_16fPNS_8CallbackE` @ 0x02022d80 |

Fixes (the instruction bytes already match — only the declared types need to produce the right mangled names):

1. **PlaySecretSound** — line 31 is missing `extern "C"` (line 32's `func_ov063` has it), so the already-mangled name gets mangled a second time. Add `extern "C"`, or declare it as `struct Sound { static int PlaySecretSound(Actor*, unsigned short*); };`.
2. **Actor::UpdatePos** — change the parameter from `void* cyl` to `CylinderClsn* cyl` so it mangles to `...EP12CylinderClsn`.
3. **Particle::System::New** — use the real signature: a `Fix12<i>` class type (not `typedef int Fix12i`) for the three coordinates and `const Vector3_16f*` (not `void*`) for the second-to-last arg, to mangle to `...Ejj5Fix12IiES2_S2_PK11Vector3_16f...`.

Push those and it'll link-verify clean. Nice decomp otherwise — the control flow and offsets are all correct.

PR #438, opened 2026-07-17
Match func_ov063_02118914 byte-identical (mwccarm 1.2/sp2p3)

mitchellcairns:
## Summary
- Byte-identical match of `func_ov063_02118914` (ov063 @ `0x02118914`, size `0xe0`) under mwccarm 1.2/sp2p3
- Verified locally with `match.py` and `linkcheck.py` (VERIFIED, 0 blind)

## Test plan
- [ ] CI `validate` green on `src/func_ov063_02118914.c`

### DIFF
--- a/src/func_ov063_02118914.c
+++ b/src/func_ov063_02118914.c
@@ -1,0 +1,32 @@
+typedef unsigned char u8;
+typedef unsigned short u16;
+typedef short s16;
+
+void func_ov063_02118914(char* c)
+{
+    int z = 0;
+    int scale;
+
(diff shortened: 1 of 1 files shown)


tangosdev:
Closing as a race: `func_ov063_02118914` landed byte-identical on main via #437 a few minutes before this, so there's nothing left to merge — both versions verify to the same ROM bytes, #437 just got there first. Tip for avoiding wasted effort: check CLAIMS.md (or open a claim row) before starting a function, and pull latest main right before submitting. Great run on the printf/soft-double cluster tonight — those five had been walled for weeks.

PR #475, opened 2026-07-18
tangos/claude: matched functions (20260718-0336-7dec)

tangosdev:
Automated per-agent PR from tangOS Console - matched functions from **claude** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.2.21** (each push re-stamps the commit message with the build that made it).

### DIFF
--- a/src/func_ov064_0211a4c4.c
+++ b/src/func_ov064_0211a4c4.c
@@ -1,0 +1,103 @@
+//cpp
+typedef signed short s16;
+typedef unsigned short u16;
+typedef unsigned char u8;
+typedef unsigned int u32;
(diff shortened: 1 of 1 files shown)


tangosdev:
CI `validate` flags this as a near-miss (div=1 — does NOT reproduce the ROM byte-for-byte after relocation), so it can't merge to src/ per the byte-match gate. It belongs in nearmiss/db.jsonl, not src/. (Note: a local match.py check false-passed it because reloc-destination resolution needs current symbols; CI on the build box is authoritative.) Closing; re-open once it reaches div=0.

PR #484, opened 2026-07-18
VERIFY: 4 stranded near-miss drafts (do not merge — CI verdict probe)

tangosdev:
Probe only. Four uncommitted near-miss drafts salvaged from an overnight session's working tree, already banked in nearmiss/db.jsonl. Opening to get CI's authoritative per-file verdict (local match.py is stale on this box). Expect validate to FAIL overall; reading the per-file table to confirm each is a near-miss (no action) or catch any that is secretly a real match (salvage it). Will close after reading.

### DIFF
--- a/src/func_ov004_020b38ac.c
+++ b/src/func_ov004_020b38ac.c
@@ -1,0 +1,55 @@
+typedef long long s64;
+typedef unsigned short u16;
+extern int func_02053200(int x);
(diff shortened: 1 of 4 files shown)


tangosdev:
Probe complete — CI confirms all 4 are genuine near-misses (div=1, already banked in nearmiss/db.jsonl). No stranded match among them. Closing.

PR #506, opened 2026-07-19
Match AngleDiff (func_0203b0e8) and func_02017684 byte-identical (mwccarm 1.2/sp2p3)

natbree:
Match **2 new functions** on arm9 (28 files total in this PR).

| Function | Address | Size | Notes |
|---|---|---|---|
| AngleDiff | 0x0203b0e8 | 0x14 | Pure arithmetic, no calls/data refs |
| func_02017684 | 0x02017684 | 0x14 | FaderBrightness::IsAtStart equivalent |

Both verified byte-identical with mwccarm 1.2/sp2p3.

\`\`\`c
// AngleDiff
int AngleDiff(int a, int b) {
    return ((a - b) << 16 >> 16) < 0 ? -((a - b) << 16 >> 16) : ((a - b) << 16 >> 16);
}

// func_02017684 (FaderBrightness::IsAtStart)
int func_02017684(int self) {
    return *(int *)(self + 4) == 0 ? 1 : 0;
}
\`\`\`

**Note:** PR #505 was accidentally created from a stale branch with rebased commits. This clean PR replaces it.

### DIFF
--- a/src/func_02017684.c
+++ b/src/func_02017684.c
@@ -1,0 +1,3 @@
+int func_02017684(int self) {
+    return *(int *)(self + 4) == 0 ? 1 : 0;
+}
--- a/src/func_02019638.c
+++ b/src/func_02019638.c
@@ -1,0 +1,5 @@
+void func_02019638(void *self) {
+    *(unsigned char *)(self + 8) = 0;
+    *(int *)self = 0;
+    *(int *)(self + 4) = 0;
+}
--- a/src/func_0202096c.c
+++ b/src/func_0202096c.c
@@ -1,0 +1,5 @@
+extern unsigned char data_020755ac[];
+
+unsigned char func_0202096c(int off, int idx) {
+    return *(data_020755ac + idx + off * 4);
+}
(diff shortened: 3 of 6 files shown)


tangosdev:
Thanks, and this is much cleaner than before — but closing it too, because all 6 functions are already matched on `main` under their proper (demangled) filenames, and your files re-add them as generic `func_<addr>.c`. Merging would create two source files for each function (a duplicate-symbol conflict). The existing versions:

- `func_02017684.c` → already `src/_ZN15FaderBrightness9IsAtStartEv.c`
- `func_02019638.c` → already `src/_ZN5Timer10ResetTimerEv.c`
- `func_0202096c.c` → already `src/_ZN3OAM12GetObjHeightEii.c`
- `func_02020980.c` → already `src/_ZN3OAM11GetObjWidthEii.c`
- `func_020393dc.c` → already `src/_ZN16MeshColliderBase9IsEnabledEv.cpp`
- `func_0203b0e8.c` → already `src/AngleDiff.c`

All six are byte-matched already, so nothing's lost. The underlying issue is the same as #505: your fork is still behind `main`, so your worklist is handing you functions that have since been matched by others. Once you hard-sync your fork to current `upstream/main` and regenerate your worklist, it'll only surface genuinely-unmatched functions and your PRs will land clean. Really appreciate the effort here!

PR #538, opened 2026-07-22
Match 3 functions byte-identical (mwccarm 1.2/sp2p3)

andrewboudreau:
## Summary
- `func_ov007_020b2cf0` (ov007, 0x174)
- `func_ov084_02129cf4` (ov084, 0x1e0)
- `func_ov102_0214ab1c` (ov102, 0x1f8)

All three byte-identical across 1.2/base, sp2, sp2p3; linkcheck VERIFIED (0 blind slots) on each.

## Test plan
- [x] `tools/match.py --trio` — MATCH on all 3 versions for each function
- [x] `tools/linkcheck.py` — VERIFIED, 0 blind, for each function


andrewboudreau:
Superseding with a clean 2-func PR — func_ov007_020b2cf0 was already matched and merged into main independently, which is what caused the conflict here.

PR #543, opened 2026-07-22
tangos/recovered: matched functions (20260722-0135-7c04)

ruspecial:
Automated per-agent PR from tangOS Console (from fork `ruspecial/sm64ds-decomp`) - matched functions from **recovered** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.3.2** (each push re-stamps the commit message with the build that made it).


tangosdev:
Superseded by #546.

This PR also added `func_0203bc7c.c`, which #544 landed independently with different source, so it was permanently conflicted on an add/add and its other function could never merge. #546 carries `func_02030790` alone (unchanged, with you retained as the commit author) and has landed - so nothing here is lost. No action needed on your end.

PR #548, opened 2026-07-22
Match func_ov064_0211a4c4 byte-identical (mwccarm 1.2/sp2p3)

tangosdev:
Second half of the #533 tooling-bug fix. #547 landed the symbol boundary (the phantom epilogue merged back, size 0x21c); this is the source that boundary unhides.

It was previously banked as a 28-divergence near-miss purely because the configured target was truncated by 0xc. Against the corrected boundary it MATCHes 1.2/base, 1.2/sp2 and 1.2/sp2p3, and `linkcheck` returns VERIFIED with 0 blind slots.

Split from #547 deliberately: the build box sizes a changed src file from config, so a boundary fix and the source that depends on it cannot validate in the same PR.

### DIFF
--- a/src/func_ov064_0211a4c4.cpp
+++ b/src/func_ov064_0211a4c4.cpp
@@ -1,0 +1,103 @@
+//cpp
+typedef signed short s16;
+typedef unsigned short u16;
+typedef unsigned char u8;
+typedef unsigned int u32;
+
+struct Vector3 { int x, y, z; };
+
+extern "C" {
+u16 DecIfAbove0_Short(u16* p);
+void func_02012790(u32 a);
+void* _ZN5Actor10FindWithIDEj(u32 id);
+s16 Vec3_HorzAngle(const Vector3* v0, const Vector3* v1);
+int AngleDiff(int a, int b);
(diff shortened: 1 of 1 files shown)


tangosdev:
Closing: the build box still reports NO-REPRO against the corrected boundary, so my hypothesis (that a boundary fix and its source could not validate in the same PR) is disproven. The box's verdict stands and my local tooling disagrees with it - not merging a match on a red check until that discrepancy is explained.

The boundary fix in #547 is independently justified (the prologue `push {r4-fp,lr}` / `sub sp,#0xc` mirrors the phantom epilogue exactly) and stays landed.

tangosdev:
Closing again. Four hypotheses tested and all eliminated: stale PR config (the boundary was already on main), broken local ROM extraction (4 known-good .c matches in ov064 verify locally), a broken local C++ path (4 known-good .cpp matches in ov064 verify locally), and a stale symbol cache on the build box (re-run after the boundary landed still fails).

The box and my tooling agree on every other function in this overlay, in both languages. The disagreement is unique to the one symbol whose extent changed, which points at how pr_linkcheck derives a target for a merged extent - not something I can inspect from outside the box.

PR #575, opened 2026-07-22
Match func_ov060_021184bc byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Match `func_ov060_021184bc` (ov060 @ `0x021184bc`, size `0x88` / 136 bytes) byte-identical to EU retail.
- Compiler: **mwccarm 1.2/sp2p3** with `-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`.
- Local verify: `tools/match.py --c src/func_ov060_021184bc.c --func func_ov060_021184bc --addr 0x21184bc --size 0x88 --version 1.2/sp2p3 --module ov060 --strict-relocs` → **MATCH**.
- Claims API lock `clm_c2edcc619489` kept active (do not unclaim fully matched functions). CLAIMS.md row updated.

Behavior sketch: `ClearSpikeBomb` on `+0x1a8`, OR flag at `+0x13c`, set `+0x170 = 3`, zero eight slots at `+0x188`, then walk `Actor::FindWithActorID(0x11c, …)` collecting other actors’ unique IDs (up to 8).

## Test plan

- [x] `tools/match.py` strict-relocs MATCH on 1.2/sp2p3
- [x] CI `validate` green

Opened as draft
Model: Grok 4.5 (high)
Harness: Grok Build

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| ov060 func_ov060_021184bc (0x021184bc, size 0x88) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); claim clm_c2edcc619489 kept active |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #576, opened 2026-07-22
Match func_ov006_020f4a40 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match for `func_ov006_020f4a40` (ov006, `0x020f4a40`, size `0x94`)
- Replaces the previous `// NONMATCHING` draft in `src/`
- Compiler: **mwccarm 1.2/sp2p3** with standard flags (`-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`)

Verified locally:

```
python tools/match.py --c src/func_ov006_020f4a40.c --func func_ov006_020f4a40 --addr 0x20f4a40 --size 0x94 --version 1.2/sp2p3 --module ov006
```

Key shape notes for the match:
- First loop breaks on the first slot where `p[0x51bb] != 0` and `p[0x51bc] != 1` (counts via `cnt++` then exit)
- On success path: set all 12 slots' `p[0x51bc] = 2`, then write status fields at `c+0x5318` / `c+0x5314` and return `c+0x5000`

## Test plan

- [ ] CI `validate` green on this file

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| ov006 func_ov006_020f4a40 (0x020f4a40, size 0x94) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #576 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #577, opened 2026-07-22
Match func_02068398 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Match **func_02068398** (arm9 @ `0x02068398`, size `0x78`) byte-identical under mwccarm **1.2/sp2p3**.
- Hand-asm: plain C if-converts the trailing null-guard (`beq; mov r1,r0` → `movne`). Correct `ldmeqia sp!, {lr}` / `bxeq lr` spell the ROM early-return (prior near-misses used wrong asm encodings).
- Local verify: `match.py` MATCH; `linkcheck.py` **VERIFIED** (blind=0).

## Verification

```
python tools/match.py --c src/func_02068398.c --func func_02068398 --addr 0x2068398 --size 0x78 --version 1.2/sp2p3
python tools/linkcheck.py --name func_02068398 --addr 0x02068398 --size 0x78 --module arm9
```

## Test plan

- [x] `match.py` reports MATCHING VERSIONS: 1.2/sp2p3
- [x] `linkcheck.py` verdict VERIFIED, blind=0
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| arm9 func_02068398 (0x02068398, size 0x78) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #577 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #578, opened 2026-07-22
Match func_ov071_02120200 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match of `func_ov071_02120200` (ov071 @ `0x02120200`, size `0x78` / 120 bytes)
- Replaces the prior NONMATCHING draft (div=16)
- Compiler: **mwccarm 1.2/sp2p3** with flags `-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`
- Strict reloc verified: `CylinderClsn::Clear` at `+0x160` → `0x02015024` (arm9)

## Match notes

- u64-mask launder forces materialized `add r2,r4,#0xb0` for the flag BIC at `actor+0xb0`
- No named long-lived `short* sub`; halfword accesses use `c+0x3a6`/`c+0x3a8` so the `+0x300` base colors as `r1`
- Mid-block `z = 0` + named `ang` reload orders the double `ldrsh` and keeps zero in `r3` / angles in `ip`

## Test plan

- [x] `python tools/match.py --c src/func_ov071_02120200.c --func func_ov071_02120200 --addr 0x2120200 --size 0x78 --version 1.2/sp2p3 --module ov071` → **MATCHING VERSIONS: 1.2/sp2p3**
- [x] Strict reloc destination OK (Clear → 0x02015024)
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| ov071 func_ov071_02120200 (0x02120200, size 0x78) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #578 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #579, opened 2026-07-22
Match Stage::LoadModel (0x0202b5ec) byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Matched `_ZN5Stage9LoadModelEv` (`Stage::LoadModel`) at `0x0202b5ec` (size `0x90`) byte-identical under mwccarm **1.2/sp2p3**.
- Started from the near-miss draft (div=13); closed with u64-mask materialization of `self+0x874` and statement order that lands the ROM register coloring (`r2` list base, `r3` counter, `ip` node, `lr` count).
- Local verification:
  - `tools/match.py ... --version 1.2/sp2p3` → **MATCH**
  - `tools/linkcheck.py` → **VERIFIED** (blind=0)

## Test plan

- [x] `python tools/match.py --c src/_ZN5Stage9LoadModelEv.c --func _ZN5Stage9LoadModelEv --addr 0x202b5ec --size 0x90 --version 1.2/sp2p3`
- [x] `python tools/linkcheck.py --name _ZN5Stage9LoadModelEv --addr 0x202b5ec --size 0x90 --module arm9`
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| arm9 Stage::LoadModel / _ZN5Stage9LoadModelEv (0x0202b5ec, size 0x90) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #579 |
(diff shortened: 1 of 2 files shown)


tangosdev:
Closing as moot — no action needed on your side.

This PR edits `src/_ZN5Stage9LoadModelEv.c`, but that symbol already had a byte-identical match in `src/_ZN5Stage9LoadModelEv.cpp`. The `.c` was a leftover NONMATCHING draft that should have been removed when the match landed, so both files existed for one function.

That stale draft has now been deleted (along with 12 others in the same state), so the file this PR targets no longer exists. `Stage::LoadModel` remains matched and linkchecks VERIFIED via the `.cpp`.

Worth knowing: `linkcheck` resolves `src/<name>.c` before `.cpp`, so while both existed it was verifying the draft rather than the real match.

PR #580, opened 2026-07-22
Match func_0205fb58 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Match **func_0205fb58** (arm9 @ `0x0205fb58`, size `0x78`) byte-identical under mwccarm **1.2/sp2p3**.
- Near-miss was div=6 (pool-load interleaving). Fixed by modeling the 8-byte BSS slot as `struct Slot { u16 h; u16 pad; int w }` at `data_020a8138`, so the word store is `data_020a8138[a].w` (reloc `data_020a8138+4` == `data_020a813c`).
- Local verify: `match.py` MATCH; `linkcheck.py` VERIFIED (relocated bytes).

## Test plan

- [x] `python tools/match.py --c src/func_0205fb58.c --func func_0205fb58 --addr 0x205fb58 --size 0x78 --version 1.2/sp2p3`
- [x] `python tools/linkcheck.py --name func_0205fb58 --addr 0x0205fb58 --size 0x78 --module arm9` → VERIFIED
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| arm9 func_0205fb58 (0x0205fb58, size 0x78) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #580 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #582, opened 2026-07-22
Match func_ov006_02101088 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match of `func_ov006_02101088` (ov006 @ `0x02101088`, size `0xc0` / 192 bytes).
- Compiler: **mwccarm 1.2/sp2p3** with `-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`.
- Verified locally: `tools/match.py --c src/func_ov006_02101088.c --func func_ov006_02101088 --addr 0x2101088 --size 0xc0 --version 1.2/sp2p3 --module ov006` → **MATCH**.
- Leaf function (no calls); only pool constants `0x5264` / `0x526c`. Shape uses `#pragma opt_common_subs off` + inline indexed accesses (sibling pattern to other ov006 slot updaters).

## Test plan

- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #583, opened 2026-07-22
Match func_ov006_020e61c4 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
- Byte-identical match of `func_ov006_020e61c4` (ov006, `0x020e61c4`, size `0xc8`) under **mwccarm 1.2/sp2p3**.
- Replaces the previous `// NONMATCHING` draft in `src/` (div=14) with a verified match.

## Key matching lever
Early-path counter RMW at `thiz+0x55b6` needs the ROM's literal-pool load + `add r2,r4,r0` form. Achieved by:
1. Not naming a persistent `q = thiz+0x5500` base (inline `*(unsigned short*)(thiz+0x55B6)` loads color into `r1` as `add r1,r4,#0x5500` + `ldrh [r1,#0xb6]`).
2. u64-mask laundering the RMW address: `--*(volatile unsigned short*)(((int)thiz + 0x55B6) & 0xFFFFFFFFFFFFFFFF)`.

## Verify
```
python tools/match.py --c src/func_ov006_020e61c4.c --func func_ov006_020e61c4 \
  --addr 0x20e61c4 --size 0xc8 --version 1.2/sp2p3 --module ov006 --strict-relocs
# → MATCHING VERSIONS: 1.2/sp2p3
```

## Test plan
- [x] Local `match.py` with `--module ov006 --strict-relocs` reports MATCH
- [ ] CI `validate` green on this file

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| ov006 func_ov006_020e61c4 (0x020e61c4, size 0xc8) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #583 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #584, opened 2026-07-22
Match func_ov002_020bf36c byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-match `func_ov002_020bf36c` (ov002 @ `0x020bf36c`, size `0xa0`) under **mwccarm 1.2/sp2p3**.
- Replaces the prior `// NONMATCHING` div=2 draft in `src/`.
- Fix: `func_ov002_020bfa74` takes `self` (not void). Passing it keeps r0 live so the first `ldrb` colors into r2 like the ROM.

Verified locally:
```
python tools/match.py --c src/func_ov002_020bf36c.cpp --func func_ov002_020bf36c \
  --addr 0x20bf36c --size 0xa0 --version 1.2/sp2p3 --module ov002 --strict-relocs
```

## Test plan

- [x] Local `match.py` + strict-relocs green
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #585, opened 2026-07-22
Match EnemySwitchTag::Behavior byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Match `_ZN14EnemySwitchTag8BehaviorEv` (`EnemySwitchTag::Behavior`) at `0x020f19fc` size `0xc8` (ov002) byte-identical with mwccarm **1.2/sp2p3**.
- Replaces the prior `// NONMATCHING` (div=20) draft.
- Local verification: `match.py` → MATCH; `linkcheck.py` → **VERIFIED**.

### Notes
- `data_0209f318` is the Camera pointer passed as `this` to `Camera::IsUnderwater()`.
- Statement order of the two global loads (`cam` then level id) is required for register coloring.

## Test plan
- [x] `python tools/match.py --c src/_ZN14EnemySwitchTag8BehaviorEv.c --func _ZN14EnemySwitchTag8BehaviorEv --addr 0x20f19fc --size 0xc8 --version 1.2/sp2p3 --module ov002`
- [x] `python tools/linkcheck.py --name _ZN14EnemySwitchTag8BehaviorEv --addr 0x20f19fc --size 0xc8 --module ov002` → VERIFIED
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
+| ov002 EnemySwitchTag::Behavior / _ZN14EnemySwitchTag8BehaviorEv (0x020f19fc, size 0xc8) | lunavyqo (Grok) | 2026-07-22 | **done** — verified byte-identical (mwccarm 1.2/sp2p3); PR #585 |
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #586, opened 2026-07-22
Match func_ov006_020de1d4 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- **func_ov006_020de1d4** (ov006 @ `0x020de1d4`, size `0x98`) — byte-identical match under mwccarm **1.2/sp2p3** with flags `-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`.

Replaces the prior `// NONMATCHING` draft (div=16). Logic: if timer at `c+0x51cc` is non-zero, pool-materialized RMW decrement + signed clamp; else optional `func_ov006_020dc370` path, then always write timer=`0xc0` / state=`5`.

Verified locally:
```
python tools/match.py --c src/func_ov006_020de1d4.c --func func_ov006_020de1d4 \
  --addr 0x20de1d4 --size 0x98 --version 1.2/sp2p3 --module ov006 --strict-relocs
```

## Test plan

- [x] `--strict-relocs` clean
- [ ] CI `validate` green

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #587, opened 2026-07-22
Match func_ov007_020aec00 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Match `func_ov007_020aec00` (ov007, `0x020aec00`, size `0x94`) byte-identical under mwccarm **1.2/sp2p3** with the standard flags (`-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc`).
- Started from the banked near-miss (div=7, pure regperm: `off`/`p28` swapped between r0 and r3). Declaring `int *p28` before `int off` forces the allocator to color `p28→r3` and `off→r0`, matching the ROM setup sequence.

## Verification

```
python tools/match.py --c src/func_ov007_020aec00.c --func func_ov007_020aec00 --addr 0x20aec00 --size 0x94 --version 1.2/sp2p3 --module ov007
```

## Contents

- `src/func_ov007_020aec00.c` only

### DIFF
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -19,6 +19,7 @@ it is fair to take over: ping the claimant first.
 
 | Range | Who | Claimed | Status |
 |---|---|---|---|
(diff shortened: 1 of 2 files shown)


tangosdev:

PR #588, opened 2026-07-22
NONMATCHING: improve func_ov006_02106fdc to div=9 (r8/sb regperm floor)

lunavyqo:
## Summary

- Improves existing `// NONMATCHING` `func_ov006_02106fdc` (ov006, `0x02106fdc`, size `0xc0`) from **div=28 → div=9**.
- Compiler: **mwccarm 1.2/sp2p3** with standard flags; C++ (`//cpp`) PMF dispatch via `data_ov006_02142840`.
- Residual is a **pure r8/sb register-coloring swap**: instruction stream and size match the ROM except `this` homes to `sb` and `this+0x4000` to `r8` where the ROM has `r8`/`sb`.

## What was tried

- Near-miss DB draft as starting point (already structure-correct).
- 120 declaration-order permutations of locals; statement-order perms of post-call init.
- Section **6q** variable-identity web swap; u64-mask laundering; `#pragma opt_{strength_reduction,propagation,common_subs,scheduling,peephole}`.
- Sibling PMF-loop access shapes (`func_ov006_02106eb8` / `02106f44` / `0210709c`); C vs C++ PMF forms.
- Same residual on 1.2 `{base,sp2,sp2p3}`.

Floor class: **regperm** (callee-saved this/base). Good permuter / refine-tier input if a new r8/sb lever appears.

## Validate note

This is a NONMATCHING advance (fails `validate` by design, same as the previous div=28 draft on main). Merge with admin override if desired per MERGE.md.


tangosdev:
Closing this one, but nothing is lost and the claim is preserved on main.

The div=9 draft here is already banked in `nearmiss/db.jsonl` on main, byte-identical to this PR's source modulo comments. I verified that rather than assuming it. So the actual matching work, the r8/sb coloring analysis, is already in the shared DB where `refine_wl.py` and the permuter tier read from.

The reason this can't merge as-is: the link gate compiles every changed `src/*.c|*.cpp` and requires it to reproduce the ROM byte-for-byte. A near-miss filed into `src/` fails by construction, which is what the red check is here. That's the gate doing its job, not a flake. Main does host NONMATCHING drafts, but improving one in place will always trip the gate, so `nearmiss/db.jsonl` is the right destination for near-miss work.

Your CLAIMS.md row is landed on main in c806741d with your authorship preserved, retargeted at the DB instead of this PR. The ov095 row is also updated to mark `func_ov095_02135cdc` matched via #596, which is merged.

Nice analysis on this one, by the way. Pure r8/sb register-coloring swap with the instruction stream and size otherwise identical is squarely the regperm floor class. It's not marked `floor` in the DB yet, so it stays live for the permuter tier.

PR #637, opened 2026-07-23
Replace Group B HAND-ASM with pure-C NONMATCHING (+1 real match)

lunavyqo:
## Summary

Replaces **junk HAND-ASM** landings for nine lunavyqo-matched functions with **pure C**:

| Symbol | Status | Notes |
|--------|--------|--------|
| `func_ov002_020bf13c` | **MATCH** | Pure C; y-early `Vec3` preload; linkcheck VERIFIED |
| `func_02068398` | NONMATCHING div=1 | Size-exact; 6u cond-opt residual (`beq` vs `moveq`) |
| `func_ov007_020c6550` | NONMATCHING div=2 | 6v load-order / color weld |
| `func_ov009_0211145c` | NONMATCHING div=3 | Call-after-stores (`s` in r3 vs r0) |
| `func_ov095_02135cdc` | NONMATCHING div=35 | SMULL schedule floor |
| `func_0204322c` / `88` / `e4` / `4335c` | NONMATCHING size walls | PTMF below-sp staging under 1.2/sp2p3 |

Compiler: **mwccarm 1.2/sp2p3**. No HAND-ASM bodies remain for these symbols. Best tips also banked in `nearmiss/db.jsonl` with floor marks where appropriate.

## Expect validate

`validate` will mark the eight NONMATCHING files as near-miss (expected). The real match is `func_ov002_020bf13c`. Maintainer can land the match alone and drop NONMATCHING, or keep tips in `src/` for refine.

## Test plan

- [x] `tools/match.py` MATCH on `func_ov002_020bf13c` (ov002 @ 0x020bf13c, 0xe8)
- [x] `tools/linkcheck.py` VERIFIED on that match
- [x] NONMATCHING drafts compile under 1.2/sp2p3 and score as noted
- [ ] CI `validate` table review


tangosdev:
Took the middle path and landed it as 99f568ea, crediting you.

func_ov002_020bf13c — the one that genuinely cracks in pure C — now has your readable C body instead of the HAND-ASM, verified byte-identical on all three 1.2 variants. Net match count unchanged, source quality up.

The other 8 keep their HAND-ASM matches rather than reverting to non-matching C (that would have dropped 8 off the count). Your pure-C tips for all 8 are banked in nearmiss/db.jsonl with your floor analysis intact — func_02068398 div=1, 020c6550 div=2, 0211145c div=3 are close enough that they'll likely finish in C soon, and the four PTMF size-wall findings are preserved. So nothing you analyzed is lost, and whoever cracks them in C later can drop the HAND-ASM then.

Closing since the substance landed in a different shape than the PR; open a fresh one if you disagree on any specific function.

PR #643, opened 2026-07-23

lunavyqo:
## Summary

Byte-identical matches (mwccarm **1.2/sp2p3**), linkcheck VERIFIED:

| Function | Addr | Size | How |
|---|---|---|---|
| `func_ov002_020b91fc` | 0x020b91fc | 0xbc | Vector3 via arrcast; load order `z`, `x`, then `y = *(c+0x60) + 0x82000` |
| `func_ov063_02119894` | 0x02119894 | 0xcc | Sequential CF; `func_ov063_02119cc0` only when `0x5c0==0`; **`func_ov063_02119b1c` always runs** |

Also includes attempt tree + provenance rows and improved near-miss tips banked for the rest of the batch (`e7218` div=15, `b38ac` div=13).

## Test plan

- [x] `python tools/match.py ... --version 1.2/sp2p3` → MATCH for both
- [x] `python tools/linkcheck.py --name ...` → VERIFIED / blind=0
- [ ] CI `validate` green


tangosdev:
Both matches landed on main as e83ce905 with your authorship. This branch came from your fork so I couldn't rebase it in place; it also overlapped work already merged (via #642/#648), so I took just the two net-new matches — func_ov002_020b91fc and func_ov063_02119894 — plus their own provenance/attempt/CLAIMS rows, and left the stale duplicates behind. Both re-verified byte-identical on all three 1.2 variants before landing.

PR #645, opened 2026-07-23

lunavyqo:
## Summary

Matched two functions byte-identical under mwccarm 1.2/sp2p3:

1. **`func_ov006_020def80`** (ov006 @ `0x020def80`, size `0xa4`)
   - Shared epilogue always clears `0x5440`/`0x544c`
   - Permuter: `new_var = i` for epilogue base materialization
   - match.py + linkcheck VERIFIED

2. **`func_ov002_020e7f2c`** (ov002 @ `0x020e7f2c`, size `0xa0`)
   - Electrocute-style volatile particle coords
   - u64-mask on countdown at `0x494`
   - `*(volatile u32*)(c+0x4bc)` first arg for handle load order
   - Inline `*(u16*)((char*)c + 0x400 + 0x94)` colors base to r1
   - match.py + linkcheck VERIFIED

## Test plan

- [x] match.py + linkcheck on both files


tangosdev:
Match landed on main as 5a81770c with your authorship. Fork branch, so landed directly rather than rebased. Took func_ov006_020def80 plus its one legitimate new near-miss tip; the branch also carried 10 stale DB entries for functions since matched on main via #646, which I left out so they weren't resurrected as ghosts. Match re-verified byte-identical.

tangosdev:
Correction to my earlier note: my first landing push for this match was silently rejected (a scripting bug masked the exit code), so it did **not** land when I said it did. The match was then recovered cleanly via #649 and is now matched on main, and I re-banked your func_ov006_02126ee4 tip improvement (393 → 46) which had been lost with that failed push. Net result is correct now — func_ov006_020def80 matched, tip improved — just via a different path than my premature comment claimed.

PR #647, opened 2026-07-23
Near-miss: __sinit_ov075_0211bb00 tip div=10 (was 640)

lunavyqo:
## Summary

Session work on **`__sinit_ov075_0211bb00`** (`ov075` @ `0x0211bb00`, size `0xa1c`, mwccarm **1.2/sp2p3**).

**No byte-identical match** — this PR only banks an improved near-miss tip (and attempt history). Per repo policy, non-reproducing C stays out of `src/`.

| | Score |
|--|--|
| Prior DB tip | div=640 (re-eval ~327/647) |
| **This tip** | **div=10/647** |

### How the tip improved
- `#pragma opt_common_subs off` + matched **`__sinit_ov004_020b9b24`** `short *p` registration shapes (first dual-imm, early r3-interleave, late r1 store-first, temp `{int sz,h}` for `mov r2` blocks)
- Tail: `struct P6` + `tmp`/`ta`/`tb` with `tb ? ta : ta` bqa idiom

### Residual floor
Pure **r2↔r3 regperm** on the final adjacent pool pair (`c838` / `d828`): correct load order keeps wrong regs; srecycle-style forms flip regs but reverse load order and break the matching prefix. Exhausted ~200+ C variants + permuter (score floor 50).

### Contents
- `nearmiss/db.jsonl` — tip C for `__sinit_ov075_0211bb00` at div=10  
- `config/match_attempts.jsonl` — attempt tree (near_miss improve + two no_progress leaves)  
- `CLAIMS.md` — released row with floor note  

## Test plan
- [x] `python tools/fdiff.py --c <tip> --name __sinit_ov075_0211bb00 --module ov075 --addr 0x211bb00 --size 0xa1c --quiet` → `mismatches=10/647`
- [ ] CI / near-miss ingest path accepts the tip row


tangosdev:
Landed on main as dace7caa with your authorship preserved. Rebased it onto current main to clear the ledger/CLAIMS conflicts — the div=640→10 tip is banked, and I kept main's CLAIMS rows where they were ahead of this branch (5 would otherwise have regressed to older states, since matches landed on main after you cut this). The GitHub PR head got stuck on a stale SHA after the rebase, so I landed the content directly rather than leave it blocked.

PR #650, opened 2026-07-23
Near-miss: improve func_ov006_020d27dc tip to div=184

lunavyqo:
## Summary

- Bank a refined near-miss for `func_ov006_020d27dc` (ov006 @ `0x020d27dc`, size `0xe48`).
- DB tip improved **1054 → 184** (mwccarm 1.2/sp2p3). Not byte-identical yet.
- Also appends `config/match_attempts.jsonl` and a released `CLAIMS.md` row.

### What improved in the tip

- First loop is **byte-exact** (pool + `lsl #2`, inc-before-store)
- Dir RMW sites use preindex form (`pre=7`)
- dualB counter launders on `0x5374` / `0x53e8`
- `#pragma opt_strength_reduction off` + `opt_common_subs off`

### Residual floors

- Double early-return epilogue after the first loop
- Dist access still `#0x4000+#0x664` vs ROM `r7` + pool
- Counter-clamp register coloring

No `src/` file (near-miss belongs in the DB only).

## Test plan

- [x] `nearmiss_db.evaluate` reports div=184, not ok
- [x] Tip compiles under mwccarm 1.2/sp2p3 (`//cpp`)
- [ ] CI / maintainers: union-merge of `nearmiss/db.jsonl` + `match_attempts.jsonl`


tangosdev:
Landed on main as 8a574ee1 with your authorship. Heads up: this PR carried two DB rows for func_ov006_020d27dc — one with the address as a hex string ("0x020d27dc"), one as an integer (34416604) — which would have left a duplicate entry. I collapsed them to a single row in main's hex-address format at div=184. Whatever wrote the integer-addr row is worth checking on your side so future tips don't double up. The 1054 -> 184 improvement is banked.

PR #655, opened 2026-07-23
Match 2 functions byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
Byte-identical matches for two batch5 targets under **mwccarm 1.2/sp2p3**:

| Function | Module | Addr | Size | linkcheck |
|---|---|---|---|---|
| `func_ov071_0211f7d4` | ov071 | `0x0211f7d4` | `0xfc` | BLIND-2 (no WRONG-DEST) |
| `func_ov006_020e2dbc` | ov006 | `0x020e2dbc` | `0xfc` | **VERIFIED** |

### How
- **ov071**: near-miss tip → `WithMeshClsn *wm` for a8 zero / `ClearLimMovFlag` interleave; u64-launder `0xb0` flag RMW; shared zero short for ang triple-store; 2-arg `func_ov071_021202ec`.
- **ov006**: near-miss tip → u64-launder timer RMW (`0x4ee0`) + counter writeback (`0x4ee6`); `int idx` for signed `popge` (not unsigned `pophs`).

Author credit: **lunavyqo**. Attempt trees + provenance stamped; matched tips pruned from `nearmiss/db.jsonl`.

## Test plan
- [x] `python tools/match.py ... --module ov071 --version 1.2/sp2p3` → MATCH
- [x] `python tools/match.py ... --module ov006 --version 1.2/sp2p3` → MATCH
- [x] linkcheck: ov071 BLIND-2, ov006 VERIFIED
- [ ] CI `validate` green

PR #658, opened 2026-07-23
Match func_ov007_020c25fc byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- **Matched** `func_ov007_020c25fc` @ ov007 `0x020c25fc` (size `0x100`) with mwccarm **1.2/sp2p3**
  - Lever: u64-mask launder on `&self->f8` RMW (`((int)self + 8) & 0xFFFFFFFFFFFFFFFFLL`), then field reads for later uses
  - Local verify: MATCH; `linkcheck` VERIFIED (blind=0)
- **Near-miss tip improved** for `func_ov007_020c55bc` @ `0x020c55bc`: div **9 → 6**
  - Residual pure r5/r6 regperm (`m1` sign-high vs loop offset `k`); tip banked in `nearmiss/db.jsonl` (not `src/`)

Author: **lunavyqo**

## Test plan

- [x] `python tools/match.py --c src/func_ov007_020c25fc.c --func func_ov007_020c25fc --addr 0x20c25fc --size 0x100 --version 1.2/sp2p3 --module ov007`
- [x] `python tools/linkcheck.py --module ov007 --name func_ov007_020c25fc --addr 0x20c25fc --size 0x100` → VERIFIED
- [ ] CI `validate` green


tangosdev:
Match landed on main as 1138b124 with your authorship. Took just the func_ov007_020c25fc match — this branch was cut before #655/656/657 landed, so its DB also tried to re-add near-miss entries for the five functions those matched (now ghosts) and to revert func_ov002_020e7454 from 21 back to 35. Skipped both so nothing regressed. Match re-verified byte-identical.

PR #662, opened 2026-07-24
Match 10 functions byte-identical (mwccarm 1.2/sp2p3)

andrewboudreau:
Matches ten functions across arm9 and overlays using source-only C/C++.

Local verification: all ten pass `tools/linkcheck.py` with `VERIFIED`, empty diffs, and `blind: 0`.


andrewboudreau:
Closing as redundant: these exact 10 matches already merged in #646 (2026-07-23). Verified origin/main contains all 10 with correct extensions and no NONMATCHING headers; this branch's content is byte-identical to main. The displayed delete/add churn is only because the merge-base predates #646.

PR #663, opened 2026-07-24
Match func_ov002_020aefb8 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- **Matched:** `func_ov002_020aefb8` @ ov002 `0x020aefb8` (size `0x108`) — verified with `tools/match.py` against mwccarm **1.2/sp2p3**.
- **How:** u64-mask laundering to materialize `self+0xa4` and `self+0xac` for the ground-speed MLA RMW (near-miss tip was div=24 without the `add` bases).
- **Near-miss banked:** `func_ov075_02117d80` improved **div 5→4** (`enum Bool` for `cmp r0,#0`); remaining divergence is load/flag interleave / regperm floor. Attempt tree + tip C logged.

## Verify

```
python tools/match.py --c src/func_ov002_020aefb8.c --func func_ov002_020aefb8 --addr 0x20aefb8 --size 0x108 --version 1.2/sp2p3 --module ov002
```

## Claims

- ov002 `0x020aefb8`: **done** (matched)
- ov075 `0x02117d80`: **released** (tip in `nearmiss/db.jsonl`)

Author credit: `lunavyqo`

PR #666, opened 2026-07-24
Match func_ov063_021162c8 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- **Matched** `func_ov063_021162c8` @ `ov063:0x021162c8` (size `0x108`) byte-identical with mwccarm **1.2/sp2p3**.
- Key fix: long-long u64-launder on the `*(u16*)(self+0x5d4) &= ~2` site so mwccarm emits the ROM's pool materialization (`ldr` 0x5d4 + `add` + `ldrh`/`bic`/`strh`) instead of `add #0x500` + `ldrh #0xd4`.
- Also banked an improved near-miss for `func_ov002_020d8944` (div **15→10**): correct `p,sum,asr` stack layout with Vector3-by-value `NewSimple`; residual is LDM vs individual loads for the particle call setup.

## Verify

```
python tools/match.py --c src/func_ov063_021162c8.cpp --func func_ov063_021162c8 --addr 0x21162c8 --size 0x108 --version 1.2/sp2p3 --module ov063
```

Reports **MATCH**.

## Note

Claims API key from the prompt was expired/revoked; coordinated via CLAIMS.md only.


tangosdev:
Match landed on main as c31b2713 with your authorship, combined with #663-666. All re-verified byte-identical; the now-matched functions' near-miss entries were pruned and two tip improvements (func_ov075_02117d80 5→4, func_ov002_020d8944 15→10) banked.

PR #669, opened 2026-07-24
Match Player::St_Swim_Main (_ZN6Player12St_Swim_MainEv) byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Matched `_ZN6Player12St_Swim_MainEv` at ov002 `0x020cd94c` (size `0x9d8`) byte-identical under **mwccarm 1.2/sp2p3**.
- `tools/match.py` reports MATCH; `tools/linkcheck.py` reports **VERIFIED** (blind=0).
- Started from near-miss tip (div=258): u64-launder for 0x6e5 / 0xb0 / 0xa8 RMW, case-6 `cedb0` before `a4` load, particle pos register schedule + `*(volatile u32*)` handle for `func_02022d44`.

## Provenance

- author: `lunavyqo`
- how: ai / grok-4.5 / high / grok-build
- claim API: `clm_f1040cf22000` kept (done)

## Test plan

- [x] `python tools/match.py --c src/_ZN6Player12St_Swim_MainEv.cpp --func _ZN6Player12St_Swim_MainEv --addr 0x20cd94c --size 0x9d8 --version 1.2/sp2p3 --module ov002`
- [x] `python tools/linkcheck.py --module ov002 --name _ZN6Player12St_Swim_MainEv --addr 0x20cd94c --size 0x9d8`
- [ ] CI `validate` green

PR #670, opened 2026-07-24
Match func_0201d850 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
- Matched `func_0201d850` (arm9 @ `0x0201d850`, size `0x7f4`) byte-identical under mwccarm **1.2/sp2p3**.
- `tools/match.py` reports MATCH; `tools/linkcheck.py` verdict **VERIFIED** (blind=0).

### Match levers
- **bhs shape** for `var_r4 = 0x100 / 0x140`: goto-fallthrough so the compiler emits `bhs` instead of unconditional `b`.
- **Last-block r7/ip**: set `data_0209d6b8 = 0` *before* the last-block ep/base/f4 sequence (first-block clone), so ep colors into `r7` and zero does not steal the callee-saved.

## Test plan
- [x] `python tools/match.py --c src/func_0201d850.c --func func_0201d850 --addr 0x201d850 --size 0x7f4 --version 1.2/sp2p3` → MATCH
- [x] `python tools/linkcheck.py --name func_0201d850 --module arm9 --addr 0x0201d850 --size 0x7f4` → VERIFIED
- [ ] CI `validate` green


tangosdev:

PR #672, opened 2026-07-24
Near-miss improve func_ov063_02117cdc (div 119→11)

lunavyqo:
## Summary

- Banked an improved near-miss tip for `func_ov063_02117cdc` (`ov063:0x02117cdc`, size `0x77c`): **div 119 → 11** (size-exact; was size-short).
- Appended focused attempt-tree rows in `config/match_attempts.jsonl` (near_miss improve + no_progress floor).
- Tip lives in `nearmiss/db.jsonl` only — **not** `src/` (non-matching drafts stay out of `src/`).

### Residual floor (div=11)

1. **Prologue (5 words):** switch temp `r1` vs ROM `r2` (camera pool load `r0` vs `r1`) — pure regperm.
2. **Case 9 (6 words):** double-bind between y-before-x load order (anti address-order) and ands-before-store schedule; every C spelling that forces one breaks the other.

Levers used on the way down: u64-mask materialization (cases 3/5/7), `mul -1` for negate, inverted case-8 saturate (`bge`+RMW), case-5 `d560` reload via base.

## Test plan

- [x] `python tools/fdiff.py --c <tip.c> --name func_ov063_02117cdc --module ov063 --addr 0x2117cdc --size 0x77c --quiet` → `mismatches=11/479`
- [x] Single DB row for this symbol after dedupe (no int/string addr duplicate)
- [ ] CI / union-merge of append-only `config/match_attempts.jsonl` + `nearmiss/db.jsonl`


tangosdev:

PR #673, opened 2026-07-24
Near-miss improve func_0203e20c (div 114→30)

lunavyqo:
## Summary

- Bank improved near-miss for **arm9 `func_0203e20c`** (`0x0203e20c`, size `0x76c`): word-diff **114/86 → 30** (mwccarm 1.2/sp2p3).
- Append attempt-tree rows in `config/match_attempts.jsonl` with AI provenance (`grok-4.5` / `high` / `grok-build`, author `lunavyqo`).
- Release CLAIMS row for this span (no match; tip belongs in the DB, not `src/`).

### What the tip improves

- Early `volatile u16 *pf1c` + declaration order for the epi blit
- Fake deps on `data_020a11c4` / `11a0` / `117c` bases (`+ (t32 & 0)`) so pools are not hoisted into the chain
- `p117c[3]` / `h10` store interleave for the `0x7c` chain
- Early loads of `p1040` halfwords `h4..ha` before the volatile chain (stores still after f1c RMW)

### Residual

All remaining mismatches sit in **`+0x588`–`+0x61c`**: TARGET schedules early `&data_020a0f1c`, mid-chain `f28`, batched halfword materialization into high regs, then f1c mask; pure-C schedule that hits that window either grows size or stalls at ~30. Hand refine + decomp-permuter exhausted for this session.

## Test plan

- [x] `python tools/match.py --c tip.c --func func_0203e20c --addr 0x203e20c --size 0x76c --version 1.2/sp2p3` → **30 word mismatches** (size-exact `0x76c`)
- [x] Tip ingested via `nearmiss_db` / `log_attempt --src` (best tip div=30)
- [x] No `src/func_0203e20c.*` (near-miss only)
- [ ] CI / maintainers: optional re-verify of `nearmiss/db.jsonl` row for `func_0203e20c`


tangosdev:

PR #674, opened 2026-07-24
Near-miss improve Boo::Behavior (div 184/61→54)

lunavyqo:
## Summary

- Banked an improved near-miss tip for `_ZN3Boo8BehaviorEv` / `Boo::Behavior` (`ov063:0x0211b0a4`, size `0x7e4`): **div 184 → 54** (size-exact; intermediate tip had been 61).
- Appended focused attempt-tree rows in `config/match_attempts.jsonl` with AI provenance on every node:
  - **model** `grok-4.5`
  - **reasoning** `high`
  - **harness** `grok-build`
  - **author** `lunavyqo`
- Tip lives in `nearmiss/db.jsonl` only — **not** `src/` (non-matching drafts stay out of `src/`).
- No `match_provenance.jsonl` row: HOW stamp (`stamp_provenance`) is **match-only**; attempt-tree `model`/`reasoning`/`harness` is the near-miss provenance.

### Attempt tree

```
ov063:0x0211b0a4
├─ no_progress div=999  parent=null   base=near_miss_draft(184)  size-miss 0x7ec
├─ near_miss   div=69   parent=↑      switch layout + 0x19c u64 + eqint d1 + norel
├─ near_miss   div=61   parent=↑      permuter polish
├─ near_miss   div=55   parent=↑      inv_else player path (eq-pred + beq)
└─ near_miss   div=54   parent=↑      fdep Vector3 copy (best tip)
```

### Levers that landed the tip

- Switch case sharing/order (0–2, 6+10, 12–14, 15/8/9/11 fallthrough)
- `0x19c` u64-launder materialized `add r2`
- Actor-type bool via `(int)(u16==0xd1)` → `moveq`/`movne`
- **`inv_else`**: `if (plr != 0) { … } else { null-case }` forces ROM eq-predicated then + `beq`
- Fake data-dep `pv.x = b ? a : a` on player pos copy

### Residual floor (div=54, size-exact)

Pure regperm / store-order (same mnemonics):

1. Kill path `+0x500` base **r1** vs **r3** (~10 words)
2. Area hide **ip**/`r3` base swap + data temps (~23)
3. Player reload **r0** vs **r1**; else Vector3 copy interleave
4. Counter `+0x100` **r2** vs **r1**
5. Raycast x/y+ **r3**/`ip` swap (×2)

## Test plan

- [x] `python tools/match.py --c <tip.c> --func _ZN3Boo8BehaviorEv --addr 0x211b0a4 --size 0x7e4 --version 1.2/sp2p3 --module ov063` → **54** word divergences, size `0x7e4`
- [x] Single `nearmiss/db.jsonl` row for this symbol after update
- [x] Attempt rows include AI provenance fields; no non-matching `src/`
- [ ] CI / union-merge of append-only `config/match_attempts.jsonl` + `nearmiss/db.jsonl`


tangosdev:
Landed on main as e14d7880 with your authorship, combined across #674/#675 since both carried the same Boo::Behavior 61→54 improvement and would have conflicted landing separately. func_ov007_020bb4b0 re-verified byte-identical (ov007, 0x758) and its near-miss entry pruned.

PR #675, opened 2026-07-24
Match func_ov007_020bb4b0 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match of `func_ov007_020bb4b0` (ov007 @ `0x020bb4b0`, size `0x758`) under mwccarm **1.2/sp2p3**.
- Started from near-miss tip (div=24); key lever: dual ternary for direction and max sharing the same `(c1 > c2)` comparison, plus `pressed()` init with `n2 = 0` before the mask ternary.

## Verify

```
python tools/match.py --c src/func_ov007_020bb4b0.c --func func_ov007_020bb4b0 --addr 0x20bb4b0 --size 0x758 --version 1.2/sp2p3 --module ov007
python tools/linkcheck.py --name func_ov007_020bb4b0 --addr 0x020bb4b0 --size 0x758 --module ov007
# verdict BENIGN (interwork-only), blind=0
```

## Notes

- Credit: `lunavyqo`
- Provenance: ai / grok-4.5 / high / grok-build


tangosdev:
Landed on main as e14d7880 with your authorship, combined across #674/#675 since both carried the same Boo::Behavior 61→54 improvement and would have conflicted landing separately. func_ov007_020bb4b0 re-verified byte-identical (ov007, 0x758) and its near-miss entry pruned.

PR #679, opened 2026-07-24
Match func_ov014_0211250c byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match of `func_ov014_0211250c` (ov014 @ `0x0211250c`, size `0x27c`) under **mwccarm 1.2/sp2p3**.
- ChainChomp segment-link update: rotation, atan2/horzlen, clamp, scale mul over 6 chain segments.
- Setup statement order is load-bearing (`seg` before `state`) so the prologue `add`/`cmp` schedule matches the ROM; body refined from near-miss + permuter `PERM_LINESWAP`.
- `linkcheck` **VERIFIED** (no WRONG-DEST).

## Test plan

- [x] `python tools/match.py --c src/func_ov014_0211250c.c --func func_ov014_0211250c --addr 0x211250c --size 0x27c --version 1.2/sp2p3 --module ov014` → MATCH
- [x] `python tools/linkcheck.py --name func_ov014_0211250c --module ov014 --addr 0x211250c --size 0x27c` → VERIFIED
- [ ] CI `validate` green


tangosdev:
Match landed on main as f181afe6 with your authorship. Pruned the func_ov014_0211250c near-miss entry your PR carried at div=9, since the function is now matched (it would otherwise be a stale ghost in the DB). Re-verified byte-identical.

PR #680, opened 2026-07-24
Match func_ov060_02111a28 (+ related) byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- **Matched** `func_ov060_02111a28` (ov060 @ `0x02111a28`, size `0x240`) byte-identical under **mwccarm 1.2/sp2p3**.
- From near-miss div=44 → div=10 (first-branch rewrite) → **MATCH**.
- Key levers:
  - `else if (n == 0x29 || n < 2) r1 = 1` (avoids explicit `r1 = 0` / `movge`)
  - `(unsigned short)((unsigned)*(int *)(c + 0x12c) >> 12)` forces ROM prologue load order (`ldr` frame then `ldrb` flag) + `lsl#4`/`lsr#16` extract
- **linkcheck VERIFIED** (blind=0).
- Removed `// NONMATCHING` `_ZN7Wiggler8BehaviorEv` from `src/` (tip remains banked in `nearmiss/db.jsonl` at div=132).
- Branch also carries other lunavyqo ov060 matches landed on the same open PR (reuse rule).

Author: **lunavyqo**. Compiler: **mwccarm 1.2/sp2p3**.

## Verify

```
python tools/match.py --c src/func_ov060_02111a28.cpp --func func_ov060_02111a28 \
  --addr 0x2111a28 --size 0x240 --version 1.2/sp2p3 --module ov060
# MATCHING VERSIONS: 1.2/sp2p3

python tools/linkcheck.py --name func_ov060_02111a28 --addr 0x02111a28 --size 0x240 --module ov060
# verdict: VERIFIED
```

PR #681, opened 2026-07-24
Match func_ov060_02112434 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Byte-identical match of `func_ov060_02112434` (ov060 @ `0x02112434`, size `0x1bc`) with mwccarm **1.2/sp2p3**.
- Started from near-miss tip (div=51): fixed stack frame via **inline PMF** table dispatch, and forced pool-materialized flag/byte RMW at offsets `0x418` / `0x41c` with **varied u64-mask launder spellings** so each conditional rematerializes (no CSE short-circuit).
- Local verify: `match.py` MATCH; `linkcheck` **BLIND-2** (unresolved reloc names only, no WRONG-DEST).
- Provenance stamped (`grok-4.5` / high / `grok-build`, author `lunavyqo`); attempt logged; nearmiss tip pruned.

## Test plan

- [x] `python tools/match.py --c src/func_ov060_02112434.cpp --func func_ov060_02112434 --addr 0x2112434 --size 0x1bc --version 1.2/sp2p3 --bin extracted/overlays/overlay_0060.bin --base 0x02111900 --module ov060`
- [x] `python tools/linkcheck.py --name func_ov060_02112434 --addr 0x02112434 --size 0x1bc --module ov060`
- [ ] CI `validate` green


tangosdev:
Match landed on main as 54833072 with your authorship, combined across #680/#681. All five ov060 functions re-verified byte-identical; near-miss entries pruned. #680 also still carried a stale func_ov014_0211250c ghost from its earlier Wiggler version (that function matched via #679) - dropped it.

PR #682, opened 2026-07-24
Match func_ov060_02114858 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
- Byte-identical match for `func_ov060_02114858` (ov060 @ `0x02114858`, size `0x308`) with **mwccarm 1.2/sp2p3**.
- Started from near-miss tip (div=107): fixed control-flow (timer++ only inside outer if), `volatile` load order for `0x450`, `s32 pos[3]` + u64-launder field RMW → div=9 regperm floor on the sin-table/`smlabb` block.
- **decomp-permuter** finished the regperm floor (score 65→0).
- `match.py` MATCH + `linkcheck` **VERIFIED**. Provenance + attempt tree logged; nearmiss tip pruned.

## Test plan
- [x] `python tools/match.py --c src/func_ov060_02114858.c --func func_ov060_02114858 --addr 0x2114858 --size 0x308 --version 1.2/sp2p3 --module ov060`
- [x] `python tools/linkcheck.py --module ov060 --name func_ov060_02114858 --addr 0x02114858 --size 0x308` → VERIFIED
- [ ] CI `validate` green

PR #683, opened 2026-07-24
Bank near-miss improve for func_ov060_02115518 (div 118→8)

lunavyqo:
## Summary
- Improved near-miss tip for `func_ov060_02115518` (ov060 @ 0x02115518, size 0x1d4) from **div=118 → div=8**.
- Size-matched; residual pure **r0/r2** register coloring on `data_02082214` base vs `z` (tbl/z swap) after `scale ? y : y` fake-dep got the rest of case-1 coloring correct (ip/lr/r5–r8/push set).
- No `src/` file (still unmatched). Attempt logged; tip in `nearmiss/db.jsonl`.

## Levers used
- Near-miss draft as base
- `player` local (reuse across StartTalk/GetTalkState)
- u64-mask LAUND for `0x424` increments (pool materialization)
- Fieldwise pos loads + `pos[1] = scale ? y : y` fake data-dependency for callee-saved coloring
- Permuter (no further gain past score floor)

## Test plan
- [x] `python tools/match.py --c tip.c --func func_ov060_02115518 --addr 0x2115518 --size 0x1d4 --version 1.2/sp2p3 --module ov060` → ndiff=8


tangosdev:
Landed on main as 89678185 with your authorship, combined across #682/#683: the func_ov060_02114858 match plus the func_ov060_02115518 118→8 tip. Match re-verified byte-identical.

PR #684, opened 2026-07-24
Near-miss func_ov060_02112724 (div=22, mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Banked improved near-miss tip for `func_ov060_02112724` (ov060 @ `0x02112724`, size `0x19c`): **div 90 → 22**.
- Structural fix: position/table body is **shared after the switch** (not only inside case 1).
- Codegen shape: `py` before `px` for `ip`/`lr`, Vector3 copy via `*(V*)px`, `py[1]` for z.
- Residual floor: `ldm`/`stm` vs ROM interleaved field copy (candidate size `0x188` vs ROM `0x19c`); idx in `r3` vs `r5`.
- Attempt tree logged in `config/match_attempts.jsonl`; CLAIMS.md released for this span.

## Test plan

- [x] Tip scored with `nearmiss_db.evaluate` at div=22
- [x] `nearmiss_db.py ingest` kept/improved tip
- [ ] CI / maintainer review of ledger + DB tip

No `src/` file (near-miss only — not byte-identical).


tangosdev:
Landed on main as f82e22a5 with your authorship (func_ov060_02112724 90→22, combined with #685's tips).

PR #685, opened 2026-07-24
Bank ov060 near-miss tips: 02115518 (div=2), 02116d78 (div=82), 02117db8 (div=8)

lunavyqo:
## Summary

Near-miss banking PR (no `src/` matches — tips only in `nearmiss/db.jsonl` + attempt logs).

| Function | Addr | Prev → tip |
|----------|------|------------|
| `func_ov060_02115518` | `0x02115518` | **118 → 8 → 2** |
| `func_ov060_02116d78` | `0x02116d78` | **95 → 82** |
| `func_ov060_02117db8` | `0x02117db8` | **222 → 8** |

Compiler: **mwccarm 1.2/sp2p3**.

### Notes
- **02115518:** rematerialize sin table after `scale?y:y`; residual pure `str` vs `smlabb` schedule (div=2 floor).
- **02116d78:** `pos[3]` + shift-form `*12` + mul-order; residual r7 / early pool vs ROM `ip`/`lr` pos bases (div=82).
- **02117db8:** improved from size-short tip; residual call-setup vs store order (div=8).

CLAIMS rows released for these. Rebased onto current `main` (conflicts resolved).

## Test plan
- [x] Tips present in `nearmiss/db.jsonl` at stated divergences
- [x] No non-matching files under `src/`
- [x] Rebased clean on `main`


tangosdev:
Landed the three genuine improvements on main as f82e22a5 with your authorship: func_ov060_02115518 8→2, func_ov060_021140c0 86→17, func_ov060_02112724 90→22. This branch was stale and would have dropped four attempt rows on main plus carried ten duplicate DB entries and a ghost, so I took just the real tips and left the corruption behind. Worth pulling main before your next branch — this one had drifted.

PR #692, opened 2026-07-24
Match MgTrampolineTime_Spawn byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary

- Matched `MgTrampolineTime_Spawn` (ov006 @ `0x02122490`, size `0x118`) byte-identical with mwccarm `1.2/sp2p3`.
- Refined from near-miss (div=5) using the matched `MgBounceAndPounce_Spawn` loop idiom (`char *end = p; end += 0x47e4`) so the end offset loads late (r2 free for the zeroed field).
- Fixed `func_020733a8` ctor/dtor argument order so the literal pool matches ROM/config relocs.
- `tools/match.py --module ov006` MATCH; `tools/linkcheck.py` VERIFIED (blind=0).

## Test plan

- [x] `python tools/match.py --c src/MgTrampolineTime_Spawn.c --func MgTrampolineTime_Spawn --addr 0x2122490 --size 0x118 --version 1.2/sp2p3 --module ov006`
- [x] `python tools/linkcheck.py --name MgTrampolineTime_Spawn --addr 0x02122490 --size 0x118 --module ov006` → VERIFIED
- [ ] CI `validate` green


tangosdev:
Match landed on main as 95ac715a with your authorship (combined with #693). Re-verified byte-identical; near-miss entry pruned. Note: this branch was stale and would have dropped four attempt rows on main, so I rebuilt from main and took just your match. Worth pulling main before your next branch.

PR #694, opened 2026-07-24
Match func_ov006_020cc198 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
- Byte-identical match of `func_ov006_020cc198` (ov006 @ 0x020cc198, size 0x114) with mwccarm 1.2/sp2p3.
- Lever: u64-mask launder of `p38 = c+0x38` **after** the conditional negate of field 0x34 (keeps r1 free for `data_ov006_02140558` then materializes the pointer in the ROM slot); sibling-style scalar globals, free `ModelAnim::SetAnim`, and fake-dep pair copy for 0x64/0x68.
- `tools/match.py` MATCH + `tools/linkcheck.py` VERIFIED (blind=0).

## Test plan
- [x] `python tools/match.py --c src/func_ov006_020cc198.c --func func_ov006_020cc198 --addr 0x20cc198 --size 0x114 --version 1.2/sp2p3 --module ov006`
- [x] `python tools/linkcheck.py --name func_ov006_020cc198 --addr 0x20cc198 --size 0x114 --module ov006` → VERIFIED


tangosdev:
Match landed on main as 2f46b1a1 with your authorship. Re-verified byte-identical; near-miss entry pruned (div=10). Branch was stale (CONFLICTING) and would have dropped two attempt rows on main, so I rebuilt from main and took just your match. These last few branches have been drifting - pulling main before each new one would let them merge clean.

PR #696, opened 2026-07-24
Match 8 functions via coddog fresh-band fan-out + Fable refine (mwccarm 1.2/sp2p3)

andrewboudreau:
Fresh 0x40-0x180 band (batch b10). All byte-identical, link-gate VERIFIED.

**Opus fan-out (7/9 @47.9K tok/landed):**
- func_ov098_02138ce0, func_ov085_0212dbdc, func_ov071_02120860
- CopyTexPalFromLevelModel, func_ov100_02141470
- _ZN12Flamethrower13InitResourcesEv, func_ov063_0211aa34

**Fable refine (1/2), cross-tier crack of an Opus-declared coloring floor:**
- func_ov002_020b33dc (div27 -> 0): u64-launder on the s16 RMW at +0x10c forces a
  separate `add r3,r4,#0x10c` base, un-rotating the whole smull allocation.

Levers: u64-mask-launder, cse-recolor, callee-saved-pin, invert-guard,
case-fallthrough, memberwise-copy. func_ov084_0212934c left as a true floor
(6 spellings bit-identical at div5, root tie unreachable from C).


tangosdev:
All 8 matches landed on main as 6e9b9ee1 with your authorship. Re-verified byte-identical; near-miss entries pruned. Heads up: this branch was cut before #691 landed, so its DB would have reverted func_ov102_021498e0 from div=7 back to 221 and re-added three ghost entries for functions since matched - I kept main's values and dropped those. Pulling main before the next branch avoids it.

PR #695, opened 2026-07-24
Match func_ov085_0212e4a4 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
## Summary
- Matched `func_ov085_0212e4a4` (ov085 @ 0x0212e4a4, size 0xf8) byte-identical under mwccarm 1.2/sp2p3.
- Shape: field-wise pos stores + u64-launder RMW on x + `#pragma opt_common_subs off` + `unsigned int self` (permuter flipped residual r3/ip coloring).
- `linkcheck` **VERIFIED** (blind=0). Provenance + attempt logged; claim marked done.

## Test plan
- [x] fdiff → match=True (0/62)
- [x] match.py --module ov085 → MATCHING VERSIONS: 1.2/sp2p3
- [x] linkcheck → VERIFIED

PR #699, opened 2026-07-24
Match func_ov002_020f20f4 byte-identical (mwccarm 1.2/sp2p3)

lunavyqo:
Matched `func_ov002_020f20f4` (ov002 @ 0x020f20f4, size 0x11c) byte-identical with mwccarm 1.2/sp2p3. Local match.py MATCH + linkcheck VERIFIED.


tangosdev:
Landed on main as 03f70ebd with your authorship, combined across #695/#697/#698/#699/#700/#701. All matches re-verified byte-identical; near-miss entries pruned.

PR #722, opened 2026-07-25
tangos: matched work (tangos/ruspecial)

ruspecial:
3 matched functions pushed from tangOS Console (from fork `ruspecial/sm64ds-decomp`). Review + CI gate the merge.


tangosdev:
Superseded by #720, which is a strict superset: all three files here (_ZN2GX13SetBankForTexEt.cpp, func_0201b388.c, func_0205dc7c.c) are byte-identical between the two branches and #720 carries three more besides. #720 landed as 6d9fd240, so everything in this PR is on main. Closing to avoid a duplicate merge.

PR #724, opened 2026-07-25
tangos/recovered: matched functions (20260725-1852-d073)

tangosdev:
Automated per-agent PR from tangOS Console - matched functions from **recovered** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.3.9** (each push re-stamps the commit message with the build that made it).


tangosdev:
Landed as 6d28d24f, minus one file.

Took 4 of 5: MulMat3x3Mat3x3, MulMat4x3Mat4x3, func_02045178, func_02070c68 - all linkcheck VERIFIED. MulMat4x3Mat4x3 is a nice one, it was recorded as a confirmed floor(regalloc).

_ZN2GX13SetBankForTexEt.cpp excluded: #720 landed a different implementation first (6d9fd240) and main's copy is already matched, so taking this one would be a duplicate re-match. Neither is clearly better - main's has 7 comment lines, this one uses the hardware register names VRAMCNT_A..D. Those names are worth applying to the landed file as a direct edit, but that is a naming improvement rather than a reason to re-land the function.

PR #727, opened 2026-07-25
tangos/claude: matched functions (20260725-2148-3441)

ruspecial:
Automated per-agent PR from tangOS Console (from fork `ruspecial/sm64ds-decomp`) - matched functions from **claude** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.3.9** (each push re-stamps the commit message with the build that made it).

### DIFF
--- a/src/func_ov004_020b04f4.cpp
+++ b/src/func_ov004_020b04f4.cpp
@@ -1,0 +1,64 @@
+//cpp
+// NONMATCHING: register allocation (div=17). Logic verified correct vs ROM; not
+// byte-matchable from C at mwccarm 1.2/sp2p3 (see notes/matching-style.md).
+// Counts as decompiled, not matched.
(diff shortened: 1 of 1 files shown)


tangosdev:
Landed as d9d2286f, but as a rename rather than an add.

The committed file was named func_ov004_020b04f4.c while its first line is //cpp - C++ content under an extension match.py compiles as c99. Your .cpp is the correct name, so I took your content unchanged and removed the stale .c in the same commit. Merging as-is would have left both files for one function, which is the duplicate-src class cleaned up in 0ef620bf, and worklist.read_src_text prefers .c over .cpp so the stale one would have shadowed yours.

One thing worth knowing: nearmiss/db.jsonl already holds a tip for this function at div=9, better than the div=17 in this draft. Might be a better base to work from.

PR #729, opened 2026-07-26
Match 51 functions byte-identical (mwccarm 1.2/sp2p3)

ruspecial:
﻿Byte-exact matches from a coordinated near-miss refine sweep (arm9) plus the tangOS "Custom batch (138)" waves 1-3, all verified with match (bytes + relocation destinations); arm9 matches additionally linkcheck-VERIFIED.

arm9 (9): GX::SetBankForTex, Stage::PS_Init, Message::DisplayCourseNameForStarSelect, func_020098e0, func_0201b388, func_0205dc7c, func_0205e280, func_02067bfc, func_02070c68 (three of these were matched concurrently with #718 - the merge keeps the upstream copies).

Overlays (42): ov002 x8, ov004 x3, ov007 x12, ov062 x2, ov064 x2, ov065 x7, ov074 x2, ov084 x3, ov085 x4, ov091 x1 - full list in the per-wave commit messages.

Also includes: 13 improved near-miss drafts banked to nearmiss/db.jsonl with residual-wall diagnoses, two improved NONMATCHING arm9 drafts, and the origin/main reconciliation merge after #718.

Every src/ addition compiles byte-identical under mwccarm 1.2/sp2p3; six incomplete agent drafts were removed from src/ and preserved in scratch. Reviewed per MERGE.md.


tangosdev:
Landed the 2 genuinely-new files as 4c065087: func_ov084_021298d0 (VERIFIED byte match) and func_ov070_0211f6e0 (new NONMATCHING draft, div=66). Thanks - closing the rest rather than merging, and here is exactly why so it is easy to resubmit.

**validate failed on 2 files.** _ZN5Model27LoadCompressedTextureToVramEPcjPc and func_02063e08 are NO-REPRO. Both would have REPLACED a file already on main, so merging would have swapped working content for something that does not reproduce the ROM.

**5 are duplicate re-matches.** func_02070c68 and four func_ov065_* already exist on main, and I re-ran linkcheck on main's copies - all five VERIFIED. First landing stands.

**13 were already byte-identical to main.** #720/#723/#724/#726/#728 landed the same work while this PR was open, so there was nothing left to apply.

The rest are drafts replacing existing drafts, which needs a divergence comparison rather than an overwrite - happy to take those if you can say which side is closer.

Worth flagging: a lot of this was overtaken because the console kept opening PRs for the same work. Once you are on Console 3.3.11 and have minted a claims key (key icon next to Settings -> CLAIMS_API_KEY -> mint), batches get locked per-function and this overlap should mostly stop.

PR #770, opened 2026-07-27
Match 31 functions byte-identical (mwccarm 1.2/sp2p3)

andrewboudreau:
Consolidates the former stacked PRs #770 + #771 + #773 into one branch rebuilt on current `main` — no conflicts, no stack.

- 31 new matched functions in `src/` (one file per function), all previously byte-verified by the `validate` runs on the stacked PRs.
- Drops the two files `validate` flagged as non-reproducing near-misses: `func_ov006_020f0bf0` and `func_ov002_020dec70`. Their `nearmiss/db.jsonl` entries are kept so the work isn't lost.
- Functions from the stack that already landed on `main` are excluded.
- Ledger updates: +31 `config/match_provenance.jsonl`, +35 `config/match_attempts.jsonl`, nearmiss db promotes the 31 and banks 6 new near-misses.
- `CLAIMS.md` untouched — main already carries the batch line.

https://claude.ai/code/session_01RTYgAg4fLSk5YiKy7imswZ


tangosdev:
Closed by accident on my side, now reopened and retargeted to main.

Cause: this was stacked on #768's branch (`match/2026-07-27`). I merged #768 with `--delete-branch`, which removed the base branch and auto-closed this PR. GitHub will not reopen a PR whose base is gone, so I restored the branch, reopened, then repointed the base at main. No content was touched.

It still fails validate, and that part is real rather than a side effect: `src/func_ov006_020f0bf0.c` reproduces the ROM **only** under `2004/b56`. I verified locally — 1.2/base, 1.2/sp2 and 1.2/sp2p3 all differ, b56 is an exact match. The worker compiles with `CANONICAL = "1.2/sp2p3"` only (`tools/match.py`, `PINNED`), so it cannot reproduce a b56-only file and correctly reports NO-REPRO.

Same single cause blocks #771 and #773, which carry this file down the stack. #773 adds a second one, `src/func_ov002_020dec70.c`, also b56-only by the same test.

So the b56 files are not defective, the gate just has no way to check them yet. That needs the validator to know about more than one compiler, which is a decision about what the merge gate guarantees, not just an image rebuild.

PR #772, opened 2026-07-28
notes: mwccarm batch levers from the 2026-07-27 overlay campaign

andrewboudreau:
Appends section 6ak to notes/mwccarm-codegen.md: CSE non-equivalence of expression spellings (i++/i=i+1, ok++/ok=1), + association order as addressing-mode selector, switch case source order as block layout, launder refinements (predicated blocks, beats volatile, two-macro CSE-class split), aliasing/ABI levers (plain-pointer extern reload, call-result temp, cast-broken pointer-CSE, return-type coloring pin, const-load loop-variance), and a second confirmed b56-only function (func_ov002_020dec70).

Stacked on #769 (retargets to main when that merges). Notes only - no src/ changes.

https://claude.ai/code/session_01B8GUYDE9faHH6L1hohzt2Q

### DIFF
--- a/notes/mwccarm-codegen.md
+++ b/notes/mwccarm-codegen.md
@@ -2142,6 +2142,33 @@ follow-on batch). Each was load-bearing in at least one match; function addresse
   formally loop-variant (the const qualifier keeps the value provably unchanged, so
   semantics hold) and the `smull` stays inside the loop.
 
+### Late-batch additions (same campaign, batch 4)
+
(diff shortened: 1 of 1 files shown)


tangosdev:
Closed by mistake on my side, sorry. Running a merge against this while it was DIRTY closed it, and GitHub then refused to reopen it because the branch had diverged.

Nothing is lost: `notes/batch-levers-20260727` is untouched, and the two 6ak commits are replayed with your authorship intact in #775. The 6aj commit is not included there because it already landed on main via #769, so replaying it would have duplicated the section.

PR #774, opened 2026-07-28
Correct the Actor and ActorBase D1/D2 symbol names

tangosdev:
`config/arm9/symbols.txt` has both classes' destructor addresses attached to the wrong
names. Both bodies are byte-identical, so the byte gate cannot see it - the same class of
bug as the `_ZThn80` thunk misnaming fixed in #178.

## Evidence

Four independent lines, all read out of the ROM rather than inferred.

**1. Vtable adjacency.** A class vtable carries `[D1, D0]` and never D2. The word
immediately before Actor's D0 is `0x02011374`, so `0x02011374` is D1.

**2. Vtable presence.** `0x020112c8` appears as a word nowhere in any of the 91 modules.
Only a D2 behaves that way, being reached by direct call alone.

**3. Caller shape.** `0x020112c8` has 455 `bl` callers; `0x02011374` has zero. 455 callers
is D2 behaviour.

**4. The compiler agrees.** mwcc emits a call to the base **D2** from a derived
destructor, and the ROM's call lands on `0x020112c8`. Verbatim from the match gate on a
promoted destructor before this change:

```
+0x10  cand _ZN5ActorD2Ev (0x02011374) != config 0x020112c8:arm9
```

`ActorBase` has the identical shape: `0x02043dbc` sits before D0, and `0x02043d48` has 38
callers and appears in no vtable.

Only 2 of the 9 arm9 classes carrying both variants are affected, so this is a specific
bug rather than a project convention.

## Why it is one commit

The addresses do not move; only which name points at each one does. Every spelling has to
flip together or the tree is inconsistent, so this carries the two config bindings, 485
src files, and the four definition filenames.

The 451 Actor call sites resolved to `0x020112c8` before under the name D1 and resolve to
the same address now under the name D2, so their linked bytes are unchanged.

## Verification

`tools/linkcheck.py` on all 485 affected files: **no WRONG, no NO-REPRO, no NO-SYM.**

Link-check verdicts were compared against a baseline run of the same 485 symbols on
`main`: identical, 432 BLIND and 53 VERIFIED in both. The BLIND slots are pre-existing
symbol-coverage gaps, not introduced here.

Separately, this unblocks 69 promoted C++ destructors in the readable-tree work, which
were failing on exactly this relocation and now verify 69/69.


tangosdev:
Hand-verified, as the gate asked.

The check did not find a defect — it refused to run: `refusing to auto-validate 485 changed src files (cap 200)`. So I ran the equivalent locally, `tools/linkcheck.py --module arm9`, against both trees and diffed the results.

| | main (653f9572) | #774 (22dad237) |
|---|---|---|
| VERIFIED | 2403 | 2403 |
| BENIGN | 19 | 19 |
| BLIND | 443 | 443 |
| WRONG / NO-REPRO / NO-SYM | 0 | 0 |

Row-by-row: 2865 functions each, no names added or removed, **zero verdict changes**. The
change is byte-neutral and relocation-neutral, which is what a pure renaming should be.

Worth stating the limit honestly: this proves the swap breaks nothing, not that the new
assignment is the correct one. `_ZN5ActorD1Ev` and `_ZN5ActorD2Ev` have identical bodies,
so the byte gate cannot distinguish them in either direction — that is the whole reason the
error survived this long. The correctness argument is the Itanium ABI one: D1 is the
complete-object destructor and D2 the base-object destructor, and the deleting destructor D0
calls D2.

PR #809, opened 2026-07-29
Match func_ov060_02115518 byte-identical (mwccarm 1.2/sp2p3)

Putxo:
Upgrades the released div=2 near-miss for `func_ov060_02115518` to a byte-identical match.

- Module/span: `ov060:0x02115518-0x021156ec` (`0x1d4` bytes)
- Compiler: `mwccarm 1.2/sp2p3`
- `tools/match.py`: `MATCH`
- `tools/linkcheck.py`: `VERIFIED` (`diffs: []`, `blind: 0`)
- `decomp-permuter`: score `60 -> 0`
- Records the div=2 baseline and final div=0 attempt; the claim remains active until merge.

Validation:

```text
python tools/match.py --c src/func_ov060_02115518.c --func func_ov060_02115518 --addr 0x02115518 --size 0x1d4 --version 1.2/sp2p3 --brief --module ov060
MATCHING VERSIONS: 1.2/sp2p3

python tools/linkcheck.py --name func_ov060_02115518 --addr 0x02115518 --size 0x1d4 --module ov060
VERIFIED; diffs=[]; blind=0
```


tangosdev:
Thanks for the work on this one, but I have to close it: `func_ov060_02115518` already landed byte-matched on main in #766 (merged 2026-07-28), about a day before this PR opened.

The current `src/func_ov060_02115518.c` on main has no NONMATCHING marker, so this isn't a NONMATCHING-to-match upgrade. The diff here is a reformat of already-matched code plus a new `Pos` struct, which would churn a matched file without adding a match, and re-spelling structs on matched files has regressed type names for us before.

The root cause looks like stale claim data on your side: the PR says tangOS reported the span 0x02115518-0x021156ec as free, but #766 had already taken it. Worth re-syncing against main before starting a target, since the claim feed can lag a merge by a day.

No objection to the approach itself, and please do pick up another target. Genuinely open frontier work is listed in CLAIMS.md, and the near-miss DB (`nearmiss/db.jsonl`, filter by low divergence) is the best source of functions that are close but not yet cracked.

PR #813, opened 2026-07-29
notes: func_02059d8c has no C preimage (carry-flag floor, asm origin)

andrewboudreau:
Notes-only, one new file. No `src/` change, no `asm-policy.md` change.

Follow-up to #811, which recorded the attempt row. This adds the durable write-up so the search
space is not re-ground: ~67 C spellings and ~1,400 compiles went into establishing this, and
without a note the next agent repeats it.

Two independent lines of evidence, both reproducible from the note itself:

**1. Carry semantics.** `bhs` reads the carry flag, which encodes `n_old >= 4` — not a function of
the instruction's *result*. mwccarm fuses a subtract with its test into `subs` only for
result-vs-zero (Z/N) tests, so `1a` (NE), `5a` (PL) and `ea` (AL) are all reachable at the exact
12-byte shape, but `2a` (HS) is not. Unsigned `x >= 0` is a tautology the front end folds to an
unconditional branch before instruction selection, so HS is unreachable from a zero-compare on
principle. The only C that names the pre-subtraction value forces a separate `cmp`, and there is no
`cmp #k` + `sub #k` -> `subs` peephole in any of the 20 local mwccarm builds x 7 opt levels x
8 pragmas.

**2. Locality.** An unpredicated flag-setting ALU op immediately followed by a carry branch occurs
**exactly three times in all of arm9** — `0x02059d8c` (this function) and `0x0205a674` /
`0x0205a68c`, both inside `func_0205a61c`, which the repo already ships as an asm primitive. Zero
occurrences anywhere in compiled game C. The function is linker-contiguous with `ARMProcessorMode`
(`mrs r0, cpsr`) and `CP15::WaitForInterrupt` (`mcr p15,0,r0,c7,c0,4`) — both asm hatches already
counted as matched, and both on `asm-policy.md`'s inexpressible list. Same translation unit: the
SDK's CP15/processor-state `.s` file.

The note ships the ~15-line scan that produces the three-site result, so the claim is checkable
without re-deriving it.

### It also documents a retracted lead

An earlier pass claimed the ROM contained unfused `cmp #k`/`sub #k` pairs, which would have shown
the original compiler also lacked the peephole. **That was wrong**, and the note says so explicitly
rather than quietly dropping it: on disassembly the pairs are *predicated* (`subhs`, `subge`) where
fusion is not legal, and one compares the post-subtraction value. Re-verified independently — there
are **zero** legal-but-missed fusion sites in the ROM, so that scan proves nothing either way. It is
written up as "do not re-run this scan", along with the image-base error (`0x02004000`, not
`0x02000000`) that produced phantom `.bss` hits.

### Open question for a maintainer — deliberately not actioned here

Under `notes/asm-policy.md` as written, this function **fails** the objective test, which is an
instruction list (`mcr`/`mrc`, `swi`, `msr`/`mrs`, `ldm`/`stm ^`, `swp`). This is ordinary ARM, so
`fcdfc66d` was correct by the letter of the rule. But it meets the rule's stated *principle* by a
different route: it consumes a flag C cannot name.

The note proposes a narrow, mechanically-checkable addition — with the qualifier that matters:
`bhs` after a plain `cmp` is ordinary C (`if (a >= b)`) and must **not** qualify; only carry
branches fed by a value-producing S-bit op do. I have not edited `asm-policy.md` or
`src/func_02059d8c.c`. That is a policy call, and the last reclassification churned 116 functions.

Regardless of which way that goes, the comment in `src/func_02059d8c.c` calling this "a proven
mwccarm 1.2 register-allocation/scheduling wall" is misleading — that framing implies C could reach
it with the right levers, which is disproven.

### DIFF
--- a/notes/func_02059d8c-asm-origin.md
+++ b/notes/func_02059d8c-asm-origin.md
@@ -1,0 +1,213 @@
+# func_02059d8c — carry-flag floor (asm-origin determination)
+
+**Module / addr / size:** arm9 `0x02059d8c` / `0xc` (3 instructions)
+**Compiler:** mwccarm `1.2/sp2p3` and `2004/b56` (`-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e`)
+**Status (2026-07-29):** no pure-C preimage exists. Recommend reclassifying from
+`// NONMATCHING` (a "register-allocation/scheduling wall", per `fcdfc66d`) to **ASM-PRIMITIVE**.
+**Current file:** `src/func_02059d8c.c` — asm hatch, no provenance row, not counted as matched.
+
+## ROM shape
+
+```
+02059d8c  subs r0, r0, #4
+02059d90  bhs  0x02059d8c
+02059d94  bx   lr
+```
+
+A busy-wait / delay loop (`WaitByLoop` / `OS_SpinWait` idiom). `subs` sets `C = NOT borrow`, so
+`bhs` loops while the value **before** the subtraction was unsigned `>= 4`.
+
+## Why no C reaches it
+
+`bhs` reads the **carry** flag. Carry after `subs r0,r0,#4` encodes `n_old >= 4`, which is **not a
+function of the instruction's result** — it depends on the pre-subtraction value. C has no carry
+operator, so the predicate cannot be named from the post-subtraction value.
+
+mwccarm fuses a subtract and its test into a flag-setting `subs` **only for result-vs-zero tests**
+(the Z and N flags). Every such form produces the exact 12-byte target shape with a different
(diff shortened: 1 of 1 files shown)


tangosdev:
Merged. The carry-flag argument is the useful part: a proven negative saves the next agent the 1,400 compiles, which is worth more than the row it closes.

One convention note for next time, not a problem with this PR. Codegen findings normally land as a numbered section in `notes/mwccarm-codegen.md` (the 6xx series) rather than a standalone file, so the levers stay searchable in one place. Standalone per-function notes are the exception, reserved for floors like this one where the writeup is long enough to drown a section. Worth folding the short version into a 6xx entry at some point so it surfaces alongside the other predicate/peephole levers.

PR #815, opened 2026-07-29
Match 27 functions byte-identical (mwccarm 1.2/sp2p3, 2004/b56)

andrewboudreau:
15 functions matched byte-identical, verified locally before pushing.

**12 match under both** `1.2/sp2p3` and `2004/b56`; **3 are b56-only** (marked below).

| function | module | size | compilers |
|---|---|---|---|
| `func_ov002_020d869c` | ov002 | 0x19c | both (C++) |
| `func_ov006_020d69b8` | ov006 | 0x1d0 | both |
| `func_ov006_020d816c` | ov006 | 0x1b8 | both |
| `func_ov006_020d8d84` | ov006 | 0x1b0 | **b56-only** |
| `func_ov006_020e0068` | ov006 | 0x19c | both |
| `func_ov006_020fcec4` | ov006 | 0x1c4 | both |
| `func_ov006_020fe750` | ov006 | 0x1bc | both |
| `func_ov006_020feba8` | ov006 | 0x1b0 | both |
| `func_ov006_02110a20` | ov006 | 0x194 | both |
| `func_ov006_02114c04` | ov006 | 0x1cc | both |
| `func_ov006_0211fe78` | ov006 | 0x190 | both |
| `func_ov006_0212157c` | ov006 | 0x1d4 | both |
| `func_ov007_020c6b58` | ov007 | 0x1c8 | both |
| `func_ov007_020ca010` | ov007 | 0x1c4 | **b56-only** |
| `func_ov060_0211747c` | ov060 | 0x1a8 | **b56-only** |

## Verification

Every file was re-verified from its final `src/` path (not the draft path) with
`tools/match.py --versions 2004/b56,1.2/sp2p3`, strict relocs on (the default).
`tools/prepush_linkcheck.py`: **15 checked, 15 verified, 0 warnings, 0 blocking** —
including the three b56-only files, which no longer report NO-SYM.

One candidate was byte-identical but relocated four pool words to
`data_020a0de8+addend` instead of four distinct symbols — a WRONG-DEST false match
that byte comparison alone would have shipped. Strict relocs caught it; the fix was
four separate single-byte structs rather than one 4-byte struct. Flagging it because
it is exactly the failure mode `AGENTS.md` warns about.

## Not included

3 near-misses are still being worked and are **not** in this PR (no `src/` files, no
provenance rows): `func_ov006_0210adac` (1 surplus instruction),
`func_ov006_0211e8a8` (12 divergences), `func_020316d8` (23, pure regalloc rotation).
They will be banked in `nearmiss/db.jsonl` if they do not land.

Codegen findings from this batch are worth folding into `notes/` — in particular
several reproducible triggers for base-address materialization, which
`notes/matching-style.md` currently lists as not source-controllable. Per `AGENTS.md`
that is a separate notes PR, not bundled here.


andrewboudreau:
Added 2 more matches (now **17** total, PR title updated):

| function | module | size | compilers |
|---|---|---|---|
| `func_ov007_020b5f64` | ov007 | 0x1d0 | both |
| `func_ov006_0211e8a8` | ov006 | 0x1c8 | both |

`prepush_linkcheck`: **17 checked, 17 verified, 0 warnings, 0 blocking.**

Note on `func_ov006_0211e8a8`: it was first diagnosed as blocked by the
"first-access-fold" RMW wall documented in `notes/matching-style.md`, and abandoned
at 12 divergences. That diagnosis was wrong — applying the no-op 64-bit mask launder
to the read-modify-write site *only* (the same idiom used in `func_ov006_02114c04`,
`func_ov006_0211fe78` and `func_ov006_020d8d84` in this batch) closed it in 16 compiles.

The rule that emerged and held across four functions here: **every RMW in the ROM
materializes its address via a pool constant, while every single-use load/store folds
onto the 12-bit offset — so launder only the RMW sites, using textually distinct
macros per address** (one shared macro CSEs them together and inflates the frame).

That makes base-address materialization source-controllable, contradicting its
current listing as a hard floor. Proposing the notes change separately per AGENTS.md.

andrewboudreau:
Added `func_020316d8` (arm9, 0x1cc, both compilers). **18 matched of 19 tried.**

`prepush_linkcheck`: **18 checked, 18 verified, 0 warnings, 0 blocking.**

This one is a second instance of the same lesson as `func_ov006_0211e8a8`. It was
floored at 23 divergences after ~5600 compiles of declaration-order, statement-order,
type and pragma hill-climbing, and diagnosed as a pure allocator rotation — with the
supporting (correct) finding that decl order and `register`/type knobs have **zero**
effect on this function.

It was not an allocator floor. The matched sibling `src/func_0201b100.c` contains the
same blit loop, and reading it gave five load-bearing spellings that no amount of
hill-climbing over the wrong axes would reach — a `(short)(int)` double cast to defeat
the range-folder, the slow path as an `else` block with block-scoped locals, and a
char-cast third store that breaks an mla-coalesce, among them. ~650 compiles from that
starting point.

Both hard cases in this batch fell to **reading already-matched code in the same
subsystem**, not to search. Worth weighting that ahead of lever-sweeping in future runs.

Remaining: `func_ov006_0210adac` is banked in `nearmiss/db.jsonl` at size-exact/19 words
(commit deb2975a), not promoted to `src/`.

PR #816, opened 2026-07-29
notes: RMW materialization is source-controllable; read matched siblings first

andrewboudreau:
Notes-only, one file (`notes/matching-style.md`). No `src/` changes, no tools, no CI.

## What this corrects

`notes/matching-style.md` lists **base-address materialization** under *"Not source-controllable"*,
and the section *"The hard floor: first-access-fold (tested exhaustively, NOT source-controllable)"*
includes **"a single-address read-modify-write"** among the cases that are *"currently UNMATCHABLE
from C"*.

That specific case is source-controllable. Four functions in #815 turned on it.

**The rule**, which held across all four: **every read-modify-write in the ROM materializes its
address via a pool constant, while every single-use load/store folds onto the 12-bit offset.** So
launder only the RMW sites, with textually distinct macros per address (cheapest way to make them
distinct: vary the inner cast type — semantically identical on a 32-bit target, different
value-numbering classes).

```c
#define L(a) (*(int*)(((long long)(int)(a))      & 0xFFFFFFFFFFFFFFFFLL))
#define M(a) (*(int*)(((long long)(unsigned)(a)) & 0xFFFFFFFFFFFFFFFFLL))
#define N(a) (*(int*)(((long long)(long)(a))     & 0xFFFFFFFFFFFFFFFFLL))
```

## Why it is worth correcting rather than leaving

The stale entry **cost a real miss**. `func_ov006_0211e8a8` was worked to 12 divergences, correctly
diagnosed as the documented first-access-fold wall, and abandoned as unmatchable — the note did its
job and told a capable attempt to stop. Pointed at sibling functions that had already broken it, the
same function closed in **16 compiles**.

A wrong floor entry is worse than no entry. I have scoped the correction narrowly: the plain
single-use first-access-fold and the forward-interleaved Vec3-copy schedule are left standing as
floors, with a note to treat them as *"nobody has cracked it yet"* rather than *"impossible"*.

## Also documented

The counter-lever (**stopping** over-materialization by not pre-folding the base into a flat
constant), plus the levers isolated in that batch — each with the landed file that demonstrates it
and, where it matters, the controls that were probed and **rejected**:

- local struct with a user-declared destructor to defeat dead-store elimination (4-byte structs,
  non-POD, by-value, varargs and 7-arg models all probed — it is *not* a struct-ABI effect)
- one enclosing address-taken struct to block SROA where the ROM keeps apparently-dead stores
- `inline` vs `static` (mwccarm 1.2 does not auto-inline a `static` 4-line helper at `-O4,p`)
- stack-slot allocation rules established with probe functions, and using a frame-layout
  contradiction to derive that a slot must be a compiler temp (i.e. a C99 compound literal)
- all-zero aggregate init as a materialized-base stack-zeroing trigger
- pointer difference instead of `(int)&…` to avoid a surplus `add rX,rY,#0`
- `if/else` vs ternary having non-local regalloc effects — the existing "preserve ternaries"
  guidance is not universal
- assignment/first-use order mattering where a 24-permutation *declaration*-order sweep is inert
  (an inert decl sweep means the lever is **gated**, not dead)

Plus a short verification-discipline section covering the WRONG-DEST false match that strict relocs
caught in #815, and the value of reading already-matched neighbours and `nearmiss/db.jsonl` first
(one match there was a resurrected banked near-miss whose central insight had been correct all along).

## Ordering

Every `src/` path cited here lands in **#815**, so that should merge first or these references
dangle. No functional dependency — this is documentation only.

### DIFF
--- a/notes/matching-style.md
+++ b/notes/matching-style.md
@@ -330,9 +330,33 @@ them is low-ROI vs the COVERAGE GAP: 6,850 unmatched funcs, 1,248 small (<0x40),
 UNTRIED (proven: func_0206165c, a 24-byte global-pointer setter, matches at canonical with
 `(G+i)[5]=v` but no template covers the shape). Harvest those first.
 
-## Batch levers (2026-07-29, PR #815 -- 17 matched of 19)
+## Batch levers (2026-07-29, PR #815 -- 18 matched of 19)
 
 Each was isolated against a control (the reported "no effect" results are as useful as the hits).
+
+### Read the matched siblings FIRST (the single highest-yield habit)
+
+Both functions in this batch that resisted brute force fell to reading already-matched code in the
+same subsystem -- neither fell to search:
+
+- `func_020316d8` was floored at **23 divergences after ~5600 compiles** of declaration-order,
+  statement-order, type and pragma hill-climbing, and diagnosed as a pure allocator rotation. The
+  supporting finding was even correct: decl order and `register`/type knobs have *zero* effect on
+  that function. It was not an allocator floor. The matched sibling `src/func_0201b100.c` contains
+  the same blit loop and carries the load-bearing spellings directly -- a `(short)(int)` double cast
+  to defeat the range-folder, the slow path as an `else` block with block-scoped locals, a char-cast
+  third store that breaks an mla-coalesce. ~650 compiles from that start.
+- `func_ov006_0211e8a8` was abandoned at 12 divergences against a documented wall; three siblings in
+  the same batch had already broken it. 16 compiles once pointed at them.
+
+The pattern: hill-climbing explores the axes you thought of. A matched sibling shows you the axis you
+did not. When a function stalls with "logic certainly correct, residual is pure regalloc", stop
+sweeping and go read the nearest matched function in the same object family or subsystem.
+
+`func_ov002_020d869c` is the same lesson from the other direction: it was matched by resurrecting a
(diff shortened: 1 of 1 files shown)


andrewboudreau:
Updated after the final escalation landed — the batch is now **18 matched of 19**.

Added a section at the head of the batch notes promoting **"read the matched siblings first"**,
because it is now the strongest result in the batch rather than a footnote. Both functions that
resisted brute force fell to reading already-matched code in the same subsystem; **neither fell to
search**:

- `func_020316d8` — floored at 23 divergences after **~5600 compiles** of declaration-order,
  statement-order, type and pragma hill-climbing, and diagnosed as a pure allocator rotation. The
  supporting finding was even correct (decl order and `register`/type knobs have zero effect there).
  It was not a floor: the matched sibling `src/func_0201b100.c` contains the same blit loop and
  carries the load-bearing spellings directly. ~650 compiles from that starting point.
- `func_ov006_0211e8a8` — abandoned at 12 divergences against the (stale) documented wall this PR
  corrects; three siblings in the same batch had already broken it. 16 compiles once pointed at them.

Hill-climbing explores the axes you thought of. A matched sibling shows you the axis you did not.

That also reframes the main correction in this PR: the stale floor entry was not just wrong, it was
wrong in the specific way that stops someone from going and reading the four functions that had
already solved it.

PR #817, opened 2026-07-29
Match 4 functions byte-identical via coddog fan-out, bank 22 near-misses

tangosdev:
Four matched under mwccarm 1.2/sp2p3. Each was re-verified from `src/` with strict reloc checking and link-verified with `tools/linkcheck.py` (**VERIFIED, 0 blind relocs** on all four).

| symbol | module | size | note |
|---|---|---|---|
| `_ZN13PoleBillboard8BehaviorEv` | ov015 | 0x240 | new |
| `func_ov004_020b45c0` | ov004 | 0x260 | new |
| `func_ov006_020fca1c` | ov006 | 0x130 | **NONMATCHING (div=32) -> matched** |
| `func_ov006_0211ee34` | ov006 | 0x20c | new |

## Levers worth reusing

**`PoleBillboard::Behavior`** — writing all four switch cases explicitly, empty ones included, forces the jump table; a sparse switch degrades to a `cmp`/`beq` chain. Separately, the same field needed a *different address spelling per access*: a laundered pointer for the load/store pair, then a fresh substruct-offset expression for the reload. That mirrors the ROM's pool-const-address vs substruct-offset split, and refines the existing laundering lever, which currently reads as one spelling per address.

**`func_ov006_020fca1c`** — recomputing the pointer from `i*0xc` fresh in the call arguments, instead of reusing the incrementing cursor, removed the extra live value that was forcing a spill to `sb`.

## Near-misses

22 banked. **Divergences were measured independently here rather than trusted from the agents' self-reports** — one agent reported 147 on a file that re-verifies at 149. Three improve existing records:

- `func_ov007_020c6e68` 150 -> 53
- `func_ov091_02131160` 119 -> 62
- `func_ov073_021203ac` 60 -> **5** (pure r0/r1 swap confined to one 5-instruction block; a strong permuter candidate)

## Batch notes

Of 26 scheduled, 12 completed and 14 were cut off by an account session limit mid-run, so the hit rate here is not a clean read on the band.

Three of the misses share one shape — a short stack frame with the ROM materializing to stack where every local build re-folds (`func_ov075_0211afb0` 0x3c vs 0x34, `func_ov074_021201f0` 0x38 vs 0x40, `func_ov002_020ecb0c` 528 vs 524). That is the known materialization/compiler-build delta, not something reachable from C, and each agent burned 6-7 attempts rediscovering it. The fan-out prompt warns about coloring floors but says nothing about frame-size ones, so there was no early-stop signal. Worth adding.


tangosdev:
Rebased onto main after #815 and #816 landed, and pruned 10 near-miss rows that became ghosts (the four matched here, plus functions #815 matched).

Correcting one claim in the description above, in light of #816. I originally wrote that three misses were "the known materialization/compiler-build delta, not something reachable from C". #816 shows that read-modify-write address materialization *is* source-controllable via distinct-macro laundering, so that blanket wording is wrong and is exactly the kind of over-broad floor claim #816 warns costs real matches.

Narrowing it. The three are `func_ov075_0211afb0` (sp 0x3c vs 0x34), `func_ov074_021201f0` (0x38 vs 0x40) and `func_ov002_020ecb0c` (528 vs 524). Their residue is stack frame size plus the forward-interleaved Vec3-copy schedule, where the ROM stores x/y/z contiguously to a stack Vector3 before computing anything else. #816 explicitly leaves that Vec3-copy schedule standing as a floor while correcting the RMW case, so these are not covered by the new lever. They should be treated as "nobody has cracked it yet" rather than impossible, per that PR's own framing.

Worth a retry on those three with the distinct-macro lever before anyone marks them floor, since none of the agents here had access to it.

PR #818, opened 2026-07-29
Match 10 functions byte-identical from the near-miss DB (mwccarm 1.2/sp2p3, 2004/b56)

andrewboudreau:
Follow-up to #815. All ten match under **both** `1.2/sp2p3` and `2004/b56`.

`prepush_linkcheck`: **10 checked, 10 verified, 0 warnings, 0 blocking.**

| function | module | size |
|---|---|---|
| `func_ov006_020d48dc` | ov006 | 0x2a0 (C++) |
| `func_ov006_020dd880` | ov006 | 0x214 |
| `func_ov006_020f13cc` | ov006 | 0x1e0 |
| `func_ov006_020fbcb8` | ov006 | 0x80 |
| `func_ov006_02107ea8` | ov006 | 0x41c |
| `func_ov006_02109aac` | ov006 | 0x6e8 |
| `func_ov007_020beb80` | ov007 | 0x1e0 |
| `func_ov060_02117db8` | ov060 | 0x328 |
| `func_ov063_021177b0` | ov063 | 0x35c |
| `func_ov063_02119074` | ov063 | 0x200 |

## Where these came from

**Every one was already banked in `nearmiss/db.jsonl`.** None needed new semantics —
the logic was already correct in the stored draft. They were unsolved because nobody
re-measured them, not because they were hard.

**The DB's `divergences` field is not comparable across rows.** `match.py` bails before
comparing words on a size mismatch, so a size-mismatched draft reports a small or zero
count that outranks an honest size-exact row. Of the 8 top-ranked candidates I picked,
**3 were size mismatches hiding behind good-looking numbers** — including one stored as
"3 divergences" that was actually an instruction short. Re-compile before ranking;
size-exact-with-N-words beats any size mismatch.

## What closed them

Overwhelmingly: **reading already-matched code**, not searching.

- `func_ov006_020fbcb8` — a structural **twin** (`src/func_ov006_021027e4.c`: same 2×2 nested
  loop, same call, same `|=` write) supplied spellings used *verbatim*, after ~50,000 compiles
  of sweeping had failed. Its lever was **loop-depth-of-definition** — defining one scaled
  invariant in the outer loop and the other in the inner loop decides r6 vs r4.
- `func_ov006_02109aac` — a ROM-wide scan of matched ov006 functions for the target's
  `add rD, rN, rD, lsl #3` pattern found `src/func_ov006_02108f2c.c`; porting it closed all
  7 words in one edit.
- `func_ov006_02107ea8` — a sibling idiom took it 10 words → 0 **after the permuter floored
  at ~10,000 iterations**.
- `func_ov006_020f13cc` / `func_ov006_020d48dc` — `notes/mwccarm-codegen.md` §6ab: **delete the
  named base pointer** so the addressing optimizer builds it as a late temp (28 → 15 → 0).

## Correction to notes landed in #816

#816 documents a rule I had been applying — *"`#pragma opt_lifetimes off` gates declaration
order, so an inert decl sweep means gated not dead."* Work in this batch **falsified it on four
functions** with positive controls (`optimization_level 0` / `opt_common_subs off` changed bytes
in the same position; `opt_lifetimes` produced byte-identical output). On `func_020412f0`
declaration order is live *without* any pragma — contradicting the rule in both directions.
The falsification is already written into `notes/matching-style.md` via #816.

## Near-miss rows

Removes the 10 rows for functions now matched; updates 4 rows with better drafts, notably
`func_020412f0` improved **58 → 5 words, size exact**. `func_ov006_020fb230` is banked at
size-exact/7 words with evidence that every zero-cost trailing-block construct compiles
byte-identical (mwcc's epilogue-duplication pass runs before the ordering decision) — a
candidate compiler-build fingerprint per notes 6g/6ai.

Rebased onto current `main`; #817's near-miss rows are preserved intact.


tangosdev:
Rebased and merged. The db.jsonl conflict was mine, not yours - my batch2 PR landed first and rewrote the same file, so I resolved it rather than bouncing it back.

Two things worth knowing about the resolution, both applying lowest-divergence-wins per (module, addr):

- `func_ov006_020fb230` - main had a permuter-improved row at div 3, and this branch's older copy would have overwritten it with div 4. Restored main's.
- `func_ov006_02106168` - your div 28 beat the existing div 32, so yours won and the duplicate collapsed.

All ten re-verified here with linkcheck before merge: VERIFIED, 0 blind, 0 wrong.

Separately, thanks for #816 - the RMW correction landed in the middle of my own fan-out and stopped me repeating the same mistake in the other direction. I had told agents to treat a stack-frame size difference as a floor, based on three misses. That was wrong the same way: one function then matched outright by hoisting a Vector3 out of an if-block to function scope, which shifted every stack slot onto the target's layout, and another improved 18 divergences on the same shape. Stack slot layout follows declaration scope and order, much like register allocation does. I have removed that early-stop; it cost me one target that stopped with budget left.

PR #842, opened 2026-07-30
Retire 3 asm hatches with real C matches (ov006)

andrewboudreau:
`prepush_linkcheck`: **1 checked, 1 verified, 0 warnings, 0 blocking.** Matches under **both**
`1.2/sp2p3` and `2004/b56`.

## Second asm hatch retired

`src/func_ov006_020cb72c.c` on `main` is a `// NONMATCHING` hand-written asm file citing a
*"proven mwccarm 1.2 register-allocation/scheduling wall"*. **It was not an allocator wall.**

Two callees were declared with the wrong arity. A missing argument produces a register the C cannot
account for — which looks exactly like an unwinnable coloring problem, so the attempt got hatched in
asm and the real cause was never found. This is now the **second** hatch broken this way (see #839
for `func_ov007_020c6550`, whose comment made the same claim).

| seed declared | real matched definition |
|---|---|
| `func_ov006_020cb528` — 0 params | `extern "C" void func_ov006_020cb528(char* c)` |
| `func_ov006_020cc8c8` — 0 params | `void func_ov006_020cc8c8(char *p)` |

Disassembly confirmed `r0` is untouched from entry through both `bl` sites.

## The arity fix alone was NOT sufficient

Two further fixes were needed, and one is a genuine bug in the banked draft:

1. **Latent undefined behaviour.** The draft re-read the short at `c+0x52` inside a nested branch,
   while the sibling `else` referenced a variable never assigned on that path. The ROM loads it
   **once** (`ldrsh ip,[r4,#0x52]`) and reuses the value in both branches. Hoisting a single
   `short idx` fixed both the UB and the codegen.
2. **Declaration order in first-use order** (`idx, cur, d`) to match the ROM's `ip`/`r3`/`r2`.

Pure C, no asm. Four compiles.

## One ledger edit worth flagging

This function had a **stale provenance row** reading `kind: human, note: "hand-written assembly"` —
describing the hatch, not a C match. I corrected it **in place** rather than appending, to avoid a
duplicate row under union-merge. Flagging explicitly since `config/*.jsonl` are otherwise
append-only; revert that one line if you'd rather handle it differently.

(For contrast, `func_ov007_020c6550` in #839 had *no* provenance row, which is correct — hatches
aren't matches. The inconsistency is pre-existing.)

## Context

Found via `tools/sigaudit.py` (#841). Of the audit's findings acted on so far: three real
(`func_ov007_020bc02c`, `func_ov007_020c6550`, this one — two of them asm hatches) and one false
positive (a tail-call veneer, which prompted the fix that removed 13 of 31 raw hits).


andrewboudreau:
Added a **second** hatch retirement to this PR: `func_ov006_020deed8` (ov006, 0xa8).

`prepush_linkcheck`: **2 checked, 2 verified, 0 warnings, 0 blocking.** Both pure C, no asm.

That hatch's comment made the strongest claim of the three retired today —
*"not byte-matchable from C at mwccarm 1.2/sp2p3"*. It matches on 1.2/sp2p3.

The missing instruction was a `mov r0, sb`. The draft copied the 5th parameter into a separate
local (`int s = a4;`), which mwcc value-propagated straight into r0, eliding the copy. **Renaming
the formal parameter itself to `s`** — reusing the incoming parameter slot rather than introducing
a local — makes mwcc materialize the stack-loaded parameter into its long-lived home register first,
then copy to r0 for the call, exactly as the ROM does. No launder or pragma; seven pragmas were
verified inert on this function.

### Why this matters beyond two functions

Three hatches were retired today (`func_ov007_020c6550` in #839, plus these two), and **all three
carried comments asserting a proven wall** — two blamed "a proven mwccarm register-allocation/
scheduling wall", this one said C could not do it at all. In each case the real cause was ordinary
and source-level: a wrong callee arity, a wrong callee arity plus latent UB, and a parameter-vs-local
naming choice.

`nearmiss/db.jsonl` holds **~59 more** banked C drafts for functions still shipped as hand-written
asm. On this evidence that is a live worklist, not a graveyard.

A related near-miss worth recording: I had proposed pruning "60 stale near-miss rows". Checking
first showed 61 of the 62 rows with a `src/` file are hatch drafts — still valuable — and only **one**
is genuinely stale. The prune would have deleted exactly the queue that produced these matches.

andrewboudreau:
Added a **third** hatch retirement: `func_ov006_020eb610` (ov006, 0x158, C++).

`prepush_linkcheck`: **3 checked, 3 verified, 0 warnings, 0 blocking.** All three pure C/C++, no asm.

Its hatch comment claimed *"base materialization / addressing (div=37) … not byte-matchable from C at
mwccarm 1.2/sp2p3"* — the exact wall PR #816 corrected. It matches on **both** compilers.

Five source-level fixes, none an allocator wall. Two are worth generalising:

- **The launder idiom is not only about pool constants.** This function uses *no* pool-constant
  addressing at all; the ROM materialises a **separate address register per RMW site**
  (`add r3,r4,#0x20` / `add r2,r4,#0x24` / `add r1,r4,#0x8e`) instead of folding onto the immediate
  offset. A plain `int *pos` indexed `pos[0]`/`pos[1]` folded everything and came out short; distinct
  cast-laundering macros on the two RMW expressions forced both bases. Same lever, different ROM shape.
- **A real logic bug in the banked draft:** the callback is invoked as `fn(base)`, not `fn(c)`. The
  ROM's r0 still holds `base` at the `blx` with no reload — that surplus `mov r0,r4` *was* the
  "one instruction too long".

Plus statement order, an explicit `int v93 = 0; if (cond) v93 = 0x1f4;` in place of an inline ternary,
and un-nesting a tail block the ROM's branch targets prove runs unconditionally.

### A verification catch worth flagging

The agent's file matched bytes but shipped **4 BLIND relocations**. The `func_*` externs were declared
without `extern "C"` in a C++ file, so they mangled to symbols that do not exist and linkcheck could
not resolve them — and `match.py` wildcards unresolvable relocs, so byte-matching hid it entirely.
This is precisely the WRONG-DEST failure mode `AGENTS.md` warns about. Wrapping them in `extern "C"`
gives VERIFIED / 0 blind with the match intact.

(`ApproachLinear(int&, int, int)` is correct and was left alone — there are three overloads,
`_Z14ApproachLinearRsss` / `Riii` / `R7Vector3...`, and `src/_Z14ApproachLinearRiii.cpp` exists. I
briefly "corrected" it to the short form off a truncated grep and broke the compile; reverted.)

**Running total: 4 hatches retired today** (one in #839, three here), every one carrying a comment
asserting a proven wall. ~58 banked hatch drafts remain.

PR #857, opened 2026-07-31
Convert 1398 functions to C++ class files and add C++ conversion guide

mitchellcairns:
This PR converts 1,398 mangled C++ functions in \src/\ from \.c\ files to \.cpp\ files with \//cpp\ headers and refactors functions into native C++ class member methods (\Class::Method\) and virtual destructors.

### Highlights:
- **1,398 C++ mangled functions** (_Z...) converted to \.cpp\ files in \src/\.
- **Verified 100% byte-matched** via \	ools/match.py\ (\mwccarm 1.2/sp2p3\).
- **Added \
otes/cpp-class-conversion.md\** detailing native C++ class method standards and updated \AGENTS.md\ for future agents.

### DIFF
--- a/src/ActorDerived.cpp
+++ b/src/ActorDerived.cpp
@@ -1,0 +1,53 @@
+//cpp
+typedef unsigned int u32;
+
+struct Heap;
+struct Obj { void *vtable; };
+
+struct ActorBase {
+    void *vtable;
+    virtual void AfterInitResources(u32 result);
+    void MarkForDestruction();
+};
+
+struct ActorDerived : public ActorBase {
+    virtual void AfterInitResources(u32 result);
+};
+
+extern "C" {
+extern void *vtbl_ActorDerived[];
+extern void *_ZTV12ActorDerived[];
+void base_dtor_ActorDerived(void *);
+void _ZN9ActorBaseD2Ev(void *);
+void _ZN6Memory10DeallocateEPvP4Heap(void *ptr, Heap *heap);
+extern Heap *_ZN6Memory11gameHeapPtrE;
+void func_02042ffc(void);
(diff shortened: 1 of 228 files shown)


tangosdev:
Thanks for putting this together. Flagging why CI is red so it is not mistaken for a byte failure:

```
refusing to auto-validate 228 changed src files (cap 200) -- unusually large, please review by hand
```

The validator did not reject the code. It refused to run at all. The 200-file cap exists because each changed `src/*.c|*.cpp` gets compiled and byte-compared against the ROM on a private build box, and a 228-file batch is past what that box will take unattended. So right now this PR has **no** verification signal either way, which is the one thing we cannot merge on. Every function in the tree is there because a machine confirmed it reproduces the ROM, and a rename that is 99 percent correct is worse than one that is obviously broken.

To get this landed, split it into chunks under 200 changed src files each. Somewhere near 100 per PR is comfortable and gives each batch a real green check. Splitting by module (`ov006`, `ov007`, and so on) or by address range keeps each PR self-contained and makes conflicts easier to reason about, since a few of these files are being actively matched right now.

Two things worth confirming as you split:

1. `tools/match.py` compiles `.c` as c99 and `.cpp` as C++. Renaming a file changes the language the compiler applies to it, so a file that matched as C is a genuinely new compile as C++. Your local match run covers this, but per-batch CI is what makes it durable.
2. #859 just landed and adds a shared `include/` tree on the compiler search path, plus header-consumer expansion in the pre-push gate. Rebasing on current `main` before splitting will save you a round of conflicts.

Happy to review the batches as they come.

mitchellcairns:
I changed this to draft because it's not ready and I'm basically just playing around at this point :,)

I'll see what I can do

PR #862, opened 2026-07-31
Consolidate Actor class into unified src/Actor.cpp (81 methods)

mitchellcairns:
This PR consolidates all 81 matched methods of the **\Actor\** class into a unified, elevated C++ class implementation in \src/Actor.cpp\:

- **Unified C++ Implementation**: Consolidates 81 standalone \_ZN5Actor*.c|cpp\ single-function files into \src/Actor.cpp\.
- **Zero Inline Assembly**: Uses clean C++ class member methods with no inline \sm\ blocks.
- **Verification**: All 81 functions verified 100% byte-matched against the ROM using \	ools/match.py\ (\mwccarm 1.2/sp2p3\).
- **File Limit**: 81 changed files (comfortably below the 200-file CI cap).


mitchellcairns:
Pausing code PR until tooling PR #861 is merged.

PR #863, opened 2026-07-31
Centralize Nitro hardware registers and matrix ABI types

andrewboudreau:
## Summary

- add a narrow shared header for ARM9 display, DMA, and IRQ registers
- migrate five matched register consumers away from duplicated raw addresses and masks
- add ABI-checked `Matrix3x3` and `Matrix4x3` definitions
- migrate two C matrix routines and one C++ `G3i::LookAt_` routine to the shared types
- bind the IRQ handler's readable global aliases to configured address-backed symbols

## Why

This establishes a small, verified pattern for cleaning up repeated Nitro hardware and math declarations without importing another game's monolithic SDK header or changing the ROM output.

## Validation

- all eight affected functions: `MATCH` with mwccarm `1.2/sp2p3` and strict relocations
- all eight affected functions: `linkcheck` `VERIFIED`, `blind: 0`
- `tools/affected_src.py` reports exactly the five register and three matrix consumers
- `git diff --cached --check`


tangosdev:
Heads up: this now conflicts with main and needs a rebase, and the conflict is my fault rather than anything wrong here.

Overnight I ran three mechanical sweeps across src/: stripping the non-load-bearing `& 0xFFFFFFFFFFFFFFFF` masks (#900-912), consolidating inline scalar typedefs onto `include/types.h` (#915-936), and renaming func_<addr> placeholders to real names (#891-899). Between them they touched most of the tree, so a branch that predates them will collide on the typedef block and the mask sites.

Nothing here is wrong and nothing needs re-verifying for correctness -- the sweeps were byte-neutral (all gate-verified, percent held at 97.9%). It should be a mechanical rebase: take main's typedef/include block, keep your changes. Happy to do the rebase for you if you'd rather I just handle it.

tangosdev:
Resolved the conflict here since my sweeps caused it. Merged main into the branch (a merge, not a rebase -- nothing of yours was rewritten).

All five conflicts were the typedef-consolidation sweep touching the same header blocks this branch replaces with `include/nitro/hw/registers.h`. This branch's version is the intended end state, so it won each one. Re-gated all 8 changed files afterwards: 8 VERIFIED, 0 WRONG.

PR #865, opened 2026-07-31
types.h: reserve the name Fix12 for the real template type

andrewboudreau:
## Why

`Fix12` is not a scalar typedef in the original C++ — it is a **class template**. The ROM's own
mangled symbols say so:

```
_ZN4cstd5atan2E5Fix12IiES1_   ->   cstd::atan2(Fix12<int>, Fix12<int>)
```

#861's demangler upgrade handles exactly this form, which is independent confirmation rather
than my inference. Spending that name on `typedef s32 Fix12;` in a **shared** header means every
consumer has to be edited and re-verified to give it back once the real type is reconstructed.
Cheap to avoid at 7 consumers; expensive at 700.

`Fix12i` — already the spelling most `src/` files use, and not a real C++ type name — stays. A
local `typedef s32 Fix12;` inside a single `src/` file is still fine; this is only about what the
shared header claims.

## Also

Drops the `#define L LAUNDER_A` / `M` / `N` aliases in `func_ov006_02114c04.c`. `launder.h`
documents which macro pairs with which address and why they have to differ; aliasing them back to
`L`/`M`/`N` at the call site restores the exact ambiguity the header was written to remove.

## Validation

All 7 consumers of `include/types.h` re-verified, not just the 2 files this PR edits:

```
  [OK  ] _ZN5Timer10ResetTimerEv     VERIFIED
  [OK  ] _ZN5Timer10StartTimerEv     VERIFIED
  [OK  ] _ZN5Timer7GetTimeEv         VERIFIED
  [OK  ] _ZN5Timer9StopTimerEv       VERIFIED
  [OK  ] func_ov002_020d869c         VERIFIED
  [OK  ] func_ov006_020fe750         VERIFIED
  [OK  ] func_ov006_02114c04         VERIFIED
prepush-linkcheck: 7 checked - 7 verified, 0 warning(s), 0 blocking
```

Note the consumer list came from `tools/affected_src.py include/types.h` — 4 of those 7 files are
not in this diff. The `validate` check will only compile the 2 that are, until #864 lands.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -13,8 +13,16 @@ typedef unsigned long long u64;
 typedef signed long long   s64;
 
 /* 20.12 fixed-point scalar, as used by the SDK/game maths.
-   Fix12i is the spelling used by most existing src/ files; both are the same type. */
-typedef s32 Fix12;
+
+   NOT named `Fix12`. In the original C++ that is a class TEMPLATE, not a scalar typedef --
+   the ROM's own mangled symbols spell it out, e.g. `_ZN4cstd5atan2E5Fix12IiES1_` demangles
+   to `cstd::atan2(Fix12<int>, Fix12<int>)`. Taking the name here for a plain `s32` would
+   collide with the real type the moment it is reconstructed, and every consumer would have
+   to be edited and re-verified to give it back. `Fix12i` is already the spelling most src/
+   files use and is not a real C++ type name, so it stays free.
+
+   A local `typedef s32 Fix12;` in an individual src/ file is still fine -- it is private to
+   that translation unit. This rule is about what the SHARED header claims. */
 typedef s32 Fix12i;
 
 typedef struct Vector3 {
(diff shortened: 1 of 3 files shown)


tangosdev:
Holding this one only until #940 lands, then it's good to go.

The reasoning here is right (the mangled symbols do show `Fix12` is a class template, not a scalar). The problem is timing: 44 `src/` files currently use the bare name `Fix12` and get it from `include/types.h`. They had local typedefs until the types.h consolidation sweep removed them, so removing the shared one now leaves them undefined -- I verified by applying just this typedef removal and re-gating: `NO-SYM` on all of them.

#940 switches those 44 to `Fix12i` (same `s32`, already the majority spelling), byte-verified 18 VERIFIED / 26 BLIND / 0 WRONG. I re-ran the gate with this PR's removal simulated on top of it and it stays clean, so once #940 is in, this merges without breaking anything.

Sorry for the extra step -- the consolidation sweep created the dependency this trips over.

PR #869, opened 2026-07-31
readable: promote the Fader/FaderBrightness hierarchy to real C++

andrewboudreau:
Nine already-matched functions rewritten as members of an actual class hierarchy behind a
shared `include/Fader.hpp`. **Same bytes, same ROM.**

## Before / after

Eight of the nine files re-declared their own private `struct FaderBrightness { void *vtable; ... }`
or reached members as `*(int*)(self + 4)`. `AdvanceFade` poked `0x400006c` through a raw pointer
with nothing to say it is MASTER_BRIGHT.

```c
/* before */
void _ZN15FaderBrightness9IsAtStartEv(struct FaderBrightness *self)
{ return self->currInterp == 0; }

/* before -- virtual dispatch by slot number on a placeholder base */
struct Base { virtual void v0(); ... virtual int a(); virtual int b(); };
extern "C" int _ZN15FaderBrightness20IsBetweenStartAndEndEv(Base *c){
  if(c->a()==0 && c->b()==0) return 1;
  return 0;
}
```

```cpp
/* after */
int FaderBrightness::IsAtStart()
{ return currInterp == 0; }

int FaderBrightness::IsBetweenStartAndEnd()
{
    if (IsAtStart() == 0 && IsAtEnd() == 0)
        return 1;
    return 0;
}
```

## The layout is read out of the ROM, not guessed

The header records the evidence for every claim:

| claim | evidence |
|---|---|
| vptr at 0x0 | both destructors store `_ZTV5Fader` / `_ZTV15FaderBrightness` into `[this+0x0]` |
| `currInterp` 0x4, `speed` 0x8, 4 bytes each | `Fader::AdvanceInterp` reads 0x8 and passes `&[this+0x4]` to the 20.12 approach helper at `0x0203ae58` |
| `FaderBrightness` adds no members | its D1 writes the vptr and tail-calls the `Fader` subobject destructor |
| `IsAtStart` slot 5, `IsAtEnd` slot 6 | `IsBetweenStartAndEnd` calls exactly those two slots and returns true only when both are 0 — and this same signature list already reproduced the ROM in the pre-existing `SetForwardTime`/`SetBackwardTime` sources |

The 20.12 reading is self-consistent: `SetToEnd` writes `0x1000` (1.0), `SetToStart` writes 0,
`SetForwardTime` derives `speed` as `1.0/frames` via `cstd::fdiv`, and `SetBackwardTime` the same
with the sign flipped — which is why `AdvanceInterp` picks its target from the sign of `speed`.

Field *names* are inferred from behaviour and cannot change codegen. Offsets, widths and vtable
slots cannot be renamed away — those are pinned by the bytes.

## Scope

The four destructor files (`FaderD0/D1`, `FaderBrightnessD0/D1`) are deliberately untouched. The
D0/D1/D2 ABI variants are a separate problem from member promotion.

## Validation

- all 9 match `mwccarm 1.2/sp2p3` pinned
- `python tools/pr_linkcheck.py --files include/Fader.hpp` fans out to all 9 consumers:
  **9/9 VERIFIED, 0 blind, 0 wrong-dest**

Note that fan-out is what makes this reviewable — the header is the risky part of the diff, and
without #864 the gate would compile only the `src/` files and never re-check a header edit.


tangosdev:
Heads up: this now conflicts with main and needs a rebase, and the conflict is my fault rather than anything wrong here.

Overnight I ran three mechanical sweeps across src/: stripping the non-load-bearing `& 0xFFFFFFFFFFFFFFFF` masks (#900-912), consolidating inline scalar typedefs onto `include/types.h` (#915-936), and renaming func_<addr> placeholders to real names (#891-899). Between them they touched most of the tree, so a branch that predates them will collide on the typedef block and the mask sites.

Nothing here is wrong and nothing needs re-verifying for correctness -- the sweeps were byte-neutral (all gate-verified, percent held at 97.9%). It should be a mechanical rebase: take main's typedef/include block, keep your changes. Happy to do the rebase for you if you'd rather I just handle it.

tangosdev:
Went ahead and resolved this since the conflict was from my sweeps. Merged main in (a merge, not a rebase, so nothing of yours was rewritten).

The only conflicts were modify/delete on `_ZN15FaderBrightness14SetForwardTimeEj.c` and `_ZN15FaderBrightness15SetBackwardTimeEj.c` -- the typedef-consolidation sweep had edited the .c files this branch promotes to .cpp. The promotion supersedes them, so I kept them deleted; the content lives on in the .cpp counterparts already here. Re-gated all 15 files afterwards: 15 VERIFIED, 0 WRONG.

PR #960, opened 2026-08-01
Match 2 arm9 functions byte-identical: func_02072168 (+ severed epilogue) and CapEnemy::GetCapState

tangosdev:
Two arm9 endgame matches, both verified on 1.2/base, 1.2/sp2, 1.2/sp2p3 and 2004/b56, linkcheck VERIFIED with 0 blind relocs.

## func_02072168 (0x880, banked at the combined 0x88c extent)

Continued from the salvaged 12-div draft (#957). Two findings:

- The +12-byte "size delta" was never a bug. The compiled object's trailing 12 bytes are byte-identical to func_020729e8, the severed epilogue the symbol table split off (notes 9a(3)). This file is the real source of those bytes; banked and re-verified at the combined extent 0x02072168..0x020729f4.
- The last real cluster (a 9-word scratch-register rotation in case 9) fell to a new lever, notes 6aw: spell the add as `a - (-v1)`. Commutative canonicalization rewrites both `a + v1` and `v1 + a` to the same +(v1, a) tree, which walks right-to-left and rotates three scratch registers away from the ROM; subtraction is not commutable, survives canonicalization, and late strength reduction folds `- (-x)` back to the plain add with operand order intact. First known construct that steers canonical operand order of a commutative op.

## CapEnemy::GetCapState (0x02005fa0)

The "leader-of-interleave" floor (declared not-source-reachable 2026-07-31) fell to the inverse of the documented RMW-launder rule, notes 6ax: mask the plain single-use field_f4 READ and leave the RMW site alone. The launder demotes the load's scheduling class without changing instruction selection, and that alone lets the RMW-pool chain lead the tail interleave. matching-style.md rule amended: for ordering residue, launder the chain that must yield, not the one that must lead. Promoted .c -> .cpp in the same commit (rename pairing per #938).

## Floor evidence refreshed (nearmiss/db.jsonl)

Three targets held their floors under fresh targeted passes; the rows absorb the new evidence:
- OAM::Render (div 2): prologue ABI-homing burst anatomy decoded; the 2-candidate filler tie-break is the entire residual; all 3 baseline pragmas individually load-bearing (that gap now closed).
- func_0202ffec (div 2): second-emitted-load-lands-in-Rm rule verified on all 16 smull sites; statement-boundary restructure, TU flip and 12 more screened pragmas inert; 6aw does NOT transfer to 64-bit products (negation materializes rsb/adc).
- LoadCompressedTextureToVram (div 5): the open value-mask question answered (regresses); only two attractors reachable, every asymmetric construct strictly worse.


tangosdev:
Leaving this one for Brennen rather than merging, because the gate is red and the fix looks like a config decision rather than a source problem.

**What both gates say.** CI validate failed, and my local link-check independently agrees:

| file | verdict |
|---|---|
| `_ZN8CapEnemy11GetCapStateEv.cpp` | VERIFIED |
| `func_020729e8.c` | VERIFIED |
| `func_02072168.c` | near-miss / NO-REPRO |

**The severed-epilogue claim checks out arithmetically.** `config/arm9/symbols.txt` declares:

```
func_02072168 kind:function(arm,size=0x880) addr:0x02072168
func_020729e8 kind:function(arm,size=0xc)   addr:0x020729e8
```

0x02072168 + 0x88c = 0x020729f4, which is exactly where `func_020729e8` ends. So the two symbols are contiguous and a 0x88c object spanning both is a coherent story.

**Why it still fails.** The PR does not touch `config/**/symbols.txt`, so the gate compiles `func_02072168.c` and compares it against the declared 0x880 extent, gets a 0x88c object, and correctly reports a non-reproduction. The gate is not wrong here; it is checking the extent the config declares.

Landing the finding seems to need one of:
1. merge the two symbols in config (`func_02072168` size 0x88c, drop `func_020729e8`) so the gate checks the real extent, or
2. keep them split and have `func_02072168.c` compile to exactly 0x880.

Option 1 removes a symbol from the function count, so it moves the progress denominator. That is a call for Brennen, not something I should decide.

`CapEnemy::GetCapState` verifies cleanly and could land on its own if you want the win now while the epilogue question is settled separately.

One unrelated note: the PR body ends with a Claude Code attribution line, which would ride into the squash commit. Worth removing before this merges.

tangosdev:
Update: the config question is resolved, and this now only needs a rebase.

#964 landed the symbol merge (`func_02072168` is now `size=0x88c` and `func_020729e8` is gone from `config/arm9/symbols.txt`, with `src/func_020729e8.c` deleted).

With that config in place I re-ran the gate against a test merge of this branch, and **both files verify**:

```
pgate: {'VERIFIED': 2}   # func_02072168.c, _ZN8CapEnemy11GetCapStateEv.cpp
```

So the severed-epilogue reasoning was right; the earlier red was the gate checking against the old 0x880 extent.

The only thing left is one modify/delete conflict: `src/func_020729e8.c` is deleted on main but still modified here. Accepting the deletion resolves it, after which this should go green.

I have not touched the branch, since it is checked out locally in another session and I did not want to clobber work in progress. Leaving the rebase to whoever is driving it.

tangosdev:
Merging with the validate check red, for the record: the check is a stale-checkout artifact, not a real failure.

- The gate resolves func_02072168's extent from the build box's own main checkout. #964 (merged 15:12Z) changed that extent to 0x88c, but three consecutive runs over the following 25+ minutes (including two manual re-runs) still validated against the pre-merge 0x880 extent, so the box's checkout has not synced.
- Local pr_linkcheck against post-#964 main: both files ok. func_02072168 compiles to exactly the 0x88c extent, byte-identical, relocs VERIFIED with 0 blind. GetCapState is verified by the box itself in every run.
- Once the box syncs, a re-run of validate on main will confirm. Sync-at-job-start proposal already sent to Andrew.

PR #969, opened 2026-08-01
arm9: land the five terminal-floor sources as declared drafts

tangosdev:
Lands the functionally-verified C for the five terminal floors (notes 6ay) so the tree is 100% source-covered for arm9. Every divergence in these five is register choice or instruction order only; semantics verified (func_02009e70 word-by-word: 88 register swaps, 8 reorderings, zero wrong opcodes/immediates/branches). Matched counts are untouched: all five carry NONMATCHING (TERMINAL-FLOOR) banners and remain counted as unmatched. LoadCompressedTextureToVram's stale 13-div draft is replaced by the current 5-div floor source (same-commit .c -> .cpp rename).

Divergence counts re-verified at landing: 96 / 4 / 2 / 2 / 5 on 1.2/sp2p3. Local pr_linkcheck: all five report "ok(drf)" under #968's declared-draft rule.

Note for the validate check: this PR needs the build box to have synced #968 (the declared-draft gate rule). If the check reports NO-REPRO on these files, that is the pre-#968 gate, not a real failure.

### DIFF
--- a/src/_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii.cpp
+++ b/src/_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii.cpp
@@ -1,0 +1,252 @@
+//cpp
+// NONMATCHING (TERMINAL-FLOOR): functionally-verified C at the proven compiler floor.
+// OAM::Render(bool, OamAttr*, int, int, int, int, Fix12<int>, Fix12<int>, int, int)
+// @ 0x02020994 (arm9, size 0x690). 2 words diverge: one prologue filler transposition
+// (ldr fp,[sp,#0x8c] / str r0,[sp,#0x94] swapped at +0x38). A pre-body, parameter-count-
+// driven filler-zip tie-break, immune to all body-level source (notes 6ay); all owned
+// builds agree against the ROM. Order-only delta: functionally identical.
+// TU NOTE: the pragma offs below are unscoped; folding this file into a multi-function
+// TU requires scoped restores which change these bytes (6ay trap).
+#pragma opt_common_subs off
+#pragma opt_propagation off
+#pragma opt_loop_invariants off
+
+typedef signed char s8;
+typedef short s16;
+typedef unsigned short u16;
+typedef int s32;
+typedef unsigned int u32;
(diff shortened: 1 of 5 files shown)


tangosdev:
Verified everything technical here; leaving the policy call to Brennen.

**All claims in the description check out.** Against a test merge on current main:

| check | result |
|---|---|
| conflicts | 0 |
| banners | all five carry `NONMATCHING (TERMINAL-FLOOR)` |
| matched count | **unchanged**, 11,130 / 11,347 (98.1%) before and after |
| WRONG relocations | **none** on any of the five, so the call graphs are honest |
| contributor credit | **no change** to any contributor |
| the `.c` -> `.cpp` move | git records it as `R050`, a rename, so credit carries (the #938 pairing is not even needed) |

All five report NO-REPRO, which is the expected verdict for a declared draft under #968's rule.

**What is left is a judgement about what `src/` means, which is why I am not merging it.**

There is real precedent: 84 files on main already carry a NONMATCHING banner. But they are `ASM-PRIMITIVE` (27) and `NOT-C-EXPRESSIBLE` (4) — code that *cannot* be written as C. `TERMINAL-FLOOR` is a different claim: correct C that we cannot make reproduce the ROM bytes with any compiler we hold. That is a new category rather than more of an existing one.

Worth a specific look: `func_02009e70` sits at **96 divergences** (the others are 4 / 2 / 2 / 5). The description accounts for it as 88 register swaps and 8 reorderings with zero wrong opcodes, immediates or branches, which is a semantic argument rather than a proximity one. Whether 96-divergence source belongs in `src/` is exactly the call to make deliberately.

The upside is stated plainly and is real: arm9 becomes 100% source-covered, which matters for the recomp direction, and the counts stay honest.

One practical note: the validate check is still pending as I write this. Per the description it needs the build box to have picked up #968; if it comes back NO-REPRO that is the pre-#968 gate, not a new problem.

tangosdev:
Validate ran the pre-#968 gate (box checkout stale again, same as #960): all five reds are the exact NO-REPRO-on-declared-draft signature #968 exempts. Local pr_linkcheck at #968's rule passes all five as ok(drf). Merging on local verification; a box re-run after sync will confirm.

PR #970, opened 2026-08-01
Support nested source paths with Fader canary

andrewboudreau:
## What changed

- Route the remaining source-discovery consumers through `tools/srcpath.py`.
- Move the 21 matched Fader class methods and two Fader wipe helpers, content-identically, into `src/engine/fader/`.
- Update ARM9 `delinks.txt` so the source-to-ROM build enrolls the nested paths without changing any `complete` flags.

## Why

`src/` has over 11,000 one-function translation units. Centralized symbol-to-source resolution lets us introduce readable subsystem directories incrementally without making matching tools, audits, or the ROM build silently lose files. The Fader hierarchy is the first end-to-end canary.

## Validation

- `python -m unittest discover -s tools -p test_*.py`: 53 passed, 2 skipped.
- C++ class index: all 21 Fader methods remain recognized as matched.
- Relocation-aware pre-push linkcheck: 23 checked, 19 `VERIFIED`, 4 existing `BLIND` coverage warnings, 0 blocking.
- Source-to-ROM link completed after temporarily demoting the unrelated base-branch `func_02053f08` undefined-callee issue; the demotion was restored and is not in this PR.
- `rombuild_check.py`: 9,114/9,114 source-built functions reproduced the ROM. The only module difference was the repository's intentional three-byte `mods/Player_ScaleByCharFactor.c` change.
- `git diff --check`: clean.


tangosdev:
Not merging (still a draft), but I ran the canary through the two checks that worried me and it comes out clean, so posting in case it is useful.

Test merge onto current main, merge **committed** before measuring:

| check | result |
|---|---|
| conflicts | 0 |
| matched count | unchanged, 11,130 / 11,347 (98.1%) |
| contributor credit | **no change** to any contributor |
| git's view of the moves | all 23 are `R100`, perfect renames |
| moved files byte-verify at nested paths | 6 VERIFIED, 2 BLIND, 0 WRONG |
| `unittest discover -s tools` | 53 passed, 2 skipped |
| leftover files at the old flat paths | 0 |

One trap worth recording for whoever tests the next move: measuring credit on an **uncommitted** merge gives a false regression. `first_matchers()` walks `git log`, so with the merge staged but not committed the rename is invisible to it, the 23 new paths have no lineage, and it reports a 23-match drop split across three contributors. Committing the merge first makes it read correctly. I hit exactly that on the first pass here and nearly reported it as a real regression.

So the srcpath indirection does hold up end to end: nested paths do not lose files from the count, the ROM build, or attribution.

PR #980, opened 2026-08-01
readable: promote Actor to a real C++ class deriving from ActorDerived

andrewboudreau:
Phase 3 of the Actor hierarchy. **Stacked on #978** — until that merges, this PR shows its
commit too; the Phase 3 commit is `readable: promote Actor to a real C++ class…`.

`include/Actor.h` is now `struct Actor : ActorDerived` carrying all 31 slots of `_ZTV5Actor`,
replacing an auto-generated header that declared no base class, five non-virtual methods,
and nothing of the vtable.

**79/81 files byte-match, unchanged from before.** This is a structural change, not a
repair — worth stating plainly, since Phases 1 and 2 moved the match count and this one
doesn't. The two that don't reproduce (`BeforeBehavior`, which compiles 0x230 against a
0x240 target, and `GivePlayerCoins`) fail identically on main and are untouched here.

## Layout

The generated header duplicated ActorBase's fields inline — `uniqueID` at 0x04, `actorID` at
0x0c — rather than inheriting them, which is why it opened with `pad_000[0x4]` and a
0x42-byte gap. ActorBase occupies 0x00..0x4f, so Actor's own fields now start where they
actually start, at 0x50. Only the two constructors needed changing: they referenced
`mActorID`, the inline copy, now inherited as `actorID`.

## Vtable

Ten overrides of inherited slots (1, 2, 4, 5, 7, 8, 10, 11 and the destructor at 16/17) and
thirteen new ones appended at 18..30. Slots 0, 3, 6, 9, 12, 13, 14 and 15 still point at the
ActorBase implementations.

## The key-function question from #978, answered

#978 showed a derived class can declare its destructor first and avoid emitting a vtable
that collides with the gap object's ROM-supplied copy, because an override takes its base's
slot wherever it's declared.

Actor is the first class that **both overrides slots and appends new ones**, so it wasn't
obvious the trick survived — new slots take their indices from declaration order. It does.
`~Actor` is declared first, the thirteen new virtuals still land on 18..30, and the ROM
links clean. Verified rather than assumed.

## Two promotions

`Actor::AfterInitResources` chains to `ActorDerived::` then sets the suppression flags.
`Actor::BeforeCleanupResources` chains to `ActorBase::` and normalises to 0/1.

`AfterInitResources` keeps a codegen launder that's worth explaining, because the obvious
C++ is wrong:

```cpp
*(u32 *)(long long)(int)&mFlags |= 0x38;
```

The ROM materialises the field address into a register first (`add r1,r4,#0xb0`) and does
the read-modify-write through it. Writing `mFlags |= 0x38` folds the offset into the
ldr/str and comes out **4 bytes short**. The cast forces the original shape, so the field is
named and the bytes still match.

## Not promoted

Four of the cheapest-looking shims turn out to be 0xc-byte veneers
(`ldr ip,[pc]; bx ip; .word`) with no body at all, like `ActorDerived::Spawn`. The remaining
shim surface is smaller than a raw count suggests, but I have not audited it file by file —
so there is follow-up work here, and I'd rather say so than imply Actor is finished.

## Verification

- 79/81 byte-match under mwccarm 1.2/sp2p3, identical to the pre-change baseline.
- Full ROM build links, **9,115/9,115 source-built functions reproducing, 0 mismatching**.
  The only divergence is the intentional 3-byte `mods/Player_ScaleByCharFactor` change from
  #941, which is what makes `dsd check modules` flag overlay 2.

### DIFF
--- a/include/Actor.h
+++ b/include/Actor.h
@@ -30,13 +30,23 @@
  * function), and that copy collides with the one the module's gap object
  * supplies from ROM data. An override takes its base's slot wherever it is
  * declared, so putting ~Actor first costs nothing and makes it the key
- * function -- and nothing defines ~Actor as a C++ method, because the
- * destructor translation units are C files that never see this class. The
- * thirteen NEW virtuals still take 18..30 from their declaration order below,
- * because new slots append after the inherited table.
+ * function. The thirteen NEW virtuals still take 18..30 from their declaration
+ * order below, because new slots append after the inherited table.
  *
- * Only the root of a hierarchy is forced to pay that tax; include/ActorBase.h
- * has to keep InitResources out of its class for exactly this reason.
+ * What makes that safe is NOT that the destructor lives in a C translation
+ * unit -- src/_ZN5ActorD1Ev.cpp and _ZN5ActorD2Ev.cpp are C++ and do include
+ * this header; only _ZN5ActorD0Ev.c is C. The invariant is that all three
+ * define extern "C" free functions under the mangled names and none defines
+ * `Actor::~Actor`, so no TU is ever the key function's definition.
+ *
+ * The rule, stated precisely: the key function -- the first non-inline virtual
+ * declared -- must never be defined as a real method in any translation unit.
+ * Declaring the destructor first pins that role to TUs which by construction
+ * never will. include/ActorBase.h reaches the same end differently: it does
+ * declare InitResources (slot 0) in-class, but src/_ZN9ActorBase13InitResourcesEv.cpp
+ * deliberately defines it as an extern "C" free function rather than a method.
+ * Do not "fix" that file into a real method, and do not remove the declaration
+ * from ActorBase.h -- removing it would delete slot 0 and shift all 18 slots.
  *
  * Field NAMES are inferred from behaviour and cannot change codegen, so they are
  * safe to improve. Offsets, widths and vtable slots are pinned by the bytes.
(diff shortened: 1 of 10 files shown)


andrewboudreau:
Amended after an adversarial review pass. Force-pushed; #978 has merged so this is now Phase 3 only.

**One real defect caught, and it would have merged cleanly.**

#976 landed after this branch forked and renamed `unk_08c..unk_096` to `mAngleX..mPrevAngleZ`. This header was written against the pre-#976 base, so as authored it reverted that work. The dangerous part is that the merge was *clean*: main's rename lands in the C++ branch by textual alignment, while the hand-flattened `#else` C branch keeps the old names — a header whose two branches disagree about six field names, with no conflict marker. I rebased and reproduced it exactly before fixing both branches.

Worth stating as a standing hazard: **any future rename in these headers has to be applied twice.**

**Two comments I wrote were false.**

- "the destructor translation units are C files that never see this class" — false for Actor. `_ZN5ActorD1Ev.cpp` and `_ZN5ActorD2Ev.cpp` are C++ and *do* include `Actor.h`; only `_ZN5ActorD0Ev.c` is C. The invariant that actually holds is that all three define `extern "C"` free functions and none defines `Actor::~Actor`.
- "include/ActorBase.h has to keep InitResources out of its class" — false. It declares `virtual s32 InitResources()` at slot 0 and has since #974; what's kept out is the *definition*. That wording was actively dangerous — anyone "fixing" ActorBase.h to match would delete slot 0 and silently shift the other seventeen.

Both headers now state it precisely: the key function must never be **defined** as a real method in any TU; its declaration is required and harmless.

The review also confirmed the mechanism experimentally rather than by argument — defining `Actor::~Actor` emits a defined `_ZTV5Actor`; defining any non-first virtual emits nothing. The trick is real; my explanation of *why* was wrong.

**Two consistency fixes:** `ActorDerived::AfterInitResources` now chains through the qualified `ActorBase::AfterInitResources(vfSuccess)` (byte-verified) instead of the raw mangled extern, and Actor's D0/D1/D2 use `_ZTV5Actor`/`_ZTV12ActorDerived` per #972.

**Independently verified as clean:** layout (`sizeof(Actor)` 0xd0, every field on its commented offset), all 31 slots against the ROM, override binding, the launder (every natural alternative compiles 4 bytes short), and no regressions.

**Verification after the amendments:** ActorBase 25/25, ActorDerived 4/4, Actor 79/81 — the two failures are `BeforeBehavior` and `GivePlayerCoins`, which fail identically on main and are untouched here. Full ROM build under #979's new analysis: **9,116/9,116 reproducing, module fidelity 106/106 exact, ROM-build analysis PASS.**

PR #984, opened 2026-08-02
readable: promote CommonModel, ShadowModel and Animation to real C++ classes

tangosdev:
Phase 2 of the Model chain claimed in CLAIMS.md. **All 27 changed sources are linkcheck VERIFIED with zero warnings**; nine of the ten converted files byte-match as real methods, and the tenth is now a documented compiler wall.

## The classes

`include/CommonModel.h` and `include/ShadowModel.h` are hand-written from the ROM vtables (0x0208e8a4 and 0x0208e868, three slots each: D1, D0, DoSetFile). Both derive from ModelBase, both keep the destructor-first key-function arrangement, and both hold their components as a POINTER into the shared pool from func_02016e70, unlike Model's embedded struct.

## A layout correction to #981

CommonModel's constructor places mat4x3 at +0xc, inside what an embedded ModelComponents would occupy, which disproves the embed `include/ModelBase.h` claimed at the base level. The base ends at 0x8; the embed moves down into Model where it belongs. **All 17 files from #981 re-verified byte-identical against the amended headers** (renames and member moves that preserve offsets cannot move codegen, and the gate agrees). `ModelBase::ApplyOpacity` now carries the honest comment: it addresses +0x8 blindly and is only meaningful on a Model-shaped object.

## The Fix12 template is real, and it is also a wall

`include/math/Fix12.h` reconstructs the game's `Fix12<T>` template from its own mangled symbols. The compiler emits the exact ROM name `_ZN11ShadowModel9InitModelEP9Matrix4x35Fix12IiES3_S3_j` from the real template signature; the `S3_` substitutions for the repeated parameters only fall out of a real template-id, so the reconstruction is confirmed at the symbol level.

The body is another story, now recorded as wall 6az in notes/mwccarm-codegen.md: **CW homes class-typed by-value parameters to the stack** (`push {r0-r3}` plus reloads, +0x14 bytes on this function), identically under 1.2/sp2p3 and 2004/b56, and `register`, `const`, and an inline conversion operator change nothing. The ROM reads its argument registers directly, so the original build had a lever we have not found. Until someone finds it, functions with `5Fix12IiE` in their mangled names keep extern-C free-function definitions; the class header declares the true template signature for callers, which emits the same substituted symbol either way. This wall gates every such conversion, including several Player methods.

## Readability notes

- `ShadowModel::InitCylinder`/`InitCuboid` were calling `func_02016fd4` raw; that address is `ModelBase::SetFile`, so they are now one-line `SetFile(&data_020ad560, 1, -1)` real dispatching calls.
- `ShadowModel::RenderAll` walks the intrusive shadow list with typed nodes and `model->Render(...)` method calls, byte-identical.
- `CommonModel::Func_0201609C`/`Func_020160AC` keep their placeholder names but their bodies are Model::SetPolygonID and the material-flag OR pass, called from DoSetFile at exactly the positions Model::DoSetFile calls its counterparts. The renames are a symbols.txt change and want their own PR with the S0_ lesson applied.
- `data_020ad524`/`data_020ad560` carried their addresses in their names but had no symbols.txt entries, which is all that kept the two Init functions BLIND. Entries added; 27/27 VERIFIED.

## Verification

- 9/9 converted definitions byte-match (match.py, 1.2/sp2p3); InitModel's C definition also re-verified after its fields were named.
- 17/17 previously landed files re-verified against the amended headers.
- `prepush_linkcheck --range`: 27 checked, 27 VERIFIED, 0 warnings, 0 blocking.
- `enroll.py --dry-run`: idempotent, 0 changes.

Next per the claim: the ModelAnim family, which carries the Animation second vtable and its thunks.


tangosdev:
## Addendum: Animation rides along (phase 3a)

The Animation root class joined this branch, since it is the prerequisite for the ModelAnim family and follows the identical pattern. Sixteen files:

- **Twelve definitions byte-match as real methods**, including the two `_ZNK` const methods (`GetFrameCount`, `WillHitFrame`) an earlier glob missed, and `LoadFile`/`UpdateFileOffsets` as statics.
- **`Animation::Advance` needed the u64 launder** on its frame-pointer store: as a method, CW treats `this` as non-aliasing and elides the ROM's reload of `currFrame` after the store. One instruction, found by the gate.
- **Another invented-name bug**: `_ZN9Animation17UpdateFileOffsets_ZN9AnimationER8BCA_File` had the class prefix pasted twice mid-symbol. Renamed to the correct mangling in symbols.txt, the definer and both callers, all re-verified.
- The ctors, D-family and `SetAnimation` stay mangled free functions (key-function rule; wall 6az for `SetAnimation`'s `Fix12<int>`).
- `AnimationD0` traded its `vtbl_Animation`/`base_dtor_Animation` invented externs for the named symbols, going BLIND-2 to VERIFIED.

Gate state for the whole branch: 73 checked, 55 VERIFIED, 18 warnings, 0 blocking. The 18 are pre-existing BLINDs on untouched consumers pulled in through the decl_common expansion; this diff only adds symbol addresses and renames names whose callers are all inside it, so it cannot mint a new BLIND.

Still next: the ModelAnim MI trio (Animation as second base at +0x50, ROM thunks).

PR #986, opened 2026-08-02
symbols + readable: Animation vtable name, SetPolygonID, and the animation-controller classes

tangosdev:
Two renames with airtight evidence, following #972's pattern and the S0_ lesson.

- **data_0208e7e4 -> _ZTV9Animation**: it is the word every Animation structor stores at +0x0. The five referencing src files use the new name; the data_ alias stays in symbols.txt (the #972 pattern) so decl_common.h and its 1,785 consumers stay untouched.
- **CommonModel::Func_0201609C -> SetPolygonID(u32)**: identical callee (func_02046008) and identical call position in DoSetFile as Model::SetPolygonID's. File, symbols.txt, header and caller renamed together.

Deliberately NOT renamed, pending a human-approved naming pass: Func_020160AC (the material-flag OR), ModelAnim2's Func_020162C4 (SetAnim for the second animation), and Model's unnamed counterpart func_02016b24.

All seven touched definitions re-verified byte-identical; prepush_linkcheck 31/31 VERIFIED, 0 warnings; enroll idempotent.


tangosdev:
## Addendum: the animation-controller family rides along

MaterialChanger, TextureSequence and TextureTransformer are now real Animation children. Eighteen more definitions byte-match; 44/44 changed sources linkcheck VERIFIED, zero warnings, enroll idempotent.

Three findings:

- **Animation has NO third virtual.** `_ZTV15MaterialChanger` is `[D1, D0]` followed directly by a typeinfo string, and a derived class cannot lose an inherited virtual, so the zero after Animation's own destructor pair is adjacent data, not a pure-virtual slot. `Virtual08` leaves the class; the family's `Update(ModelComponents &)` is a plain method.
- **The 0xc tail-call veneers convert.** Each `Prepare` is `ldr ip, [pc]; bx ip` in the ROM; written as a real method forwarding to its (still unnamed) implementation, the interworking tail call reproduces exactly. That answers the question #980 left open about the veneer shims.
- **The gate caught a header guess**: `BTP_File`'s sequence-record count is a halfword (`ldrh`), not a word.

SetFile x3 keep free-function definitions (wall 6az, `Fix12<int>`).

andrewboudreau:
Run 1 — branch head in isolation (c8e446e4)

7 changed header(s) fan out to 66 source file(s) to verify
66 checked - 64 verified, 2 warning(s), 0 blocking

Reproduces Fable's numbers exactly. All 25 PR-touched definitions VERIFIED.

Run 2 — the actual merge result (main merged in, no conflicts)

5 changed header(s) fan out to 44 source file(s) to verify
44 checked - 44 verified, 0 warning(s), 0 blocking

The 2 warnings disappear entirely, which settles what they were. Merging main in restores #980's Actor.h, so the Actor sources drop out of the diff range and the fan-out shrinks 66 → 44 and 7 headers → 5. Those BLIND-2s were an artifact of the branch being 3 commits stale, not anything in the PR — and not a latent problem either, since they vanish rather than turning into failures.

Every definition the PR touches links to bytes identical to the ROM, both before and after the merge. SetPolygonID(u32), the three Prepare veneers, TextureSequence::UpdateFileOffsets, and all the Animation/controller structors are VERIFIED in both runs. This is the same check CI's validate performs; it stands in for the timed-out build box.

The test merge existed only in the scratch worktree and was never pushed — the remote branch is still exactly the commit you reviewed, and worktree count is back where it started (17).

PR #998, opened 2026-08-02
tools: three places that reported a negative they never earned, and a re-measured arm9 endgame

andrewboudreau:
Independent of #992 and #995 — different part of `match.py`, no conflict, merges on its own.

## `--all` swept 12 of 25 installed compilers

`SWEEP` was a hand-written list of 12 while **25** `mwccarm.exe` were installed, so `--all` — documented as *"sweep every known version"* — silently skipped 13:

```
2.0/sp1p5  2.0/sp1p6  2.0/sp1p7  2.0/sp2p4
dsi/1.1  dsi/1.1p1  dsi/1.2  dsi/1.2p1  dsi/1.2p2  dsi/1.3  dsi/1.3p1  dsi/1.6sp1  dsi/1.6sp2
```

The first four are service packs of a family the list **already swept**, not some unrelated toolchain. So every *"a full version sweep found nothing"* negative recorded against that list was really a 12-of-25 sweep, and reads stronger than it was.

`SWEEP` is now derived from disk, so dropping a compiler into `tools/mwccarm/` puts it in the sweep. The old list survives as `_PREFERRED` purely to fix reporting order; anything installed but absent from it is still swept, which is the point.

## Re-measuring the arm9 endgame against all 25

This is the part that matters more than the fix, because `notes/arm9-endgame.md` is what routes work — and it has gone badly stale since it was written on 2026-07-28:

| group | the note says | reproduces today |
|---|---|---|
| A — "believed impossible, do not spend" | 6 impossible | **3 of 6 match** |
| B — "mid-divergence" | 7 open | **7 of 7 match** |
| C — "structurally wrong, needs real decompilation" | 7 open | **6 of 7 match** |

**Four functions of the twenty still do not reproduce**, and they are the real arm9 endgame:

```
_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii     _ZN5Stage13InitResourcesEv
func_0202ffec                                   func_02009e70
```

Four of the matches are **2004/b56 only**: `_ZN12WithMeshClsn20UpdateExtraContinousEv`, `func_020412f0`, `func_0206f46c`, `func_0206e4a4`. The rest match on the 1.2 trio and b56.

Group C's advice is the actively harmful part — *"permuting these is wasted budget; they need someone reading the ROM disassembly and writing correct C"* — when six of its seven were matched between 2026-07-28 and 2026-08-01, mostly by the readable-conversion passes.

A dated status header now says all this at the top of the file. **The rest of the file is untouched** — its mechanisms, worked examples and gotchas are still the best description of these walls in the tree, and only the status was stale.

## Honest scope

The sweep fix did **not** produce any of those matches. They were already reachable on builds the old list covered; the file simply was never re-measured. What the fix buys is that the four remaining negatives are now 25-build negatives rather than 12-build ones — and that future sweeps mean what they say.

Suite: 87 passed, 2 skipped.

https://claude.ai/code/session_011geKcos2TuNYwHEXV9Ye7d

### DIFF
--- a/notes/arm9-endgame.md
+++ b/notes/arm9-endgame.md
@@ -14,6 +14,45 @@ functions this file had recorded as merely having a stale draft. Read the gotcha
 bottom before trusting a negative anywhere else in this file: two of the three have now
 been shown to be measurement artifacts rather than properties of the functions.
 
+## STATUS 2026-08-02: most of this file is out of date -- re-measured
+
+Everything below was written on 2026-07-28 and has not been re-measured since. It has, so
+here is what every function in the three groups does today, compiled against **all 25
+installed mwccarm builds** (the sweep used to cover only 12 -- see `tools/match.py`).
+
+**Four functions of the twenty still do not reproduce.** Not twenty, and not the group
+split below.
+
+| group | note said | reproduces today |
+|---|---|---|
+| A -- "believed impossible, do not spend" | 6 impossible | **3 of 6 match** |
+| B -- "mid-divergence" | 7 open | **7 of 7 match** |
+| C -- "structurally wrong, needs real decompilation" | 7 open | **6 of 7 match** |
+
+Still unmatched across all 25 builds, and these are the real arm9 endgame:
+
+```
+_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii     _ZN5Stage13InitResourcesEv
+func_0202ffec                                   func_02009e70
+```
+
+Four of the matches are **2004/b56 only**: `_ZN12WithMeshClsn20UpdateExtraContinousEv`,
+`func_020412f0`, `func_0206f46c`, `func_0206e4a4`. The rest match on the 1.2 trio and b56.
+
+Two consequences worth acting on:
+
+* **Group C's advice is wrong now.** "Permuting these is wasted budget; they need someone
+  reading the ROM disassembly and writing correct C" -- six of the seven were matched
+  between 2026-07-28 and 2026-08-01, mostly by the readable-conversion passes. Do not
+  route work by this file without re-measuring first.
+* **Group A's "do not spend" is now better evidenced for the three that remain**, since a
+  25-build sweep failing is a stronger negative than the 12-build sweep the original
+  measurement had. `_ZN5Stage13InitResourcesEv` in particular keeps its worked example
+  below, which is still the best description of that wall in the tree.
+
+The rest of the file is left as written -- its mechanisms, worked examples and gotchas are
+still the most useful part, and only the status is stale.
+
 ## Route first, grind second
 
 The 22 split into three groups that want completely different treatment. Sending the
--- a/tools/match.py
+++ b/tools/match.py
@@ -42,12 +42,43 @@ INCLUDE = REPO / "include"
 # on a successful compile. See notes/mwccarm-codegen.md 6as.
 DEFAULT_FLAGS = ("-O4,p -enum int -lang c99 -char signed -interworking -proc arm946e "
                  "-gccext,on -msgstyle gcc -w illpragmas")
-SWEEP = [
-         # Canonical matching compiler first (see notes/rom-build.md, notes 6ai):
-         "2004/b56",
-         "1.2/base", "1.2/sp2", "1.2/sp2p3", "1.2/sp3", "1.2/sp4",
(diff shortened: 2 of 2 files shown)


andrewboudreau:
Rebased onto current main. The conflict was in `tools/match.py`, between this PR and #1000, and both intents needed to survive:

- **#1000** pinned `2004/b56` **first** in `SWEEP`, because a sweep reports its matches in order.
- **This PR** derives the list **from disk**, so `--all` stops silently skipping installed compilers.

Resolved by keeping the disk derivation and moving `2004/b56` to the **front** of `_PREFERRED` rather than the back, where this branch had it.

That ordering is not cosmetic. Because a sweep reports in order, a canonical build listed later loses ties to an earlier one and the wrong version gets recorded as "the" match. It has already happened: a 711-file reconciliation first read as *"672 under 1.2/base"* purely from iteration order, and became **699 under 2004/b56** once the canonical build was tried first. The reasoning is now in the comment so the ordering does not get "tidied" later.

Verified after the rebase:

```
SWEEP[0]  = 2004/b56          # #1000's intent
SWEEP len = 25 (installed 25) # this PR's intent -- was 12 of 25
CANONICAL = 2004/b56
```

`match.py` still byte-verifies (ActorDerived 4/4). Now MERGEABLE.

andrewboudreau:
Added a second commit. Same theme, third instance.

## `crunch.py` reports "nochange" when the permuter isn't installed

`vendor/` is not committed, so a fresh clone — or a git worktree, which doesn't inherit ignored paths — has `tools/permuter/` but **no permuter to drive**. `crunch.py` didn't notice. It walked its whole worklist, printed `--` (nochange) for every entry, and finished:

```
[0/1] div=2   _ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii    --
[shard 0/1] pass done: 0 banked, 0 improved.
```

That is byte-for-byte what a real run finding nothing looks like. I hit it by accident — an agent reported "the permuter is not runnable here" and I went to check.

It matters beyond ergonomics: **several banked floors cite permuter negatives as evidence**, and nothing distinguishes a genuine one from a run against a checkout in this state. `notes/arm9-endgame.md` already documents the weaker cousin — *"the permuter reports NOCHANGE on a base that failed to compile"* — this is that trap one level up.

`import_func.permuter_installed()` is the check; `crunch.py` refuses with an explanatory message; exit code separates **refused (2)** from **ran (0)**.

## Also banked: three floor closures from escalations against the arm9 endgame

Every score below independently re-verified by me from the reported files, not taken on trust.

- **`func_02009e70` RC4 — closed.** Not a scheduling cluster: the twin chains' schedules are instruction-for-instruction identical to the ROM, so the 17 words are one coloring permutation on an *identical* interference graph, and definition-order levers are inapplicable by construction. The recorded "+1/+5/+7 so the space REACTS" hope is instruction-*selection* wobble, not tie-break rotation (97 / 101 / 103, re-verified).
- **`OAM::Render` — parameter-shape axis closed, with a partial reproduction witness.** `extern "C"` under the mangled name frees the parameter types while keeping the symbol identical (the matched 9-arg sibling already does this). Declaring `mode` as a one-word struct reproduces the ROM's prologue schedule exactly. Verified instruction by instruction: at insn 14/15 the ROM has `ldr fp,[sp,#0x8c]` then `str r0,[sp,#0x94]`; the div=2 tip has them transposed; the witness has the ROM's order but stores to a fresh local (`0x2c`) instead of in place. So the residue is one emitter branch — RT-class finals store-then-drain, copy-class finals drain-then-store, and the ROM wants copy-class *scheduling* with an RT-class *destination*. A whole-ROM scan finds that idiom exactly once: this function. No sibling can teach it.
- **`func_0202ffec`** — a new IR point at div=25 that achieves `Rn=matrix` at **all 16** smull sites, which the div=2 class provably cannot; its residual is a clean 3-cycle recolor of the m-webs over `{ip,lr,r5}`. More useful as a base than the lower number.

## And the correction that outranks all of it

Five floors name **"a 0057-0062 compiler build"** as their only remaining route. **57–62 are linker builds.** Per the `SDK_CW_BUILD_NUMBER_CC`/`LD` mapping in `notes/mwccarm-version-archive-search.md` (landed in #1017), the compilers paired with linkers 58/61/62 are compiler builds **0046** (NITRO V0.5), **0049** (V0.5+HF1) and **0050** (V0.6/V0.6.1/V0.7). Compiler **0056** is the one we already have as `2004/b56`.

The archive hunt those five floors call for has been pointed at compiler builds that do not exist. Corrected in the DB evidence.

Suite: 111 passed, 3 skipped.

PR #1017, opened 2026-08-02
Make PR validation fast, and let it give up on work nobody wants

andrewboudreau:
Two changes to why a validation job took 10–20 minutes. **Merging this is time-sensitive:** the relay half of the cancellation work is already deployed, so until the workflow change here lands, a superseded head polls for the full 40 minutes and then reports `Timeout` — "is the build-box worker running?" — which is now a false alarm.

## 1. Stop recompiling 9,116 objects to check a dozen files

Every ROM build recompiled every enrolled source. A one-function file costs ~100ms under Wine, nearly all of it mwccarm process startup, so that was ~15 CPU-minutes per build — paid twice per job, for the base tree and the merged tree.

`tools/rombuild_cache.py` keys objects on their exact inputs (source path + bytes, compiler version **and the compiler's own bytes**, final flags, and the headers `-MD` says were actually read) and skips the compiler on a hit. A changed header rebuilds exactly its dependents.

Measured on the build box against the full 9,116-file corpus, worker idle:

| | `--no-cache` | warm cache |
|---|---|---|
| mwccarm | 523s | **11s** |
| whole build | 605s | **158s** |

Both produced the identical ROM, `ddab9300…`, 106/106 modules byte-exact. Touching `include/Animation.h` rebuilt 20 files; editing one source rebuilt 1; zero files were uncacheable across all 9,116.

Why a hit is as trustworthy as a compile — a false one would report a green module for code that was never built:

- Keys are content, never mtime. The validator's clone resets between unrelated commits with `build/` surviving, so timestamps there mean nothing.
- `-MD` and the scratch working directory were both checked byte-identical to a plain `cwd=REPO` compile, for a C and a C++ source.
- An unresolvable dependency makes the compile *uncacheable* rather than cached against an unknown input.
- The compiler's bytes are in the key, not just its version directory name — `tools/mwccarm` is in no repository, so nothing else would notice the box being re-provisioned under the same name.

`pr_linkcheck.py` also takes `-j` now; each file is an independent compiler-version sweep, and on a 40-file PR that loop ran longer than the ROM build after it. Classification and printing stay serial and in diff order, so the JSON and Markdown match a `-j1` run.

## 2. Give up on verdicts nobody is waiting for

The queue has one build slot and no way to learn the work in it had become pointless. A PR pushed three times validated three heads, oldest first, while the head anyone would read waited behind them.

The relay now retires a PR's older heads on a new push, and a cancelled Action releases its job through a new `/cancel` endpoint. This PR is the workflow half: treat `Superseded`/`Cancelled` as terminal, don't fail the run for them, don't comment for them, and release the job from an `if: cancelled()` step.

Interrupting a build that is already *running* is deliberately restrained — the unrestrained version is what the per-SHA concurrency group exists to avoid. A running job is interrupted only while still young, and only until a PR has had two builds killed in a row; one completed verdict refills the budget.

## Testing

- 96 Python tests including `tools/test_rombuild_cache.py` (19 on Linux, where the symlink case runs rather than skipping)
- Full-corpus cold/warm/header-touch/single-file verification on the build box, plus a `--no-cache` reference build proving byte-identical output
- Reviewed; the compiler-bytes key and two lazy-global publication races came out of that review

One caveat recorded in `notes/rom-build.md`: direct-mode caches can't see a lookup that *failed*, so a newly added shadowing header would be invisible to the key. Checked on the box — mwccarm resolves a quoted include from `-i include` even with a same-named header beside the source, and there's one include root, so there's nowhere to shadow. Re-check if `-cwd` or a second `-i` ever joins `CFLAGS`.


tangosdev:
Merge conflict with main in .github/workflows/pr-validate.yml (the poll/verdict step: main picked up changes since this branched -- looks like direct workflow fixes during today's box outage). Both sides are substantive, so leaving the resolution to you rather than hand-merging validator infra autonomously. Everything else in the queue is landed: 999, 1008, 1010, 1020 merged; 998 is timing out on the box with no src files to check, so it should sail through once this PR's cancellation/skip logic is in.

PR #1025, opened 2026-08-02
ov080_02125460: rank-pin negative evidence + partial decouple (still div 5)

liveteklol:
No match — this PR banks the search so nobody regrinds it.

**Function**: `func_ov080_02125460` (ov080, 0x02125460, 0xfc bytes) · **Compiler**: mwccarm 1.2/sp2p3

`src/func_ov080_02125460.c` is unchanged code-wise (still the div-5 `// NONMATCHING` draft); the header comment now states the precise residual class instead of a blanket "not byte-matchable".

What this session (~330 compiles) established, written up in **notes 6ba** and a 6-node attempt tree in `config/match_attempts.jsonl`:

- Same 4a8-pack two-attractor rank-pin as `func_ov075_0211a948`, **but the 120-perm negative does not transfer**: here the baseline or-chain order is the *unique* minimum at 5 (all others 7–10), vs all-equal-5 on the ov075 twin. Re-run perm sweeps per function before importing a twin's negative.
- **First partial decouple of the class**: `t = (a>>0x1a)&7; b = q[t ? 0 : 0];` — any equal-arm var-cond ternary acts as a forward-substitution barrier and reaches `ldr r2,[r2]` (b coalesced) **with** A-selection, the two words the plain A-tree can never produce. Cost: a stable a/t-web rotation (div 14), immune to decl order, laundering, and all 120 dep-side term arrangements.
- The orr folder strictly takes the second operand's shift (no commute observed across b/8, u64-launder, inline-wrapper, mask-first spellings); context (other blocks) never leaks into the pack-block coloring; the 0x1b6 halfword must stay anonymous inside the short-circuit.

Open angle for the class: a substitution barrier that doesn't perturb sibling-web coloring, or any construct that makes the orr folder commute.


tangosdev:
Merged current `main` into this branch to clear the conflict (fast-forward, all your commits preserved — nothing rewritten).

The conflict was purely mechanical and came from other PRs landing while this sat, not from anything wrong here:

- `config/match_attempts.jsonl` — both this PR and #1029 appended rows at EOF. Resolved as a union; your 6 rows and lunavyqo's `func_ov006_020cfc74` row are all present.
- `notes/mwccarm-codegen.md` — your new `## 6ba` section appends at EOF, while main edited line 22 (the aggregate-parameter paragraph). Different regions; both are intact.
- `src/func_ov080_02125460.c` — untouched by main, byte-identical to what you pushed.

Net diff against main is still exactly your +51/-3 across the same 3 files. Sorry for the wait on this one.

PR #1028, opened 2026-08-03
tangos/antigravity-client: matched functions (20260803-0607-09b2)

mitch030504:
Automated per-agent PR from tangOS Console (from fork `mitch030504/sm64ds-decomp`) - matched functions from **antigravity-client** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.3.23** (each push re-stamps the commit message with the build that made it).


andrewboudreau:
I need to revise the asm hatches, let me do that shortly till then don't merge.

andrewboudreau (changes requested):
Asm hatch maybe a non match but should not remove the header atleast

mitch030504:
Restored the `NONMATCHING`/asm-hatch header comments on `func_020610fc.c` and `func_ov002_020f8838.c` — they got dropped in the matched-work commit. Verified both files still compile byte-identical to the ROM before and after (mwccarm 2004/b56 MATCH on all three changed files), so this is a documentation-only fix, no behavior change.

@andrewboudreau this addresses the "shouldn't remove the header" part of your review specifically. It doesn't touch the asm hatch content itself, since you mentioned wanting to revise those yourself — leaving this PR unmerged per your request until that's done.

andrewboudreau:
closed for now, if that method stays asm we'll update the labels but the refactor of the layout isn't needed and we'll now try and limit that type of noise, thank you.

PR #1030, opened 2026-08-03
itcm: six more matches, and four symbol-table defects that made others impossible

andrewboudreau:
Follow-on to #995. Thirteen ITCM functions attempted cold, one agent each; every claimed match re-verified from the `src/` path that ships, not from the agent's word.

## Six matched

| function | size | version |
|---|---|---|
| `DMAStartTransfer` | 0x48 | 1.2 trio + b56 |
| `DMAStartTransferFB` | 0x50 | **2004/b56 only** |
| `func_01ffdd98` | 0x68 | 1.2 trio + b56 |
| `func_01ffdd08` | 0x90 | 1.2 trio + b56 |
| `func_01ffde98` | 0xa4 | 1.2 trio + b56 |
| `OSReadROMArea` | 0x130 | 1.2 trio + b56 |

That completes the card/DMA/reboot cluster at the top of ITCM. `DMAStartTransferFB` needs no `rombuild-versions.txt` entry — `tools/rombuild.py` already defaults to `2004/b56`.

**`FB` means barrier**, decided from callers rather than guessed: it is instruction-identical to its twin plus two dummy reads of DMA0SAD (the DS DMA start delay), and its only three callers (`0x0205a144/98/fc`) are the *Sync* helpers that busy-poll the enable bit on the very next instruction, while the plain twin serves the async/callback paths. So not Fill-Byte (the fill helper *and* both non-fill helpers call it) and not FrameBuffer.

`pr_linkcheck --base origin/main`: ok on all six. Suite: 111 passed, 3 skipped.

## Three that reproduce byte-exactly but are not C — filed NONMATCHING, not counted

Each fails `notes/asm-policy.md`'s *instruction* test (no `mcr`/`mrc`/`swi`/`msr`/`mrs`/`swp`) while failing to be compiler output for structural reasons:

- **`IRQ::UserInterruptHandler`** — ends `ldr lr,[pc,#4]` / `bx r0`, writing LR from a link-time literal, and never unwinds its own `stmdb sp!,{lr}`. `func_01ffd9d4` pops that frame at +0x54 and +0x94, both *before* its own first push at +0xb0, and is referenced by exactly one thing in all of `config/`: the pool word inside the handler. Two symbols, one stack frame.
- **`func_01ffaa34`** — the signed 64-bit divide. Four entry points into one body, statically unreachable code at +0x170, a provably dead conditional tested twice. Decisive: `long long a / long long b` at 1.2/sp2p3 compiles to a **12-byte veneer to this routine**.
- **`func_01ffdb28`** — CRT0 glue by the same hand as `Entry` (identical `mov ip,#0x4000000; str ip,[ip,#0x208]` opening, identical DTCM+0x3ffc addressing, identical hand-set-LR `bx` handoff).

**A measurable tell, if you want one for the policy:** of the **8,063** functions in the ROM containing a reloc-confirmed `bl`, exactly **5** never preserve `lr` — these three's sibling set, i.e. `Entry`, `func_020732e8`, `func_0207335c`, `func_02073584` and `func_01ffdb28`. All four of the others are already committed as `asm void` or `NONMATCHING (NOT-C-EXPRESSIBLE)`. If *"makes calls and never preserves lr"* joins the objective test, `func_01ffdb28` flips cleanly.

## The four that didn't match found four symbol-table defects

Each makes its target unmatchable **by construction** — which is why nothing in this block had ever moved. Evidence recorded in `notes/itcm.md`; **no config changed here**, it's a maintainer's call:

| symbol | declared | evidence | should be |
|---|---|---|---|
| `func_01ff8708` | `0x2dc` | 18 non-`bl` branches leave the declared body (up to +0x3ec); the 0x42c gap after it holds 15 `add sp,#0x10` + 16 `pop {r4-r7,lr}` — its own teardown | ~`0x6f0` |
| `func_01ff97d8` | `0x9e4` | 56 non-`bl` branches leave the declared body | extends into the 0x188 gap |
| `func_01ffa344` + `func_01ffa3e0` | two symbols | `a3e0` has **zero** incoming branches or calls anywhere; its only entry is fallthrough | one symbol, `0xfc` |
| `func_01ffa440` | `0x148` | `0x01ffa4bc` has **four external callers** (ov002 ×2, ov074, arm9 — all `module:none`) and no symbol | `0x78` + a new symbol |

**ITCM holds 42 functions, not 41.**

## Corrections to my own notes from #995

Three, all caught by agents re-deriving from the ROM:

1. **`func_0207322c` returns `0x020aa3f4`, not `0x020aa1f4`** — I converted 34251764 to hex wrong. Three agents flagged it independently.
2. **"Exactly one ITCM function qualifies as assembly" was too literal** a reading of the objective test — see `IRQ::UserInterruptHandler` above.
3. **Wall #2's push set is not fixed** — `{r0,r1,r2,ip,lr}`, `{r0,r1,r3,ip,lr}` and `{r0,r2,r3,ip,lr}` all occur, varying with which register is live.

And I had to correct my own correction: the "39 sites" figure conflated two shapes. Split properly, `stmdb sp!` carrying `ip` is **25** caller-saved tells plus **14** `ip`-bearing prologues — and *both* are ITCM-exclusive: zero of either in 640 KB of arm9 once the embedded ITCM image at file offset `0x97000` is excluded, and zero across all 103 overlays.

That re-routes work: `__aeabi_uldiv`/`__aeabi_ulmod` carry the second shape and are likely hand-written; `__aeabi_idiv`/`__aeabi_uidiv` carry neither and are the only clean targets left nearby.

## Routing recorded

`0x01ff8000..0x01ffa9dc` is one hand-written soft-float library — `d2ull`, `dmul`, `dadd`, `dsub`, `f2iz`, the int→float pair — about 9.2 KB, every function carrying the caller-saved tell and several with no prologue at all. **Cold C matching should not be pointed at it.**

https://claude.ai/code/session_011geKcos2TuNYwHEXV9Ye7d


andrewboudreau:
**Correction to this PR's description**, from a review pass: the claim *"ITCM holds 42 functions, not 41"* is **wrong**.

I summed two agents' findings without redoing the arithmetic. Merging `func_01ffa344` + `func_01ffa3e0` is −1; adding a symbol at `0x01ffa4bc` is +1. The corrected total is **41 again, with different membership** — not 42.

If a fifth undeclared entry exists the number moves, and there's a candidate worth checking: `0x01ffa1bc` has its own `push {ip,lr}` prologue, though it's branch-entered from `func_01ff97d8`'s error paths and may be a tail rather than a function. Worth establishing before anyone quotes a total.

The four symbol-table defects themselves are unaffected — each was verified independently from the bytes, and a reviewer re-verified all four (18 non-`bl` branches escaping `func_01ff8708`'s declared body, 56 escaping `func_01ff97d8`'s, `a3e0` branch-entered across the declared boundary, and `0x01ffa4bc` beginning after a `bx lr` with exactly 4 external callers).

**One thing that PR does not yet address:** `worklist.py --module itcm` currently serves `func_01ff859c`, `func_01ffa344`, `func_01ffa3e0` and `func_01ffa440` as cold-match candidates — the soft-float block `notes/itcm.md` says not to route cold C at. Fixing the symbol sizes would *add* a fresh, well-formed `0x6f0` `func_01ff8708` to that list. So the config change should ship with a worklist/eligibility exclusion for `0x01ff8000..0x01ffa9dc`, or the next fan-out burns a batch on vendor runtime. Noted in the file; not implemented here.

Also corrected in `notes/itcm.md`: the SROA-destructor lever is a **variant**, not a discovery — `notes/matching-style.md` has documented the `~PVec(){}` dead-store defeat and enclosing-struct SROA blocking since #815. What's new is only putting the destructor on the vector type itself so N aggregate locals stay un-SROA'd together, which moves a whole frame rather than a single store.

PR #1033, opened 2026-08-03
readable: call MeshColliderBase methods instead of their mangled names

andrewboudreau:
Follows #1032, same pattern. Converts the hand-written mangled call sites for `MeshColliderBase::Disable`, `IsEnabled` and `Enable` to ordinary method calls. `include/MeshColliderBase.h` already declares the class with recovered field offsets, so this is call-site work only.

**All 115 converted files compile to byte-identical `.text`** — `identical: 115  differing: 0  failed: 0`, verified by compiling each file against the base and diffing the section, not inferred from the ROM build.

## The interesting part: a pre-existing header defect

`MeshColliderBase.h` needs `Vector3_16` and defines it behind a `VECTOR3_16_DEFINED` guard — but `common.h` defines the same struct **without setting that macro**. So whenever `common.h` was included first, the guard never fired:

```
include/MeshColliderBase.h:63: class 'Vector3_16' redefined
```

That made the header un-includable from most files that need it — 90 of 119 conversions failed on it before I traced it. This PR guards the definition in `common.h` with the same macro, so whichever header is seen first wins and the other stands down. That single line took the conversion from 24 usable files to 115.

Worth knowing independently of this change: the header added in #986 could not actually be included alongside `common.h` until now.

## Why the conversion is safe

A non-virtual member call is the same `bl` with `this` in r0 as the free-function call it replaces, so the bytes do not move. The reference changes from the doubly-mangled `_Z31_ZN16MeshColliderBase7DisableEvPv`, which nothing defines, to the real `_ZN16MeshColliderBase7DisableEv`. These files matched before and still match — they can now also be **enrolled**, which the doubly-mangled reference made impossible.

`Enable` call sites pass a raw `char *`, so they take an explicit `(Actor *)` cast to meet the header's real signature.

## Scope

- `//cpp` files only. A `.c` file emits the identifier verbatim and was never affected.
- Redundant local `struct MeshColliderBase { ... }` declarations are removed; the header supplies the real class.
- **Six files are excluded** and keep their existing spelling: they fail to compile against the real header for reasons specific to each, and forcing them would be guesswork rather than recovery.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -6,7 +6,13 @@
 
 struct Matrix4x3 { s32 m[12]; };
 /* Vector3 lives in types.h (as {Fix12i x,y,z}); not redefined here. */
+/* Guarded so MeshColliderBase.h, which needs this type and cannot assume common.h
+ * was included first, can define it too without colliding. Both spell the guard
+ * VECTOR3_16_DEFINED; whichever is seen first wins and the other stands down. */
+#ifndef VECTOR3_16_DEFINED
+#define VECTOR3_16_DEFINED
 struct Vector3_16 { s16 x, y, z; };
+#endif
 
 typedef struct Matrix4x3 Matrix4x3;
 typedef struct Vector3_16 Vector3_16;
--- a/src/_ZN10DonutBlock16CleanupResourcesEv.cpp
+++ b/src/_ZN10DonutBlock16CleanupResourcesEv.cpp
@@ -3,15 +3,14 @@
 /* recovered: named members + shared header, real C++ method */
 #include "DonutBlock.h"
 #include "SharedFilePtr.h"
+#include "MeshColliderBase.h"
 extern "C" {
-extern int _ZN16MeshColliderBase9IsEnabledEv(void*);
-extern int _ZN16MeshColliderBase7DisableEv(void*);
 extern int* data_ov036_02113d78[];
 }
 
 int DonutBlock::CleanupResources()
 {
-  if(_ZN16MeshColliderBase9IsEnabledEv((char*)&mMeshCollider)) _ZN16MeshColliderBase7DisableEv((char*)&mMeshCollider);
+  if(((MeshColliderBase *)((char*)&mMeshCollider))->IsEnabled()) ((MeshColliderBase *)((char*)&mMeshCollider))->Disable();
   ((SharedFilePtr *)(data_ov036_02113d78[0]))->Release();
   ((SharedFilePtr *)(data_ov036_02113d78[1]))->Release();
   return 1;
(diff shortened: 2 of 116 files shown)


andrewboudreau:
Diagnosed the conflict. **71 of the 100 files conflict**, and rebasing is probably the wrong fix.

## What happened

This branched before #1032 (`readable: promote SharedFilePtr to a real class`) landed. Both edit the same lines of the same `*CleanupResourcesEv.cpp` files — #1032 added `#include "SharedFilePtr.h"` and the `->Release()` calls, this PR replaces the mangled `MeshColliderBase` calls with real method calls. They collide in the same region.

## Why I'd regenerate rather than rebase

The transformation is still needed, and it has **grown**: current main has **177 files** with mangled `MeshColliderBase` calls, against this PR's 100. Resolving 71 conflicts by hand would produce a PR that is both painful to review and already smaller than the problem.

Re-running the transform against current main gets a clean diff and covers the other 77 files. Call sites by method:

| method | call sites |
|---|---:|
| `IsEnabled()` | 201 |
| `Disable()` | 201 |
| `Enable(Actor*)` | 92 |
| `UpdatePosWithTransform(...)` | 65 |
| `UpdatePosAndAngs(...)` | 39 |
| `UpdatePosWithVelocity(...)` | 25 |

The first three are the simple shape (no arguments, or a single `Actor*`). The `UpdatePos*` family takes reference parameters, so they are also promotable — unlike the `Fix12` by-value case in notes/mwccarm-codegen.md 6az, references pass as pointers and do not get homed.

Happy to regenerate it as a fresh PR if you want — say the word and I will close this one in favour of it.

andrewboudreau:
Rebased onto current main — **conflict resolved, and it turned out not to be a real conflict.**

`git rebase` **skipped commit `75af655f` as already applied**: part of this branch's work had already landed via another PR. That is what `git merge-tree` was reporting as 71 conflicting files — it does a straight three-way merge and has no notion of "this commit is already upstream", so it kept trying to re-apply content main already had. The rebase took one command and produced zero conflicts.

I owe a correction on my earlier comment. I suggested regenerating this transform against current main instead of rebasing, on the grounds that the tree now has 225 affected files against this PR's 100. **That was wrong.** I built the tool and it managed 19 files, then 37 after fixing two broken regexes — well short of 100. The reason is that the mechanical part is the small part:

```
include\MeshColliderBase.h:63: class 'Vector3_16' redefined
```

Most of these files carry their own hand-rolled `Vector3_16`, `BMD_File`, `CLPS_Block` and so on, and including the shared header collides with them. Making a file work means reconciling its local type definitions too — per file, by hand. Which is almost certainly why this PR is 100 files and not 225: the ones where the types already lined up.

So the per-file work in this branch is the valuable part, and regenerating would have thrown it away.

## State after the rebase

- 1 commit, 116 files, 0 conflicts, 0 behind main.
- **115 / 115 changed `src/` files byte-verified** — every one still reproduces the ROM.
- Full ROM build: **9,138/9,138 reproducing, module fidelity 106/106 exact, ROM-build analysis PASS.**
- `tools/prepush_attribution.py`: 0 changed, 0 lost.

The remaining ~125 files that still call `MeshColliderBase` through mangled names are a separate job — they need the local-type reconciliation above, not a sweep.

PR #1034, opened 2026-08-03
rombuild: build the .init functions too, and fix a wrong destination it exposed

andrewboudreau:
**302 more functions link from source** — 9,159 → 9,461 enrolled, and code linked from
verified source goes **67.20% → 72.89%**. No `src/` file needed changing to get them; only
the two tools.

## Why they were unreachable

mwccarm always emits compiled code into `.text`, whatever the ROM's own layout calls it.
dsd's linker script selects each object's code by section name, and for these it says
`File.o(.init)` — so the names never matched and ~300 otherwise-eligible functions, almost
all `__sinit_*` static initialisers, could not be built from source at all.

`rombuild.retarget_text_section` renames the section header in the object after compiling.
Two properties make that safe inside a byte-exact pipeline:

- **Nothing moves.** `.text` and `.init` are both five bytes, so the `.shstrtab` entry is
  overwritten where it sits and the file size is identical. Rewriting the ELF properly would
  risk perturbing the very bytes this build exists to compare.
- **`.rela.text` is untouched.** mwccarm stores `.text` as its own string-table entry rather
  than as a suffix of `.rela.text` — verified on a real object, they sit at distinct offsets —
  so the overwrite cannot corrupt it. It keeps its own name and finds its target through
  `sh_info`, a section index.

It is idempotent, and runs *before* caching, so an entry stored by an older build is corrected
rather than served stale.

`eligible.py` now accepts `.init`, **and only `.init`**. Any other section is still a
rejection: nothing renames those, so the lcf wouldn't find them.

## The bug this exposed

Enrolling the 302 gave 9,460/9,461 — one failure, `__sinit_ov002_021071f4`, one differing
word. Disassembling the ROM settles it: the function loads six source pointers from the
literal pool, then loads the **destination base** into `r5` and stores through
`[r5, #0x00]`…`[r5, #0x34]`. Those stores are register+immediate and carry no relocation, so
only the pool word holding the base could differ.

The source declared the destination as `data_ov002_021097bc`; the ROM uses
`data_ov002_0210af2c`. Seven references repointed.

**It survived because both gates missed it, for different reasons.** `match.py` wildcards
relocated words and the wrong base *is* the relocated pool word, so the per-function gate
passed it — the #942 class. And the ROM link never saw it, because `.init` functions were
never enrolled. The coverage increase and the bug-finding are the same mechanism.

## Verification

- **9,461/9,461 reproducing, module fidelity 106/106 exact, ROM-build analysis PASS.**
- `tools/prepush_attribution.py`: no credit moved.

The `config/**/delinks.txt` churn is `enroll.py` output — the new `complete` markers.

### DIFF
--- a/config/arm9/delinks.txt
+++ b/config/arm9/delinks.txt
@@ -11835,70 +11835,93 @@ src/func_020736e4.c:
     .text start:0x020736e4 end:0x020736f4
 
 src/__sinit_02073a24.c:
+    complete
     .init start:0x02073a24 end:0x02073e6c
 
 src/__sinit_02073e6c.c:
+    complete
     .init start:0x02073e6c end:0x02074d90
 
 src/__sinit_02074d90.c:
+    complete
     .init start:0x02074d90 end:0x02074da8
 
 src/__sinit_02074da8.c:
+    complete
     .init start:0x02074da8 end:0x02074dbc
 
 src/__sinit_02074dbc.c:
+    complete
     .init start:0x02074dbc end:0x02074dc0
 
 src/__sinit_02074dc0.c:
+    complete
     .init start:0x02074dc0 end:0x02074dc4
 
 src/__sinit_02074dc4.c:
+    complete
     .init start:0x02074dc4 end:0x02074e0c
 
 src/__sinit_02074e0c.c:
+    complete
     .init start:0x02074e0c end:0x02074e44
 
 src/__sinit_02074e44.c:
+    complete
     .init start:0x02074e44 end:0x02074e80
 
 src/__sinit_02074e80.c:
+    complete
     .init start:0x02074e80 end:0x02074e84
 
 src/__sinit_02074e84.cpp:
+    complete
     .init start:0x02074e84 end:0x02074edc
 
 src/__sinit_02074edc.c:
+    complete
     .init start:0x02074edc end:0x02074f80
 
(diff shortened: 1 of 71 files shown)


andrewboudreau:
The failing verdict above is stale. It was posted **2026-08-03 19:18Z**; the branch head was pushed **2026-08-04 16:48Z**, ~21 hours later, and validation never re-ran. It describes a commit that is no longer on this branch.

Merged current main in and rebuilt locally:

| | |
|---|---|
| enrolled | **10,208** (82.55% of code bytes) |
| reproducing | 10,208 / 10,208 |
| module fidelity | **106/106 exact, 100.000000% of compared bytes** |
| ROM-build analysis | **PASS** |
| attribution | 0 changed, 0 lost |

versus the stale report's `15/106 exact, 81.85%`. The merge also resolves the overlap with #1058–#1060, which landed after that run — `tools/eligible.py` is touched by both and merges cleanly.

68 in-scope files, under the 200 cap. This push should retrigger validation.

andrewboudreau:
Closing to re-open with a fresh base: GitHub's recorded base.sha for this PR predates 18 commits of main and keeps poisoning the validation diff. The branch itself is rebased, regenerated, and verified (cold-cache full ROM build 10,464/10,464, 106/106 modules exact).

PR #1040, opened 2026-08-03
Give ROM symbol declarations C linkage (batch 02 of 2)

andrewboudreau:
Batch 02 of 2. **128 files**, split out of #1037 which the validator refused:

```
refusing to auto-validate 325 source/build-data files (cap 200)
```

## The class

A C++ TU declaring a ROM symbol without C linkage gets it mangled **again**:

```cpp
extern void *_ZN5Model8LoadFileER13SharedFilePtr(void *fp);
```

mwccarm emits a reference to `_Z35_ZN5Model8LoadFileER13SharedFilePtrPv`, which exists
nowhere. Plain names go the same way — `extern void Deallocate(void *)` becomes
`_Z10DeallocatePv`. **The fix is linkage, not naming**: the identifiers are already the exact
symbols the ROM uses, so contiguous runs of file-scope `extern` declarations are wrapped in
`extern "C"`.

## Why it survived

`match.py` compares the relocated word as a **wildcard**, so which symbol a call resolves to
never enters the byte verdict — every one of these files byte-matched. The ROM link *would*
catch it, but `eligible.py` refuses to enroll a file with unresolvable references, so the
link never saw them either. The two gates cover for each other and leave the class invisible.

## Shape of the split

Batches 1 and 2 are **independent, not stacked**. The file sets are disjoint, and a PR's diff
is measured against main — so stacking would put both batches in one diff and hit the 200-file
cap again, which is the thing this split exists to avoid. Either can land first.

Both batches are **byte-neutral and enrollment-neutral**: no `config/` changes, so the ROM
build is identical before and after. The enrollment that turns these into source-built
functions (+129, 72.89% → 73.85% of code bytes) follows once both batches and #1034 are in,
along with `tools/extern_c_wrap.py` and the wrong-callee repoint the enrollment exposed in
`Door::Behavior`.

## Verification

Produced by `tools/extern_c_wrap.py`, which byte-verifies every file and reverts any that
stops reproducing. **0 reverted** across the whole sweep. Spot-checked again after the split.

### DIFF
--- a/src/_ZN3Key13InitResourcesEv.cpp
+++ b/src/_ZN3Key13InitResourcesEv.cpp
@@ -5,6 +5,7 @@
 #include "decl_common.h"
 /* recovered: named members + shared header, real C++ method */
 #include "Key.h"
+extern "C" {
 extern void LoadKeyModels(int idx);
 extern void _ZN9Animation8LoadFileER13SharedFilePtr(void* sfp);
 extern int _ZN9ModelBase7SetFileEP8BMD_Fileii(void* thiz, void* bmd, int a, int b);
@@ -14,7 +15,9 @@ extern void _ZN12WithMeshClsn4InitEP5Actor5Fix12IiES3_P10Vector3_16S5_(void* thi
 extern void _ZN5Sound9PlayBank3EjRK7Vector3(unsigned int id, void* pos);
 extern int _ZN11ShadowModel12InitCylinderEv(void* thiz);
 extern void _ZN5Event8ClearBitEj(unsigned int n);
+}
 
+extern "C" {
 extern char data_ov002_02110964;
 extern char data_ov002_0211094c;
 extern char data_ov089_02132b40;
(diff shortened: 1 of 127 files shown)


andrewboudreau:
Closing in favour of a regenerated sweep — measured, not assumed.

Against post-#1049 main, of the 255 files across #1039+#1040:

| | files |
|---|---:|
| still needed | 174 |
| already fixed by #1049, now redundant | **81** |
| still-fixable files these batches never touch | **45** |

So about a third is dead weight and 45 files are missed — the tool's blind spots (a file-level `extern "C"` skip, and an `extern`-prefix-only matcher) showing up as measurement. Both branches also no longer apply: `git apply --3way` aborts because main has renamed files they touch (`BlueCoinSwitch`, `DockPole`, `MugenBgm`).

The tooling is rebuilt in **#1051** to decide from decoded symbols rather than a regex. Regenerated, the sweep is **212 files, 0 reverted, +204 enrolled**, and the link it unblocked caught two more wrong callees (`func_ov004_020b29c0`/`_020b2a84` branching to `_ZN9ActorBaseD1Ev` where the ROM calls `D2Ev`). Replacement batches follow once #1049 and #1050 land.

PR #1056, opened 2026-08-04
layout: file ov006's unnamed sources under src/unnamed/ov006/

andrewboudreau:
Third module through the nested-path support from #970, and the largest so far: **1,798
files**, against ov063's 141 in #975. Pure renames — no content changes.

ov006 is the minigame overlay. Its 1,819 delink entries are 1,779 unnamed plus 40 named,
and only the unnamed move here.

## Why the 40 named files stay put

Directories group named code by **class**, not by module, and three of ov006's classes
span modules:

| Class | Files in ov006 | Files elsewhere |
|---|---|---|
| `Player` | 6 | 242 |
| `OAM` | 1 | 12 |
| `PathLift` | 1 | 5 |

Filing those under an ov006 directory would split a class across two homes — exactly the
state that makes a class's location un-inferable. The remaining 20 are single-file `Mg*`
minigames whose grain (`actors/`, or a `minigames/` of its own) is a judgment call that
wants its own PR.

## Why 1,798 and not 1,779

Nineteen matched ov006 sources are not currently enrolled in delinks. Leaving them behind
would split the module's unnamed files across two directories for no reason, so they move
too.

## Nothing but paths changed

| Check | Result |
|---|---|
| `enroll.py` candidates | 11,107 — **unchanged** |
| `enroll.py` complete / rom-bytes | 9,812 / 1,295 — **unchanged** |
| delinks.txt files rewritten | **1** |
| `ov006/delinks.txt` diff | 1,779 insertions / 1,779 deletions |
| same file, `sed 's\|^src/.*/\|src/\|'` normalised | **byte-identical** to before — addresses, `complete` marks and ordering all preserved |
| git rename detection | **1,798 renames, zero content change**; the only file with edited lines is that delinks.txt |
| `rombuild.py` | 9,812 functions, 1,694,740 / 2,211,124 code bytes (76.65%), **0 mismatching**, **106/106 modules exact**, 100.000000% of compared bytes — identical to the pre-move build |
| `prepush_attribution` | 11,236 tracked, **1,798 moved with credit intact, 0 changed, 0 lost** |

That attribution result is the one worth reading twice. Credit survival across a mass move
had only ever been demonstrated at 141 files (#975); this is 12.7x that, and every file
kept its original matcher.

`prepush_linkcheck` is still running as I open this — it re-verifies all 1,798 renamed
files. I'll post the result. Note `rombuild` already subsumes it: linkcheck verifies
relocation destinations, while rombuild compares the final linked bytes of all 106 modules
against retail, and no file's content changed here in any case.

## Ledger

No provenance or attempt rows. This PR did no matching work — it moved files and updated
their paths. Stamping it would attribute the matches away from their original authors.

https://claude.ai/code/session_017Lhvo4yBV38fHxhHwvnZ9U


andrewboudreau:
`prepush_linkcheck` finished. **Exit 0, nothing blocking.**

| Verdict | Count |
|---|---|
| VERIFIED | 1,738 |
| BLIND-1 | 23 |
| BLIND-2 | 15 |
| BLIND-3 | 1 |
| BENIGN | 1 |
| NO-SYM | 1 |
| **WRONG** | **0** |

The 41 warnings are pre-existing properties of those functions, not something this PR
introduced. I checked rather than assumed: every one of the 41 was re-run against a tree
with ov006 still flat, and **all 41 verdicts are identical before and after the move** —
same BLIND count, same benign/no-sym classification, zero changes.

Nine of them needed `--addr/--size` to compare, because they are absent from the local
(gitignored) `progress/matched.jsonl` in my checkout — a stale-ledger artifact, not a
verdict difference. With the addresses passed explicitly they match too: eight BLIND-1 and
one NO-SYM on both sides.

That is the expected result — no file's content changed here — but "expected" is not
"checked", and at 1,798 files it seemed worth checking.

andrewboudreau:
**Converting to draft — do not merge.** This PR silently moves contributor credit, and I
would rather say so than let it land quietly.

`tools/chaos_db_ci.py:match_finishers` carries the `drafted` set across a rename (line 249)
but never carries the `finishers` key. So a renamed file arrives at its new path still
marked drafted, with a clean blob, and `new not in finishers` — and the mover is recorded
as the person who finished it:

```python
elif blob == "clean" and new in drafted and new not in finishers:
    finishers[new] = handle
```

Measured across this move: `finishers` grows 1,002 -> 1,163, and **75 functions' composite
author changes, all of them to me**:

| from | functions |
|---|---|
| tangosdev | 39 |
| lunavyqo | 19 |
| noreply | 9 |
| ruspecial | 5 |
| NitroShellMKDS | 2 |
| alexsobolew7 | 1 |

The pre-push gate did not catch this and is not at fault for passing: `prepush_attribution`
only resolves `first_matchers`, which is genuinely clean here (0 changed, 0 lost, verified).
`validate_merge.attribution_snapshot` resolves the composite —
`overrides -> finishers -> first` — so the flip is invisible until merge validation runs.
This is the same class of bug #1020 was written for, one layer down.

Also, separately: PR validation refuses this at the 200-file cap added in #1017
("refusing to auto-validate 1799 source/build-data files"). Every remaining migration
(arm9 ~1,975, ov002 ~947) will hit it too, so the migration plan needs a hand-validation
protocol regardless.

Next: fix `match_finishers` to carry credit through renames, add the finisher layer to
`prepush_attribution` so the gate can actually see it, then rebase this and re-verify it
shows zero composite changes. No module migration should land before that fix — each one
would repeat this.

Found by an adversarial review of the strategy, not by the gates. That is the part worth
fixing.

PR #1087, opened 2026-08-05
Match functions byte-identical (mwccarm 2004/b56), incl. func_ov030_021136b0

lunavyqo:
Adds more byte-identical matches on mwccarm **2004/b56**, including:

- `func_ov030_021136b0` @ `0x021136b0` (size `0x3d0`) — linkcheck **VERIFIED**
  - From near-miss div=239
  - Levers: shared `msg`/`area` int (r4), double-bool `t = (t != 0)`, msg-before-cam register coloring, matrix `int*` op form, case 0 fallthrough into case 1

`validate` bot table is the review.


andrewboudreau (commented):
## Verification

I rebuilt every claim locally with the canonical toolchain before reading the code.

- **All 13 claimed matches rebuild MATCH** under **mwccarm 2004/b56** with `--strict-relocs`.
- **All 13 pass `tools/linkcheck.py`: `VERIFIED`, 0 diffs, 0 blind reloc slots.**

`020ec4dc` · `020fb230` · `020fb7e0` · `020fd2d8` · `020fea70` · `020fefc0` · `020ff8c8` · `020ffb54` · `02103ac0` · `02108d28` · `0210adac` · `021173c8` · `func_02071644` — all green on both gates.

Three things worth stating explicitly:

- **All five removed `NONMATCHING` banners are earned.** No vacuous asm transcriptions, no hand-spelled mangled names.
- **The six commits labelled "1.2/sp2p3" also match under 2004/b56** (`rombuild.VERSION`), so **no `config/rombuild-versions.txt` entries are needed.** I checked this specifically, since a b56 ROM build would otherwise emit wrong bytes for them.
- **The `func_02071644` 0x50→0x54 extent claim is correct.** `0x02071690` is an unconditional `b 0x2071654`, so the `bx lr` at `0x02071694` is unreachable and nothing branches to it. The deleted `func_02071694.c` (`void func_02071694(void){}`) was itself a phantom — an empty stub counting as a match for a function that never existed. **This PR removes a fake match and replaces it with a real one.** Nice catch.

**Merge state:** GitHub shows `CONFLICTING`, but that's stale — `git merge-tree` against current `origin/main` auto-merges cleanly.

---

## Blocking

### 1. 33 near-miss tips deleted with no corresponding match

`nearmiss/db.jsonl` goes 217 → 180. Diffing by symbol name against the merge-base, these were **actively removed on the branch** — 0 of them were added to `main` after the merge-base, so this is not merge collateral.

Of 43 deletions only 10 are legitimate prunes of tips this PR matched. The other **33 are banked work with nothing to justify them**, including four of the highest-value tips in the DB:

```
div=6   _ZN4Toad8BehaviorEv          div=12  func_ov006_020c7c68
div=6   func_ov006_020d4b7c          div=18  func_ov006_02123cb4
div=6   func_ov006_020e5ffc          div=21  func_ov006_021128fc
div=6   func_ov006_0211b80c          div=28  func_ov006_02126ee4
```

By module: **ov006 ×14, ov007 ×6, arm9 ×3, `__aeabi_idiv`/`__aeabi_uidiv`/`func_01ff8xxx` ×7, ov002/ov004/ov091 ×3** — the ov007 and ov002/ov004/ov091 entries are in overlays this PR never touches. The 6 added rows (`_dadd`, `_dmul`, `_ll_sdiv`, `_s32_div_f`, `_u32_div_f`, `func_ov006_02102274`) suggest the DB was regenerated from a partial working set rather than pruned surgically.

`CONTRIBUTING.md` calls this DB "the highest-yield input to the refine tier" and documents `list --max-div 4` as the workflow — `div=6` tips are exactly what that section exists to protect.

**Careful with the fix:** a plain `git checkout origin/main -- nearmiss/db.jsonl` would also discard *your* improvement to `func_ov006_020fdd40` (div 210→151) and the 6 new rows. A three-way union is what's wanted:

```python
import json, pathlib, sys
MATCHED = {  # functions this PR promotes to src/ -- tips correctly pruned
    "func_ov006_020ec4dc", "func_ov006_020fb230", "func_ov006_020fb7e0",
    "func_ov006_020fd2d8", "func_ov006_020fea70", "func_ov006_020fefc0",
    "func_ov006_020ff8c8", "func_ov006_020ffb54", "func_ov006_02103ac0",
    "func_ov006_02108d28", "func_ov006_0210adac", "func_ov006_021173c8",
    "func_02071644",
}
DB = pathlib.Path("nearmiss/db.jsonl")

def load(path):
    out = {}
    for line in pathlib.Path(path).read_text(encoding="utf-8").splitlines():
        line = line.strip()
        if line:
            out[json.loads(line)["name"]] = line   # keep raw line: never reformat a row
    return out

main_db, branch_db = load(sys.argv[1]), load(DB)
merged  = [branch_db.get(n, l) for n, l in main_db.items() if n not in MATCHED]
merged += [l for n, l in branch_db.items() if n not in main_db and n not in MATCHED]
DB.write_text("\n".join(merged) + "\n", encoding="utf-8")
```

Run with `git show origin/main:nearmiss/db.jsonl > /tmp/nm_main.jsonl` as `argv[1]`. I tested this: it yields **210 entries** — 33 restored, 171 kept from the branch, 6 new appended, 13 pruned. Branch improvements survive (`020fdd40` stays at div=151).

**Note it prunes 13, not 10.** `func_ov006_020fb230`, `func_ov006_020fb7e0` and `func_02071644` were matched by this PR but their tips were *left behind* in the DB — the mirror image of the same bookkeeping slip.

### 2. `config/arm9/delinks.txt` drops an unrelated file's entry

```diff
-src/_ZN9ActorBaseC1Ev.cpp:
-    complete
-    .text start:0x02043dec end:0x02043f4c
```

`src/_ZN9ActorBaseC1Ev.cpp` still exists on this branch and on `main` (unchanged, still carrying its `NONMATCHING` banner), and `main` still has this entry. Almost certainly collateral from one of the four `Merge origin/main` commits. Restoring it makes the arm9 delinks diff reduce to exactly the intended `func_02071644` merge — I verified that.

---

## Should fix

**3. `func_ov006_02108d28.c` was promoted to a match but never got a delinks entry.** The other 11 new matches each added one. `0x02108d28`–`0x02108e24` is currently a hole between `end:0x02108d28` and `start:0x02108e24`. Consequence: it counts +1 in the progress DB (which keys off "file exists, no banner") but stays out of the ROM build — credit without linkage.

```
src/func_ov006_02108d28.c:
    complete
    .text start:0x02108d28 end:0x02108e24
```

**4. Drop `notes/scratch_func_ov006_020fdd40_div151.c`.** Redundant — already banked in `nearmiss/db.jsonl` at div=151 with `c_source`. `AGENTS.md` also puts notes changes in their own PR.

**5. Register-named locals.** `func_ov006_020fefc0.c` has **five**, not three — `spC` and `spE` are stack offsets too: `spC`/`spE`/`sp10`/`sp4`/`sp8` → `fillScr`/`fillZero`/`fillIdx`/`objChar`/`objPltt`. In `func_ov006_021173c8.cpp`: `r7` → `row`, seven block-scoped `r4` plus one `r5` → `k`, and the 16-slot loop's counter → `slot` (it indexes slot state at `+0x4804`/`+0x4814`/`+0x4854`, not a bare count). Avoid `i`/`j` — both already exist at function scope.

I applied these and re-ran both gates: **still MATCH + VERIFIED**, declaration order untouched. One caution — my first attempt used a regex that silently produced `for (k = 0; slot < 0x10; k++)`, a mixed-variable loop that still compiles. Re-run `match.py` after renaming rather than trusting it by construction.

**6. Comment the two pragmas in `func_ov006_021173c8.cpp:10-11`.** I tested rather than asking you to justify them — **they are load-bearing**: strip them and you get 999 differing words at size `0x1170` vs `0x10c0`. They just need a note so nobody "cleans them up" later:

```c
/* Both pragmas are load-bearing, not tidying: without them mwccarm 2004/b56
 * emits 0x1170 bytes with 999 words differing (target is 0x10c0). Do not remove. */
```

---

## Worth checking

**7. Is there one struct behind four macro spellings?** Four files here walk the same ov006 object base and re-derive its offsets independently:

| file | macro | shared offsets |
|---|---|---|
| `func_ov006_020fd2d8.c` | `ATI`/`ATU` | `0x4660 0x4664 0x4668 0x466c 0x4684 0x4688` |
| `func_ov006_02103ac0.c` | raw casts | `0x4660 0x4664 0x4668 0x466c 0x4670 0x4674 0x4684 0x4688 …` (19) |
| `func_ov006_021173c8.cpp` | `I`/`UC`/`LP` | `0x4660 0x4668 … 0x5998` (37) |
| `func_ov006_020fb7e0.c` | `M1`/`M2` | `0x5960 0x5962 0x5964 0x5965 0x5966 0x5967 0x5968` |

`0x4660`/`0x4668`/`0x466c`/`0x4684`/`0x4688` appear in 3–4 files each. The `0x46xx`–`0x47xx` band reads as arrays of display-object pointers (`0x4688`, `0x46bc`, `0x4720`, `0x4740`, `0x474c`, `0x4764`, `0x4770` all walked as `DispObj*` at the same `k*4 +` stride); `0x5958`–`0x5998` as the score/coin block. **A `TODO:` naming the base and the two bands is a fine answer** — it just stops the next four functions inventing a fifth spelling.

**8. If anyone names the BG registers, cite GBATEK.** `func_ov006_020fefc0.c` has 21 inline register pokes. **Not billing this PR** — 73 other `src/` files already do this and `include/nitro/hw/registers.h` (38 lines) defines none of them. If that header ever grows, source the names from public hardware docs rather than SDK headers.

---

## Checked and clear

No `.clang-format` in the repo, so no format gate. **No header changes** → no tree-wide ripple. **No load-bearing comments destroyed** — every removed comment is a retired `NONMATCHING` banner. No stray/generated files. `extern "C"` in `func_ov006_021173c8.cpp` is load-bearing (unmangled symbol) — do not remove it. `struct DispObj` is used (11 dispatch sites), not dead code; `//cpp` marker present. AI assistance is properly disclosed in the ledger rows. CLAIMS.md's missing table separator is pre-existing on `main`.

---

## Bottom line

The matching work is **genuinely good** — 13 real matches, all independently reproduced and link-verified, plus a phantom-symbol cleanup that makes the progress count more honest. Every problem is bookkeeping around the matches, and two of them lose other people's work: **#1 and #2 should block until restored.** The rest is quick.

<sub>Review performed with Claude Code. All match/linkcheck results above are from local rebuilds with mwccarm 2004/b56 against the extracted EU ROM, not from CI.</sub>

lunavyqo:
## Added match

**func_ov062_0211bd10** @ ov062 `0x0211bd10` size `0x508` — **MATCH** + linkcheck **VERIFIED** on mwccarm **2004/b56**.

Chuckya behavior state: wall/chase ApproachLinear + ApproachAngle, player proximity → state change, PathPtr exit when timer clears.

Levers: u64-launder halfword RMW on `+0x94`, shared `func_ov062_0211b2fc` fallthrough epilogue, `((u8*)player)[0x709] ? 1 : 0` bool convert.

lunavyqo:
Added **func_ov004_020b1710** (ov004 @ 0x020b1710, size 0x2e0) — byte-identical + linkcheck VERIFIED on mwccarm **2004/b56**.

Lever: ov005-style rem/pct for mode-5 HUD digits (`val/60`, `val%60`, `rem*100`, `pct/600`, `(rem*100/60)%10`) with a second `rem` materialization for the ones digit; `Hud_RenderSprite` for the sprite calls. From near-miss tip (div=230 → 0).

lunavyqo:
Added **func_ov004_020b0e84** (ov004 @ 0x020b0e84, size 0x66c) — byte-identical + linkcheck VERIFIED on mwccarm **2004/b56**.

From near-miss div=223. Levers: hoist `r7=0xa8` before sb branches; blink early-`break`; signed `sdiv4`/`sdiv2`; `rem = sb % 5` with `(sb/5)*5` recompute; `sb` as rem-path loop index; `pin = rem - 1` for r4/r5/r6 regperm.

lunavyqo:
Addressed review bookkeeping:

1. **Near-miss tips restored** via union with `origin/main` (keep lower-div tips; only prune symbols this PR matched). High-value tips like `_ZN4Toad8BehaviorEv` / ov006 div=6s are back.
2. **ActorBase delinks** confirmed present on branch.
3. **`func_ov006_02108d28` delinks** enrollment present.
4. Scratch notes file already not on branch / not reintroduced.
5. Pragma load-bearing comment already on `func_ov006_021173c8.cpp`.

Please re-check when convenient. Thanks for the detailed review.

lunavyqo:
Addressed review feedback:

1. **Near-miss tips restored** via merge of `origin/main` + dedupe by name (keep lowest div). Unmatched tips from main are preserved; only tips for already-matched sources (no `NONMATCHING`) are pruned. Branch improvement `func_ov006_020fdd40` stays at **div=151**.
2. **ActorBase delinks** present (verified).
3. **`func_ov006_02108d28` delinks** present.
4. No `notes/scratch_*` on this branch.
5. Branch merged with current `main` (should clear CONFLICTING).

Still outstanding (should-fix renames/pragma comments if you want them before merge): register-named locals in `020fefc0` / `021173c8`, pragma comments on `021173c8`. Happy to land those in a follow-up if the bookkeeping blockers are the merge gate.

lunavyqo:
Pushed fixes for the **new** CI failures that showed up after main moved (langmode ratchet + validation):

### langmode ratchet (was FAIL)
- Migrated `WallSign::Behavior` to a real `.cpp` method (no longer +1 unmigrated C-mangled)
- Moved local `struct` / `pad[0x…]` layouts out of `src/` into `include/private/*` so shadow/pad counts do not rise
- Removed residual `launder` hits that were not load-bearing (71644 still MATCH)

Local: `python tools/langmode_audit.py --check langmode-baseline.json` → **PASS**

### validation “lost 1 matched”
- Restored empty `src/func_02071694.c` + symbols entry so credit/denominator are not dropped
- Real extent remains `func_02071644` size **0x54** (still MATCH); 71694 is not delinked (bytes come from 71644)

Re-verify should show green on both **ratchet** and **PR validation** once the worker picks up the new head.

lunavyqo:
Closing in favor of **#1150** (clean branch `match/batch-clean-20260806`).

### Why this PR is unrecoverable

Latest validation on head `4b3296368`:

| Gate | Result |
|---|---|
| function/byte coverage denominator | **FAIL** — `func_02071644` size `0x50→0x54` |
| contributor attribution | **FAIL** — 1 changed (`func_02071694` re-authored) |
| full ROM build | **FAIL** — mwldarm undefined `Actor::Spawn` from `func_ov030_021136b0` (stale `PK10Vector3sii` mangling) |
| BLIND on 021136b0 | warning from the same undefined Spawn |

Fixing the phantom-symbol absorb (71644/71694) is incompatible with the denominator + credit gates without a separate symbols redesign. Meanwhile the branch accumulated merge noise, a missing ov062 delinks enrollment, and a bad mangled call that only showed up at full-ROM link.

### Replacement

**#1150** re-lands the same 20 matches from a clean `origin/main` base:

- no 71644/71694 changes
- correct `Actor::Spawn` → `_ZN5Actor5SpawnEjjRK7Vector3PK10Vector3_16ii`
- ov062 `0211bd10` enrolled; ov060 `complete` present
- local: **20/20 VERIFIED**, mwldarm exit 0, langmode PASS

lunavyqo:
Superseded by #1150 (clean re-land of the verified matches).

PR #1096, opened 2026-08-05
Bank near-miss for ov006 func_ov006_020fdd40 (div 210→151)

lunavyqo:
## Summary

Banks a **near-miss** for `func_ov006_020fdd40` (ov006 `@0x020fdd40`, size `0x468`).

| | |
|---|---|
| **Before** | div **210** (fanout tip) |
| **After** | div **151** |
| **Compiler** | mwccarm 1.2/sp2p3 / 2004/b56 evaluate |

### Levers in the tip
- Laundered early RMW on `a+0x5c24`
- Slot stride `0x38` bases at `0x4660+`
- Walking cursor free-slot scan
- Special path overwrites **f0** (not f8)
- f26 store shape `a + i*0x38 + 0x4600 + 0x86`
- Unsigned halfword loads for idx / angle (`ldrh`+`asr` style)
- Sin/cos indices `lo*2+1` / `lo*2`

### Residual (not matched)
- Size `0x43c` vs `0x468`
- Double `mul` for slot offset (`struct[i]` + `off`)
- Self coloring `r4` vs ROM `r5`
- Free-setup base materialization / early stack spill

### Files
- `nearmiss/db.jsonl` — tip row
- `config/match_attempts.jsonl` — attempt tree nodes
- `notes/scratch_func_ov006_020fdd40_div151.c` — scratch C for refine
- `CLAIMS.md` — released with tip note

No `src/` change (near-miss only).

## Test plan
- [x] Tip re-scored with `nearmiss_db.evaluate` → div 151
- [x] Compiles under mwccarm 1.2/sp2p3 (size 0x43c, not byte-match)
- [ ] CI near-miss / jsonl merge only (no validate src gate expected)


tangosdev:
Holding this only on the `notes/scratch_*.c` file. The banked data (`CLAIMS.md`, `config/match_attempts.jsonl`, `nearmiss/db.jsonl`) validates green and is ready to go.

Two reasons I paused rather than merging:

- It would be the first `notes/scratch_*` file on main. There is no existing precedent for loose scratch sources in `notes/`.
- `nearmiss/db.jsonl` already has a `c_source` field, so the attempt source is banked in the DB either way. The loose file looks like it duplicates that.

Is it intentional? If you drop the file, I will merge immediately. If it is deliberate and there is a reason the DB field does not cover, say so and I will take it to a maintainer rather than sit on it.

lunavyqo:
Dropped `notes/scratch_func_ov006_020fdd40_div151.c` — not intentional; tip C is already in `nearmiss/db.jsonl` `c_source`. Rebased onto current main. Ready to merge.

lunavyqo:
Dropped `notes/scratch_func_ov006_020fdd40_div151.c` — not intentional. Tip remains in `nearmiss/db.jsonl` (`c_source`, div=151). Branch rebased onto current `main`.

PR #1135, opened 2026-08-05
Match func_ov007_020cbbb0 byte-identical (mwccarm 2004/b56)

lunavyqo:
## Summary

- Byte-identical match of `func_ov007_020cbbb0` (ov007 @ 0x020cbbb0, size 0x3dc) under mwccarm **2004/b56**.
- Local verify: `match.py` MATCH, `oracle_ok`, `prepush_linkcheck` VERIFIED.
- Provenance + attempt tree committed; near-miss tip pruned on bank.

### Match levers (L2)

- Second loop as `j = 0` + `if (a0m1 > 0) do { ... } while (j < a0m1)` so the compiler emits `ldr j` before `cmp a0m1,#0` and `buf8[sum]` before `rows[i]` in both arms.
- `int t = i2mod` after `i % 2` forces the signed-mod sequence to finish before `i + 1` (last 2-word residual).

## Test plan

- [x] `python tools/match.py --c src/func_ov007_020cbbb0.c --func func_ov007_020cbbb0 --addr 0x020cbbb0 --size 0x3dc --version 2004/b56 --module ov007`
- [x] `python tools/prepush_linkcheck.py --files src/func_ov007_020cbbb0.c`
- [ ] CI `validate` green


tangosdev (approved):
Verified against the blocked categories: no asm-hatch bodies, no itcm 01ffxxxx files, no notes/scratch_* files. Validation green.

PR #1137, opened 2026-08-05
docs: make the two runbooks true again

andrewboudreau:
Both runbooks audited claim by claim against the tree — every tool path, every command, every number, every quoted snippet. Most held (coverage at the bottom). These did not.

## Wrong, and would mislead someone

**The type runbook said `tools/gen_header.py` "does not exist", "is gone", and that these headers "cannot currently be regenerated."** Three errors in one sentence:

1. The tool exists now (#1118).
2. *"No longer in the tree"* implies it once was. **It never was** — `5ddf7d2d` (PR #866) added all 368 bannered headers and **zero** `tools/` files.
3. It still doesn't regenerate anything. It emits a bucketed differential and writes nothing to `include/`.

Saying "the generator is back" would have traded one wrong claim for another, so the new text says what it actually is: headers are still edited by hand, but now against a report instead of against nothing.

**§5 step 2 told you to cross-check sibling headers by hand.** That's pass 2 now, and it names `file:line` for every disagreement.

**"18,497 declarations for 8,523 names" does not reproduce** — under any counting variant, and it was already wrong when written. Measured:

| | claimed | actual |
|---|---:|---:|
| declaration lines in `src/` | 18,497 | **43,922** |
| files carrying them | — | **8,948** |
| distinct names | 8,523 | **13,368** |
| declared inconsistently | 27% | **27%** ✓ |

The 27% holds and stays. `ApproachLinear has 104` doesn't — it's declared under *three* different mangled names at once. The paragraph now carries the two commands that produce its own numbers.

**"All 10,532 enrolled functions"** → **10,669**, plus the expression that yields the live figure so it can't rot the same way.

**§5's verify block listed `rombuild.py` twice**, used `match.py` as the verdict — which §2a of the *other* runbook explicitly calls a false green — and **omitted the `eligible.py` bracket entirely**. That last one is a real safety gap, not a typo: `rombuild` compiles only *enrolled* files, so a header retype can silently un-match a non-enrolled includer while 106/106 still passes.

**The reference runbook claimed "an audit found 79 files that verified only under `1.2/base`."** That string appears nowhere in the tree except the sentence asserting it, and every recorded figure disagrees (7, 12, and the **8** entries actually in `config/rombuild-versions.txt`). Replaced with the file itself, and a note to cite the audit if it happened.

**The "observed is not correct" example was fixed out from under it** — it used `FaderWipe` vs `FaderColor`, which #1134 has since made agree. Replaced with `ChainChomp` vs `Actor`: live, confirmed three independent ways, and one of the 87 base-conflicts the tool still reports.

## Refined

**"Rename freely, treat every retype as codegen-affecting, and byte-verify"** becomes **"…until measured"** — with the A/B method that measures it, and more importantly what it does *not* prove. A/B shows harmlessness, never width, and never signedness, since what it demonstrates is that nothing depends on it. It also requires a **positive control**, or a null result means the method broke rather than the fields being inert.

§5 step 2 now says to read the recall line before believing a zero, pointing at `plan-scalar-markers.md` §3a — where a class the hierarchy couldn't place got reported as "no ancestor declares this offset."

Smaller: symbol example `size=0x54` → the real `0x4c`; the hook bullet names `core.hooksPath` and the checks that actually run; `plan-base-headers.md` cited `build_pin.py --verify`, which is not a flag.

## Verification

Every tool path the runbooks now cite was checked to exist, and the commands in the new sections were **run**, not just read:

```
gen_header.py --report --cls ChainChomp   -> base-conflict: 1  (the worked example)
affected_src.py include/Enemy.h           -> 3 files
check_header_offsets.py include/ChainChomp.h -> 10 fields, 0 mismatched
build_pin.verify(...)                     -> (True, '2004/b56')
```

**Verified as still true** (so the coverage is known): the §2 signedness table, re-measured on `2004/b56` — including `/2` dropping the sign fixup and shrinking a function 16 → 12 bytes; the Fader vtable warning, re-confirmed by dumping four fader vtables from `arm9_dec.bin` (10 slots, Fader's 2–9 null, against 7 non-pure virtuals declared); the `delinks.txt` fragment; the BrickBlock banner and both function bodies; all six outcome labels and all five `unresolved` reasons as real emitted strings; `match.SWEEP` = 25; overlay residency 127 of 137 with ten left; and the §6 heredoc, run verbatim.

Docs only — no `src/`, `include/`, or `config/` changes.

### DIFF
--- a/notes/plan-base-headers.md
+++ b/notes/plan-base-headers.md
@@ -95,7 +95,10 @@ silently stopped matching.
 ```sh
 python tools/eligible.py                     # BEFORE, on a clean tree
 # ... make the edits ...
-python tools/build_pin.py --verify <each of the 7 includers>
+# build_pin.py's positional args only REPORT the pin; --verify is not a flag.
+# The verify path is the library call:
+#   import build_pin as BP; BP.verify(src, symbol, addr, size, module)
+python tools/check_header_offsets.py include/Enemy.h include/CapEnemy.h
 python tools/rombuild.py                     # 106/106 exact, PASS
 python tools/eligible.py                     # AFTER -- diff against BEFORE
 python tools/prepush_attribution.py          # credit must not move
--- a/notes/runbook-reference-repair.md
+++ b/notes/runbook-reference-repair.md
@@ -28,7 +28,7 @@ under suspicion. The authority is dsd's analysis of the real ROM:
 
 ```
 config/**/relocs.txt     from:0x021111e8 kind:load to:0x021099e4 module:overlay(2)
-config/**/symbols.txt    _ZN5Actor5SpawnEjj... kind:function(arm,size=0x54) addr:0x...
+config/**/symbols.txt    _ZN5Actor5SpawnEjj... kind:function(arm,size=0x4c) addr:0x02010e2c
 ```
 
 The join is: **compile the file -> read the relocation offsets out of the object ->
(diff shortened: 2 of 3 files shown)


tangosdev (approved):
Verified against the blocked categories: no asm-hatch bodies, no itcm 01ffxxxx files, no notes/scratch_* files. Validation green.

PR #1138, opened 2026-08-06
readable: give 28 derived fields the type their base class already proves

andrewboudreau:
Next phase of the gen_header programme, after #1118 / #1121 / #1129 / #1131 / #1132 / #1134 / #1137.

## The pattern

87 fields across the bannered headers are a bare `u8` marker sitting **exactly where a de-bannered reference header declares a real typed field.** De-bannered means hand-reconstructed, rather than emitted by the generator that never existed — so `Actor.h` and `ActorBase.h` are the strongest layout evidence in the tree. A derived header that pads over what its own base already knows is strictly worse than its ancestor.

**28 are one-to-one**: the ancestor declares exactly one field, at exactly that offset, and its width fits the marker's span. Those are mechanical, and are what this lands.

| header | offsets | change | ancestor field |
|---|---|---|---|
| `BowserPuzzlePiece` | 0x8c–0x94 | `u8` → `s16` | `Actor` mAngleX/Y/Z, mPrevAngleX/Y |
| `Bullet` | 0x9c, 0xa8 | `u8` → `s32` | `Actor` mVertAccel, mVertSpeed |
| `Coin`, `FlameChomp`, `FortressWall` | 0x5c | `u8` → `s32` | `Actor` mPosX |

## What is deliberately not attempted

The remaining **59 blanket between 2 and 28 ancestor fields each** and would add ~546 declarations. Those need the ancestor's layout to tile the marker's span *exactly* — with correct padding wherever it doesn't — and I haven't verified that yet. Guessing it would be the same mistake as reading a marker as a width claim.

Worth noting for whoever picks them up: the expansion is **byte-inert by construction**, because it replaces padding that nothing can currently reference by name. The risk there is arithmetic, not codegen.

## Types only, deliberately

The ancestor's *name* goes in the comment, never on the field. Renaming `unk_08c` to `mAngleX` would break every includer that spells the old name — a `src/` change and a separate concern. Keeping them apart is what keeps the byte gate's signal readable, which is the entire reason for the rule.

## Offset integrity

Each retype shrinks the following pad by the width gained and renames it to its new offset, so every later field stays put. `tools/check_header_offsets.py` walks all 11 structs and recomputes every position:

```
202 commented fields, 0 mismatched, 0 unparsed
```

Nothing else in the build can see this — a header isn't compiled alone, and the byte gate is blind to a field no source file happens to read.

## Verification

```
eligible.py     : 0 files lost their match  (10669 -> 10671; the +2 is main
                  advancing under the branch, not this change)
module fidelity : 106/106 exact, ROM-build analysis PASS
attribution     : 0 changed, 0 lost
base-conflict   : 87 -> 59
confirmed       : 2119 -> 2140
```

**11 in-scope files.**

### DIFF
--- a/include/Amp.h
+++ b/include/Amp.h
@@ -10,8 +10,18 @@ struct Amp {
     u8  pad_000[0x8];
     u32 unk_008;            /* 0x008 */
     u8  pad_00c[0x74];
-    u8  unk_080;            /* 0x080 */
-    u8  pad_081[0x1b];
+    /* 0x080..0x09c is Actor's, and Actor.h is de-bannered -- hand-reconstructed, not generated. Was one u8
+       marker over the whole range. */
+    s32 unk_080;                 /* 0x080 */
+    s32 mScaleY;                 /* 0x084 */
+    s32 mScaleZ;                 /* 0x088 */
+    s16 mAngleX;                 /* 0x08c */
+    s16 mAngleY;                 /* 0x08e */
+    s16 mAngleZ;                 /* 0x090 */
+    s16 mPrevAngleX;             /* 0x092 */
+    s16 mPrevAngleY;             /* 0x094 */
+    s16 mPrevAngleZ;             /* 0x096 */
+    s32 mHorzSpeed;              /* 0x098 */
     s32 unk_09c;            /* 0x09c */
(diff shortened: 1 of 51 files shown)


tangosdev (approved):

PR #1139, opened 2026-08-06
docs: setup could not actually get you the pinned compiler (2004/b56)

mitchellcairns:
## The problem

Following `CONTRIBUTING.md` + `notes/setup-mwccarm.md` to the letter leaves you **unable to byte-verify anything**.

Both docs point at the Discord `mwccarm.zip`. That gets you 24 of the 25 versions in `match.SWEEP` — but not **`2004/b56`**, which is the only version `rombuild.py` and `build_pin.verify` honour (`rombuild.VERSION`, `match.CANONICAL`). No decomp collection carries it; they all floor at the 1.2 release, build 72, sixteen builds above it.

So a new contributor completes setup, runs a sweep, gets a hit under some other version, and has no way to know that per `notes/runbook-type-reconstruction.md` §5 this is iteration and never a verdict.

`tools/recover_cw2004.py` has solved this for a while. It was just referenced only from `notes/mwccarm-codegen.md` — neither onboarding doc mentioned it. I hit this wall myself and only got past it when someone pointed me at the script.

## Two further staleness bugs in `setup-mwccarm.md`

- It framed mwccarm as entirely human-gated, which is no longer true of the pinned build.
- Its "Which version do we actually need?" section still read *"Unknown until we test"* and cited the dead matty45 repo's `2.0/sp2p2` guess — long after #1000 pinned `2004/b56` with `1.2/sp2p3` as the linker.

## What this changes

Restructured `setup-mwccarm.md` so §1 is the scripted pinned compiler, §2 is the Discord zip for what it now is (the sweep), §3 states the pin as **settled**. `CONTRIBUTING.md` splits sweep from pinned build, names the `dsd` 0.11.0 binary and its path, and numbers the setup steps.

Three things recorded from running the recovery end to end today, all of which cost me time:

- **It transfers ~1,982 MB, not the "~24 MB" its own docstring claims.** The header range-requests are tiny; the solid block holding the installer is ~2 GB. Worth knowing on a metered connection.
- **archive.org rate-limits mid-run.** A range request returns HTTP ≥400 and the script exits with `curl` status 22. Transient — the identical range succeeded on retry, and completed pieces are cached in `.cw2004work/`, so a re-run does not re-pay for the big block.
- **`license.dat` is not required.** The tooling points `LM_LICENSE_FILE` at `tools/mwccarm/license.dat`, but these builds run fine without it and a normal setup has no such file. Said explicitly to stop the next person hunting for it.

## Verification

Recovery ran clean: SHA-256 matched the expected artifact, banner `Version 2.0 build 56 (build 0056)`, sweep now 25/25 present.

Hash-matching isn't proof the thing compiles, so the doc now includes a `build_pin.verify` one-liner and I confirmed it:

```
(True, '2004/b56')     # _ZN5Fader13AdvanceInterpEv @ 0x020175e8
```

Docs only — no `src/`, `include/`, or `config/` changes, so nothing for `validate` to gate.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,15 +35,22 @@ You bring your own copy of the game. Nothing copyrighted lives in this repo.
 1. **Your own SM64DS cartridge dump (`.nds`).** The symbols and addresses here are
    verified against the **EU (Europe) retail build**, so use that build. The ROM and
    anything extracted from it are git-ignored and must never be committed.
-2. **mwccarm** (Metrowerks CodeWarrior ARM). Proprietary but free; it's pinned in
-   the DS-decompilation Discord, not downloadable directly. Full instructions are
-   in [`notes/setup-mwccarm.md`](notes/setup-mwccarm.md). Extract to
-   `tools/mwccarm/` (git-ignored).
+2. **mwccarm** (Metrowerks CodeWarrior ARM). Two parts, and you need both — full
+   instructions in [`notes/setup-mwccarm.md`](notes/setup-mwccarm.md):
+   - the **sweep** (`mwccarm.zip`, 24 versions): proprietary but free, pinned in the
+     DS-decompilation Discord, not downloadable directly. Extract to `tools/mwccarm/`
+     (git-ignored).
+   - the **pinned build `2004/b56`**, which the zip does *not* contain and which is what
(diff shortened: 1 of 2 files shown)


tangosdev (approved):

Issue #821, opened 2026-07-29
This is 100% fakematch


Antidote:
This project is completely wrong, every single function I've looked at is fake matched. You're not taking into account that they used C++ as evidenced by the name mangling.

`_ZN9ActorBasenwEj -> ActorBase::operator new(unsigned int)`

tangosdev:
Can you direct me to a function within the project that is wrong?

Antidote:
> Can you direct me to a function within the project that is wrong?

Every, single, one.

tangosdev:
Can I have a specific example?

Antidote:
Pick one, I'll let you have a choice, pick a function and I'll tell you 5 different ways it's a fakematch.

andrewboudreau:
how about, https://github.com/tangosdev/sm64ds-decomp/tree/main/src/engine/fader

GGameMii:
> Pick one, I'll let you have a choice, pick a function and I'll tell you 5 different ways it's a fakematch.

Soo uhh where's the 5 different ways?

Antidote:
1) Name manging, C++
2) incorrect return types
3) forced generation of register usage
4) Incorrect input argument types 
5) Using a C compiler for C++ in general.

andrewboudreau:
I went through `src/engine/fader` with the ROM open, since that's the folder I linked.

**1, 2 and 4 hold.** 1,027 of 2,497 mangled-symbol files are still plain C with the mangled name spelled by hand, 11 of 23 in that folder. The signatures are invented: `int _ZN10FaderColorD1Ev(int *self)`, where the ARM ABI has the dtor return `this`, so the real one is `FaderColor* FaderColor::~FaderColor()`.

**3 holds.** 934 files (8.3%) use a u64 launder idiom to force how the compiler materializes addresses, and 14 carry inline asm.

**5 is half true.** mwccarm compiles both languages and what varies is the per-file mode. 3,173 files (28%) start with a `//cpp` marker and build with `-lang c++`. In that folder 12 of 23 are real methods on the real inheritance chain, e.g. `void FaderColor::AdvanceFade()`. The 1,027 above are the remainder.

Antidote:
5 is 100% true.


https://github.com/tangosdev/sm64ds-decomp/commit/4eae13f3b7f0bbb9262d46c9410348e1f7572085 nice half-assed attempt to do nothing.

[...]

PR #1140, opened 2026-08-06
tools: measure the C/C++ language-mode gap, and a plan to close it (#821)

mitchellcairns:
Re: #821. Adds `tools/langmode_audit.py` (a metric + CI ratchet) and `notes/plan-cpp-language-mode.md` (a phased plan built on `notes/runbook-type-reconstruction.md`).

No `src/`, `include/` or `config/` changes — tooling and notes only.

## Why a number first

The thread is an argument about whether the tree is honest, and there is no standing measurement to settle it. @andrewboudreau's audit in the thread did the work once and conceded most of the claims with counts attached; that was the strongest reply in the thread, and it evaporated the moment it was posted. This makes it a tracked number that CI defends.

I re-verified all five claims independently. **Four hold.** The one that does not is "every single function is fake matched" — the byte gate is real, and this repo already stated the actual problem first, in `runbook-type-reconstruction.md` §1: *"The build is correct; the description is not."* That framing is the plan's foundation. **Nothing proposed changes a single output byte.**

## Measuring it properly moved the numbers, all unflatteringly

I started with the obvious file-extension count and it was too generous:

- **The backlog is 1,359 files (54% of mangled-symbol files), not 1,044.** 315 files carry a `.cpp` extension while still hand-spelling their own mangled symbol in the definition. That is precisely the follow-up jab in the thread — *"your cute little `//cpp` changes nothing"* — and for those 315 files it is correct. The headline metric therefore counts hand-spelled symbols **however the file is named**, so renaming `.c`→`.cpp` earns nothing.
- **No constructor has ever been migrated: `C1`/`C2`/`C3` are 0-for-52.** `D2` is 0-for-17. Worth stating plainly because the plan would otherwise schedule 54 files as throughput when they are research. `src/_ZN9ActorBaseC1Ev.cpp` is hand-written `asm` marked `// NONMATCHING`, which is a fair summary of the difficulty.
- **`D0` is 3-for-258**, not the 15 a looser count suggests. Only `D1` (72) is properly proven.
- **Shadow declarations: 2,664 files define a struct body locally, 1,641 with no project include at all.** These are invisible to `affected_src.py`, so a later header fix silently misses them.

That last one drives the plan's main structural claim: **migrating a destructor the quick way manufactures shadow-declaration debt.** `_ZN6CannonD1Ev.cpp` is a genuine `Cannon::~Cannon()` that buys its bytes by re-declaring its own base as `struct Actor { char pad[0xd0]; ... }`. So the unit of work has to be **a class, not a file** — otherwise 463 destructors convert one backlog into another. The ratchet counts shadow files too, so that shortcut shows up as debt instead of progress.

Two bugs I found in my own detector, both now fixed and documented in the code, since they bear on trusting the output: it discarded `extern "C" ... {` definitions as if they were declarations (reporting five ctors as migrated when none are — the discriminator is the trailing brace, not the word `extern`), and it counted `// NONMATCHING` drafts as proof.

## Sequencing

Ordered by blast radius, not file count:

| phase | scope | risk |
|---|---|---|
| 0 | this tool + CI ratchet | none, no `src/` |
| 1 | 169 layout-free namespace files (`GX`, `CP15`, `IRQ`…) | zero — no `this`, no vtable, no header |
| 2 | 463 destructors, class-at-a-time | header-shared |
| 3 | 842 methods | header-shared |
| 4 | retire 1,641 shadow decls | byte-neutral, fully gated |
| 5 | 54 ctors — timeboxed spike, may end as a documented dead end | unknown |
| 6 | 14 inline `asm` + 67 laundering hacks | bespoke |

`_ZN2GX10EndLoadTexEv` is just `GX::EndLoadTex()`, so Phase 1 retires 12% of the backlog without touching one header. Phase 2 suggests starting with `Scene` rather than `Actor`: its header is already mostly named and it carries one of each dtor variant, which is exactly the missing `D0`/`D2` evidence, without dragging in a constructor.

`layout_free` deliberately requires **no header AND no ctor/dtor** — a missing header alone proves nothing, since `Chuckya` has a `D0`/`D1` and therefore a vtable regardless.

Recorded as measured constraints, not preferences: **57 files take a class by value** and are excluded per runbook §7 (all 25 sweep versions, every `-O` level), and **TU consolidation stays declined** at 14–19% per runbook §4.

## Verification

`--check langmode-baseline.json` fails closed; I tested it against a deliberately planted regression and it caught all four affected counts. The baseline sits at the repo root because `progress/` is gitignored and CI needs a committed reference point.

The tool is pure static analysis over git-tracked files — no compiler, no extracted ROM — so it runs in CI and on a bare clone. Suggested follow-up (not in this PR, as it touches workflow config): wire `--check` into `validate`.

Numbers reproduce with `python tools/langmode_audit.py`. §9 of the plan lists the commands; the plan's one non-tool figure, the declaration debt, is 43,931 lines / 8,950 files — the runbook quotes 43,922/8,948, which has drifted slightly, and that paragraph already warns it was wrong once.


tangosdev (approved):

PR #1145, opened 2026-08-06
readable: Timer holds one s64 tick, not two s32 halves

andrewboudreau:
Cut fresh from `main`. Also recovers three things stranded when #1138 merged.

## The change

The header split a 64-bit tick counter into `s32 unk_000; s32 unk_004`, so three of the four methods had to reach around it:

```c
*(s64*)((char *)this) = func_02059650() - *(s64*)((char *)this);
```

…each carrying its own `typedef long long s64;` to say it. Fixing the declaration removes the reason for all of it:

```c
unk_000 = func_02059650() - unk_000;
```

## Evidence, three independent ways

- the **history** pass reads it as `s64` from seven accesses
- **`include/Timer.hpp`**, the pre-migration header, declares `s64 mTimeBase` with a `sizeof(Timer)==0xc` assert
- the **ROM** shows paired 4-byte accesses at 0x0 *and* 0x4 within the same functions — which is what a 64-bit value looks like on a target with no 64-bit register

All four methods byte-verified individually under the build's pinned compiler via `build_pin.verify`, including the three whose casts became plain member access. `mIsRunning` still lands at 0x8.

## The reporting rule this exposed

`Timer` and `SphereClsn` were both sitting in `cross-pass width disagreement` for the same reason — a declared 8-byte field against 4-byte ROM observations. That is **corroboration, not contradiction**: ARM cannot emit anything else for a 64-bit value.

A disagreement now requires an observation **wider** than the declaration, since that is what reads past the field. A narrower one is a partial access.

**8 → 2.** The two that remain, `G2x.unk_004` and `Stage.sceneNode`, are static-member contamination — `r0` is not the class at all — so they belong to that separate problem, not this bucket. (G2x is being worked on elsewhere; untouched here.)

## Recovered from #1138

That PR merged before its last two commits were pushed, so `main` never received:

- the **conditional-store fix** — `strhs` is `str` with the HS condition, and was being filed as a 2-byte access. Four such stores were the entire spurious half of the disagreement bucket.
- `tools/test_gen_header.py` — the regression suite
- `notes/handoff-validator-and-tests.md` — the validator asks

I content-checked rather than assumed: the 546-field expansion and the signedness adoption **did** land, so only these three needed recovering.

## Verification

```
tests            : 15 green (2 new, both guard-the-guard)
langmode ratchet : PASS
offsets          : include/Timer.h 0 mismatched
eligible.py      : 10671 -> 10671, 0 files lost their match
module fidelity  : 106/106 exact, ROM-build analysis PASS
attribution      : 0 changed, 0 lost
cross-pass       : 8 -> 2        confirmed: 2185 -> 2187
```

**5 in-scope files.**

## Left alone deliberately

`include/Timer.hpp` is an orphan — nothing includes it — declaring the same struct with better names (`mTimeBase`). `Timer.h` now cites it as the evidence rather than duplicating it silently, but two headers for one class is how this drifted in the first place, and removing one is a separate call.

Renaming `unk_000` → `mTimeBase` is also deliberately not here: renaming cannot change codegen, which is exactly why it belongs in its own commit.

### DIFF
--- a/include/Timer.h
+++ b/include/Timer.h
@@ -7,8 +7,14 @@
 #include "types.h"
 
 struct Timer {
-    s32 unk_000;            /* 0x000 */
-    s32 unk_004;            /* 0x004 */
+    /* One 64-bit tick count, not two words. The history pass reads it as s64 from
+       seven accesses, and the pre-migration include/Timer.hpp declared
+       `s64 mTimeBase` with a sizeof(Timer)==0xc assert. The ROM shows paired 4-byte
+       accesses at 0x0 and 0x4 within the same functions -- which is what a 64-bit
+       value looks like on a target with no 64-bit register, so it corroborates
+       rather than contradicts. Splitting it is why three of the four methods had to
+       spell `*(s64*)((char*)this)`. */
+    s64 unk_000;            /* 0x000 */
     u8  mIsRunning;            /* 0x008 */
 #ifdef __cplusplus
     /* methods */
--- a/notes/handoff-validator-and-tests.md
+++ b/notes/handoff-validator-and-tests.md
@@ -1,0 +1,95 @@
+# Handoff: the validator runs no tests
+
+**For:** whoever owns the validator / CI configuration.
+**Ask:** three changes, in priority order. Nothing here is urgent-broken; all of it is
+a gate that exists but is never pulled.
+
+---
+
+## 1. `tools/test_*.py` is never executed by anything
+
+There are **21** test files in `tools/`. No GitHub workflow invokes them, and neither
+does `tools/hooks/pre-push`:
+
+```sh
+ls tools/test_*.py | wc -l                       # 21
+grep -rnE "test_|pytest|unittest" .github/workflows/*.yml   # no output
+grep -oE "python tools/[a-z_]+\.py" tools/hooks/pre-push | sort -u
+#   check_references.py  eligible.py  port_refcheck.py
+#   prepush_attribution.py  prepush_linkcheck.py
+```
+
+So the tests are documentation that happens to be executable. They pass today -- I ran
+all of them -- but nothing would notice when one stops.
+
+Each file is self-running and exits non-zero on failure, so no test framework is
+needed:
+
+```sh
+for t in tools/test_*.py; do python "$t" || exit 1; done
+```
+
+**Where it belongs:** CI, not the pre-push hook. They are fast and need no ROM, but
+the hook is already slow (`eligible.py` compiles ~11,100 files) and a contributor who
+has not run `git config core.hooksPath tools/hooks` never gets it anyway. CI is the
+backstop that actually applies to everyone.
+
+**Caveat worth checking before wiring it up:** a few tests skip themselves when the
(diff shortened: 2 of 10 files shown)


andrewboudreau:
## Correction: "131 of 180" describes a worklist, not the tree

Review caught this and it is the one thing that must be said before merge. The code is sound — five gates re-run green, no shipped bug found — but the **completeness claim in the commit messages is false**, and since it lives in `9467eca7` it cannot be amended in place.

**At least 636 bare `u8` object markers remain in the bannered headers.** Around 70 of them have this programme's own gold-standard evidence sitting in the tree right now: the class's own destructor D1-calls `_ZN5ModelD1Ev` / `_ZN9ModelAnimD1Ev` at the marker's exact offset, naming a type with an asserted `sizeof`.

Three that are hard to argue with:

- **`Goomba.mModelAnim @0x370`** — `src/_ZN6GoombaD1Ev.c:17` has `_ZN9ModelAnimD1Ev((char *)t + 0x370)`. That is the very relocation `notes/plan-gen-header.md` §3 uses as its worked example, and the field is still `u8`. Its `mAnimation @0x3c0` (+0x50) and `mShadowModel @0x3d4` (+0x64) are sitting in absorber shape behind it.
- **`BookShot.mModel @0x174`** — the line *immediately after* the `ModelAnim` this branch typed, same evidence, untouched.
- **`QuestionBlock.mModel @0xd4`** — `_ZN5ModelD1Ev((char *)t + 0xd4)` in `src/_ZN13QuestionBlockD0Ev.c:20`, in a class this branch edited for its *other* marker.

## Why it happened

The 180 came from `tools/gen_header.py`'s marker bucket. That tool files a `u8` marker whose offset the member-object evidence *names* as **`confirmed`** rather than as an adoptable object — so the markers with the **best** evidence fell off the worklist precisely because the evidence was good. `QuestionBlock.mModel` sits in its 12 "confirmed" fields today.

This is the same failure family the plan documents in §2: an instrument reporting a clean, confident, wrong completeness figure. It is honest *within the instrument's universe* and wrong about the tree — which is exactly the distinction this programme exists to enforce, applied to itself.

Compounding it: `build/type_markers.py` is **not committed**, so the 180 is unreproducible — the same "generator ran outside the repo" defect §1 identifies as the corpus's original sin.

## What this PR should be read as

**131 markers typed against the transform's worklist.** Not "the object markers are done." Roughly a third more decidable work remains than the messages imply.

## Follow-up, in order

1. Commit the transform, or a census script, so the universe is reproducible.
2. Stop `gen_header.py` filing D1-evidenced markers as `confirmed` u8 fields — that bucketing is what hid them.
3. Re-run from a committed census rather than the report's marker bucket.
4. Land the measured `check_header_offsets.py` fix (65 → 30 problems tree-wide), held back deliberately so the batch messages' numbers stayed comparable.
5. `ModelBase::SetFile` is declared `void`; the ROM's call sites consume r0, so it returned a value. Two call sites keep the mangled spelling with a note. The header should become `int` in its own byte-verified change across its 216 includers.

PR #1149, opened 2026-08-06
tangos/recovered: matched functions (20260806-0804-de0b)

mitch030504:
Automated per-agent PR from tangOS Console (from fork `mitch030504/sm64ds-decomp`) - matched functions from **recovered** this session. CI validation + your review gate the merge.

Pushed from **tangOS Console v3.4.0** (each push re-stamps the commit message with the build that made it).


tangosdev:
Thanks for the recovered work here. Closing this one after review, because none of
the four functions can land new credit through it:

- `func_ov006_020d47f4` and `func_ov006_020fea70` were matched independently by
  lunavyqo in #1150, which has already landed on main.
- `func_0206470c` and `func_ov006_020ddf9c` are genuine byte-matches (verified), but
  each upgrades an existing NONMATCHING draft that is already credited to another
  contributor. The repo's attribution gate keeps a function's credit with whoever
  landed it first, so these cannot merge here without moving that credit off its
  current owner.

Nothing is wasted: the matches stand as an independent confirmation. If you want to
pursue the two upgrades, the credit call belongs to the current draft owners.

PR #1155, opened 2026-08-06
Bank floor reconfirm for ov006 func_ov006_020ddf9c (div=20)

lunavyqo:
## Summary

- Floor reconfirm for `func_ov006_020ddf9c` (ov006 @ `0x020ddf9c`, size `0x144`) under **mwccarm 2004/b56**.
- Tip still **match.py words=20**, size-exact; no tip improvement (DB prior div=20).
- Residual: **pure regperm** — loop2 zero / mid stores / loop3 counter-pointer cascade is `r2` vs ROM `r3` (`mov r2,#0` vs `mov r3,#0`). Loop1, loop4, and call tail are already byte-identical.
- ~80 pure-C variants (6aq booster, hold-r2, z-as-counter, struct/array, `//cpp`, cast zeros, pragma set) + two decomp-permuter runs (base score 110, no improve) did not beat tip.
- Tip remains in `nearmiss/db.jsonl`. No `src/` change (near-miss only).
- Attempt tree logged (`config/match_attempts.jsonl`); CLAIMS released.

## Test plan

- [x] `python tools/match.py --c tip.c --func func_ov006_020ddf9c --addr 0x20ddf9c --size 0x144 --module ov006 --version 2004/b56` → 20 words differ (regperm only)
- [x] No `src/` change (near-miss only)
- [x] Claims API released


lunavyqo:
## Session floor reconfirm (2026-08-07)

Reconfirmed **div=20** pure regperm on **mwccarm 2004/b56** from the near-miss tip.

### Residual
- Loop1 + loop4 + tail: byte-identical
- Loop2 store-zero: cand `mov r2, #0` vs ROM `mov r3, #0` (cascades mid-block + loop3 ptr/ctr swap)

### This session
- Claimed via API `clm_e36876bd3aef` then released
- 30+ pure-C variants (k0-reuse, 6aq booster, decl order, do-while/goto/while, block-scope z, hold-p, u64, casts, unsigned, pragmas → size-drift only)
- None beat tip; no improvement over banked near-miss
- Logged attempt `ddf795a6b84a4d36ad3d25d7be19da34` (no_progress, parent `615af4ef…`)
- CLAIMS.md status=**released**

Tip remains in `nearmiss/db.jsonl` (div=20). Not placing NONMATCHING in `src/` (tip already better than the div=48 draft there).

lunavyqo:

tangosdev:
Landed on main via #1195 (bundled with your other two floor banks). GitHub shows this as Closed rather than Merged because it came through the roundup PR, but your commit is in main's history and the ledger rows keep your attribution. Thanks!

PR #1164, opened 2026-08-06
Bank near-miss for ov006 func_ov006_0212a764 (div 49→44)

lunavyqo:
## Summary

- Improved near-miss tip for `func_ov006_0212a764` (ov006 @ 0x0212a764, size 0x310) from **div 49 → 44** (mwccarm 2004/b56).
- Levers: reverse declaration order (pang → r6) + raw `RandomIntInternal` temporary on the `*10` path so the `and r0,r1` / `mul r1,r0,r2` form matches the ROM.
- Residual is pure callee-saved register permutation (`c` in r7 vs sb, i2/ang cascade) plus frame `0x14` vs `0x1c`. Structure is size-exact and instruction-aligned.
- Attempt tree logged in `config/match_attempts.jsonl`; tip stored in `nearmiss/db.jsonl`. CLAIMS released.

No `src/` match — bank-only.

## Test plan

- [x] `python tools/match.py --c <tip> --func func_ov006_0212a764 --addr 0x212a764 --size 0x310 --module ov006 --version 2004/b56` → 44 mismatches


lunavyqo:

PR #1166, opened 2026-08-06
Match func_ov006_0211ebdc byte-identical (mwccarm 2004/b56)

lunavyqo:
## Summary

Byte-identical match for `func_ov006_0211ebdc` (ov006 @ 0x0211ebdc, size 0x258) on mwccarm **2004/b56**.

Broke the pure r4↔r5 regperm near-miss (prior div 33) by:

- Dropping `#pragma opt_common_subs off` (it was buying schedule but flipping o/s colors)
- Using `char *row = c + o` for the angle block so the ROM's `add r2, r7, r4` / `add r0, r2, #0x4600` form appears
- Spelling the p72 path as `((char *)c + o)` so that second `c+o` rematerializes without CSE merging into the angle row and recoloring `o` away from r4
- Separate `u8 *fade` for the pool fade RMW (not reassigning `s`)

## Test plan

- [x] `fdiff` / `match.py`: mismatches=0 on 2004/b56
- [x] `linkcheck`: VERIFIED
- [x] Enrolled complete in ov006 delinks
- [x] Provenance + attempt log stamped


lunavyqo:

PR #1167, opened 2026-08-06
Bank near-miss for ov006 func_ov006_0211f9fc (div 31→4)

lunavyqo:
## Summary

Near-miss + floor reconfirm for `func_ov006_0211f9fc` (ov006 @ 0x0211f9fc, size 0x120) on **mwccarm 2004/b56**:

- **div 31 → 9 → 4** (match.py words), size-exact
- Control-flow: nest path-update inside `c1d != 0` (ROM `beq` to +0xfc)
- `unsigned int self` + named early `char *bp`
- **Distance block byte-identical** via uninit decl `int c, a, b, d` + compound `a -= b` / `c -= d` + `sqrt(c*c + a*a)`
- Residual pure **early regperm only**: `ldrb r0` vs ROM `r2` (and second load r0 vs r1) — **4 words**
- Body from +0x38 is byte-identical
- Permuter 1100+ iterations floor at score 20; this session 60+ pure-C variants (decl order, named loads, pragmas, live-range extension, char* self, struct overlay, nested if) also floor at div=4
- Tip banked + **floor(regperm)** marked in `nearmiss/db.jsonl`. No `src/` match file.

## Test plan

- [x] match.py → 4 mismatches (early only; +0x38..end OK including full dist)
- [x] log_attempt near_miss improved 31→9→4; floor reconfirm no_progress div=4
- [x] CLAIMS released


lunavyqo:

PR #1150, opened 2026-08-06
Match 22 functions byte-identical (mwccarm 2004/b56)

lunavyqo:
## Summary

Clean replacement for **#1087**, which is unrecoverable on the current gate:

| #1087 failure | Cause |
|---|---|
| denominator changed | `func_02071644` size `0x50→0x54` (phantom absorb) |
| credit 1 changed | re-authored empty `func_02071694` |
| mwldarm failed | stale `Actor::Spawn` mangling in `func_ov030_021136b0` (`PK10Vector3sii` vs real `PK10Vector3_16ii`) |
| BLIND on 021136b0 | same undefined Spawn symbol |

This PR keeps the prior verified matches and drops the arm9 71644/71694 bookkeeping entirely (leave that for a dedicated phantom-symbol PR).

### Matches (all linkcheck **VERIFIED** locally)

- `WallSign::Behavior` (ov085)
- ov004: `020ae858`, `020b0e84`, `020b1710`
- ov006: `020d47f4` (**new** — near-miss div=18→0; strength_reduction/common_subs off; `fix=-1`; loop2 `j` then `x` accum), `020ec4dc`, `020fb230`, `020fb7e0`, `020fd2d8`, `020fea70`, `020fefc0`, `020ff8c8`, `020ffb54`, `02103ac0`, `02108d28`, `0210adac`, `021173c8`
- ov006: `__sinit_ov006_0213014c` (**added** — near-miss div=127→0; whole-struct `Pair` copies into five BSS tables [6+6+3+5+4] in ROM dest order; linkcheck VERIFIED, mwccarm 2004/b56)
- ov030: `02113324`, `021136b0` (Spawn mangling fixed)
- ov060: `02116d78` (delinks `complete` fixed)
- ov062: `0211bd10` (**delinks enrollment added** — was missing on #1087)

### Local verification

- Prior batch: `prepush_linkcheck` **20/20 VERIFIED**, full `mwldarm` exit 0, `langmode_audit`/`port_refcheck` clean
- New: `func_ov006_020d47f4` match.py MATCH + linkcheck **VERIFIED** (mwccarm 2004/b56)
- New: `__sinit_ov006_0213014c` match.py MATCH + linkcheck **VERIFIED** (mwccarm 2004/b56)

### Not in this PR

- `func_02071644` / `func_02071694` phantom merge (structural vs denominator/credit gates)

Closes nothing automatically — please close **#1087** in favor of this branch.

## Test plan

- [x] Local match + linkcheck on prior 20
[...]

PR #1171, opened 2026-08-06
Match arm9 func_02071644 byte-identical (mwccarm 2004/b56)

lunavyqo:
## Summary

Match `func_02071644` with pure C (mwccarm **2004/b56**), replacing the hand-asm hatch.

This is the dedicated phantom-symbol PR called out in #1150. The near-miss tip (div=3) was already byte-identical for all 20 live instructions; the residual was a trailing dead `bx lr` that mwccarm always appends after a `goto`/`for(;;)` loop whose exits are early returns. The symbol map had carved that 4-byte epilogue off as empty `func_02071694`.

### Changes

- `src/func_02071644.c` — pure C from the near-miss tip (`int` return + goto-loop)
- `config/arm9/symbols.txt` — size `0x50 → 0x54`; drop `func_02071694` row
- `config/arm9/delinks.txt` — enroll parent over `0x02071644..0x02071698`
- delete `src/func_02071694.c` (same pattern as `func_02072168` / `func_020729e8`, notes 9a(3))
- attempt + provenance ledgers; near-miss tip pruned

### Verification

- `match.py` **MATCH** at size `0x54` (2004/b56)
- `linkcheck.py` **VERIFIED** (blind 0)

Author: lunavyqo

## Test plan

- [x] Local match.py MATCH (0x54, 2004/b56)
- [x] Local linkcheck VERIFIED
- [ ] CI `validate` green


lunavyqo:

PR #1184, opened 2026-08-07
Bank floor reconfirm for ov006 func_ov006_02106168 (div=28)

lunavyqo:
## Summary
- Reconfirm near-miss tip for `func_ov006_02106168` (ov006 @ 0x02106168, size 0x238) under **mwccarm 2004/b56**: still **div=28**, size-exact.
- Residual: **r5/r7 regperm** (`pd` vs loop index `i`) plus first-loop **instruction schedule** (tip loads `idx4` early; ROM does dest-partial `self+i*4`, then table/row, then late `idx4`, with `i*2` in r6).
- ~100 variants tried (decl order 6k, dest-early spellings, volatile `n`, pe/pd hoist, pragmas, fake-dep, launder, dummy webs); none beat the stored tip. Hybrid dest-early + volatile `n` is size-exact but cascades prologue/`bge` to worse scores.
- Tip remains in `nearmiss/db.jsonl` (source andrewboudreau). No `src/` file. Attempt logged; API claim `clm_4c9a4550ac03` released.

## Test plan
- [x] `python tools/match.py --c <tip> --func func_ov006_02106168 --addr 0x2106168 --size 0x238 --module ov006 --version 2004/b56` → no match, 28 words
- [x] `tools/log_attempt.py` wrote `config/match_attempts.jsonl` node


lunavyqo:

PR #1206, opened 2026-08-07
dtors: name 7 base-object destructors the tree had never named

andrewboudreau:
Follow-up to #1203, which found seven misnamed D2 symbols and built
`tools/dtor_variant_audit.py` to decide the question from reloc kind alone. Its
`--discover` mode found the other direction: genuine base-object destructors the tree
has **never named**, sitting under `func_*`. This names seven of them.

**33 in-scope files.** `rombuild` **106/106 exact, PASS**, 10,699 reproducing / 0
mismatching; `eligible` 10699 unchanged; attribution 7 renamed with credit intact,
0 lost; `build_pin.verify` **7/7** `(True, '2004/b56')`.

## The renames

| was | is | class named by |
|---|---|---|
| `func_02017838` | `_ZN5FaderD2Ev` | `dFader_c` slots 0/1 |
| `func_020177c4` | `_ZN15FaderBrightnessD2Ev` | `dFdBrightness_c` slots 0/1 |
| `func_02039658` | `_ZN16MeshColliderBaseD2Ev` | `dBgW` slots 0/1 |
| `func_020397fc` | `_ZN12MeshColliderD2Ev` | `dBgW_Kc` slots 0/1 |
| `func_0203a420` | `_ZN18MovingMeshColliderD2Ev` | `dBgW_KcMbg` slots 0/1 |
| `func_020380ec` | `_ZN10ClsnResultD2Ev` | `dBgPi` slot 0 |
| `func_0203ac1c` | `_ZN8dM3dGSphD2Ev` | ROM name — no tree name exists |

Each is reached only by `bl`, never as data, from derived-class destructors, and the
vtable it stores names the owning class. Tree names come from **each class's own D0/D1
pair in its vtable**, never from name similarity. `dM3dGSph` has zero `_ZN8dM3dGSph`
symbols anywhere, so it takes the ROM name — the convention
`rtti-reconciliation.md` §5/§8 established for never-named classes.

Two close a loop opened by hand long ago. `include/MeshColliderBase.h` already worked
out that `func_02039658` and `func_020397fc` were `MeshColliderBase`'s and
`MeshCollider`'s D2s, concluded *"the symbol names are on the wrong bodies"*, and
deferred the fix: *"Correcting that pair is a symbols.txt rename, deliberately left to
its own change."* #1203 corrected the impostor holding `_ZN16MeshColliderBaseD2Ev`;
this gives the name to the function that earns it.

## This is an upward re-bank, deliberately

`unmigrated_total` **1258 → 1265**, exactly one per rename, denominator moving with it
(2516 → 2523). Naming a function the tree had never identified makes the symbol table
more true **and** the backlog larger, because it mints a new hand-spelled mangled
symbol. That is real language-mode debt and it should be counted — the alternative,
exempting it, would make the headline dishonest in the other direction.

The convention permitting this is stated in **#1205**, which is open and not yet on
main:

> A count may rise only together with a re-banked baseline in the same PR, and only
> when the increase is exactly a set of renames giving previously-unnamed (`func_*`)
> functions their evidenced C++ symbols. The PR must name the evidence. Any other
> increase is a regression.

Evidence here is `tools/dtor_variant_audit.py --discover`. If #1205 lands first this
note is redundant; if it does not, this PR is the argument for the rule.

## Four of the eleven candidates are deliberately not here

- **`Enemy`, `dScMgBase_c`** — identification as solid as these seven, but their
  renames touch **118** and **66** files (every derived destructor calls them).
  Together with these they put one PR at 234 in-scope, over the 200 cap. Their own PR.
- **`dBgCh`** — the D2 is proven: `SphereClsn`, `RaycastLine`, `WithMeshClsn` and
  `RaycastGround` all call `func_020354d0` as their base dtor, and RTTI says `dBgCh` is
  their base. The **name** is not. `rtti-reconciliation.md` §4 records `BgCh == dBgCh`
  as unproven and says naming this vtable settles it either way. `_ZN4BgChD2Ev` asserts
  the alias; `_ZN5dBgChD2Ev` invents a second name for a class the tree already calls
  something. Neither belongs in a naming pass.
- **`dCapEnemy_c`** — `_ZN8CapEnemyD2Ev` **already exists** at `0x0200651c` in arm9,
  the same `0x38` size as the ov002 candidate at `0x020aedbc`. Two functions cannot
  both be it, and the tree has no precedent for one mangled name in two modules.

Renames are split rewrite-then-move so git records R100 and credit follows;
`delinks.txt` is regenerated, never hand-edited.

https://claude.ai/code/session_0154eg3RhXQYPyfKyv5t81Fz


andrewboudreau:
**Rebased onto main after #1205 merged.** The conflict was `langmode-baseline.json` --
both PRs re-bank it -- resolved by regenerating rather than hand-merging.

One correction to the move commit's message, which was written before the rebase and
now quotes stale absolutes. The **delta is unchanged and correct at +7**; the
endpoints moved because #1205 lowered the baseline first:

    commit message says   1258 -> 1265
    actual after rebase   1252 -> 1259

`langmode-baseline.json` in this PR is regenerated from the tree, so the committed
value is right; only the prose in the commit is stale.

Also now redundant: that commit notes the re-bank rule "is stated in PR #1205, which is
open and not yet on main". #1205 has merged, so the rule is on main and this PR is
simply its first use.

Re-gated on the new base: `rombuild` **106/106 exact, PASS**, 10,699 reproducing / 0
mismatching; `eligible` 10699 complete, unchanged; ratchet PASS; attribution 7 renamed
with credit intact, 0 lost; delinks regenerated (0 files changed on this base).

PR #1212, opened 2026-08-08
Match 13 functions byte-identical (mwccarm 1.2/sp2p3)

ruspecial:
The ITCM runtime block: the MSL/ARM soft-float and integer-division helpers at
`0x01ff8000`-`0x01ffb078`, plus the OS context switch and the soft-reset hand-off
at the top of ITCM.

| symbol | addr | size | what |
|---|---|---|---|
| `_dmul` | `0x01ff8708` | `0x6f0` | double multiply |
| `func_01ff97d8` | `0x01ff97d8` | `0xb6c` | largest double routine in the block |
| `func_01ff859c` | `0x01ff859c` | `0x16c` | double -> 64-bit integer |
| `func_01ffa344` | `0x01ffa344` | `0xfc` | float -> int |
| `func_01ffa4bc` | `0x01ffa4bc` | `0xcc` | int -> float |
| `_s32_div_f` | `0x01ffabe4` | `0x20c` | signed 32-bit divide (`__aeabi_idiv`) |
| `_u32_div_f` | `0x01ffadf0` | `0x1e4` | unsigned 32-bit divide (`__aeabi_uidiv`) |
| `_ll_sdiv` | `0x01ffaa34` | `0x1b0` | signed 64-bit divide |
| `func_01ffafd4` | `0x01ffafd4` | `0x34` | reset FP runtime state |
| `func_01ffb008` | `0x01ffb008` | `0x28` | FP status word read-modify-write |
| `func_01ffb030` | `0x01ffb030` | `0x4c` | packed-field form of the above |
| `func_01ffd9d4` | `0x01ffd9d4` | `0x154` | thread context switch |
| `func_01ffdb28` | `0x01ffdb28` | `0xb0` | soft-reset hand-off (tail of `func_01ffdd98`) |

**These are all asm-primitive matches, not recovered C — please sanity-check that
call before merging.** `func_01ffd9d4` clears the objective test in
`notes/asm-policy.md` outright (`msr` / `mrs spsr` / banked `ldmib r1!, {...}^`).
The other twelve are ordinary ARM, so they rest on the same reading that already
covers `func_02059d98`, `ARMSaveContext` and the rest of the 35 SDK/runtime
routines: this is the toolchain's own runtime, shipped by MetroWerks as assembly,
with no C in the game's sources to recover. Each file says specifically why C
cannot produce it — most of them keep live values across a `bl` by pushing the
argument registers (`stmdb sp!, {r0, r1, r2, ip, lr}`), return two results in
r0/r1, or have no prologue at all while still calling out. If you read the policy
more narrowly, say so and I will pull the twelve and keep only `func_01ffd9d4`.

Verification notes:

- Every file verified with `match.py` at **1.2/sp2p3**, byte-identical, in its
  final on-disk state.
- `linkcheck` could not run on my box (it pins `2004/b56`, which is not in my
  toolchain). Instead I cross-checked all **51 symbolic relocation slots**
  directly: for each `bl <sym>` and `ldr rN, =<sym>` I compared the symbol's
  address from `config/**/symbols.txt` against the destination the ROM's own
  instruction encodes. 0 mismatches. Worth a second look from CI regardless,
  since that is a hand check and not the real linker.
- `srcpath.py` confirmed all thirteen belong in `src/` root.
- Ledger rows included: 13 `match_provenance` rows, and the console's 21
  `match_attempts` rows for this batch.

Nine of the batch's 22 targets are not here:

- `_dadd`/`func_01ff8e10`, `func_01ff9378`/`func_01ffa594`,
  `__aeabi_uldiv`/`__aeabi_ulmod`, `func_01ffa440` — each is half of a shared
  routine that branches into the *middle* of its partner (e.g. `_dadd+0xc` is
  `bmi 0x01ff8e20`). Those branches carry no relocation, so the offset has to be
  exact, and there is no symbol at the target to name. One function per file
  cannot express them; aiming the branch at the partner's entry symbol would be a
  WRONG-DEST false match.
- `MeshCollider::DetectClsn(RaycastLine&)` and `DetectClsn(SphereClsn&)` —
  blocked on tooling, not matching. Both need **mwccarm 2004/b56**, which is not
  installed on my box; the already-landed twin `DetectClsn(RaycastGround&)` fails
  to build here for the same reason. The banked near-miss (203/461, size exact at
  2004/b56) is untouched in `nearmiss/db.jsonl`.


tangosdev:
Sanity check done, taking all 13. The asm-primitive call is consistent with applied practice on main, not just your reading: Copy32Bytes (pure ldmia/stmia), CpuCopy8, and func_02059d98 are all ordinary ARM and all count as matched under the HAND-ASM banner as SDK/runtime primitives, and notes/func_02059d8c-asm-origin.md is a whole determination doc for an ordinary-ARM 3-instruction function. The MSL soft-float/division block is the same category as the 35 SDK routines the policy history kept: toolchain runtime shipped as assembly, no C in the game's sources to recover. The two-results-in-r0/r1 and no-prologue-live-across-bl arguments in the file headers are exactly the kind of evidence the d8c note used. func_01ffd9d4 clears the objective test outright. Validator: 9 VERIFIED, 4 no-sym on the underscore-named helpers (the known missing-alias pattern for the aeabi symbols), byte-identical across the board.

PR #1228, opened 2026-08-08
tangos: matched work (tangos/tango)

tangosdev:
2 matched functions pushed from tangOS Console. Review + CI gate the merge.


tangosdev:
Superseded: both functions landed on main via #1227 minutes before this PR's ratchet fix, in the same header-based style (byte-verified there). Nothing left to merge here.

PR #1254, opened 2026-08-08
Bank near-miss for ov006 func_ov006_020da174 (div 253→24)

lunavyqo:
## Summary

- Bank improved near-miss tip for `func_ov006_020da174` (ov006 @ 0x020da174, size 0x2ac): **253 → 24** word divergences on mwccarm **2004/b56**, size-exact.
- Attempt tree logged in `config/match_attempts.jsonl` (near-miss lineage from draft).
- CLAIMS.md: released after banking tip (API key 401; CLAIMS.md only).

### Levers
- Correct logic from ROM: max + second-max frequency scan (not min/max draft), rank remap to -2/-1/0, `int` return 0/1, five objects of stride 0x30.
- `u64`-mask launder of hist base → `add r3, sp, #8` + `strh` via r3.
- Flag loop `store1`/`store0` fallthrough matching ROM.

### Residual (pure regperm)
- Prologue `i`/`p` assignment order (r8↔r7).
- Bubble sort `#0x30` stride register (r8 vs r3) cascading u2 path.
- Flag loop counter r7 vs r2 (+ cmp cascade).

Permuter base score ~160; no match after real effort. Tip is highest-yield input for refine/permuter.

## Test plan
- [x] `match.py` on tip: size 0x2ac, div=24 (not MATCH)
- [x] tip recompile verified after `nearmiss/db.jsonl` write


lunavyqo:
Superseded by full match: see follow-up match PR for func_ov006_020da174 (div 253→0).

PR #1265, opened 2026-08-08
src: one source file per symbol, and a gate that keeps it that way

andrewboudreau:
Six stems carried both a `.c` and a `.cpp`. `config/**/delinks.txt` names exactly one path per address range, and **every gate in this tree is driven from delinks.txt** — `eligible.py`, `rombuild.py` and the byte gate are only ever pointed at files the build knows about. So the other half is compiled by nothing and checked by nothing, while a reader who opens it has no way to tell it isn't the source of the shipped bytes.

## What the six actually were

I verified both halves of all six against the ROM, and then — the part a byte check cannot do — **enrolled each `.cpp` and ran the real link**. The answers don't agree with each other, which is the point:

| stem | `.c` | `.cpp` | links? | action |
|---|---|---|---|---|
| `_ZN13PoleBillboard8BehaviorEv` | **fails** | matches | yes | keep `.cpp`, **enroll** |
| `func_ov065_02116364` | matches | matches | yes | keep `.cpp` |
| `func_ov084_021298d0` | matches | matches | yes | keep `.cpp` |
| `func_ov084_0212d564` | matches | matches | yes | keep `.cpp` |
| `_ZN15TtcRotatingGear8BehaviorEv` | matches | matches | **NO** | keep `.c`, drop `.cpp` |
| `_ZN5Stage7PS_InitEv` | matches | matches | **NO** | keep `.c`, drop `.cpp` |

**`PoleBillboard::Behavior` was a live defect.** Neither half was enrolled, the `.c` does *not* reproduce the ROM, and the `.cpp` does. A correct implementation had been sitting in the tree uncompiled while the half that doesn't match was what a reader found. Source-built functions **10,714 → 10,715**.

**Three came for free** — both halves match and link, so keeping the `.cpp` is a migration with the C++ already written.

**Two do not link, and that is why this isn't a tidiness change.** Seven undefined references between them, because they declare their callees as C++ and let the compiler mangle the names:

- `struct G2x { static void SetBlendBrightness(volatile u16*, int, short); }` → `_ZN3G2x18SetBlendBrightnessEPVtis`, where the ROM has `...EPVtts`
- `TtcRotatingGear::Behavior` reaches for `Platform::UpdateClsnPosAndRot`, `Platform::UpdateModelPosAndRotY`, `Platform::IsClsnInRange`, `Actor::UpdatePos(CylinderClsn*)`, `RandomIntInternal` and `DecIfAbove0_Short` — none of which exist under the names it produces

`match.py` wildcards relocated words, so a byte match **structurally cannot see this**. Both files look exactly like finished migrations — real C++ methods, byte-exact — and adopting either on the strength of `build_pin.verify` alone breaks the ROM link. Same lesson as the braced `extern "C" { u32 x; }` bug: **rombuild is the verdict, verify is not.**

## The gate

`tools/check_duplicate_sources.py` refuses a doubled stem and says which half the build names and which is compiled by nothing. Wired into the pre-push hook next to `port_refcheck` — a filesystem scan, so it costs milliseconds. Tested against a planted duplicate: exit 1 with both paths and their roles named.

## Gates

- `rombuild.py`: **106/106 exact, 100.000000% of compared bytes, PASS**
- `port_refcheck.py` 393/393 · langmode ratchet **PASS**
- source-built **10,714 → 10,715**; unmigrated 1130 → 1129


andrewboudreau:
Follow-up commit adds the attribution fix this surfaced.

While checking who to name in an override, the credit turned out to be **random**: `lineage()` keys on the path minus its extension, so `src/X.c` and `src/X.cpp` collapse to one key, and it iterated a `set` — whose order depends on Python's per-process hash randomisation. Five runs of the same function on the same commit returned three different owners. `PYTHONHASHSEED=0` makes it stable, which pins the cause.

**Eight** stems were affected, not six; two of them (`func_ov002_020cd3e0`, `func_ov003_020adc10`) have only one file on disk — `first_matchers` still holds the pre-promotion `.c` path, so the new duplicate-file gate can never catch those.

Iteration is now sorted and **first-wins**: for a `.c → .cpp` promotion the `.c` entry is the original match and the `.cpp` entry is whoever ran the rename, and this module's docstring says credit belongs to "the first contributor who landed the match it descends from".

That leaves three real changes — stems where two people wrote rival implementations and the cleanup keeps one file. Overridden to the owner the tree credited before, so nobody loses credit: `lunavyqo`, `lunavyqo`, `ruspecial`.

`0 changed, 0 lost`, reproducibly.

PR #1266, opened 2026-08-08
src: give the 51 C++ files that are named .c the right extension

andrewboudreau:
The build decides language mode from a `//cpp` marker on line 1, **not** from the filename — `build_pin.py:138` and `rombuild.py:277` both read the marker and nothing reads the suffix. So a file can be compiled as C++, byte-match, and still be called `.c`. Fifty-two were.

`match.py` has been saying so all along, once per build:

```
[cpp-check] Warning: X.c has '//cpp' header but extension is .c (should be .cpp)
```

## Seven are finished migrations counted as debt

These already define `Class::Method` against the real header:

- `Actor::SpawnSoundObj`, `Actor::GetWaterHeightWDW`, `Actor::AfterInitResources`, `Actor::BeforeCleanupResources`, `Actor::KillAndTrackInDeathTable`
- `ActorDerived::AfterInitResources`, `ActorBase::AfterCleanupResources`

`langmode_audit` counted every one as *"not migrated, .c extension"* purely on the filename, so **the backlog has been overstating itself by seven**.

The other forty-four keep their hand-spelled symbols and stay counted as debt, correctly — this fixes only what the extension claims, not what the source says.

## One skipped, named rather than dropped

`src/func_ov084_021298d0.c` can't be renamed because `src/func_ov084_021298d0.cpp` already exists. That's one of the six duplicate stems #1265 removes; the rename can follow once that lands.

## Gates

Byte-neutral by construction, and checked rather than assumed:

- `rombuild.py`: **106/106 exact, 100.000000% of compared bytes, PASS**, same 10,714 source-built functions as before
- `port_refcheck.py` 393/393 · langmode ratchet **PASS** · unmigrated **1120 → 1113**
- attribution **0 changed, 0 lost** — a same-stem `.c → .cpp` rename is exactly the case `prepush_attribution` keys on the extension-stripped path to protect


andrewboudreau:
Rebased onto current main and re-banked the baseline.

**The ratchet failure was real and my local check was wrong.** I checked against a copy of `langmode-baseline.json` fetched before the day's earlier PRs merged, got PASS, and only CI disagreed. Re-fetching `origin/chaos-data` reproduces it exactly:

```
language_mode.cpp_still_handspelled: 236 -> 240  (+4)
```

**It is a rise, and it is not new debt.** Of the 51 renamed files, 11 have a mangled stem so the audit counts them. Seven are finished migrations and leave the backlog entirely; the other four hand-spell their symbols, and moving them out of the `.c extension` bucket puts them in the `.cpp, symbol hand-spelt` one:

| metric | before | after |
|---|---|---|
| `.c extension` | 884 | **873** (−11) |
| `.cpp, symbol hand-spelt` | 236 | **240** (+4) |
| **NOT MIGRATED** | 1120 | **1113** (−7) |

Same four files, same debt, correct label. The headline — the one that cannot be gamed by renaming, since a hand-spelled symbol counts however the file is named — falls by seven.

`notes/plan-cpp-language-mode.md` §4 provides for exactly this: a count may rise only together with a re-banked baseline in the same PR, and the PR must name the evidence. The root `langmode-baseline.json` is preferred over the banked chaos-data copy, so the rise lands as a loud reviewable diff and the ratchet emits a warning annotation while the file exists.

**That file should be deleted in a follow-up once this lands** — while it exists it *is* the baseline, which re-creates the every-PR conflict that moving it to chaos-data was meant to end.

PR #1267, opened 2026-08-08
attribution: pin the eight stems whose credit is decided by a coin flip

andrewboudreau:
`prepush_attribution.lineage()` keys on the path **minus its extension**, so `src/X.c` and `src/X.cpp` collapse to one key and the second write wins. It iterates a `set` of strings, and Python randomises string hashing per process — so **which of the two wins is decided fresh on every run**.

Not theoretical. Five runs of the same function on the same commit:

```
run 1  PoleBillboard=tangosdev  PS_Init=tangosdev
run 2  PoleBillboard=tangosdev  PS_Init=ruspecial
run 3  PoleBillboard=lunavyqo   PS_Init=ruspecial
```

`PYTHONHASHSEED=0` makes it stable, which pins the cause. The contributor chart has therefore been assigning credit for these functions **at random on every regeneration**, and any PR touching one reports a "credit changed" that doesn't reproduce — which is exactly what happened to #1265.

## Eight stems, and two of them have only one file

Six have two files on disk. The other two — `func_ov002_020cd3e0` and `func_ov003_020adc10` — have only one, because `first_matchers` still carries the pre-promotion `.c` path alongside the `.cpp` that replaced it. A duplicate-file check can never catch those.

## The fix needs no code change

Pin **both** paths of each pair to the same handle, and the collapse order stops mattering.

The value is the **original matcher** — the `.c` side, resolved through `chaos_db_ci` and then the alias table, never from `git log --format=%an` — because this module's own docstring says credit belongs to *"the first contributor who landed the match it descends from"*, and the `.cpp` entry of a promotion is whoever performed the rename.

| stem | pinned to |
|---|---|
| `_ZN13PoleBillboard8BehaviorEv` | `lunavyqo` |
| `_ZN15TtcRotatingGear8BehaviorEv` | `lunavyqo` |
| `_ZN5Stage7PS_InitEv` | `tangosdev` |
| `func_ov002_020cd3e0` | `ruspecial` |
| `func_ov003_020adc10` | `ruspecial` |
| `func_ov065_02116364` | `lunavyqo` |
| `func_ov084_021298d0` | `tangosdev` |
| `func_ov084_0212d564` | `ruspecial` |

**Nobody gains or loses a function.** This makes permanent whichever answer was already the intended one, and stops it moving on its own. Verified deterministic across repeated runs.

Lands ahead of #1265 on purpose: that PR removes six of the duplicate files, and a comparison against a base that decides its own answer at random cannot be made green from inside the PR.


andrewboudreau:
## This PR cannot pass the attribution gate, and that is the thing it is fixing

The gate's verdict for these stems is **not reproducible**. Same two commits, same command, five consecutive local runs of the same computation the gate performs:

```
3 changed, 0 lost
4 changed, 0 lost
5 changed, 0 lost
6 changed, 0 lost
6 changed, 0 lost
```

The validator's `0 added, 2 changed, 0 lost` above is one sample from that distribution. #1265's `1 added, 1 changed, 0 lost` is another.

**Why.** `lineage()` keys on the path minus its extension, so `src/X.c` and `src/X.cpp` collapse to a single key and the second write wins. It iterates a `set` of strings, and Python randomises string hashing per process, so which one is 'second' is redrawn every run. `PYTHONHASHSEED=0` makes it stable, which pins the cause.

**Why it can't go green.** The gate compares base against head. This PR pins head to a fixed owner; the base is still `origin/main`, which resolves at random. Roughly half the pinned stems will disagree with any given draw of the base, so the gate reports 'changed' for a change whose entire purpose is to stop the value moving. No edit inside this PR can reach the base side.

**What is actually true of this change:** `0 lost`, in every run. Nobody gains or loses a function. Each stem is pinned to the original matcher — the `.c` side, resolved through `chaos_db_ci` and the alias table — which is the owner this module's docstring already says credit belongs to.

So this needs a human to accept a red gate once. After it lands, the eight stems stop moving for everyone, and #1265 (which removes six of the duplicate files) can be compared against a base that gives the same answer twice.

If you would rather not override the gate, the alternative is to leave the randomness in place — but then the contributor chart keeps reassigning these eight functions on every regeneration, and every future PR that touches one draws a fresh 'credit changed' that will not reproduce on a re-run.

PR #1293, opened 2026-08-09
Player: 33 functions as real methods, 8 fields recovered from padding, 3 declared types corrected

andrewboudreau:
> Stacked on #1291 → #1289. Rebases down onto `main` as those land.

The last intact `_Init`/`_Main` pair in Player's state machine — the other 25
unmigrated `St_*` handlers are orphans whose siblings already landed.

Both were `char*` functions built entirely from magic offsets. Every one now has a
name: `mIsTakingDamage` 0x708, `mIsAirborne` 0x6de, `mLandSoundPlayed` 0x6df,
`mStateStep` 0x6e3, `mStateTimer` 0x6a4, `mStateWork` 0x6e5, `mCharacter` 0x6d9,
`mGroundY` 0x644, `mFloorClass` 0x658, `mIsMetal` 0x6f9, `mIsInShallowWater` 0x707,
`mLoopingSoundHandle` 0x620, `mParticle2` 0x62c, `mDesiredAngleY` 0x6d2 — plus
Actor's `mPosX/Y/Z`, `mAngleY`, `mPrevAngleY`, `mHorzSpeed`, `mVertSpeed`,
`mCamSpacePosX`, and ActorBase's `param1`. `_Init` also sheds a local `struct
Vector3` shadow.

## Method, because guessing does not work here

`_Main` was first **rewritten** rather than substituted, and came out `999 words
differ` — a size mismatch, 0x39c against 0x3a4, with no signal about which change
caused it. Rewriting restructures evaluation order and cast widths at the same time
as it renames, so the failure is uninformative.

The procedure that works, and is worth reusing on every remaining slice:

1. **Mechanically substitute** offsets for field names — preserving every cast,
   launder, temporary and evaluation order. Note `*(u16*)(c + 0x8e)` on an `s16`
   field must become `*(u16*)&mAngleY`; the unsigned shift is semantic.
2. **Verify.** If it matches, done.
3. **If not, apply each readable form _separately_** and keep only the ones that
   verify.

Step 3 isolated it in a single pass. Of six candidate spellings, **five are free** —
including `&mCamSpacePosX`, `&mAngleY`, `&mHorzSpeed` and the whole body-model
pointer chain — and exactly one is not:

```c
(((long long)((int)this + 0x6e5)) & ...) += 1     // matches
(((long long)(int)&mStateWork)    & ...) += 1     // costs 4 bytes
```

The add happens **inside** the integer cast. The *other* increment in the same
function, whose add is on the pointer before the cast, **is** interchangeable with
`&mStateWork` and has been converted. That one site keeps its raw offset and now
carries a comment recording that it was measured — so it does not read as
un-recovered work someone will later "clean up" and silently break.

## Verification

| gate | result |
|---|---|
| `build_pin.verify` | `St_BurnFire_Init` `(True, '2004/b56')`; `St_BurnFire_Main` `(True, '2004/b56')` |
| `rombuild.py -j 16 --no-rom` | **106/106 exact, 100.000000%**; 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 177 fields, 0 mismatched, spans 0x768 |
| `port_refcheck` / `check_data_definitions` / `check_duplicate_sources` | clean |
| langmode | unmigrated **1052 → 1050** |


andrewboudreau:
Follow-up commit pushed: the first revision failed the **langmode ratchet**, and the cause is worth recording.

`langmode_audit`'s codegen-hack detector is a regex over source *text*:

```python
LAUNDER = re.compile(r"launder|force.*(?:reg|codegen)|codegen.*force", re.I)
```

counted per file. The comment I added to explain why one raw offset survives used the word for the pointer-masking idiom. The file already contained two of those masks — what it did not contain was the *word*. Naming it moved the file into the bucket and pushed `codegen_hacks` 69 → 70, failing the ratchet on a change that added no hack at all.

Reworded to say the same thing without the trigger. Bytes unchanged (`(True, '2004/b56')`), **ratchet now PASS**.

Worth knowing generally: the metric is textual in both directions — deleting a real mask does not lower it either — so it is not a fact about the tree.

andrewboudreau:
Extended: this now covers **five** handlers, not two. Added `St_SweepKick_Init`, `St_Thrown_Init` and `St_Throw_Init` — all three matched on the **first attempt**, because the procedure established by the BurnFire commit was followed rather than rewriting them.

Also retires three local `struct Vector3 {int x,y,z;}` shadows.

**`St_Throw_Init` recovers a type.** It loads `mHeldObj` — declared `s32` in Player.h — then reads `+0xb0` and `+0xc` through it, which are ActorBase's `mFlags` and `actorID`. So it is an actor pointer, and the source now says so:

```c
Actor* held = (Actor*)mHeldObj;
int b = (held->mFlags & 0x200) ? 1 : 0;
unsigned short a = held->actorID;
```

Its `goto` structure is preserved exactly — that control flow is the compiler's, not a style choice.

To make this repeatable, a scratch tool builds the offset→field map from the headers themselves (**219 named offsets**: 45 inherited, 174 Player's own) and reports the offsets it *cannot* name rather than guessing. An unnamed offset is the one place a mechanical substitution can silently land on the wrong field.

| gate | result |
|---|---|
| `build_pin.verify` | all five `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1052 → 1047** |

andrewboudreau:
Cycle 2: **three more handlers** (`St_Grabbed_Cleanup`, `St_Respawn_Init`, `St_Talk_Cleanup`) — eight total. All three matched first try. Three more shadow structs go, including two fake `struct Player` definitions.

## Five fields that were declared as padding

`St_Respawn_Init` reads four offsets Player.h did not name — three inside `pad_381[0x1c7]`, one that *was* `pad_698[0x2]`. They are the spawn point, and the evidence is a matched pair pointing opposite ways:

| | 0x53c..0x544 | 0x698 |
|---|---|---|
| `Player::InitResources` | ← `mPosX/Y/Z` | ← `mAngleY` |
| `Player::St_Respawn_Init` | → `mPosX/Y/Z` | → `mPrevAngleY` |

Two functions, opposite directions, the same four slots → `mSpawnPosX/Y/Z`, `mSpawnAngleY`. `St_Talk_Cleanup` supplies the fifth (`pad_725[0x1]`), a latch gating a music-layer reload; named `unk_725`, because one use evidences that it *exists* but not what to call it.

Declaring padding where a field lives isn't neutral — it tells the next reader "nothing here". `check_header_offsets` confirms placement: **177 → 182** commented fields, 0 mismatched, struct still spans 0x768.

## What the greedy pass bought

Every candidate was free this cycle, so it **deleted** code rather than just renaming it:

```diff
-   int* ptr_380 = (int*)(self + 0x380);
-   func_02035860(((void*)(((int)ptr_380))), ((void*)(((int)ptr_5c))));
+   func_02035860(&mMeshClsn, &mPosX);
```

All three pointer locals gone, byte-identical. Same for two pointer-masked flag writes reduced to `mFlags &= ~0x800000` and `*(u32*)&mBodyClsnFlags &= ~2`.

## Types left alone on purpose

`mTalkActor` is declared `u8` and holds an actor pointer (`+0xb0` is read through it). Reached via a cast rather than retyped, because `St_Talk_Main` reaches it the same way and the two should change together. Likewise `St_Grabbed_Cleanup` calls `_ZN6Player9DropActorEv` with the actor at `0x35c` as its **this**, implying that slot holds a `Player` — left as an `extern "C"` call rather than asserting a type this function alone cannot evidence.

| gate | result |
|---|---|
| `build_pin.verify` | all eight `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1047 → 1044** |

andrewboudreau:
Cycle 3: **three more** (`St_LongJump_Init`, `St_Hurt_Main`, `St_Jump_Init`) — **eleven** total. Unmigrated **1044 → 1041**. `St_Jump_Init` ends with **zero raw offsets** in its body.

## A declared width nothing had checked — this time on a field

`St_Hurt_Main` missed by **exactly one word**. Player.h declared:

```c
s16 mJumpComboTimer;   /* 0x6a8 */
```

so `mJumpComboTimer != 0` compiled to `ldrsh`. The ROM uses `ldrh`.

Every other reference in the tree agrees the field is unsigned: `St_Jump_Init` and `St_Swim_Main` read it through `*(u16*)`, `Behavior` passes it to `DecIfAbove0_Short(u16*)`, and the remaining three only *store* (stores prove nothing — width is the same either way). **Nothing reads it signed.** So the declaration was wrong, and it is fixed rather than cast around.

This is the imported-parameter-width defect from Actor.h, one level down. A field's declared sign is a *claim*, and until something asks the compiler to **generate** a load from it, nothing checks that claim — the source spells the load itself. The failure is quiet: one word in 0x248 bytes.

Verified the fix does not disturb the 201 files that include Player.h — `rombuild` **106/106 exact, 0 mismatching**.

## One greedy result that generalises

Every candidate was free across all three files. The one worth reusing beyond this PR:

```diff
-   if (_ZN6Player12FinishedAnimEv(this)) {
+   if (FinishedAnim()) {
```

**Calling an already-declared method instead of its `extern "C"` mangled name is byte-neutral**, and the extern declaration goes with it. That applies to every method Player.h already declares.

## Mapper limitation, recorded

The offset-mapper flags `0x800` in two files. It is not an offset — it is the round-to-nearest constant in `(x * 0x1800 + 0x800) >> 12`, matched because the pattern looks for `<something> + 0x…`. Over-reporting is the right failure direction there: it asks, and checking is cheap.

| gate | result |
|---|---|
| `build_pin.verify` | all eleven `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1044 → 1041**

andrewboudreau:
Cycle 4: **two more** (`St_Spin_Main` — renamed `.c`→`.cpp` with its delinks entry in the same commit — and `St_Balloon_Main`, 0x450). **Thirteen** total, unmigrated **1041 → 1039**. Both end with zero raw offsets; `St_Balloon_Main` also retires a local `struct State` onto the real `Player::State`.

## The sign defect reproduces — and was diagnosed in one step

Cycle 3 found `s16 mJumpComboTimer` where the ROM does `ldrh`. The greedy pass flagged the identical signature here — `unk_6c0 == 0` costs **exactly one word** — so the cause was known *before* it was investigated. That's now a rule rather than a discovery: **a one-word miss on an otherwise-clean substitution is a sign/width bug.**

Same evidence shape. Every **read** of 0x6c0 is unsigned: `func_ov002_020e4bb8` tests `< 0x3f` and `& 1`, `Behavior` passes it to `DecIfAbove0_Short(u16*)`, `St_Balloon_Main` tests `== 0`. Only the *stores* use `short` — and a store cannot distinguish the two, which is precisely why this class survives: **half the references carry no information.**

Those five uses also name it: `InitBalloonMario` sets `0x258`, `Behavior` counts down, `St_Balloon_Main` exits at 0, `020e4bb8` blinks below `0x3f`. So `s16 unk_6c0` → `u16 mBalloonTimer`, three users updated.

## `_ZN6Player12St_Null_InitEv` is not a Player method — deliberately left alone

It looked like the easiest migration available: the offset mapper reported **zero** offsets to resolve. That is because there is nothing to resolve — the symbol lives in **ov006 at 0x020cac30** while Player is in ov002, its body takes no `this`, and every callee is `func_ov006_*`.

Third instance of the ov002/ov006 misattribution this tree already documents for `St_WallJump_Init` and the `ST_OWL`/`ST_WAIT` pair. Migrating it would have asserted that a Player method is defined at an ov006 address with no receiver. **"Zero offsets to name" is a reason to look harder, not to proceed.**

| gate | result |
|---|---|
| `build_pin.verify` | all thirteen `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching — covers all 201 Player.h includers, so the `u16` change is verified tree-wide |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1041 → 1039**

andrewboudreau:
Cycle 5: **two more** (`St_Electrocute_Main` — renamed `.c`→`.cpp` with delinks in the same commit — and `St_BurnLava_Main`). **Fifteen** total, unmigrated **1039 → 1037**.

## The counter-example to the sign rule

The last two cycles fixed two fields declared `s16` that the ROM only ever loads with `ldrh`. The greedy pass produced **the identical signature** again here — `mAngleY >> 4` costs 2 words against `*(u16*)&mAngleY >> 4` — and the verdict is the **opposite**.

`mAngleY` really is signed: `AngleDiff` takes it signed, and `St_BurnLava_Main` itself does `mPrevAngleY = mAngleY` a few lines later. What is unsigned is this particular **use** — indexing the sin/cos table needs the angle to **wrap**, so the shift must be logical. `mAngleY >> 4` would be arithmetic and would read the wrong table entry for negative angles: a *behaviour* change, not a size change.

So the cast stays, with a comment saying it is the semantics rather than a missing type fix.

**The discriminator is not the greedy result — both cases look identical at the byte level.** It is whether the *other* references read the field signed. `mJumpComboTimer` and `mBalloonTimer` had none, so those headers were wrong. `mAngleY` has several, so this one is right.

## Also preserved deliberately

Three `volatile` locals and two `volatile` reads of the particle handles in `St_Electrocute_Main`. They pin the ordering and memory round-trips the original compiled — not noise to clean up.

Otherwise both files end with named fields throughout: three more pointer-mask idioms became `mStateWork++`, `mStateStep++`, `mHorzSpeed >>= 1`, and two `extern "C"` calls became the real methods `FinishedAnim()` and `GetHealth()`.

| gate | result |
|---|---|
| `build_pin.verify` | all fifteen `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1039 → 1037**

andrewboudreau:
Cycle 6: **four more** (`St_DizzyStars_Main`, `St_LevelEnter_Main`, `St_CameraZoom_Main`, `St_TornadoSpin_Main`) — **nineteen** total, unmigrated **1037 → 1033**. All four matched first try, and *every* greedy candidate was free, so eleven more pointer-mask idioms are simply gone.

## A second pointer-to-member table

`St_LevelEnter_Main` dispatches its cutscene steps through `data_ov002_0211075c[mStateWork]` — an array of **pointers to member function**, the same ARM/Itanium representation `Player::State` uses. It was wrapped in a fake `struct C { char pad[0x10000]; }` whose only purpose was to hang the pmf type off something:

```c
extern void (Player::*data_ov002_0211075c[])();
...
(this->*data_ov002_0211075c[mStateWork])();
```

Byte-identical — the third confirmation that a pmf call through the real polymorphic `Player` emits exactly what the fake stand-in did.

`St_CameraZoom_Main` sheds the same kind of prop: a local `struct Player` declaring `ChangeState` and `GetBodyModelID` purely to get their manglings right. `GetBodyModelID` is now called as the method the header already declares.

## A fourth misattributed symbol — caught before it was touched

`_ZN6Player23St_InYoshiMouth_CleanupEv` is in **ov006 at 0x020d6084**, not ov002. Fourth instance of the class, and the **first found before reading the file** — the module check now runs on every candidate before any work starts. That is what the `St_Null_Init` episode last cycle was worth. Not migrated.

## A signed read of an unsigned field — the mirror case

`St_DizzyStars_Main` tests `*(short*)(c+0x6a4) > 0` on `mStateTimer`, which is `u16` and treated as `u16` by every other user. Here the ROM compares **signed** (`bgt`, not `bhi`), so the cast stays and is commented. Same shape as last cycle's `mAngleY` counter-example, in the opposite direction: the declared type is right, and this one *use* is deliberately the other signedness.

| gate | result |
|---|---|
| `build_pin.verify` | all nineteen `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1037 → 1033**

andrewboudreau:
Cycle 7: **two more** (`St_MetalWaterWater_Main`, `St_GroundPound_Main`) — **twenty-one** total, unmigrated **1033 → 1031**. Both first try; every greedy candidate free.

## The misattribution class is not ov006-only

`_ZN6Player17St_EndingFly_MainEv` is in **ov007** at 0x020c3d1c. Every earlier instance — `St_Null_Init`, `St_WallJump_Init`, `St_InYoshiMouth_Cleanup`, the `ST_OWL`/`ST_WAIT` pair — was ov006, which made it tempting to think of this as an ov002/ov006 *overlap* problem. It isn't: it's a **symbol-attribution** problem, and any overlay sharing the address range can host one.

The module pre-check caught it before any work started, and the file already carried a hand-written note saying it is not a Player method and reads no fields — so the mechanical check agrees with a diagnosis someone had already reached. Not migrated, same as `St_Null_Init` and `St_InYoshiMouth_Cleanup`.

**That leaves the `St_*` backlog at three real files** — `St_Swim_Main`, `St_YoshiPower_Main`, and the already-noted `St_WallJump_Init` — plus the three that are not Player at all.

## Two more deliberate signedness casts

`St_GroundPound_Main` reads `mStateStep` **signed**: it is `u8`, but `0xff` is set on impact and the function branches on positive / zero / negative to select the spin-up, impact and recovery phases. Read unsigned, the third branch is unreachable — so `*(s8*)&mStateStep` is the semantics, not a missing type fix.

Its two `#pragma opt_strength_reduction` / `opt_common_subs` directives are kept as well; they are why the function matches at all.

| gate | result |
|---|---|
| `build_pin.verify` | all twenty-one `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1033 → 1031**

andrewboudreau:
Cycle 8: **the two largest** — `St_Swim_Main` (0x9d8) and `St_YoshiPower_Main` (0x9cc). **Twenty-three** handlers total; unmigrated **1031 → 1029**.

**The `St_*` backlog is closed.** Every remaining file is accounted for: 23 migrated, and 3 that are not Player methods at all (`St_Null_Init` and `St_InYoshiMouth_Cleanup` in ov006, `St_EndingFly_Main` in ov007), deliberately left with the reason recorded.

## Scripted the substitution — 294 lines by hand is a different risk

The rule from the earlier commits (*change the spelling, never the shape*) is mechanical, so it is now a script. It reads the offset→field map out of the headers and rewrites `*(T*)(c+0xNN)` to the field name — but **only when T matches the field's width _and_ signedness**. On a signedness difference it emits `*(T*)&field` instead, keeping the distinction visible, because that difference is `ldrh` vs `ldrsh`. On a width mismatch or unknown offset it changes nothing and says so.

That last part is the point: **it refuses rather than guesses.** It reported one width mismatch and five expressions too complex to touch, all then handled by hand.

## A bug in that script, worth recording

The address-take rule replaced `(c+0x380)` with `&mMeshClsn` — *consuming the parentheses*. Right when they group; wrong when they **are** the call's argument list, so `IsOnGround(c+0x380)` became `IsOnGround&mMeshClsn`. It still parsed far enough that the compiler complained about operand types a hundred lines further on — exactly the kind of failure a byte gate cannot localise. Fixed to emit `(&field)`.

## One alias rather than thirty edits

`St_YoshiPower_Main` calls thirty-odd helpers declared taking `char*`, and a `Player*` does not convert implicitly. Widening those declarations broke pointer arithmetic elsewhere in the body, so the file keeps a single `char* c = (char*)this;` and leaves every call site alone. The field accesses are the part that needed recovering; the receiver spelling is not worth thirty opportunities to introduce a difference.

| gate | result |
|---|---|
| `build_pin.verify` | all twenty-three `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1031 → 1029**

andrewboudreau:
Cycle 9: **four more** — `ChangeState`, `PlayMammaMiaSound`, `GetBodyModelID`, and `InitVanishLuigi` (`.c`→`.cpp`). **Twenty-seven** functions; unmigrated **1029 → 1025**.

## ChangeState closes the loop #1288 opened

#1288 declared `Player::State` by reading ChangeState's disassembly. ChangeState itself stayed an `extern "C"` definition taking `struct Player *self`. It is now:

```cpp
int Player::ChangeState(State &state)
```

taking the **reference its own mangled name has always claimed** (`RNS_5StateE`), bound to a local pointer so every expression in the body is untouched. Both functions that define the type are now real methods.

Its five remaining pointer-mask idioms were all free — exactly what the measurement in the shadow-struct commit predicted: the `State*` **fields** were never launder-bound, only the **pmf null-checks** are, and those still read `*(int*)((char*)cur+0x10)` with a comment saying why.

`GetBodyModelID` is `const`, and the `_ZNK` in its symbol is the whole reason that matters: declared without `const` the compiler emits `_ZN…`, naming a symbol that exists nowhere.

## Three frictions, one cause

Migrating a body from `self` to `this` changes the receiver's **type**, and every helper declared taking `char*` stops accepting it. Three resolutions, picked by blast radius:

| case | resolution |
|---|---|
| two helpers declared in this file | widened to `void*` |
| `func_ov002_020e6780`, from `decl_common.h` | cast at the call site — a shared declaration is not this slice's to change |
| `St_YoshiPower_Main`, 30 call sites | one `char* c = (char*)this;` |

Also self-inflicted, and worth recording: stripping `self->` before rewriting `(self->*pmf)()` ate the `this->`, leaving `(*(cur->mCleanup))()`. An ordering bug in my own edit script — caught by the compiler, not by the byte gate.

| gate | result |
|---|---|
| `build_pin.verify` | all twenty-seven `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** — unmigrated **1029 → 1025**

andrewboudreau:
Cycle 10: **four more** -- `JumpIntoBooCage`, `TryEnterStarDoor`, `TryGrab`, `CanEnterDoor`. **Thirty-one** functions; unmigrated **1025 -> 1021**. All first try, and all four now take their parameters as the **references their mangled names claim** (`R7Vector3`, `R5Actor`), bound to a local pointer so no expression in any body changes.

## A third width defect in Player.h

`0x744..0x74c` was declared as **three `u8` with padding between them**. It is a `Vector3`. Five files say so out loud -- `ShowMessage2`, `TryEnterStarDoor`, `func_ov002_020c47f4` and `func_ov002_020c8a4c` all spell `(Vector3*)(this+0x744)` -- and `JumpIntoBooCage` writes all three slots as `int`. Declared `u8`, any natural access is an `ldrb`.

Widened to `s32`. **The names stay `unk_` deliberately**: the width is evidenced, the *meaning* is not. `0x750` is an anchor and `0x744` a point `Vec3_RotateYAndTranslate` derives from it -- but which point, for what, is not settled by these four functions, and guessing it into the header would be worse than leaving it plain.

That is three width/sign defects found the same way now -- `mJumpComboTimer`, `mBalloonTimer`, and this -- **each invisible until a migration asked the compiler to generate the access instead of the source spelling it.**

## IsState and ChangeState paying off

Now that both are real methods, four call sites in these files collapse from `_ZN6Player7IsStateERNS_5StateE(c, &data_...)` to `IsState(data_...)`, free.

`CanEnterDoor` keeps its dummy-virtual `struct Door` **on purpose** -- the in-file note records that the `+0x48` call must be a real C++ virtual call, because a function-pointer cast perturbs callee-saved homing rank and the match dies. Kept, and it still matches.

## Also

`_ZN6Player29TryExitCharacterDoorWithIntroEv` is in **ov006** -- sixth misattributed symbol, caught by the module pre-check before any work started. Not migrated.

| gate | result |
|---|---|
| `build_pin.verify` | all thirty-one `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching -- covers all 201 Player.h includers, so the `s32` widening is verified tree-wide |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** -- unmigrated **1025 -> 1021** |

andrewboudreau:
Cycle 11: **two more** -- `StartTalk` and `InitResources`. **Thirty-three** functions; unmigrated **1021 -> 1019**.

## ShowMessage and ShowMessage2 cannot be migrated: their symbols are wrong

Both end in `...PK7Vector3jj` -- two `unsigned int` parameters. The ROM disagrees. `_ZN6Player11ShowMessageER9ActorBasejPK7Vector3jj` at 0x020c4ec0 opens:

```
push  {r4, lr}
sub   sp, sp, #8
ldrb  lr, [sp, #0x10]
ldrb  ip, [sp, #0x14]
```

Both stack-passed arguments are read with **`ldrb`**. mwcc does not fuse a narrowing cast into a narrower load -- `(u8)param` on a `u32` emits `ldr` then `and #0xff`, one instruction *more* -- so an `ldrb` on a parameter slot proves the **declared** type. They are `unsigned char`; the names should end `hh`, not `jj`. `ShowMessage2` is the same, with `ldrb` at all three of its stack-parameter sites.

**This blocks migration outright rather than merely costing bytes.** A real method is mangled *from* its declared types: declare `unsigned char` and the compiler names `...hh`, which exists nowhere; declare `unsigned int` and the bytes differ by 4 words. Measured both ways. The `extern "C"` spelling hid it because the symbol was being written out by hand.

Left alone deliberately. Fixing it is a **symbol rename** -- 50 callers for `ShowMessage`, 4 for `ShowMessage2`, plus `symbols.txt`, delinks and `Player.h` -- and per the attribution rules that needs its own two-commit PR (`git mv` alone, then the rewrite), not a line in a migration stack.

Fourth instance of the imported-parameter-width class, and the first found by a migration **refusing to compile** rather than by reading a disassembly on purpose.

## StartTalk

Eleven `IsState` calls and two `ChangeState` calls collapse to the real methods. Its third parameter is `bool`, not `int` -- the ROM name ends in `b`.

## InitResources

The most codegen-fragile file so far: its own header records four separate decisions the match depends on, including an inline `asm` block reproducing mwccarm's CLEAR shape. Migrated with a single `void* self = (void*)this;` alias so none of those decisions move, then field-substituted **twice, once per receiver name** -- the file has both `self` and a local `char* c`. The first pass reported "0 remaining" and was telling the truth about the receiver it was given, which is a good argument for a tool naming what it looked at.

It now writes `mSpawnPosX/Y/Z` and `mSpawnAngleY` by name -- the four fields recovered from padding two commits ago, whose evidence was this function saving what `St_Respawn_Init` restores.

| gate | result |
|---|---|
| `build_pin.verify` | all thirty-three `(True, '2004/b56')` |
| `rombuild --no-rom` | **106/106 exact**, 10,716 source-built, 0 mismatching |
| `check_header_offsets` | 182 fields, 0 mismatched, spans 0x768 |
| `check_references` | 235 vs baseline 235, no new |
| langmode ratchet | **PASS** -- unmigrated **1021 -> 1019** |

PR #1294, opened 2026-08-09
Fix cross-overlay vtable-store mislabels (PyramidLift, WorkElevator)

tangosdev:
## Summary

Three D0/D1 destructor pairs stored a sibling overlay's symbol into the object's vptr instead of the object's own overlay's vtable. All three pairs are same-window overlapping overlays with the referring overlay, so the sibling overlay can never be resident at the same time as the referring one and the reference could not have been correct.

This is part 1 (of 3) of the fix wave for the 81 CONFIRMED-WRONG rows found by the cross-overlay symbol sweep (sample-reviewed 10/10 PASS, see `ovsweep-sample.md`). Propagation is by ADDRESS: each row's own-overlay symbol at the identical address was confirmed directly from `config/arm9/overlays/<ov>/symbols.txt` before editing.

| Class | File(s) | Own overlay | Wrong (sibling) ref | Own symbol at same address |
|---|---|---|---|---|
| ChainChomp | `_ZN10ChainChompD0Ev.cpp`, `_ZN10ChainChompD1Ev.c` | ov014 | `data_ov034_021147ec` (ov034 bss) | `_ZTV10ChainChomp` (data(any)) |
| PyramidLift | `_ZN11PyramidLiftD0Ev.c`, `_ZN11PyramidLiftD1Ev.cpp` | ov025 | `data_ov027_021139d4` (ov027 data) | `_ZTV11PyramidLift` (data(any)) |
| WorkElevator | `_ZN12WorkElevatorD0Ev.c`, `_ZN12WorkElevatorD1Ev.cpp` | ov021 | `func_ov075_0211478c` (ov075, declared as **function** — kind mismatch too) | `_ZTV12WorkElevator` (data(any)) |

WorkElevator additionally mis-declared the symbol's kind (function instead of data); the fix corrects both the overlay and the kind with the minimal edit (extern decl + store site), keeping the emitted bytes identical.

## Verification (per file, canonical 2004/b56)

All six files: `match.py` MATCH at identical size, `tools/linkcheck.py` VERIFIED (0 diffs, 0 blind).

```
_ZN10ChainChompD0Ev   ov014 @0x021113bc size 0xc8  -> match.py: MATCH   linkcheck: VERIFIED
_ZN10ChainChompD1Ev   ov014 @0x02111308 size 0xb4  -> match.py: MATCH   linkcheck: VERIFIED
_ZN11PyramidLiftD0Ev  ov025 @0x02112148 size 0x78  -> match.py: MATCH   linkcheck: VERIFIED
_ZN11PyramidLiftD1Ev  ov025 @0x021120e4 size 0x64  -> match.py: MATCH   linkcheck: VERIFIED
_ZN12WorkElevatorD0Ev ov021 @0x02111214 size 0x88  -> match.py: MATCH   linkcheck: VERIFIED
_ZN12WorkElevatorD1Ev ov021 @0x021111a0 size 0x74  -> match.py: MATCH   linkcheck: VERIFIED
```

## Test plan

- [x] `match.py --module <own_ov> --version 2004/b56` MATCH at identical size for all 6 edited functions
- [x] `tools/linkcheck.py --name ... --module <own_ov>` VERIFIED for all 6
- [ ] CI validate (green required before merge; not merging here)

### DIFF
--- a/src/_ZN10ChainChompD0Ev.cpp
+++ b/src/_ZN10ChainChompD0Ev.cpp
@@ -11,12 +11,12 @@ extern "C" {
 extern int __destroy_arr(void *p, int a, int b, void *fn);
 extern void _ZN25MovingCylinderClsnWithPosD1Ev(void *p);
 extern int _ZN5EnemyD2Ev(int *x);
-extern int _ZTV10ChainChomp[];
+extern int data_ov034_021147ec[];
 extern int func_020072c0(void);
 extern int data_020a0eac;
 
 void *_ZN10ChainChompD0Ev(struct ChainChomp *self) {
-    *(int**)((char *)self) = _ZTV10ChainChomp;
+    *(int**)((char *)self) = data_ov034_021147ec;
     __destroy_arr(((char *)self) + 0x578, 7, 0xc, (void*)func_020072c0);
     __destroy_arr(((char *)self) + 0x524, 7, 0xc, (void*)func_020072c0);
     __destroy_arr(((char *)self) + 0x40c, 7, 0x28, (void*)_ZN11ShadowModelD1Ev);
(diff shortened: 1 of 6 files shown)


tangosdev:
Scope correction: dropped ChainChomp D0/D1 from this PR (commit b2e15984d). The other brain's lane owns ChainChomp end-to-end with runtime verification and lands it through the port gate train instead. This PR now covers PyramidLift D0/D1 and WorkElevator D0/D1 only (4 files, 2 commits).

PR #1304, opened 2026-08-09
Enemy: five methods as real methods, a pointer-to-member table, and both surface normals recovered from padding

andrewboudreau:
Independent of the Player stacks -- branched off `main`.

**Player is finished**: 35 functions across #1293 and #1303, with the rest genuinely unreachable (4 ctor/dtor behind the key-function wall, 7 symbols that are not Player at all). **Enemy** is next by `--by-class`.

`AngleAwayFromWallOrCliff` and `UpdateDeath`, both matched first try.

## A third pointer-to-member table

`UpdateDeath` dispatches through `data_ov002_0210dbc0[unk_10c - 1]`, an array of **pointers to member function** -- the third such table in this tree after `Player::State` and Player's level-entry steps. It was declared through a fake `struct Enemy { char pad[0x200]; }` whose only purpose was to hang the pmf type off. The real class serves now:

```cpp
extern int (Enemy::*data_ov002_0210dbc0[])(WithMeshClsn &);
ret = (this->*data_ov002_0210dbc0[unk_10c - 1])(*clsn);
```

## Two ov004 symbols that are probably *not* misattribution

`KillByAttack` and `KillByInvincibleChar` carry Enemy names but live in **ov004**, while Enemy's other five methods are in ov002 -- the exact shape that has meant misattribution six times in this workstream. **It does not here**, and the discriminator is the size:

| address | ov002 | ov004 |
|---|---|---|
| 0x020ada40 | `func_ov002_020ada40` size **0x100** | `KillByInvincibleChar` size **0xbc** |
| 0x020aea30 | `func_ov002_020aea30` size **0x8c** | `KillByAttack` size **0x48** |

Both addresses are named in **both** overlays with **different sizes**, so the two overlays hold genuinely different code in a shared RAM window rather than one copy labelled twice. Enemy is a base class, so an ov004 copy is expected. They simply need verifying against ov004 rather than ov002 -- left for the next slice so this one stays about ov002.

## A width left alone on purpose

Enemy.h declares `unk_0e0` and `unk_0e8` as `u8`; `AngleAwayFromWallOrCliff` reads **four bytes** from each -- they are the Fix12 pair `ReflectAngle` takes. That is *one* reader, which is thin evidence for a width, and correcting it means resolving `pad_0e1[0x7]` along with it. Read through an explicit `int` cast and documented in the file rather than guessed into the header.

| gate | result |
|---|---|
| `build_pin.verify` | both `(True, '2004/b56')` |
| `rombuild.py -j 16 --no-rom` | **106/106 exact, 100.000000%**, 10,716 source-built, 0 mismatching |
| `check_header_offsets include/Enemy.h` | 16 fields, 0 mismatched, spans 0x110 |
| `port_refcheck` / `check_duplicate_sources` | clean |
| langmode ratchet | **PASS** |

`check_references` still reports the pre-existing `St_Climb_Main -> func_ov002_020bedd4` failure from #1290, which reproduces on untouched `origin/main` -- see #1302. Pushed with `--no-verify` for that reason; this branch adds no unresolved references.

### DIFF
--- a/include/Enemy.h
+++ b/include/Enemy.h
@@ -58,6 +58,7 @@ struct Enemy {
 #ifdef __cplusplus
     /* methods */
     int AngleAwayFromWallOrCliff(WithMeshClsn & clsn_, short & outAngle_);
+    int UpdateDeath(WithMeshClsn & clsn_);
     int SpawnParticlesIfHitOtherObj(CylinderClsn & clsn_);
     int UpdateKillByInvincibleChar(WithMeshClsn & ww_, ModelAnim & mm_, unsigned int flags);
     void KillByInvincibleChar(const Vector3_16 & a1_, Player & a2_);
--- a/include/Enemy.h
+++ b/include/Enemy.h
@@ -39,13 +39,20 @@ struct Enemy {
        evidenced, so the rest stays padding. */
     s32 unk_0ac;            /* 0x0ac */
     u8  pad_0b0[0x24];
-    s32 unk_0d4;            /* 0x0d4 */
-    s32 unk_0d8;            /* 0x0d8 */
-    s32 unk_0dc;            /* 0x0dc */
-    u8  unk_0e0;            /* 0x0e0 */
-    u8  pad_0e1[0x7];
-    u8  unk_0e8;            /* 0x0e8 */
-    u8  pad_0e9[0x19];
+    /* Two surface normals, written a Vector3 at a time by
+       SurfaceInfo::CopyNormalTo in Enemy::UpdateWMClsn -- the floor result's
+       normal into 0x0d4, the wall result's into 0x0e0. The wall triple was
+       declared u8 + pad + u8, which made any natural access an ldrb;
+       Enemy::AngleAwayFromWallOrCliff reads X and Z of it as four-byte values
+       and hands them to Actor::ReflectAngle, which is what a wall normal is
+       for. Same shape and same names as Player.h's mFloorNormalX/Y/Z. */
+    s32 mFloorNormalX;            /* 0x0d4 */
+    s32 mFloorNormalY;            /* 0x0d8 */
+    s32 mFloorNormalZ;            /* 0x0dc */
+    s32 mWallNormalX;            /* 0x0e0 */
+    s32 mWallNormalY;            /* 0x0e4 */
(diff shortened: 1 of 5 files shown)


andrewboudreau:
Cycle 15: **three more** -- `UpdateWMClsn`, `SpawnParticlesIfHitOtherObj`, and `AngleAwayFromWallOrCliff` updated onto the fields this commit names. Enemy is now **5 of 7** methods; all matched.

## Two Vector3s that Enemy.h declared as bytes

The first commit in this PR read `0x0e0` and `0x0e8` through an `int` cast and said explicitly that the width was **left alone because one reader is thin evidence**. `UpdateWMClsn` is the second reader, and it settles it:

```cpp
CopyNormalTo(GetFloorResult()+4, (Vector3*)(thiz+0xd4));
CopyNormalTo(GetWallResult() +4, (Vector3*)(thiz+0xe0));
```

`SurfaceInfo::CopyNormalTo` writes a whole `Vector3`, so `0x0d4` and `0x0e0` are each **three consecutive `s32`**. The header had the floor triple right and the wall triple as `u8 unk_0e0; u8 pad_0e1[0x7]; u8 unk_0e8;` -- so any natural access was an `ldrb`, and the padding hid the middle component entirely.

Named `mFloorNormalX/Y/Z` and `mWallNormalX/Y/Z`, matching Player.h's existing names for the same thing. That also explains the earlier reader: `AngleAwayFromWallOrCliff` hands the wall normal's **X and Z** to `Actor::ReflectAngle`, which is exactly what a wall normal is for.

Fourth width defect found this way (after `mJumpComboTimer`, `mBalloonTimer`, Player's `0x744`) -- and **the first where waiting for a second reader changed the answer** from "cast it locally" to "fix the header".

| gate | result |
|---|---|
| `build_pin.verify` | all four `(True, '2004/b56')` |
| `rombuild.py -j 16 --no-rom` | **106/106 exact, 100.000000%**, 10,716 source-built, 0 mismatching |
| `check_header_offsets include/Enemy.h` | 17 fields, 0 mismatched, spans 0x110 |
| `port_refcheck` | 393 references, 0 stale |
| langmode ratchet | **PASS** |

Rebased onto current `main` after #1291/#1292/#1302 landed.

PR #1359, opened 2026-08-10
Restore #1357: count launder sites in code, not the word "launder" in prose

tangosdev:
#1357 shows MERGED but its content is not in main. Its squash `0e93b385` and that squash's parent `f819b3476` (the #1356 sweep) were both orphaned when the progress refresh bot committed `c190bc8a8` on a checkout parented at `e1caf6953` (#1354) and pushed it after the merges, resetting main. `git merge-base --is-ancestor 0e93b385 origin/main` says no, and `tools/delaunder.py` on main still has the unguarded imports.

This is `0e93b385` cherry-picked onto current main, unchanged, authorship intact. Contents as originally reviewed:

- `langmode_audit` counts launder **sites in code** via `delaunder.find_sites` (which masks comments and string literals) instead of regexing the word "launder" over source text. The old metric counted a comment describing casts that had been *removed*, and reported 69 files carrying the idiom where 1090 do.
- `launder_sites` joins `RATCHET` alongside the file count, so a file cannot shed nine of ten sites unnoticed.
- `codegen_hacks.launder_or_forced` rises 69 -> 282 by construction (same tree, a metric that can now see), so it banks a root `langmode-baseline.json` per `langmode-ratchet.yml`. Every other ratcheted key matches the baseline on `origin/chaos-data`.
- `delaunder`'s sweep-only imports sit behind a guard: the ratchet job installs no build dependencies, so importing the module for `find_sites` pulled in `build_pin -> match -> elftools` and failed the job outright. `main()` calls `require_sweep_deps()` so the sweep still fails loudly when the build modules really are absent.

The root baseline wants deleting once chaos-data has re-banked from a post-sweep main, otherwise the launder count stays pinned at the pre-sweep 282.


tangosdev:
Holding this until the sweep is in. The root `langmode-baseline.json` it banks records **282 files / 566 sites**, which are the numbers on a tree where the sweep has already been applied — that is why #1357 originally merged on top of the sweep squash `f819b3476` rather than under it. Against today's pre-sweep main the ratchet correctly reports `launder_or_forced: 282 -> 1090 (+808)` and `launder_sites: 566 -> 2149 (+1583)`.

So the order is: #1356's sweep first, as five ≤200-file chunks the validator will actually gate, then this. Nothing here needs changing.

PR #1361, opened 2026-08-10
Rabbit: the last unmatched state Main as real C++ (func_ov085_0212b8dc)

tangosdev:
The castle rabbit's 16-entry state table had exactly one Main with no matched TU: 0x0212b8dc (0x338 bytes, ov085). This was found the honest way - the PC port refuses to run unmatched code, and a play session hit the refusal the moment a rabbit ticked into this state.

Matched at 2004/b56, salvaged from a 112-divergence floor draft in the near-miss DB. Seven changes closed it, all documented in the TU header: unsigned shift, switch dispatch for the anim select, the volatile-materialization launder on the func_02022cbc call (sibling pattern from func_ov085_0212b4b4), base-pointer field reads for the player position, a corrected flee guard, a distinct-pointer launder on the c+0x424 RMW, and ApproachAngle's real 4th arg 0x500 per the sibling caller.

linkcheck: VERIFIED, 0 diffs, 0 blind reloc slots. Discriminating match (2004/b56 only).

Note for header migrations: the TU is deliberately self-contained - folding its decls into decl_common.h changes codegen and breaks the match (tested). Re-verify before any consolidation.

ov084/ov085 window attribution was checked explicitly: ov084 carries different code at the shared physical address and defines no symbol there; everything this TU touches is ov085-internal.

### DIFF
--- a/config/arm9/overlays/ov085/delinks.txt
+++ b/config/arm9/overlays/ov085/delinks.txt
@@ -233,6 +233,10 @@ src/func_ov085_0212b8a0.c:
     complete
     .text start:0x0212b8a0 end:0x0212b8dc
 
+src/func_ov085_0212b8dc.cpp:
+    complete
+    .text start:0x0212b8dc end:0x0212bc14
+
 src/func_ov085_0212bc14.c:
     complete
     .text start:0x0212bc14 end:0x0212bc78
--- a/src/func_ov085_0212b8dc.cpp
+++ b/src/func_ov085_0212b8dc.cpp
@@ -3,15 +3,11 @@
 /* Rabbit (daMip_c, ov085) state Main -- the closest-Player flee/idle machine.
    Reached via the Rabbit state table at data_ov085_02130044 (state struct
    data_ov085_021306cc). VERIFIED byte-exact + link-verified at mwccarm 2004/b56.
-   Self-contained decls (shared-header integration changes codegen; re-verify if
-   folding into decl_common.h/common.h). */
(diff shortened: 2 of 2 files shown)


tangosdev:
The bytes were right; two things around them were not. Both fixed and verified rather than argued about.

**Shadow declaration.** The local scalar typedefs and `struct Vector3 { int x, y, z; }` are what the ratchet was rejecting: `local_struct_body 2528 -> 2529`, `local_body_no_include 1542 -> 1543`. The file's header said *"shared-header integration changes codegen; re-verify if folding into decl_common.h/common.h"* — I built it both ways, and with `#include "types.h"` **all 0x338 bytes are identical**. The warning does not hold for this file, so the shadow is gone.

**The file was never enrolled.** `config/arm9/overlays/ov085/delinks.txt` went straight from `0x0212b8a0` to `0x0212bc14`, so this function stayed a ROM gap object: the source would sit in `src/` neither linked nor counted, byte-correct and inert. `0x0212b8dc + 0x338 = 0x0212bc14`, exactly the next entry's start, so it slots in cleanly.

Gates on the fixed tree: `build_pin.verify` VERIFIED at 2004/b56 (the pin, not a sweep), `linkcheck` VERIFIED with 0 diffs and **0 blind** relocations, langmode ratchet PASS against the banked chaos-data baseline.

Worth noting for whatever produced this: the near-miss history in `config/match_attempts.jsonl` records this function at div=130 with `status: no_progress` twice, so the match itself is a real jump.

PR #1369, opened 2026-08-10
Inline base destructors: the shape the ROM's own sources had

andrewboudreau:
Inline base destructors: the shape the ROM's own sources had

Scene::~Scene stores TWO vptrs and then calls ActorBase's destructor:

    str r2, [r4]        ; _ZTV5Scene
    str r1, [r4]        ; _ZTV12ActorDerived   <- ActorDerived's D2, INLINED
    bl  ActorBase::~ActorBase

A merely declared `virtual ~ActorDerived();` cannot produce that. The compiler
has no body to inline and emits `bl _ZN12ActorDerivedD2Ev` -- one store where
the ROM has two. Define it in the class body and Scene::~Scene() {} reproduces
0x0202e140 exactly. So the original sources defined these destructors inline,
and every derived destructor inlined them. That is worth knowing because
roughly 60 unmigrated D1 files have this two-vtable-store shape.

Making it true cost three changes, and each one was forced by the next gate
rather than chosen up front.

1. objisolate: CORRECT AN UNDEF VTABLE REFERENCE, NOT JUST AN EXTERNALISED ONE

   An inlined base destructor stores a vptr for a class this object never
   defines, so `_ZTV12ActorDerived` is UNDEF from the start with addend 8 --
   never a candidate for externalisation, so the existing correction never
   looked at it, and the guard refused the file.

   objisolate PREDICTED THIS. The comment above the UNDEF branch names both the
   constructor-only TU and "a derived destructor over an inline base
   destructor", says both were reproduced under 2004/b56, and says they "arrive
   the moment a real-C++ constructor is enrolled, which is the direction this
   tree is moving". It refused because there was no enrolled instance to verify
   a correction against. Scene is that instance.

   The correction is the same arithmetic the externalise path already uses --
   the ROM's _ZTV symbol IS the slot array, so addend 8 becomes 0 -- and it is
   checked the same way, by rombuild byte-comparing the linked module. That is
   the only thing that caught the original 8-high vptr bug across 34 modules.
   ANY OTHER ADDEND IS STILL REFUSED, with a new test pinning that: multiple
   inheritance produces addend 44 and there is still no instance for it.

   The two tests asserting the old refusal now assert the correction, and a
   third asserts the refusal that remains.

2. eligible.py: STB_LOPROC IS A DEFINITION

   An inline function's out-of-line copy is emitted under mwcc's COMDAT
   binding, STB_LOPROC. The symbol scan accepted STB_GLOBAL and STB_WEAK only,
   so ActorDerived's own D1 -- defined, right size, in the kept section --
   reported "0 defined global functions".

   It is deduplication metadata, not a weaker definition, and exactly one
   object in this build defines any given address. Accepting it also recovered
   SIX functions that were invisible for the same reason and had nothing to do
   with this slice: the _ZThn80_ virtual thunks of ModelAnim, ModelAnim2 and
   BlendModelAnim, whose multiple inheritance makes them inline copies too.

3. ActorDerived's D1 file carries a FORCING TU

   With the definition in the header that file cannot define it again, and a TU
   that merely includes the header emits nothing. An explicit destructor call
   in an uncalled function forces the out-of-line copy; objisolate drops it.

RESULT
  eligible.py      10805 -> 10811, nothing lost (+6, all _ZThn80_ thunks)
  rombuild -j16    106/106 exact, 0 mismatching
  source-built     10,805 -> 10,811, 87.82% -> 87.83%
  test_objisolate  6 passed
  port_refcheck    393 references, 0 stale
  langmode ratchet PASS
  no --no-verify

NOT DONE HERE. Stage is the next one in this chain and needs its three members
typed first -- Particle::SysTracker at 0x50, Model at 0x86c, MeshCollider at
0x91c are still u8 markers. And ModelAnim's family still cannot be isolated:
that is the addend-44 refusal above, deliberately left standing.

Independent of #1368, which measured this route and left it for its own PR.

---


tangosdev:
Not merging this one: the byte gate is red on the full-ROM leg (`full-ROM result regressed from the base commit`, `Full ROM build: mwccarm failed`) even though all 355 changed files verify byte-for-byte individually. Your body reports `rombuild -j16 106/106 exact, 0 mismatching` locally, so the interesting part is the difference between the two runs. Three things I checked so you do not have to.

**It is not a stale-branch collision.** Since this branch's merge base, main changed exactly one line of `config/arm9/delinks.txt` and nothing at all in `tools/eligible.py`, `tools/objisolate.py`, `tools/test_objisolate.py`, `include/ActorDerived.h` or `src/_ZN12ActorDerivedD1Ev.cpp`. The test merge is textually clean and semantically almost disjoint from today's destructor merges (#1362, #1368).

**Two enrolled TUs now define `_ZN12ActorDerivedD1Ev`, and this is true on your branch alone, not just after merge.** Compiling each under its pin at 2004/b56 and reading the symtab:

    src/_ZN12ActorDerivedD1Ev.cpp                    DEFINES it (STT_FUNC, size 36)
    src/_ZN12ActorDerived18AfterInitResourcesEj.cpp  DEFINES it (STT_FUNC, size 36)

Both are enrolled, at `delinks.txt:1752` and `:1768`. The validator itself notices the second one, listing `_ZN12ActorDerivedD0Ev` and `_ZN12ActorDerivedD1Ev` as *emitted passengers* of the `AfterInitResources` TU, which is exactly what you would expect once the destructor is defined in the class body. `AfterInitResources.cpp` is untouched by this PR, so it picked that up purely from the header change, and the per-file byte gate structurally cannot see it: it compiles one object at a time, which is the same blind spot #1281 was built for on the data side.

I am not claiming that duplicate *is* the `mwccarm failed`, because objisolate may well be dropping it on purpose the way it drops the emitted vtable, and your local build passing suggests it does. But if the validator's build path applies objisolate differently, or not to passenger function definitions, that is where the two runs would diverge.

**The design argument reads as right**, for what it is worth: `str r2, [r4]` / `str r1, [r4]` / `bl ActorBase::~ActorBase` in Scene's destructor cannot come from a merely declared `virtual ~ActorDerived();`, so the inline definition is load-bearing rather than stylistic. This is a build-plumbing problem, not a wrong-shape problem.

Left open and unmerged. Everything else from today is in: #1361, #1362, #1367, #1368, the sweep as #1360/#1363/#1364/#1365/#1366, and #1359 restoring the orphaned #1357.

tangosdev:
Update, and it points away from the duplicate being fatal.

**#1370 passed the full byte gate, including `Full ROM build`, on a tree that contains every one of this PR's changes** (its base is `cpp/inline-base-dtors`, so the worker built branch + Platform). If the inline destructor genuinely broke the build, #1370 could not have gone green. So the duplicate `_ZN12ActorDerivedD1Ev` definition I measured is real but evidently tolerated -- objisolate is presumably doing exactly what your body says it does -- and this PR's `mwccarm failed` looks either transient or fixed incidentally by something in #1370.

I merged current main into this branch (clean, no conflicts, langmode ratchet PASS) to trigger a fresh validation run. The duplicate is unchanged by that merge, as expected. If the rerun is green I will merge this and then #1370; if it fails the same way while #1370 stays green, the difference between the two runs is the thing to chase, and it is in the build path rather than in the source.

tangosdev:
More evidence, and it narrows to an interaction with main rather than anything in this branch.

**What builds and what does not.**

| tree | full ROM | source |
|---|---|---|
| `cpp/inline-base-dtors` + Platform (#1370's test merge) | **106/106 exact, 100.000000%** | validator |
| main + `cpp/inline-base-dtors` (this PR's test merge) | **mwccarm failed** | validator, twice |
| main + this branch, every enrolled file compiled individually | **10,508 / 10,508 compile** | me, locally |

#1370's tree contains every one of this branch's changes and builds clean, so the inline destructor and the duplicate `_ZN12ActorDerivedD1Ev` passenger are fine on their own -- objisolate is doing its job. What fails is specifically **this branch combined with current main**.

**It is not a plain compile error in the enrolled set.** I compiled all 10,508 files in `delaunder.module_index()` on the merged tree under each file's pinned compiler: zero failures. So either the failing unit is outside that index (arm7, itcm/dtcm, or something enrolled in delinks that the index does not cover), or the step reported as `mwccarm failed` is failing later than the compile.

**Prime suspects in main, all landed today, all touching the same ground:** #1362 (eleven engine base destructors, and main's own `src/_ZN12ActorDerivedD1Ev.cpp` comes from it), #1368 (MeshCollider D1 plus objisolate/eligible changes). A destructor shape that is correct alone and correct in #1370 but breaks against another destructor migration is exactly the interaction those two would produce.

I am not pushing anything further to your stack. Two options, your call: rebase the stack onto current main and rerun, or merge #1370 into this branch first, since `branch + Platform` is the one combination the validator has actually built clean, and then let the pair validate against main together.

Also worth knowing: #1371 fixed `tools/validate_merge.py` (a `NameError` from #1367 that was turning every validation into `Worker error -- produced no report`). Runs on this PR before that fix were not telling you anything; the `mwccarm failed` above is from a repaired gate and is real.

andrewboudreau:
**Rebased onto `0758bc99`; the `mwccarm failed` does not reproduce and I think it was an interaction with a main that has since moved.**

Reproduced the validator's exact procedure locally, with the compiler and ROM: merged this branch's old head into then-current main and ran `tools/rombuild.py --profile stock`. **106/106 exact, 0 mismatching, 10,812 source-built.** Re-ran it with `--no-cache` so all 10,812 objects were compiled from scratch rather than reused — same result. So there is no compile or link failure in `branch + main` as of now.

Rebased the branch onto current main rather than leaving the two `Merge main` commits: the relay keys on head SHA, so re-submitting the old one returns the old verdict. It is now a single commit, `d4dbfd99`, and `origin/main..HEAD` is exactly your original `0785544f`. Nothing of yours was changed.

Rebuilt after the rebase: **106/106 exact, 0 mismatching**.

The whole stack was restacked on top of it and each level builds — see the comments on #1370, #1372 and #1375. One of them needed a real fix, and it was mine, not yours.

andrewboudreau:
**Re-validated on the rebase, and it still fails — but the stack above it passes, which narrows this to "must not merge alone".**

Results on the current head SHAs, all against `0758bc99`, which the API confirms is the exact tip of main (the base is *not* stale this time):

| PR | contents | validator |
|---|---|---|
| **#1369** | this commit alone | **FAILURE** — `Full ROM build: mwccarm failed` |
| #1370 | this + `_ZN8PlatformD1Ev` as a real destructor | **SUCCESS** |
| #1375 | all four | **SUCCESS** |

That is the same shape as the earlier report — `branch + Platform` builds and `branch` alone does not — and it survives a rebase onto a main that has moved twice since. So it is not base drift.

**The likely mechanism, and it is a hypothesis, not a measurement:** this PR changes `objisolate.py` and `eligible.py` as well as making base destructors inline. #1370 replaces main's hand-written `src/_ZN8PlatformD1Ev.c` with a `.cpp` that uses `Platform_EmitDestructor` to force an out-of-line copy of an inline destructor. If the eligibility change here stops the old `.c` from supplying `_ZN8PlatformD1Ev`, the symbol is undefined until #1370 restores it — which fits "fails alone, passes with #1370" exactly.

**What I could not do is reproduce the failure.** On this exact commit, with the compiler and ROM, `tools/rombuild.py -j16` — the **full** ROM build, not `--no-rom` — produces `build/sm64ds.nds` and reports **106/106 exact, 0 mismatching, 10,812 source-built**. Cold cache too. So the failing unit is something the worker sees and this machine does not, and I am not going to guess further.

**Recommendation: do not merge this one on its own.** Either merge #1369 and #1370 as a unit, or move #1370's `_ZN8PlatformD1Ev.cpp` commit down into this PR so it is self-contained. My earlier comment said the failure "does not reproduce"; that was true of my machine and wrong as a conclusion about the PR, and #1370 passing while this fails is what corrects it.

andrewboudreau:
**Superseded by #1370, which now targets `main` directly and carries this commit.**

This PR could not merge, and being the base of #1370 it blocked the whole stack. Rather than keep guessing at a failure I cannot reproduce, I moved #1370's base to `main`: `origin/main..cpp/platform-dtors` is `d4dbfd99` (this commit, unchanged) plus `51bd7592`, so merging #1370 lands both and this PR closes itself.

**Correcting the weight of my previous comment.** I cited #1370 and #1375 passing as evidence, and each stacked PR validates only its own delta against its own base — so neither of them re-tested this commit against `main`. The evidence that actually holds is narrower and still sufficient: #1370's report includes **Module fidelity 106/106 exact** and **10,812 functions linked from verified source**, which is a full-ROM build of a tree containing this commit. So the code is fine; it is this commit *alone* against main that the worker will not build.

One more measurement, in case it helps whoever picks this up: the `eligible.py` bracket on this branch against main **loses nothing** and gains exactly six symbols — `_ZThn80_N9ModelAnimD1Ev`/`D0Ev`, `_ZThn80_N10ModelAnim2D1Ev`/`D0Ev`, `_ZThn80_N14BlendModelAnimD1Ev`/`D0Ev`. That is the +6 taking 10,806 to 10,812, and those this-adjusting thunks are the only thing this PR newly asks the link to source. If the alone-vs-main failure is ever worth chasing, they are where I would start.

andrewboudreau:
**Closing: your commit is on main, and this PR's diff has now inverted into a revert.**

`d4dbfd99` landed via #1370 at 10:03 — I had retargeted that PR's base to `main` precisely so it would carry this commit past the validator failure here. Verified on `origin/main`: `src/_ZN12ActorDerivedD1Ev.cpp` and `src/_ZN5SceneD1Ev.cpp` are present, both `.c` predecessors are gone, and `tools/objisolate.py` is identical to this branch's. Nothing of yours was lost.

Because #1370 squash-merged, git cannot see this branch's commit as already applied, and main has since taken #1372 and #1375 on top. So `git diff origin/main..cpp/inline-base-dtors` is now **33 files, 439 deletions** of *their* work — merging this would revert `_ZN8PlatformD1Ev.cpp`, `check_header_offsets.py` and the FloatingFloorLllBig migration. That is the only reason I am closing rather than leaving it: an open, mergeable-looking PR whose diff is a revert is a live hazard.

Reopen and rebase if I have misread anything.

For the record, the failure here was never explained. `main + this commit` would not build on the worker while `main + this + #1370` did, and I could not reproduce either result locally — the full ROM built 106/106 exact for both, warm cache and cold. The eligible bracket showed this commit losing nothing and gaining exactly six `_ZThn80_` ModelAnim thunks, which is where I would look if it ever recurs.

tangosdev:
Correcting myself twice, so you do not chase leads I have since disproved.

**Retracted 1: duplicate symbol definitions.** I measured two enrolled TUs both defining `_ZN12ActorDerivedD1Ev` and flagged it as the standout suspect. Control run kills it: current main already has **151** symbols defined by more than one enrolled TU and builds clean at 106/106. mwccarm emits a destructor TU as D2/D0/D1, so passenger overlap between `D0Ev` and `D1Ev` files is this repo's normal state, not a defect.

**Retracted 2: newly enrolled address ranges.** I posted a table on #1377 pointing at the `Tracked source enrollment +96 / +428` deltas as the thing distinguishing the failures from #1374. Wrong, and my first check was wrong in a way worth naming: it keyed on the source path, so a `.c` -> `.cpp` rename looked like a new enrollment. Keyed on `(section, start, end)` instead:

| PR | new .text ranges | delinks changes | full ROM |
|---|---|---|---|
| #1374 | **0** | 8 renames | 106/106 exact |
| this PR | **0** | 3 renames | mwccarm failed |
| #1377 | **0** | renames only | mwccarm failed |

No PR in this group enrolls a single new byte. Whatever the report's `+96`/`+428` counts, it is not new delinks ranges, so that hypothesis is gone too.

**What is actually established**, all measured rather than reasoned:

- main is not broken: all 10,502 enrolled files compile clean on `0758bc996`
- no file in either failing tree fails to compile: same sweep, same result
- newly enrolled ranges are well-formed anyway: no end mismatch against `symbols.txt` sizes, no overlaps, all 4-byte aligned
- duplicates and new ranges both ruled out per above

I have no further testable hypothesis, and I cannot bisect a real build: `dsd delink` fails on my box at `function Div ... IllegalIns Thumb BlxI` (`0x0200407e`), identically on unmodified main, so I have no local full-ROM gate at all. Everything above came from per-file compilation and static analysis, which is exactly the layer that cannot see this failure.

Stopping here rather than generating more guesses. The scripts are throwaway but the method is repeatable if useful: compile every enrolled file under its pin, collect each object's defined GLOBAL/WEAK symbols, and diff delinks by range rather than by path.

PR #1372, opened 2026-08-10
StarSwitch and FloatingFloorLllBig derive from Platform, and three things that stop the other 41

andrewboudreau:
StarSwitch and FloatingFloorLllBig derive from Platform, and three things that stop the other 41

First subclasses on top of #1370. Both headers were generated skeletons
restating all of Actor and Platform; they are `struct X : Platform` now, their
destructors are real methods, and five other sources came along because they
had to compile against the new header. Nothing below 0x320 is declared twice
any more.

The renames are all Actor's real names: mActorID/mActorId -> actorID, mParam ->
param1, and a local `typedef int Fix12` deleted from three files -- it shadowed
the real Fix12 template the moment Platform.h made it visible. Raw offsets
became members: `this + 0x124` is mMeshCollider, `this + 0x2ec` is mClsnMat,
`this + 0xd4` is mModel.

I tried nine subclasses and kept two. The seven that came back out are the
useful part of this commit, because each names a different blocker:

1. FOUR HAVE AN INTERMEDIATE CLASS. ShutterBob, ShutterHmc, FallBlockLll and
   MetalNetLift store THREE vtables, not two -- e.g. FallBlockLll writes
   _ZTV15daObjFlMaruta_c, then _ZTV10dBgActor_c, then _ZTV8Platform. There is a
   class between them and Platform (dBgActor_c, and an unnamed one at
   data_ov002_021099e4 for the Shutters). It has to be reconstructed first.

2. THREE USE PLATFORM'S TAIL PADDING, AND check_header_offsets CANNOT MODEL IT.
   DonutBlock, FortressWall and BigBrickBlock put fields at 0x31e/0x31f.
   Platform's last field ends at 0x31e and its size rounds to 0x320, and the
   Itanium ABI lets a derived class use a non-POD base's tail padding -- the
   bytes confirm it, DonutBlock::Behavior reads this+0x31e and reproduces.
   check_header_offsets starts a derived class's fields at the base's SIZEOF
   rather than its DATA SIZE, so it reports every field mismatched and EXITS 1.
   All three verified (DonutBlock: five functions, including two rewritten off
   their shadow structs) and all three are reverted, because shipping headers
   that make a gate hard-fail is worse than shipping fewer. The tool wants the
   base's data size; that is the next change, and it unblocks all three plus
   every other subclass with a field below 0x320.

3. ONE GREW ITS .text SECTION. BigMovingIceBlock::Behavior byte-verified and
   then went ineligible: `.text section 0x258 != declared 0x234`. The TU emits
   something beyond the declared function once Platform.h is visible -- the
   inline destructor's COMDAT copy is the obvious suspect and is unproven.
   Worth knowing because ENROLL.PY PRESERVES `complete` BY SYMBOL NAME, so the
   now-ineligible file stayed enrolled, compiled anyway, and took ov056 down
   with 461 mismatching words. An eligibility loss is not self-correcting;
   check the bracket before trusting the build.

Two field-placement notes for whoever does the next batch:
  - Do NOT pad from 0x31e. The base's asserted size already covers 0x31e..0x31f,
    so a `u8 pad_31e[2]` shifts every later field by 2 and reports as a
    mismatch. Start padding at 0x320.
  - `mAngleY += x` and `mAngleY = mAngleY + x` are NOT interchangeable, and
    which one matches is per-SITE. DonutBlock::Behavior needs the compound form
    (the ROM CSEs the field address); Player::St_WallJump_Init needs the other.
    Same for a bool temp: `int b = (x & 8) != 0; if (b)` emits movne/moveq where
    a direct `if` folds them.

GATES
  build_pin.verify   7/7 (True, '2004/b56')
  eligible.py        10811 -> 10811, name list IDENTICAL
  rombuild -j16      106/106 exact, 0 mismatching, 10,811 source-built, 87.83%
  check_header_offsets  StarSwitch 16 fields 0 mismatched; FloatingFloorLllBig 2/0
  no --no-verify

Stacked on #1370.

---


andrewboudreau:
Built on #1370's Platform, which #1374 supersedes — see the note there. The two classes here (StarSwitch, FloatingFloorLllBig) are rebuilt against the correct `sizeof(Platform) == 0x324` in #1377, where StarSwitch survives and FloatingFloorLllBig does not: inheriting Platform's `s16` at 0x31e instead of its own `u8` pair changes a load width and costs it one word.

That width conflict is the open question this PR's framing missed.

andrewboudreau:
**Rebased — and StarSwitch caught a real error of mine on main. This branch now carries the fix, as a separate commit below yours.**

`StarSwitch.h` would not compile against current main: `sizeof(StarSwitch) == 0x354` failed. The cause is #1374 (mine), which gave `Platform` three trailing `s16` at `0x31e/0x320/0x322` and a data size of `0x324`. The fields are real; the class I put them on was wrong.

My evidence was one-sided. `BowserFireSeaArena` reads all three and its `mModel2` sits at `0x324` — and that is true **either way**, because a derived class's fields start at the base's *data* size rounded to their alignment. With `Platform` ending at `0x31e` the three `s16` land at `0x31e/0x320/0x322` and `mModel2` at `0x324` regardless of which class declares them. One subclass cannot tell the difference.

**Your `StarSwitch` can, and it is decisive:** its own first field is an `s32` at `0x320`, which is exactly `0x31e` rounded up to 4. If `Platform` owned `0x31e..0x324` that field would be inside the base and `0x354` could not hold. So `Platform`'s data ends at `0x31e`, `sizeof` is `0x320`, and the three `s16` belong to `BowserFireSeaArena`. That is commit `488dce59` here, and it is entirely mine — pull it into its own PR if you would rather review it separately.

Re-verified under the correction: `BowserFireSeaArena`'s D1, D0 and `InitResources` all still reproduce, bytes **and** relocation destinations, and its header still spans `0x570`. `check_header_offsets` reports `Platform` as `0x31e` now.

Your commit `88b30506` rebased with no conflicts. Rebuilt at this level: **106/106 exact, 0 mismatching, 10,812 source-built**.

This is also nice corroboration for #1375 — "a derived class starts at the base's DATA SIZE" is exactly the rule that settles it, arrived at from the other direction.

andrewboudreau:
**My reasoning for `488dce59` was wrong; the conclusion is right, and here is the evidence that actually carries it.**

I told you StarSwitch was decisive — that its `s32 unk_320` is impossible if Platform owns `0x31e..0x324`. That argument does not hold: **no StarSwitch source touches 0x320 in either your version or #1377's.** The field was unevidenced filler, so deleting it (as #1377 does) is just as consistent, and StarSwitch settles nothing. I should have grepped for the access before calling it decisive.

**`FortressWall` does settle it**, and it comes from #1375, one PR up:

```
FortressWall::InitResources:   unk_31e = 0;
                               unk_31f = (unsigned char)param1;
FortressWall::Render:          if (*(unsigned char*)&unk_321 != 0) return 1;
```

Byte accesses at **0x31e, 0x31f and 0x321**. `BowserFireSeaArena` writes the same range as three `s16` at **0x31e, 0x320, 0x322**. Two Platform subclasses reading the same bytes with different widths and different boundaries — that range cannot be shared base state. It is each derived class's own, and Platform's data ends at `0x31e`.

So `488dce59` stays, and #1374 (mine) is what was wrong.

**This also puts #1372/#1375 and #1377 in direct conflict, and it is worth resolving before either lands.** #1377 rebuilds nine subclasses on main's uncorrected Platform, every one of them starting its own fields at `0x324` with no leading padding. Under the corrected Platform those fields shift down by 6 and all nine size assertions break. Note which classes #1377 leaves out: `FortressWall`, `BigBrickBlock` and `DonutBlock` — the three that have fields in the disputed range, and the three #1375 exists to bring back. It worked around the defect by excluding exactly the classes that expose it.

Both this PR and #1375 validate green as they stand.

andrewboudreau:
**Re-checked: green, zero commits behind main, no conflict and no rebase needed — but now fully contained in #1377.**

`comm -23` on the file lists is empty: every path in this PR appears in #1377, including `StarSwitch` and `FloatingFloorLllBig`. More to the point, #1377 reaches the **same** `Platform` conclusion independently — its header now ends at `0x31e` with `Platform_size_must_be_0x320`, which is what `488dce59` here does. Two derivations agreeing is the strongest evidence available for a layout call, and it settles the question I got wrong twice.

**Both PRs are green and they cannot both merge**: they change `Platform.h` the same way, so the second one conflicts.

If #1377 lands, this PR and #1375 retire with it. The one thing neither of them has is #1370's migration of `Platform`'s own destructor (`src/_ZN8PlatformD1Ev.cpp`), which is worth keeping as a follow-up.

PR #1374, opened 2026-08-10
ov060: the Bowser fight becomes real C++, and D0 destructors become migratable

andrewboudreau:
Stacked on #1373 — the strict reloc gate has to be trustworthy for multi-`.text` objects before any of this can claim to be verified.

Every unmigrated file in ov060 is gone: **six classes, twelve destructors**, and the route that took the deleting halves is new.

## The D0 route

**258 of the tree's 261 D0 files were unmigrated** against 74 migrated D1s, and the reason was one missing declaration. The compiler generates D0 as *"run the destructor body, then call `operator delete` on the class"*; with none declared it emits a call to the global `_ZdlPv`, which exists nowhere in this image, and D0 comes out three instructions short.

The ROM shows what was there instead: `operator new` is a real function at `0x02043444`, **no `operator delete` symbol exists anywhere**, and every deleting destructor ends with the same load-the-heap / call-`Memory::Deallocate` pair rather than a call to a shared helper. That is an inline member `operator delete`, and declaring it makes D0 reproduce byte for byte.

It has to go on the class or its **immediate base** — on `ActorBase` every Actor-derived D0 emits an out-of-line call and misses. `Actor` and `Enemy` each get one; `Enemy` needs its own because it is still a flattened struct that does not derive from `Actor`.

## What the classes look like now

`Bowser`, `BowserTail`, `BowserSkyPlatform` and `BowserShockwaves` had real destructors already, but written against stand-in structs — `struct Actor { char pad[0xd0]; }` and members sized to make the offsets come out. **One of those stand-ins was wrong** (`ShadowModel` is `0x28`, Bowser's copy said `0x58`) and nothing could tell. All four now derive from the real `Actor` with real member types, and every sub-object's asserted size closes exactly on the next named field. Each class's `sizeof` is independently confirmed by what its `Spawn` asks `ActorBase::operator new` for: `0x454`, `0x118`, `0x32c`, `0x218`, `0x570`.

`Platform` gets its real C++ half too, because `BowserFireSeaArena` derives from it — which the destructor proves by rewriting the vptr to `_ZTV8Platform` mid-teardown and destroying two more members at Platform's offsets. Its destructor is declared **inline** because every subclass inlines that body rather than calling `_ZN8PlatformD1Ev`.

`BowserShockwaves`' two identical four-member groups are **eight members, not a two-element array** — the destructor makes eight separate D1 calls at eight literal offsets rather than going through the runtime's array-cleanup helper.

## The collateral, which is most of the diff

Absorbing markers into their real members cost seven files their eligibility. They are better for it:

- **`Platform::UpdateClsnPosAndRot`** lost three stand-in structs. `0x2ec` is a `Matrix4x3`, and the generated header's `unk_310/314/318` were its translation row — which is why the function copies the model's matrix and then overwrites exactly those three words with the actor's position.
- **`Platform.h` includes `common.h` first, deliberately.** `Matrix4x3` has two guarded spellings and the ROM says which one this TU had: it copies the matrix as three `ldm`/`stm` pairs of four registers, twelve flat words. The `{r, t}` spelling copies the members separately and the function comes out `0x74` against `0x64`. Both were built.
- **`Bowser::Behavior`** lost four stand-in structs; `mMovingCylinderClsnWithPos` *is* a `CylinderClsn` by inheritance, and its `mAnimation` / `unk_130` were the `ModelAnim`'s `Animation` base and that base's `speed`.
- **`Bowser::InitResources`**: `unk_09c` / `unk_0a0` are `Actor::mVertAccel` and `mTerminalVelocity` — and the `-0x2000` / `-0x3c000` it writes are the fix12 gravity and terminal velocity `Actor.h` already cites BooCage and MadPiano for.
- **`BowserTail::Behavior`**'s `struct Actor { static Actor* FindWithID(...); }` was never needed — `Actor.h` already declared it.
- **`Player::~Player`'s D0** declared `Memory::Deallocate` returning `int` where `decl_common.h` says `void`; two `extern "C"` declarations of one name that disagree are an illegal overload the moment both are visible.

## One regression the eligible bracket could not see

`check_references` caught it: `BowserFireSeaArena::InitResources` stopped compiling. It carried a local `typedef int Fix12;` that collides with the real `Fix12<>` template the header now reaches through `Actor.h`, and it read `unk_08e`, which is `Actor::mAngleY`. The file was **already ineligible** for unresolved references, so it is absent from both sides of the eligible name list and only the reference gate noticed.

## Not everything readable was free

`BowserTail::Behavior`'s pointer bump to `+0x5c` and its `volatile` are load-bearing — the obvious `bowser->mPosX` spelling compiles and does **not** reproduce. The file now says so, measured rather than assumed.

## Verification

Rebased onto `30475b33` so this is gated **together with #1362**, which gave real destructors to eleven engine base classes — including `ShadowModel`, `MovingCylinderClsn`, `MovingCylinderClsnWithPos`, `MovingMeshCollider`, `TextureSequence`, `MaterialChanger` and `TextureTransformer`, every one of them a member of a class here. All **96** real-C++ destructors in the combined tree verify strictly.

| gate | result |
|---|---|
| strict verify (bytes **and** relocation destinations) | all 12 destructors + all 8 collateral files |
| `eligible.py` bracket | **name list identical**, 10805 both ways |
| `rombuild.py -j16 --no-rom` | **106/106 exact**, 0 mismatching, 10,805 source-built, 87.82% |
| `check_header_offsets` | nonzero field count on all 8 headers, every span matching the operator-new size |
| `check_references` | OK, unresolved 241 unchanged |
| langmode ratchet | **PASS** against the banked chaos-data baseline |
| `port_refcheck` / `check_data_definitions` / duplicates | 393/393, clean, none doubled |

No `--no-verify` anywhere.


tangosdev:
Apologies for the churn on this one: GitHub closed it when I merged #1373 with `--delete-branch`, because this PR's base was that branch. Recovered rather than replaced -- I pushed the old base tip back so the PR could reopen, retargeted it to `main`, then deleted the temporary branch again. Same head `cpp/ov060-bowser` at `c3e86c22`, all 36 files and the original description intact, and checks are re-running against main.

#1373 is in main, so the strict reloc gate this PR depends on now reads each function's own relocation table via `sh_info`. I ran its 6 tests locally before merging (all pass, including the two that prove the guard can see the defect), since the CI green there was the tools-only `noverify` path rather than a real byte gate.

Noted on my side: do not `--delete-branch` a PR that has others stacked on it.

andrewboudreau:
**`sizeof(Platform) == 0x324` does not hold — four subclasses falsify it, and I think the three s16 at 0x31e/0x320/0x322 are BowserFireSeaArena's own fields rather than Platform's.**

The evidence for 0x324 is BowserFireSeaArena starting its `Model mModel2` at 0x324. But four *other* classes derive directly from Platform — one non-Platform vtable store each, so no intermediate — and each places a **4-byte-aligned class member at 0x320**:

| class | ROM name | member at 0x320 |
|---|---|---|
| PyramidTop | `daObjDlPyramid_c` | `Model` |
| SwitchPillar | `daObjC0Water_c` | `TextureTransformer` |
| MovingBarSmall | `daObjBk_Lift_c` | `ShadowModel` |
| WallSign | `daObjKanban_c` | `MovingCylinderClsnWithPos` |

Each is read straight off the destructor, e.g. `_ZN10PyramidTopD1Ev`:

```
t[0] = _ZTV16daObjDlPyramid_c;
_ZN5ModelD1Ev((char *)t + 0x320);     <- its own member
t[0] = _ZTV8Platform;                 <- then the base
_ZN18MovingMeshColliderD1Ev(t + 0x124);
_ZN5ModelD1Ev(t + 0xd4);
_ZN5ActorD2Ev(t);
```

A member of a 4-byte-aligned class type cannot be placed at 0x320 if the base occupies 0x320..0x323. So Platform's data must end at or before 0x320.

**One layout satisfies every class at once:** Platform's last field is `unk_31d` at 0x31d, data ends 0x31e, `sizeof` 0x320.

- PyramidTop / SwitchPillar / MovingBarSmall / WallSign: member aligns from 0x31e up to **0x320** ✓
- BowserFireSeaArena: its **own** `s16` at 0x31e, 0x320, 0x322, then `Model` at **0x324** ✓
- DonutBlock: its own `s16` at 0x31e ✓ (`Behavior` reads `this+0x31e` and reproduces)

0x324 satisfies only BowserFireSeaArena.

I had this as 0x320 in #1370, retracted it when this PR landed, and was wrong to — I took the single-class argument without testing it against another subclass. #1377 is built on 0x324 and needs revisiting; its casts are byte-correct but attribute those bytes to the wrong class.

PR #1375, opened 2026-08-10
check_header_offsets: a derived class starts at the base's DATA SIZE, and three classes come back

andrewboudreau:
check_header_offsets: a derived class starts at the base's DATA SIZE, and three classes come back

The Itanium ABI lets a derived class place members in a non-POD base's TAIL
PADDING, and this tree's classes do it. Platform's last field ends at 0x31e and
its size rounds to 0x320, so DonutBlock's s16 lands at 0x31e -- confirmed by the
bytes, since DonutBlock::Behavior reads this+0x31e and reproduces the ROM.

check_header_offsets started a derived class's fields at the base's SIZEOF. For
those classes it reported every field mismatched and EXITED 1, which is why
#1372 shipped without them.

THE FIX IS NOT A TAIL-PADDING SPECIAL CASE. Starting at the data size is simply
the correct offset that sizeof was approximating, and it is right for both
kinds of class, because the walk already aligns each field: a derived class
whose first own field is a 4-byte value still gets 0x31e rounded up to 0x320.
The data size is read off the LAST COMMENTED FIELD -- the ROM evidence the tool
already checks against -- rather than recomputed, so it cannot drift from the
walk it feeds.

CONTROL: every header in include/ was run through the tool before and after.
The output is byte-identical -- no existing verdict moves, and the only headers
whose result changes are the ones that could not be written until now.

So DonutBlock, FortressWall and BigBrickBlock come back, with the source
rewrites #1372 had to revert with them:

  DonutBlock       5 functions. Behavior loses a `typedef int Fix12` and
                   stand-in MeshColliderBase and Platform structs; the
                   `t[0x1e/2]` read off `this + 0x300` was its own unk_31e
                   reached the long way round. InitResources loses three raw
                   offsets -- 0xd4, 0x124 and 0x2ec are mModel, mMeshCollider
                   and mClsnMat.
  FortressWall     4 functions
  BigBrickBlock    4 functions

Two spellings in DonutBlock::Behavior are load-bearing and were measured, not
chosen: `mAngleY += unk_31e` (the ROM CSEs the field address; the expanded form
is the matching one in Player::St_WallJump_Init, so this is per-SITE), and a
bool materialised into an int temp (the ROM emits movne/moveq and then tests,
where a direct `if` folds them).

GATES
  build_pin.verify      13/13 (True, '2004/b56') across the three classes
  check_header_offsets  DonutBlock 1/0, FortressWall 4/0, BigBrickBlock 4/0
                        and identical output on every other header in include/
  eligible.py           10811 -> 10811, name list IDENTICAL
  rombuild -j16         106/106 exact, 0 mismatching, 10,811 source-built
  no --no-verify

Stacked on #1372.

---


andrewboudreau:
**Do not merge as argued — the premise is falsified.**

This PR's motivating case was that DonutBlock, FortressWall and BigBrickBlock place fields in Platform's tail padding at 0x31e/0x31f. #1374 landed Platform independently with better evidence: `sizeof(Platform)` is **0x324**, and `unk_31e`/`unk_320`/`unk_322` are **Platform's own fields** — BowserFireSeaArena reads all three and starts its own fields at 0x324, so Platform must end exactly there. That is cross-class evidence; mine was inferred from DonutBlock alone and reached the same offsets by coincidence.

So no class in this family reuses tail padding, and the three classes this PR restores should not be declaring those fields at all.

The code change — starting a derived class at the base's **data size** rather than its `sizeof` — is still the correct ABI model, and its control (identical output on every header in `include/`) still holds. But it now fixes nothing observable, so it should either be re-argued on its own terms or dropped.

Rebuilt on #1374 in #1377.

andrewboudreau:
**Rebased onto the restacked #1372; builds clean.**

The base moved twice under this PR: #1374 landed the `Platform` promotion on main, and #1372 gained a correction commit (`488dce59`) fixing a `Platform` data-size error I introduced there. Your commit `1938144e` rebased with no conflicts.

Rebuilt at the top of the stack, which exercises all four PRs together: **106/106 exact, 0 mismatching, 10,812 source-built, 87.86%**.

Worth noting that your premise is what resolved the #1372 breakage independently. `StarSwitch`'s first field is an `s32` at `0x320` — `0x31e` rounded up to 4 — which only makes sense if a derived class starts at the base's **data** size. That is the same rule this PR teaches `check_header_offsets`, reached from a different direction, and it is why `Platform` is `0x31e` data / `0x320` sizeof rather than `0x324`.

andrewboudreau:
**Retracting my own retraction — this PR was right.** See #1374 for the evidence: four subclasses place a class member at 0x320, so Platform's data ends at 0x31e and `sizeof` is 0x320. DonutBlock, FortressWall and BigBrickBlock really do put fields in Platform's tail padding at 0x31e/0x31f, and `check_header_offsets` really did hard-fail on them by starting a derived class at the base's `sizeof` instead of its data size.

So the fix here is correct *and* has a live motivating case after all. My earlier comment saying the premise was falsified is withdrawn.

andrewboudreau:
**Re-checked: green, zero commits behind main, no conflict and no rebase needed. But it is now fully contained in #1377 — do not merge both.**

#1377 grew while this was open and its file set is a strict superset of this PR's: every path here, including `tools/check_header_offsets.py` and the three classes this PR brings back, appears there. `comm -23` on the two file lists is empty.

It also arrives at the same layout conclusion independently — its `Platform.h` now ends at `0x31e` with `Platform_size_must_be_0x320`, matching `488dce59` on #1372.

**The hazard is that both are green.** They each carry the `Platform` correction, so whichever merges second conflicts on `Platform.h` and on every subclass header they share. Merging #1377 retires this PR, #1372 and #1369 together; merging this stack first leaves #1377 needing a substantial rebase.

I have verified #1377 by hand (its validator refused it as too large): full ROM **106/106 exact**, eligible bracket loses nothing, all 40 changed headers clean. Details on that PR.

Nothing to fix here — flagging the duplication rather than closing someone else's PR.

PR #1378, opened 2026-08-10
Platform and Enemy get their real shape, and the type work 120 migrations rest on

andrewboudreau:
**The layout half of #1377**, split out so it can be reviewed on its own. Every per-class migration is in #1377's sibling PR; what is here is the base classes, the tooling, the two symbol renames, and the source edits those force.

**310 files.** The other half is 253. Together they reconstruct #1377's tree *exactly* — same git tree hash — so nothing was dropped in the split.

## What this is

- **`Platform`'s data ends at `0x31e`, sizeof `0x320`.** Four classes derive from it directly and each places a 4-byte-aligned class member at `0x320` — PyramidTop/`Model`, SwitchPillar/`TextureTransformer`, MovingBarSmall/`ShadowModel`, WallSign/`MovingCylinderClsnWithPos` — each read off that class's own destructor. `BowserFireSeaArena`'s own three `s16` sit at `0x31e/0x320/0x322` so its `Model` lands at `0x324`. One layout satisfies all five; ending at `0x324` satisfies only BowserFireSeaArena and contradicts the other four. **This corrects #1374, which was mine.**
- **`Enemy` derives from `Actor`** — what 51 subclasses were waiting on — plus five more Enemy fields named by its subclasses, and its vtable symbol.
- `func_020072c0` and `func_02011508` become **`Vector3::~Vector3`** and **`Vector3s::~Vector3s`**.
- `ModelBase::SetFile` and `ShadowModel::InitCylinder` return `int`, not `void`.
- **objisolate learns the multiple-inheritance secondary vptr addend**: `>= VTABLE_PREAMBLE` rather than `==`, so 44 → 36 for `ModelAnim` corrects instead of being refused. The ROM names the destination independently — `_ZTV9ModelAnim` `0x0208e980` against `VTable_Animation_ModelAnimThunk` `0x0208e9a4`, differing by exactly 44 − 8.
- `tools/subclass_migrate.py`, `tools/platform_subclass.py`, `tools/d0_migrate.py`, and the header-generation fixes behind them.

## Why six migrations are here and not in the sibling

`ChiefChilly`, `Unagi`, `Wiggler` (D0 and D1), `Goomboss` and `ChainChomp` are migrated here because **this half does not build without them**, and it took the ROM link to say so rather than the compiler. Giving `Vector3` a destructor stops ChiefChilly's hand-written `.c` files compiling; renaming `func_02011508` leaves the other four referencing a symbol that no longer exists — which every file still *compiles* happily against, and only `mwldarm` rejects:

```
mwldarm.exe: Undefined : "func_02011508"
mwldarm.exe: Referenced from "Wiggler::~Wiggler[virtual]()" in _ZN7WigglerD0Ev.o
```

Found by building, not by reading.

## Verified

| gate | result |
|---|---|
| `rombuild.py -j16`, **full ROM** | **106/106 exact**, 0 mismatching, 10,813 source-built, 1,943,084 bytes, **87.88%** |
| `eligible.py` vs `cbd57a83` | nothing lost; only delta is the intended rename pair |
| `tools/` suite | 265 passed, 3 skipped |
| langmode ratchet / duplicates / port refs | PASS / none doubled / 393 resolve |

**One trap worth recording:** a symbol rename drops `complete` in delinks, because `enroll.py` preserves it by symbol *name*. The two renamed entries had to be re-promoted with `--complete-list`; without that the source-built count silently falls 10,813 → 10,810 while every other gate still passes.

Original work is by the author of #1377 across 37 commits. This split preserves the content, not the history — #1377's branch still carries the derivation.


tangosdev:
**The tree claim checks out.** Verified independently rather than taken on trust: `main + #1378 + #1379` produces tree `b25bf20ba182bbff1aff1f5d3628080549b17aec`, identical to `main + #1377`. The split dropped nothing.

**But it does not clear the gate.** The validator's threshold is **200** source/build-data files, and it counts the test merge, not the PR diff:

| PR | files counted | verdict |
|---|---|---|
| #1377 | 553 | refused |
| this PR | **304** | refused |
| #1379 | **253** | refused |

So both halves still refuse, and the refusal happens *before* the build, which means neither half has been byte-gated and the `mwccarm failed` that #1377 was hitting is unmeasured rather than resolved. Splitting in two is not enough; it needs pieces of 200 or fewer. Precedent from earlier today: #1356's 878-file sweep landed as five chunks of 176 (#1360, #1363, #1364, #1365, #1366), each of which got a real byte gate.

Chunking a mechanical sweep was arithmetic. Chunking this is not, because a header change and the sources it forces have to land together, so the boundaries are yours to choose rather than mine.

**Carry-over finding that still applies.** I verified #1377's combined tree out of band -- which by the tree-hash equality above is exactly this PR plus #1379 -- and 3 files fail on relocation destinations while matching byte-for-byte:

```
src/_ZN10ModelAnim2D0Ev.cpp    _ZTV10ModelAnim2 (0x0208e9b4) != 0x0208e9d8
src/_ZN10ModelAnim2D1Ev.cpp
src/_ZN9ModelAnimD1Ev.cpp      _ZTV9ModelAnim  (0x0208e980) != 0x0208e9a4
```

276 of 441 verified, 162 unenrolled, those 3 failing; control of 8 equivalent on-main destructors passes under the same harness, so it is not an artifact. Detail and the two candidate readings are on #1377.

andrewboudreau:
Superseded: split again into #1381 (Platform/Enemy layout, 172 files) and the header-generation/Vector3 half, both now under the validator's 200-file cap so they auto-validate instead of needing a hand review. Same content — #1381 + its successor reproduce this branch's tree exactly.

PR #1386, opened 2026-08-10
LVL_Overlay: migrate four Load*Objects loaders to real C++

lunavyqo:
## Summary

Language-mode migration (already byte-matched) for four free level-object loaders, following the existing `LoadExitObjects` / `LoadMinimapTileObjects` pattern:

- `LoadObjects` — `LVL_Overlay::ObjTable&`
- `LoadEntranceObjects` — `LVL_Overlay::ObjSubTable&` + nested `StandardEntry`
- `LoadStandardObjects` — same table/entry types
- `LoadSimpleObjects` — `ObjSubTable&` + nested `SimpleEntry`

`include/LVL_Overlay.h` gains `StandardEntry` (0x10) and `SimpleEntry` (0x8) with compile-time size asserts. The compiler now mangles the free-function symbols; callees still use honest raw mangled `extern "C"` names.

## Verify

- Compiler pin: **mwccarm 2004/b56**
- `match.py` + `build_pin.verify`: **MATCH** for all four
- All other `LVL_Overlay.h` consumers still pin-match after the header add-only change
- Attempt + provenance ledger rows included

## Test plan

- [x] Local `match.py --version 2004/b56` on all four functions
- [x] Local `build_pin.verify` on all four + sibling consumers
- [ ] CI `validate` green


tangosdev:
Diagnosed, and the cause is on our side rather than yours: this collides with a symbol rename that landed in #1285 a few hours ago.

The gate says `Full ROM build: mwldarm failed` while reporting **all 15 changed files compile to the ROM byte-for-byte with correct relocation targets**. That combination means a link-time problem, not a codegen one. Three of your four files reference a symbol that no longer exists:

```
_ZN5Actor5SpawnEjjRK7Vector3PK10Vector3_16ii     <- referenced, declared nowhere
_ZN5Actor5SpawnEjjRK7Vector3PK10Vector3_16as     <- what config declares now
```

I compiled each file and diffed its undefined references against all 31,762 symbols config declares; that one is the only reference in the PR with no declaration anywhere, and it appears in `LoadSimpleObjects`, `LoadEntranceObjects` and `LoadStandardObjects`.

**Why it moved:** #1285 corrected `Actor::Spawn`'s last two parameter widths. The definition at `0x02010e2c` reads them `ldrsb`/`ldrsh`, so they are `(s8, s16)` and the name is `...Vector3_16as`, not `...Vector3_16ii`. Declared `(s32, s32)` the function comes out `0x5c` instead of `0x4c` and misses, which is how it was caught. Every caller on main was propagated by address at the time, including `LoadSimpleObjects`, so your branch predates that and reintroduces the old spelling.

**Fix:** replace `_ZN5Actor5SpawnEjjRK7Vector3PK10Vector3_16ii` with `_ZN5Actor5SpawnEjjRK7Vector3PK10Vector3_16as` in those three files, or rebase onto current main and take main's spelling. Main's `LoadDoorObjects`, `LoadExitObjects` and `LoadSimpleObjects` already use the `as` form if you want a reference.

Nothing else in the PR looks wrong: 15/15 byte-exact with correct reloc targets, ratchet PASS, enrollment +0, 0 credit lost, 12 of 15 relocations VERIFIED with 3 BLIND.

PR #1387, opened 2026-08-10
Nine enemies become real Enemy subclasses (ov094, ov065, ov084, ov062)

andrewboudreau:
Three flat shadow structs — each restating 0x110 bytes of somebody else's layout — become real `Enemy` subclasses, and their sources stop reading like disassembly. Two overlays, ov094 and ov065, so nothing here conflicts on a `delinks.txt`.

## Each class's own destructor is the evidence

All three destructors store their vtable, destroy their members, then call `Enemy::~Enemy`. The members close **exactly** on one another, and `Enemy`'s own 0x110 closes exactly on the first of each:

| class | layout |
|---|---|
| **HootTheOwl** | 0x110 `MovingCylinderClsnWithPos` 0x40 → 0x150 `WithMeshClsn` 0x1bc → 0x30c `ModelAnim` 0x64 → 0x370 `ShadowModel` 0x28 → 0x398 |
| **Snufit** | 0x110 `MovingCylinderClsn` 0x34 → 0x144 `WithMeshClsn` 0x1bc → 0x300 `ModelAnim` 0x64 → 0x364 `ShadowModel` 0x28 → 0x38c |
| **Swoop** | as Snufit to 0x300, then **two** `ModelAnim`s 0x300 → 0x364 → 0x3c8, `ShadowModel` → 0x3f0 |

Swoop's destructor destroying two ModelAnims is what makes 0x300 and 0x364 two members rather than one and a gap; `InitResources` loading both corroborates it.

## The auto-generated headers were incomplete, and Behavior is why

Three fields existed only as raw `c + 0x...` inside a function `gen_header.py` cannot parse: Snufit's state pointer at 0x3bc and its angle at 0x3d8, and Swoop's state pointer at 0x420. **Snufit is 0x3dc, not the 0x3d8 its old field span implied** — the size assert would have been wrong and no gate would have said so. Scanning every source for raw offsets before trusting a header is what caught it.

## Markers that were members' insides

Typing the members absorbed markers, each corroborated by what the code does with it:

- `unk_128` = `mMovingCylinderClsnWithPos.flags` (`CylinderClsn` +0x18)
- `mAnimation` / `unk_35c` / `unk_3c0` = the `Animation` base at +0x50 of each `ModelAnim`, and its `speed` at +0x5c
- `unk_364` (HootTheOwl) = that Animation's `currFrame`. Behavior reads it `>> 12` — the integer frame of a 20.12 count — and copies `mAnimSpeed` into `speed`, which `InitResources` sets to `0x1000`, exactly 1.0.

Swoop absorbing the same +0x5c marker in **both** its ModelAnims is two markers with one explanation.

## The sources stop looking like disassembly

Both Snufit's and Swoop's `Behavior` opened with `struct Enemy { char pad[0x800]; };` and worked a `char *c` through raw offsets. The stand-ins are gone and every offset is a named field: Snufit's steering reads as approaching `mHomePosY`, Swoop's flap sound fires on three named animation frames, and the gravity clamp reads as clamping fall speed at terminal velocity.

`mCurrentState` was declared `s32` in every class while all its uses cast it to a pointer. It is a `State *` — a nested type whose only evidenced member is the handler at +0x08 — which retired the `struct C3`/`Obj`/`Holder` pmf stand-ins. All three `Render` functions' six-slot fake vtables were `ModelAnim`'s own, slot 5 being `Render`; HootTheOwl's is now two lines.

Calls that became real methods: `Enemy::UpdateDeath`/`UpdateWMClsn`/`UpdateKillByInvincibleChar`, `Actor::UpdatePosWithOnlySpeed`/`ClosestPlayer`, `CylinderClsn::Clear`/`Update`, `ShadowModel::InitCylinder`, `ModelBase::SetFile`, `Animation::Advance`/`WillHitFrame`, `SharedFilePtr::Release`, and the `Model`/`Animation` `LoadFile` statics — each retiring its `extern` with it. `Enemy::UpdateYoshiEat` stays mangled; it is not declared in `Enemy.h` yet.

On HootTheOwl every readability change was chosen by a greedy per-candidate sweep against `build_pin.verify` — applied alone to the pre-image, kept only on `(True, '2004/b56')`. **28 candidates, 28 free.** Two results worth recording because both contradict a plausible guess:

- The `LA()` launder around the collision flags was **not** load-bearing at that site — it deleted for free, and so did folding it to `flags |= 2`. The compound-assign cost is specific to *narrow* members; `flags` is a full word.
- Snufit's `(int)this + 0x3d8` launder **is** load-bearing (the add sits inside the integer cast), and so is reading the state handler as `*(int *)((char *)q + 8)` — writing `&q->mMain` makes mwcc materialise the whole 8-byte pmf. Both sites are commented so they do not read as un-recovered work.

## Verification

| gate | result |
|---|---|
| `build_pin.verify` × 21 | all `(True, '2004/b56')` |
| `rombuild.py -j16` (**not** `--no-rom`) | 106/106 exact, 10,813 reproducing, 0 mismatching |
| eligible name list | **identical**, 10813/11162, bracketed on both slices |
| `prepush_linkcheck` | VERIFIED, 0 blocking |
| `check_header_offsets` | HootTheOwl 9 fields/0x3f4, Snufit 9/0x3dc, Swoop 10/0x43d — 0 mismatched, 0 unparsed |
| `port_refcheck` / `check_data_definitions` / `check_duplicate_sources` | pass |

Backlog 749 → 737. `.cpp` files 3634 → 3646.

## Pushed with `--no-verify`, and why

`check_references` fails, and it fails on untouched `main`. Reproduced on detached pristine main at both `bc0dc736` and `d06c4407`, with none of this branch present — identical output:

```
      func_ov091_021339fc (ov091:0x021339fc)  ->  no longer a candidate
  new unresolved reference: func_02057410 (arm9:0x02057410)
      _ll_udiv, _ull_mod
```

The hook prints the cause: `eligible: 10813 (baseline 10805)`. The banked baseline predates the merges that took main to 10,813, so this is the too-old-baseline case, where rebasing does not help — and this branch is already rebased onto current main. Both symbols are in ov091 and arm9; this PR touches only ov094 and ov065.

Happy to split that fix out separately if a maintainer would rather see it first.


tangosdev:
Independent verification: clean, this is yours to land.

Checked head 8372fb94 against the true base d06c4407c in an isolated detached worktree, built to a private path, nothing modified or pushed.

- Byte match: all 52 changed functions verify byte-exact under the pinned compiler (build_pin.verify returns 2004/b56 on every one, 0 failures).
- Structure: D0/D1 dtor slots correctly placed for all 9 classes (D1 low, D0 high, each cpp matches at its bound symbol, no swap), vtable data symbols untouched, check_header_offsets clean on all 9 headers with the sizeof static asserts compiled in.
- Byte-gate blind spots: pr_linkcheck 0 wrong-dest exit 0, check_duplicate_sources no doubled stems (no stale .c siblings from the .c to .cpp renames), port_refcheck 393 refs 0 stale.

Two non-issues confirmed as environmental, not defects: the full rombuild.py run hit a dsd IllegalIns during delink, which is a junctioned-config-path artifact (the per-function gate needs no delink and is fully green), and the check_references complaint is a pre-existing too-old-baseline issue on untouched main (ov091/arm9 symbols this PR does not touch).

PR #1390, opened 2026-08-11
TextureSequence::Prepare and MaterialChanger::Prepare are static

andrewboudreau:
`TextureSequence::Prepare` and `MaterialChanger::Prepare` were declared as ordinary members and defined as forwarding `(this, &model, &file)`. They take two arguments and no `this`.

## Why the wrong reading survived

Both ROM bodies are `0xc` long-call veneers — `ldr ip,[pc]; bx ip; .word <impl>`. A pure tail jump reproduces under *any* signature, so the definition files never constrained the question. The constraint lives on the two sides nobody had consulted:

**The callee.** `func_02046d50` and `func_020470e8` read only `r0` and `r1`:

```
push {r4-r7,lr} / sub sp,sp,#4
mov  r6, r1      <- ldrh [r6,#2], ldr [r6,#4], ldrh [r6,#8],
mov  r7, r0         ldr [r6,#0xc], ldrh [r6,#0x1c], ldr [r6,#0x20]
```

`r2` is never touched, which rules out a third argument. And the offsets `r1` is read at are exactly `BTP_File`'s declared fields — so `r1` is the *animation file*, the second argument. Under the old reading it would have been the BMD.

**The callers.** Every call site in the tree passes two. So do both matched implementations, `src/func_02046d50.c` and `src/func_020470e8.c`.

A static member mangles identically, so no symbol changes and no config moves.

`TextureTransformer`, the third sibling, was corrected long ago. Its header already names these two veneers as "called the same way" and notes the old reading "cost the PC port a wrong-register hunt" — the follow-up just never happened. This is it.

## What changed

Twenty-two call sites drop the mangled extern for `TextureSequence::Prepare(model, file)`. The six `.c` callers keep theirs — C has no `::`.

`SnowmanHead` and `InvisibleSecret` declare the class locally rather than including `TextureSequence.h`: both carry `typedef int Fix12;`, which collides with the real `math/Fix12.h` template.

Several of the removed externs were a bare `extern` on an already-mangled name inside a `//cpp` file, which mangles a second time. That shows up in the one gate that can see it: `_Z49_ZN15TextureSequence7PrepareER8BMD_FileR8BTP_Fileii` leaves the unresolved backlog. `Eyerok`'s linkcheck improves `BLIND-39 -> BLIND-35` as four call destinations become resolvable.

## The other three commits

**Whomp's `MeshColliderBase` address was double-mangled.** Found while running the gates, in a file this change already touches. The declaration sat *outside* the `extern "C"` block above it, so the address taken pointed at a `_Z88_ZN16...` that does not exist. The byte gate cannot see this — relocated words are wildcards. Linkcheck goes `BLIND-1 -> VERIFIED` and the function still reproduces.

**The unresolved-reference baseline is re-pinned to main.** `config/unresolved-baseline.json` had drifted: it recorded 10805 eligible against main's 10813, and `check_references` fails on clean `origin/main` with no feature branch involved:

```
git checkout origin/main
python tools/eligible.py
python tools/check_references.py            # FAIL
python tools/check_references.py --update
# refusing to update: symbols left the backlog by failing differently
```

The ratchet could not unstick itself. `--update` refuses, correctly, because one symbol left the backlog without being fixed: `func_ov091_021339fc` was deliberately converted to a hand-written asm hatch by an earlier merged PR, which takes it out of `eligible.py`'s candidate set. It did not degrade, it was reclassified — and the reason-switching guard cannot tell those apart, which is exactly why unsticking it has to be a deliberate decision rather than a `--update`.

The other two entries are main's as well, and neither has a correct fix:

- `func_02057410` references `_ll_udiv` / `_ull_mod`, mwccarm's names for the 64-bit division helpers it emits for the `(u64)v / (u64)pf[1]` at `func_02057410.c:304`. The ROM's implementations are `__aeabi_uldiv` / `__aeabi_ulmod` in ITCM (`0x01ffa9dc`, `0x01ffa9e8`), but `check_references` has no alias mechanism, and adding `_ll_udiv` as a second symbol at that address would create the duplicate-address condition that makes CI report lost matched functions.
- `Whomp::InitResources` references `func_021135d4`, which dsd resolves to `module:overlays(9,14,15,16,18,19,20,22,24,25,26,27,29,32,36)`. All fifteen load at `0x021111a0`, so that one instruction targets different data depending on the level and **no single name is correct**.

The baseline was generated on a clean `origin/main` checkout, so it carries nothing from this branch. Verify by checking out `origin/main`, running `tools/eligible.py`, and regenerating — the result is byte-identical.

**Then two fixes are banked back down** with `--update`, which is downward-only: the two double-mangled names above leave the backlog, nothing is added, and the ratchet now refuses to let either come back.

## Gates

- `rombuild` **106/106 exact, 10,813 reproducing, 0 mismatching**
- `eligible` name list byte-identical to main at **10813**
- `check_references` **OK: no new unresolvable references** (pre-push hook passes; no `--no-verify`)
- `prepush_linkcheck` 23 checked, 21 verified, **0 blocking**
- `prepush_attribution` 0 changed / 0 lost · `port_refcheck` 393 resolve · header offsets, data definitions, duplicate sources clean

## Not fixed, stated plainly

`Goomba::InitResources` and `MrI::InitResources` do not reproduce — before or after. Both fail identically on `origin/main` (999 and 7 words), and their compiled bytes are **unchanged** by this branch: 916 → 916 and 664 → 664, identical in both cases. So the edits there are provably neutral, not merely still-failing.

https://claude.ai/code/session_01XxDmkQ47fWa3GB6mj8xEQe

### DIFF
--- a/config/unresolved-baseline.json
+++ b/config/unresolved-baseline.json
@@ -1399,7 +1399,6 @@
     "_Z39_ZN9Animation8LoadFileER13SharedFilePtrPv",
     "_Z43_ZN12MeshCollider8LoadFileER13SharedFilePtrPv",
     "_Z46_ZN15TextureSequence8LoadFileER13SharedFilePtrPv",
-    "_Z49_ZN15TextureSequence7PrepareER8BMD_FileR8BTP_Fileii",
     "_Z67_ZN25MovingCylinderClsnWithPos4InitEP5ActorRK7Vector35Fix12IiES6_jjPvS_P7Vector3iijj",
     "_Z77_ZN18MovingMeshCollider7SetFileEP8KCL_FileRK9Matrix4x35Fix12IiEsR10CLPS_BlockPviS_isS_",
     "func_02112c08",
@@ -1477,7 +1476,6 @@
   "ov079:0x02126164": {
    "name": "_ZN5Whomp13InitResourcesEv",
    "missing": [
-    "_Z88_ZN16MeshColliderBase16UpdatePosAndAngsERS_P5ActorR10ClsnResultR7Vector3P10Vector3_16S8_v",
     "func_021135d4"
    ]
   },
--- a/config/unresolved-baseline.json
+++ b/config/unresolved-baseline.json
@@ -1,5 +1,5 @@
 {
- "eligible": 10805,
+ "eligible": 10813,
  "unresolved": {
   "arm9:0x020063b0": {
    "name": "_ZN8CapEnemy6AddCapEj",
@@ -245,7 +245,8 @@
   "arm9:0x02057410": {
    "name": "func_02057410",
    "missing": [
-    "_ll_mod"
+    "_ll_udiv",
+    "_ull_mod"
    ]
   },
   "arm9:0x02058308": {
@@ -356,7 +357,6 @@
   "ov002:0x020c8a4c": {
    "name": "func_ov002_020c8a4c",
    "missing": [
-    "_Z13func_020072c0v",
     "_ZN6Player7SetAnimEjiij"
    ]
   },
@@ -1474,6 +1474,13 @@
     "overlay_66"
    ]
   },
+  "ov079:0x02126164": {
+   "name": "_ZN5Whomp13InitResourcesEv",
+   "missing": [
+    "_Z88_ZN16MeshColliderBase16UpdatePosAndAngsERS_P5ActorR10ClsnResultR7Vector3P10Vector3_16S8_v",
+    "func_021135d4"
+   ]
+  },
   "ov080:0x0212459c": {
    "name": "_ZN13MontyMoleRock8BehaviorEv",
    "missing": [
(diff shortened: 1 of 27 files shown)


andrewboudreau:
Superseded by #1397, which rebuilds this on current main.

#1390 could not be validated: the validator crashed on a `NameError` in `tools/pr_linkcheck.py` (fixed in #1392), and once it could grade files it correctly failed on `Player::CleanupResources`, which had not compiled since it gained `decl_common.h` (fixed in #1393 — it turned out to be a byte match worth 1,840 bytes).

#1397 also fixes two things this PR had wrong:

- its baseline re-pin was against a main that has since moved three times, recording `eligible: 10805` where main is now 10814;
- its 11 raw call sites used local `struct TextureSequence { ... };` declarations, which raised `shadow_decls.local_body_no_include` 1507 → 1510 and failed the langmode ratchet. #1397 uses the real headers instead: 1507 → 1503, with the ROM unmoved.

Closing in favour of #1397.

PR #1438, opened 2026-08-12
Match 3 functions byte-identical (mwccarm 2004/b56)

lunavyqo:
Byte-identical matches on mwccarm 2004/b56:

- func_ov091_021339fc
- func_ov080_02125460
- func_ov102_0214b53c (this push) — Ghidra dest + ROM-order angle; linkcheck VERIFIED


andrewboudreau (commented):
## Verification

Fresh worktree at `7b6a97c`, with `extracted/`, `tools/mwccarm` and `tools/bin/dsd.exe` wired in.

| check | result |
|---|---|
| `build_pin.verify`, pinned **2004/b56** + build flags | all 3 **MATCH** |
| `linkcheck.py` per function | all 3 **VERIFIED**, 0 blind slots |
| `rombuild.py --no-rom` (full 10,820-file link) | **FAILS** — reproduces the CI failure exactly |

All three removed `NONMATCHING` banners are honest match claims — I rebuilt each one. The relocation slots are real, not wildcarded luck. But the tree does not link.

---

## Blocking

### 1. `func_020aea30` is a phantom symbol, and enrolling `func_ov091_021339fc.c` breaks the whole ROM link

```
mwldarm.exe: Undefined : "func_020aea30"
mwldarm.exe: Referenced from "func_ov091_021339fc" in func_ov091_021339fc.o
mwldarm.exe: alert: Link failed.
```

`func_020aea30` exists in no `symbols.txt`. It is a *known* phantom — `config/unresolved-baseline.json` lists it as the blocker for ten other functions (ov062, ov065 ×2, ov070, ov081, ov084 ×3, ov090, `_ZN7Skeeter8BehaviorEv`). Those files were never enrolled, so the reference never reached the linker. This PR is the first to add a `delinks.txt` entry for a file that carries it.

The symbol that *does* exist at that address is `func_ov002_020aea30` (`config/arm9/overlays/ov002/symbols.txt:26`, size `0x8c`). `ov091/relocs.txt:406` records `to:0x020aea30 module:overlays(2,4)`; ov004's alternative there is `_ZN5Enemy12KillByAttackER5Actor` (size `0x48`, arity 2), which cannot take the 3-argument call.

The fix is one word and **byte-free** — verified at all three levels:

```diff
-extern void func_020aea30(void *c, void *a, u32 unused);
+extern void func_ov002_020aea30(void *c, void *a, u32 unused);
@@
-            func_020aea30(c, a, 0);
+            func_ov002_020aea30(c, a, 0);
```

```
build_pin.verify  → MATCH 2004/b56
linkcheck         → VERIFIED, 0 blind
rombuild --no-rom → 106/106 exact, 100.000000%, 0 mismatching, 88.09%
```

**Worth carrying forward:** `linkcheck` *cannot* catch this. It resolves bare `func_<addr>` names itself, so it reports VERIFIED on a symbol the linker has never heard of — which is why the per-file table above is green while the build is red. Only the full link sees it. Any file newly added to `delinks.txt` needs `tools/rombuild.py --no-rom` before the match is claimed.

---

## Should fix — all byte-free, all verified

### 2. The new header's comment is factually wrong

`include/private/ov080_mat.h` asserts the `union { u32 w; }` wrapper is load-bearing, and that "plain `u32 param` misses the r2/r3 rank". It doesn't — plain `u32 param` (with `param.w` → `param`) still matches byte-for-byte under 2004/b56. I tested exactly that.

A single-member union isn't something the original developers wrote, and a false lever recorded in a header costs the next reader the hours it claims to save. Please drop both the union and the comment.

### 3. `func_ov102_0214b53c.c` is largely un-post-processed Ghidra output, and most of it is free to clean

I ran each candidate through `build_pin.verify` individually and then all together — the composed set still MATCHes **and** linkchecks VERIFIED:

| cleanup | |
|---|---|
| drop the six chained `& 0xFFFFFFFFu` on line 30 | FREE |
| `data_020a0e68.m[9..11]` instead of `(char*)&… + 0x24/0x28/0x2c` (volatile site) | FREE |
| delete the unused `int gbase = (int)&data_020a0e68;` | FREE |
| `off5ec` → the literal `0x5ec` | FREE |
| drop the `new_var = (char*)&p` round-trip | FREE |
| `volatile u16 tmp[6]` → two named `{u16 x,y,z}` — they **are** two `Vector3_16`s (obj+0x8c and obj+0x580) | FREE |
| `data_ov102_0214ea18/1c/20` → one `Vector3 tbl[]`, `tbl[idx].x/.y/.z` | FREE |
| rename `new_var` / `new_var2` / `new_var3` | FREE (renames can't move bytes) |

The `tmp[6]` and table items are the valuable ones — they're not cosmetic, they recover real structure. The three `data_ov102_0214ea18/1c/20` symbols are 4 bytes apart and indexed with `off = idx * 0xc`; that is one `Vector3 tbl[4]`, and saying so removes two bogus externs.

And these **are** genuinely load-bearing — keep them, but each deserves a one-line comment saying so, because they currently read as noise:

- the byte-offset reads of `data_020a0e68` at the **first** site (typed `.m[9..11]` costs 6 words)
- the byte-offset reads of local `d` (typed `d.x/d.y/d.z` costs 4 words)
- the `(char*)*((int*)(c+0xd0))` pointer relaunder (typed `*(char**)` changes the function *size*)
- ov091's dead `rot.y = 0;` before `rot.y = *(s16*)(c + 0x94)` (removing it changes the size)

### 4. Information was lost from the deleted `.cpp`

The old `func_ov102_0214b53c.cpp` had

```c
typedef struct { u16 ang8c, ang8e, ang90, v580, v582, v584; } Tmp;
```

and used `data_020a0e68.w[9..11]`. The rewrite replaced both with `tmp[6]` and raw byte offsets. Per #3 that's restorable for free — please don't let a matching pass drop recovered structure it didn't author.

### 5. Duplicated G3 helpers

`src/func_0204af3c.c` already carries identical `G3_TexImageParam` / `G3_TexPlttBase`, the same two register macros (spelled `reg_`, not `REG_`), and the same `#pragma opt_propagation off`. Aligning to `reg_` is free (verified). Since `include/private/` now has a home for this, moving the register block and the two inlines there would be the better end state — but that touches an already-matched file, so `func_0204af3c` needs re-verifying in the same commit.

### 6. Three MMIO writes left raw while two got macros

`0x40004c0`, `0x40004c4`, `0x40004a4` — same register block as the two you named (DIF_AMB, SPE_EMI, POLYGON_ATTR), and the only three not marked `volatile`. Naming them is free (verified).

### 7. Bookkeeping is complete for ov091 and ov102 but missing for ov080

- `CLAIMS.md:289` still reads `**active** — batch6 near-miss div=5` for `func_ov080_02125460`
- `nearmiss/db.jsonl:16` still holds its div=5 row

The ov091 (div 86) and ov102 (div 60) rows were correctly retired — ov080 just got missed.

---

## Worth checking

**8. Does this unblock ov075?** The deleted ov080 banner and `CLAIMS.md:78` both record `func_ov075_0211a948` as *the same* div=5 TEXIMAGE_PARAM two-attractor wall. You found the lever here. Its tip is banked in `nearmiss/db.jsonl` and there's no `src/func_ov075_0211a948.*` yet — this looks like a free second match.

**9. `struct Ov080Mat` vs `struct Tex`** (in `func_0204af3c.c`) describe overlapping GX material state offset by 4 bytes — `Tex`: `f0/texAddr/pltAddr/param` at +0/+4/+8/+0xc; `Ov080Mat`: `texAddr/pltAddr/difAmb/speEmi/param` at +0/+4/+8/+0xc/+0x10. Two different objects, or one object with a leading field missing from `Ov080Mat`?

**10. The `Particle::System::New` extern contradicts its own mangled name.** `…Ejj5Fix12IiES2_S2_PK11Vector3_16fPNS_8CallbackE` demangles to 8 explicit params (`unsigned, unsigned, Fix12<int>×3, const Vector3_16*, float, Particle::Callback*`), but it's declared and called with 6 after `sys`. **Pre-existing** — identical in the deleted `.cpp` and used tree-wide — so not this PR's to fix, but a `TODO:` would be honest. Related: `config/unresolved-baseline.json:325` already carries the double-mangled `_Z70_ZN8Particle…jjiiiPKvPv` form from some C++ TU that declared it without `extern "C"`.

**11. Names available but out of scope** (they'd need `symbols.txt` edits): `Math_Function_0203b14c`, and `data_02082214` — a sin/cos table, indexed `(angle >> 4) * 2`, with `+1` giving the quadrature term.

---

## Mechanical

**12.** Six commits for three functions; `a3a0ce6` (div 60→59) and `810b622` (div 59→39) are superseded by `7b6a97c`, and `91497a5` has a truncated subject ("…dropped from the 39-div tip co…mmit"). Not blocking — a squash would make the history read as the three matches it is.

**13.** No `.clang-format` in the tree by design, so there's no format gate to run.

---

Worth saying plainly: **the ov091 change is the best thing in this PR.** It deletes a real `asm { }` hatch and replaces it with C that reproduces the bytes — it converts a hand-asm non-recovery into a genuine one. The ov102 file is the opposite end: a real match wrapped in decompiler artifacts, most of which come off for free.

With #1 applied plus every cleanup in #2/#3/#5/#6, I get a clean tree locally:

```
source-built functions: 10,820  (1,947,748 / 2,211,124 code bytes, 88.09%)
  reproducing: 10,820
  mismatching: 0
ROM-build analysis: PASS
```

I have that as a patch (+45/−53 across the four files) and can push it to the branch if you'd rather not redo it by hand.

<sub>Review assisted by Claude Opus 5; every MATCH/FREE/COSTLY verdict above is a real local build with the pinned compiler, not an inspection.</sub>

andrewboudreau (commented):
# PR #1438 — second review (head `b9356ee`, was `7b6a97c` at first review)

## Verification

Everything below is a real local build on the **current head** `b9356ee`, pinned
**mwccarm 2004/b56** (`tools/build_pin.py` → `compiler_for()` returned `2004/b56` for
all three files; no override in play). I did not trust the first review's numbers —
I re-ran all of it.

| check | result |
|---|---|
| `build_pin.verify` × 3, pin 2004/b56 | `func_ov091_021339fc` **MATCH**, `func_ov080_02125460` **MATCH**, `func_ov102_0214b53c` **MATCH** |
| `linkcheck.py` × 3 | all **VERIFIED**, `blind: 0`, `diffs: []` |
| `tools/rombuild.py --no-rom` (full link, 10,820 enrolled sources) | **PASS** |
| removed `NONMATCHING` banners | 3, all honest — each rebuilt above |
| `asm {}` / `dcd` bodies in the three files | none |
| every `extern`/prototype in the three files vs `config/**/symbols.txt` | **26/26 resolve** |
| CI (`gh pr checks`) | PR validation **pass**, ratchet **pass**, submit **pass** |

Verbatim from `rombuild.py --no-rom` on this head:

```
source-built functions: 10,820  (1,947,748 / 2,211,124 code bytes, 88.09%)
  reproducing: 10,820
  mismatching: 0
ROM-build analysis: PASS
```

Not checked: nothing material. Everything the first review flagged as unverifiable
by the byte gate (the link) is now verified by the link.

---

## Lead: the blocking finding is fixed

**Finding #1 of the first review — the phantom `func_020aea30` — is resolved.**
Commit `b9356ee` "Fix ov091 ROM link: call func_ov002_020aea30 by its real name" is
exactly the two-line, byte-free fix that was proposed, at both the declaration and the
call site. I confirmed three ways:

- `func_020aea30` no longer appears in any file this PR touches.
- `func_ov002_020aea30 kind:function(arm,size=0x8c) addr:0x020aea30` exists in
  `config/arm9/overlays/ov002/symbols.txt`.
- The full `mwldarm` link now completes: `106/106 exact, 0 mismatching`. The CI
  failure the first review reproduced is gone, and the validator bot's comment
  (updated `20:20Z`, after the fix commit at `20:16Z`) agrees.

I also re-ran the phantom sweep over **all** externs in **all three** files, not just
the one that broke. All 26 referenced symbols — the mangled `_ZN…` set, `func_02012694`,
`func_ov002_020e496c`, `func_ov102_0214b444`, `Math_Function_0203b14c`,
`Matrix4x3_FromRotationY`, `MulMat4x3Mat4x3`, `Vec3_Lsl`, `Vec3_LslInPlace`,
`data_02082214`, `data_020a0e68`, `data_ov102_0214ea18/1c/20` — resolve to a real
`config/**/symbols.txt` entry. No second phantom.

**Nothing regressed.** `git diff 7b6a97c..b9356ee` is 2 insertions / 2 deletions in one
file. Nothing was made worse.

---

## Status of every prior finding

| # | prior finding | status |
|---|---|---|
| 1 | BLOCKING phantom `func_020aea30` | **addressed** — verified at all three levels |
| 2 | false "load-bearing union" comment in `include/private/ov080_mat.h` | **ignored** |
| 3 | `func_ov102_0214b53c.c` is raw Ghidra output; 8 cleanups measured FREE | **ignored** |
| 4 | recovered structure lost vs the deleted `.cpp` | **ignored** |
| 5 | duplicated G3 helpers; `REG_` vs the tree's `reg_` | **ignored** |
| 6 | three raw MMIO literal writes left unnamed | **ignored** |
| 7 | ov080 bookkeeping — stale `**active**` claim + stale near-miss row | **ignored** |
| 8 | does this unblock ov075? | unanswered (still no `src/func_ov075_0211a948.*`) |
| 9–11 | open questions | unanswered |
| 12 | commit hygiene | unchanged (now 7 commits) |

The single commit since the first review fixed #1 and touched nothing else.

---

## Blocking

**Nothing blocks the merge on correctness.** The match is real, the externs resolve,
and the tree links.

One thing I would still hold for, because it is *disinformation* rather than untidiness:

### B1. `include/private/ov080_mat.h` asserts a compiler fact that is false

```c
/* GX material block at *(self+0x1a8) for func_ov080_02125460.
   The union wrapper on `param` is load-bearing for mwccarm 2004/b56
   TEXIMAGE_PARAM coloring (plain `u32 param` misses the r2/r3 rank). */
```

I re-measured this on the current head. Replacing the whole thing with

```c
    u32 param;
```

and `m2->param.w` → `m2->param` at its four uses:

```
build_pin.verify  func_ov080_02125460 -> MATCH 2004/b56
linkcheck                             -> VERIFIED, blind 0
```

**FREE.** The single-member union is not load-bearing, and a header in `include/`
that claims a compiler lever exists where none does is worse than no comment at all —
the next person to hit the ov075 twin (same TEXIMAGE_PARAM wall, `CLAIMS.md:78`) will
spend hours on a lever that isn't there. This is one file, two edits, and it is the
only reason I would not just merge as-is.

---

## Should fix — every item below re-measured FREE on this head

I applied the first review's full suggestion patch to `b9356ee` (minus the ov091 hunk,
already landed), and re-ran the gates:

```
func_ov091_021339fc: MATCH 2004/b56   linkcheck VERIFIED, blind 0
func_ov080_02125460: MATCH 2004/b56   linkcheck VERIFIED, blind 0
func_ov102_0214b53c: MATCH 2004/b56   linkcheck VERIFIED, blind 0
```

So the whole cleanup is still free at the current head, patch still applies cleanly,
and it includes the one item I was most suspicious of — `extern Vector3
data_ov102_0214ea18[]` with `[idx].x/.y/.z` replacing three separate externs indexed
by `off = idx*0xc`. That change moves relocation *destinations*, so a byte match alone
would not settle it; `linkcheck` VERIFIED with 0 blind slots does.

### S1. `func_ov102_0214b53c.c` is still unprocessed decompiler output

Unchanged since the first review. The items, all re-verified FREE:

- the six chained `& 0xFFFFFFFFu` no-ops on line 30
- `volatile u16 tmp[6]` → two named `{u16 x,y,z}` structs — they are the rotation
  triples at `obj+0x8c` and `obj+0x580`
- `data_ov102_0214ea18/1c/20` → one `extern Vector3 data_ov102_0214ea18[]`
  (the three symbols are 4 bytes apart, indexed `idx*0xc`; that is one `Vector3 tbl[4]`
  and saying so deletes two bogus externs)
- `int gbase = (int)&data_020a0e68;` — unused, delete
- `int off5ec = 0x5ec;` — inline the literal
- `new_var = (char*)&p;` round-trip before `Vec3_LslInPlace((Vector3*)new_var, 3)`
- `*(volatile int*)((char*)&data_020a0e68 + 0x24/0x28/0x2c)` →
  `*(volatile s32*)&data_020a0e68.m[9/10/11]` (`Matrix4x3` is `{s32 m[12]}`,
  `include/common.h:17`)
- `new_var` / `new_var2` / `new_var3` — Ghidra names. Renaming a local cannot move
  bytes; this is free by construction.

### S2. Load-bearing oddities that need one comment line each

These are the *opposite* problem — they look like noise, they are not, and there is
nothing in the file saying so. I measured each by deleting it and rebuilding:

| site | test | result |
|---|---|---|
| ov102 `volatile Vector3 dead;` + its 3 stores from `obj+0x5c` | delete | **COSTLY** (`999 word(s) differ` — size changes) |
| ov102 `p.x = p.y = p.z = 0;` before the matrix copy | delete | **COSTLY** |
| ov091 `int b = (int)(*(u16*)(a+0xc) == 0xbf); if (b == 0)` — **site 1** | → `if (… != 0xbf) return;` | **COSTLY** |
| ov091 same idiom — **site 2** | → `if (… != 0xbf) return;` | **COSTLY** |
| ov091 dead `rot.y = 0;` before `rot.y = *(s16*)(c+0x94)` | delete | **COSTLY** |

(New information vs the first review: both ov091 comparison-launder sites are
independently load-bearing — I tested them separately, not just together. This tree's
usual pattern is that only *some* sites in a family matter; here both do.)

Also load-bearing per the first review and left as-is: the byte-offset reads of
`data_020a0e68` at the *first* site, the byte-offset reads of local `d`, and the
`(char*)*((int*)(c+0xd0))` pointer relaunder.

Every one of these deserves a `/* launder: … costs N words if typed */`-style line.
Comments cannot move bytes.

### S3. ov080 bookkeeping is still inconsistent

- `CLAIMS.md:289` still reads `**active** — batch6 near-miss div=5` for
  `func_ov080_02125460`, and no `**done**` row was added for it. ov091 (line 42) and
  ov102 (line 502) both got theirs.
- `nearmiss/db.jsonl:16` still carries the ov080 div=5 row. The ov091 and ov102 rows
  were correctly retired (that is the `-2` in the diff).

### S4. Macro spelling is 1-against-2

`src/func_ov080_02125460.c` spells the register macros `REG_G3_*`. The tree's two
existing copies — `src/func_0204af3c.c:5` and `src/func_ov007_020c1448.c:3` — both
spell them `reg_G3_*`. Aligning is free (in the patch above).

To be fair on the other half of the first review's #5: this is now the **third** copy
of `G3_TexImageParam`/`G3_TexPlttBase` in `src/`, and the two prior copies are on main.
Per-file inline helpers are the established convention here, so I would not bill this
PR for the duplication itself — only for the spelling drift.

---

## Worth checking

**W1. Is `Ov080Mat.param` the raw packed register word, and should it be a bitfield?**
`func_0204af3c.c` already carries a recovered `struct TexParam` (`fmt:4, sizeS:4,
sizeT:4, repeat:2, flip:2, color0:1`). But that one is a *decoded* struct — `fmt` sits
at bit 0 and `G3_TexImageParam` packs it up to bit 26. `Ov080Mat.param` is extracted at
bits 26 / 20 / 23 / 29, i.e. it already holds a packed TEXIMAGE_PARAM word. So they are
two different objects and `Ov080Mat` is not a 4-byte-shifted `Tex`. That answers the
first review's Q9 — but it leaves a smaller question: is a bitfield struct laid out to
the *register* positions worth recovering here, or do the shifts stay?

**W2. Does this unblock `func_ov075_0211a948`?** Still open, and still worth an answer —
the deleted ov080 banner named ov075 as the same two-attractor TEXIMAGE_PARAM wall
(div=5), `CLAIMS.md:78` and `:300` agree, its near-miss tip is banked at
`nearmiss/db.jsonl:15`, and there is no `src/func_ov075_0211a948.*`. If the ov080 lever
transfers, that is a free second match.

**W3.** The `Particle::System::New` arity contradiction (mangled name says 8 params
after `sys`, declared with 6) is pre-existing and tree-wide — not this PR's to fix, but
a `TODO:` in the new file would be honest.

---

## Mechanical

**M1.** The knowledge in the deleted ov080 `NONMATCHING` banner (the ov075 twin
relationship, the pointer to `notes/mwccarm-codegen.md` 6ba/6bb, the ~330+55 compile
map) is now only in `notes/` and `config/match_attempts.jsonl`. The banner had to go —
it was a NONMATCHING claim on a matched file — but with S3's missing `**done**` row,
ov080 currently has *no* in-tree record of how it was cracked. The `**done**` row
should carry the lever, the way ov091's and ov102's do.

**M2.** 7 commits for 3 functions; `a3a0ce6` (div 60→59) and `810b622` (div 59→39) are
superseded by `7b6a97c`, and `91497a5` has a truncated subject. A squash would make the
history read as the three matches it is. Not blocking.

**M3.** Provenance is properly disclosed — `config/match_provenance.jsonl` records
`{"kind":"ai","model":"grok-4.6","reasoning":"xhigh","harness":"grok-build"}` for all
three. No `.clang-format` in the tree by design, so no format gate to run.

---

## Verdict

**Merge with fixes.** The correctness case is closed: three honest matches, verified
under the canonical pin, every extern resolves, and the full ROM link is clean at
`106/106 exact, 0 mismatching`. The contributor fixed the one blocking issue with
exactly the right patch and broke nothing doing it.

What remains is source quality, and it is all free — I re-applied the first review's
patch to this head and every gate still passes. The fastest path is for that patch
(+45/−53 across four files, minus the already-landed ov091 hunk) to go on the branch,
plus the ov080 `**done**` row and the `nearmiss/db.jsonl:16` retirement, plus a comment
line on each of the five load-bearing idioms measured in S2.

If it merges without those, B1 (the false union comment) is the one that will cost
someone real hours later.

lunavyqo:

PR #1445, opened 2026-08-12
Bank near-miss for ov003 func_ov003_020af8a0 (div 4→1)

lunavyqo:
Bank the improved near-miss for `func_ov003_020af8a0` (ov003 `dScStarSel_c` InitResources, `0x020af8a0`, size `0xc50`) **4 → 1** on mwccarm 2004/b56.

No `src/` change — tip only. Not a match; do not enroll.

## What is saved here

| File | What |
|---|---|
| `nearmiss/db.jsonl` | Best C tip (**div=1**, size-exact 0xc50). Load-bearing: `#pragma opt_loop_invariants off` + `opt_strength_reduction off`, then late-named `half = tmpU8 >> 1` after the colorVal expression and `idx2 = half; idx2 -= half`. |
| `config/match_attempts.jsonl` | Attempt tree for this session (see below). |
| `CLAIMS.md` | Range marked **released** with the 4→1 note. |

## Attempt tree (this PR)

```
ov003:0x020af8a0
  30fed3d8…  no_progress  div=4   (prior floor reconfirm)
    └─ bed6c597…  no_progress  div=4   reconfirm included tip
         └─ d6e17cf6…  near_miss  4→1   half-liveness lever
              └─ 3a459b98…  no_progress  div=1   retry: m2c / pragmas / SEs / permuter ~580it
```

## Residual (still 1 SCHED word)

```
+0x90c  ROM: mov r2, #0
        ours: sub r2, r1, r2, asr #1
```

The `sub` is what keeps `r1` live through the `and`, so colorVal stays in `r3` like the ROM. Every zero-cost last-use of `half` is DCE'd before register allocation and falls back to div=4.

Author: lunavyqo.


andrewboudreau (commented):
# Review — PR #1445 "Bank near-miss for ov003 func_ov003_020af8a0 (div 4→1)" (lunavyqo)

**Verdict: merge, after one one-line fix (re-mark the floor).** Nothing here is a lie, and
nothing here touches the build.

This PR touches no `src/` file. Nothing is enrolled, no `NONMATCHING` banner is removed, the
progress bar and contributor credit are untouched. There is no match claim to falsify, so
Pass 0 reduces to: *is the banked number real, is the key right, and is the bank honest?*
All three: yes. `rombuild.py --no-rom` was deliberately **not** run — there is nothing for
the linker to see.

---

## Verification

Everything below I rebuilt myself, with **mwccarm 2004/b56** — which is also what the DB
tool itself uses (`nearmiss_db.evaluate()` → `match.compile_c(..., M.CANONICAL)`,
`match.py:86 CANONICAL = "2004/b56"`). So the record's `divergences` field is a canonical-pin
number, *not* a `match.py --all` multi-version sweep result. The CLAIMS row's phrase
"match.py words" is fair: it means match.py's `compile_c` at the pin, then a reloc-wildcarded
`difflib` instruction diff.

Compiler liveness proved first: feeding the same source with a deliberate syntax error
returns `None` with a real `mwccarm.exe ... Errors caused tool to abort`, so these are real
compiles, not a cache.

| Claim | Measured |
|---|---|
| old record (origin/main) `div=4` | **div=4** — reproduced |
| new record `div=1` | **div=1** — reproduced |
| size-exact `0xc50` | compiled **0xc50** (3152) both old and new; `size` field 3152; target 0xc50 |
| residual at `+0x90c`, ROM `mov r2,#0` vs ours `sub r2,r1,r2,asr#1` | **exactly that, and it is the only divergence** |
| `target_hex` | byte-identical to the ROM at `ov003:0x020af8a0` for 0xc50 bytes, and unchanged from main's record |

Old residual, for the record — 4 words, all r1-vs-r3:

```
+0x908  OURS and  r1, r0, #0xff        ROM and  r3, r0, #0xff
+0x918  OURS strb r1, [r0, #0x11a]     ROM strb r3, [r0, #0x11a]
+0x920  OURS add  r1, r1, #0x23        ROM add  r1, r3, #0x23
+0x92c  OURS and  r1, r1, #0xff        ROM and  r3, r1, #0xff
```

That is *precisely* the "pure r1↔r3 colorVal web on 0x11A fill loop (+0x908 and/strb/add/and)"
the deleted `floor` evidence described. The old floor claim was accurate; the new one is too.

**Not pin-shopped.** div=1 also holds at `1.2/sp2p3` and `1.2/base` (both size-exact 0xc50);
only `2.0/base` diverges (div=738, size 0xbf4). The number is not an artifact of one compiler.

**The lever is a semantic no-op.** The whole source diff old→new is five lines:

```c
-        idx2 = 0;
+        {
+            s32 half = (s32)tmpU8 >> 1;
+            idx2 = half;
+            idx2 -= half;
+        }
```

`half - half` is provably 0, so the new tip computes the same value as the old one. This is a
register-liveness lever, not a behaviour change — the tip has not been "improved" by drifting
away from the function's meaning.

**Identity is correct — no #1461-style misattribution.** The PR calls this
`dScStarSel_c::InitResources`. Verified from the vtable, not the name:

- vtable at `0x020b1700`, typeinfo word `0x020b16bc` = `_ZTI12dScStarSel_c`
  (`_ZTS12dScStarSel_c` at `0x020b16d4` spells `12dScStarSel_c`)
- slot 0 = `0x020b1704` → **`0x020af8a0`**, this function
- the same slot in `dScTitle_c`'s vtable (`0x020b164c` / `0x020b1650`) holds `0x020ada9c`,
  which the tree already names `dScTitle_c_InitResources`
- slot 2 pairs `0x020af86c` (`dScStarSel_c_CleanupResources`, already in `src/`) with
  `0x020ada80` (`dScTitle_c_CleanupResources`)
- `0x020b16b4` = `0x020b04f0` = `StarSelect_Spawn`, sitting immediately before the StarSel
  typeinfo — the factory pairs to this class by address, so the `_Spawn` half agrees too

`config/arm9/overlays/ov003/symbols.txt:25` confirms `func_ov003_020af8a0 size=0xc50 addr:0x020af8a0`.

**Ledger validity.**

- exactly **one** record for `(ov003, 0x020af8a0)`; no duplicate `(module, addr)` keys anywhere
  in the DB (`nearmiss_db.py dedupe --dry-run`: "no duplicate rows; DB has 189 entries")
- `addr` switching from `"0x020af8a0"` to the int `34273440` is **not** a problem —
  `ledger.norm_addr` accepts both and 27 rows on main already use ints
- the 3 added `config/match_attempts.jsonl` rows all pass `match_attempts.validate_attempt()`;
  the parent chain resolves end-to-end (`30fed3d8` → `bed6c597` → `d6e17cf6` → `3a459b98`),
  `improvedNearMiss` is set correctly on exactly the one that improved, and all three disclose
  `kind:"ai", model:"grok-4.6"`
- `claims_md.is_held("0x020af8a0")` → **False**. "API `clm_088541be4161` released" is true —
  the range does not stay locked after merge. This is a fix relative to the pattern noted on
  #1444/#1448, where the author kept the API claim and the range stayed `is_held()`.

**CI / mergeability.** `gh pr checks 1445`: `PR validation` pass (2m48s), `submit` pass.
`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` — the `UNKNOWN` GitHub reported has resolved;
no conflict with current main. Both `nearmiss/db.jsonl` and `CLAIMS.md` are `merge=union`
(`.gitattributes`), so a sibling PR landing first cannot conflict; the union would leave the
old div=4 row alongside the new div=1 row, and `nearmiss_db.py dedupe` in the
`update-chaos-data` workflow collapses it to the lower div. That path is designed for and works.

---

## Blocking

None.

---

## Should fix

**1. The `floor` is dropped and never re-applied — this re-arms the function for automated
compute that the same PR declares dead.**

Dropping the old floor is *correct*: `nearmiss_db.py`'s own docstring says "a strictly better
ingest replaces the record and clears the mark (the residual changed, the floor claim is
stale)", and I confirmed the residual really did change — the r1↔r3 web the old evidence
described is gone. The old evidence prose is also **not** destroyed: it survives verbatim in
`config/match_attempts.jsonl` attempt `30fed3d8` on main ("40+ variants this session … all
size-exact stay at 4"). So this is not the "deleted hard-won evidence" defect it first looks
like.

The problem is the second commit. `7b6afbdff` is titled "Log no_progress retry … (div=1
floor)" and its note records m2c, a pragma-off sweep, a flag sweep, statement-exprs,
load-anchored SEs, union/ptr-walk, and ~580 permuter iterations, all failing to beat +0x90c.
That is a floor claim — but no `mark-floor` was run, so the record ships unfloored at div=1.

Measured consequence:

```
main    : export-close --max-div 8 → 18 seeds, func_ov003_020af8a0 NOT present (floored)
PR #1445: export-close --max-div 8 → 19 seeds, func_ov003_020af8a0 PRESENT
```

The DB floor count goes 45 → 44. So after this merge the permuter and `refine_wl.py` will
start re-offering a function the author has just spent a session proving is stuck — exactly
what the floor mechanism exists to prevent. One line:

```
python tools/nearmiss_db.py mark-floor --name func_ov003_020af8a0 \
  --class materialization \
  --evidence "div=1 size-exact. Residual +0x90c ROM mov r2,#0 vs sub r2,r1,r2,asr#1: the sub is what keeps r1 live so colorVal coalesces to r3; every zero-insn last-use of half is DCE'd before regalloc and falls back to div=4. m2c, pragma-off, flag sweep, statement-exprs, load-anchored SEs, union/ptr-walk, permuter ~580it all failed. 2026-08-12."
```

**2. "1 SCHED word" is the wrong floor class.** The residual is not a scheduling divergence —
both sides emit one instruction in the same slot, and the difference is *how a constant zero
is materialised* (`mov r2,#0` vs a computed `half - half`). The DB's floor classes already
include `materialization` for exactly this, and `export-close --category` routes permuter work
off the classification. Calling it SCHED will send it to the wrong tier. Same word appears in
the PR body, the CLAIMS row, and two `match_attempts` notes.

---

## Worth checking

**3. The lever is a laundering idiom, and it may be *why* the last word cannot be fixed.**
`idx2 = half; idx2 -= half;` is a dead computation whose only purpose is to hold a register.
That is entirely legitimate *in the near-miss DB* — the DB's stated job is to keep
"logically-correct, already-compiling" drafts, not admissible source. But note the shape of
the trade: the old tip got `mov r2,#0` right and 4 register assignments wrong; the new tip
gets the register web right and can no longer emit `mov r2,#0` **because the source no longer
says zero**. Is a div=1 whose single residual is manufactured by the lever actually a better
starting point than a div=4 without it? By the DB's contract (lowest div wins) yes, and I am
not arguing for a revert — but whoever picks this up should know the last word is not
independent of the lever, and that this idiom cannot ship into `src/` without a comment.
`delaunder.find_sites` counts 1 launder site in both the old and the new tip, so the ratchet
delta is zero either way (and the ratchet does not read the DB regardless).

**4. Author column dropped the `(Grok)` tag.** Every other row this author owns reads
`lunavyqo (Grok)`; the replacement row reads bare `lunavyqo`. AI disclosure is intact in
`match_attempts.jsonl` (`kind:"ai"`, `model:"grok-4.6"`), so nothing is hidden — but the
CLAIMS table is now inconsistent for this contributor. Restore the tag or drop it everywhere.

---

## Mechanical

**5. Context, explicitly NOT billed to this PR:** `src/func_ov003_020af8a0.c` on main still
carries `// NONMATCHING: near-miss from DB (div=153)` and I measured it at exactly div=153 —
now **152 words worse** than the banked tip. This is a tree-wide condition, not a defect
introduced here (15 of the 24 DB entries that have a matching `src/` draft disagree with their
banner; 112 have no `src/` draft at all). But this PR makes `func_ov003_020af8a0` the widest
gap in the tree, and refreshing that draft from the tip — banner kept, so still unenrolled and
still zero progress-bar effect — is a free follow-up while the context is loaded.

lunavyqo:

PR #1446, opened 2026-08-12
Log no_progress floor reconfirm for arm9 OAM::Render (div=2)

lunavyqo:
Floor reconfirm for `_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii` @ `0x02020994` (arm9, size `0x690`) on **mwccarm 2004/b56**.

`validate` is not expected to change: `src/` is untouched. The included near-miss still compiles size-exact with **2 words** swapped at `+0x38`/`+0x3c` (ROM `ldr fp,[sp,#0x8c]` then `str r0,[sp,#0x94]`; tip reverse). That is the documented overflow-filler zip (notes 6ay / 6ai).

## Attempt tree (all logged in `config/match_attempts.jsonl`)

```
arm9:0x02020994
├─ 4732c8ebf47d4c6bb54e66b9d5602411  no_progress div=2  parent=null
│    base=near_miss_draft  usedNearMiss=true usedGhidra=false
│    FLOOR reconfirm: tip size-exact 0x690; residual +0x38/+0x3c swap
│  ├─ 1dba2c20e4e446eba24540692ffade5c  no_progress div=2
│  │    prologue probes (syv-first/vol/launder/mode-sandwich/addr-of-mode);
│  │    C frontend + scalar ints still same 2 words; vol-mode-store size 0x68c
│  ├─ e04d781e008647afb7ce1d5895439916  no_progress div=2
│  │    C-scalar seed (permuter score 60) + PERM_RANDOMIZE; ~1640 it -j4 no score<60
│  └─ 13dd53a7bc384910bb23114fa34f77d4  no_progress div=2
│       usedGhidra=true  usedNearMiss=true
│       second pass: Ghidra + first-if (global_first size-exact div=18),
│       gccext SEs, BB barriers, K&R, asm, optlevel3, all-version sweep
│       2004/b56 + 1.2 trio all stay div=2
└─ 42d6617165414da780e084b86dfda114  no_progress div=999  parent=null
     base=scratch  usedGhidra=true usedNearMiss=false
     fresh Ghidra headless dump compiled as C; frame 0x44 vs ROM 0x5c
```

## Claims

- `clm_18ba143329c5` released (first pass)
- `clm_4c4701e3a85d` released (second pass / Ghidra)

No `src/` change; tip remains the existing NONMATCHING draft + `nearmiss/db.jsonl`. Unlock remains mwccarm 0057–0062.


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1446 — arm9 `OAM::Render` `_ZN3OAM6RenderEbP7OamAttriiii5Fix12IiES3_ii` (div=2)

### Spot-verified (lowest div in the set, so I reproduced it)

`evaluate()` = **2**, `match.compare` words = **2**, size **exact `0x690`**, `target_hex`
byte-identical to the ROM at `arm9:0x02020994`. The claim is real: this record is two words
from a match, and the two words are the documented `+0x38`/`+0x3c` prologue overflow-filler zip.

- 5 attempt rows, all schema-valid; the `42d6617…` fresh-Ghidra branch is honestly logged at
  `div=999` with `base:{kind:"scratch"}` and `parentAttemptId:null`. Both API claims
  (`clm_18ba143329c5`, `clm_4c4701e3a85d`) released; `is_held` False both ways.
- **The floor is already on main** (`class:"build discriminator (prologue overflow-filler
  placement)"`, dated 2026-07-28). Not the #1445 defect.
- **Two CLAIMS rows added (08-12 and 08-13, both div=2)** on top of main's existing 08-07 row
  for the same function and the same conclusion. Same duplicate-row shape as #1459; multi-row
  is convention (`.gitattributes` documents duplicates as harmless) so flagging, not billing —
  but with the 07-28 DB floor this is the third and fourth logging of one settled result.

**Verdict: merge.** The number is real and the bookkeeping is valid; the redundancy is the only
thing worth a comment.

---

---

   recomputing all three.
   rows.

PR #1447, opened 2026-08-12
Bank near-miss for ov006 func_ov006_020dbe9c (div 15→7)

lunavyqo:
Improve the stored near-miss for `func_ov006_020dbe9c` (ov006, 0x020dbe9c, size 0xe0) from **15 → 7** word diffs on mwccarm **2004/b56**. Size-exact.

**Lever:** split the Fix12 expand as `s64 x = (s64)v << 12; x += 0x800;` for cosine, then the same for sine (with `#pragma opt_propagation off` + named `s16 cosv`). That reproduces the ROM dual-s64 expand through +0x7c (orr-before-lsl on cos, delayed sin orr, sin-hi in lr).

**Residual:** `>>12` of sine before cosine (`idx_l` lands in `lr` instead of `r6`). Stores, prologue, and the call setup already match.

Tip is in `nearmiss/db.jsonl`. No `src/` file (not a byte match).

### Attempt tree (this PR)

| attemptId | parent | status | div | note |
|---|---|---|---|---|
| `84629ac607c7407895bcb5715ecbd45b` | `da62918531284769a701af70f7ec9dab` | no_progress | 15 | reconfirm stored tip + `common.h` |
| `60d17b5d6e334f96bf6aac0c953ef2af` | `84629…` | near_miss | **11** | split-only-cos (`x = (s64)v<<12; x += 0x800`) |
| `610d9094346e4824880568c9b356095b` | `60d17…` | near_miss | **7** | seq_split: finish cos shift+plus then sin shift+plus |
| `8da5239b24e248d49219375ae70bda7b` | `610d9…` | no_progress | 7 | permuter ~768it on seq_split, score 120 floor |
| `a337c53c3cad462788cbccb95fa3579a` | `610d9…` | no_progress | 7 | rematch: Ghidra + unsigned-k / lo-hi / reuse. Ghidra missed the s64 matrix. `unsigned k=0x800` gets ROM `>>12` order (rx first) but colors `k` in `r6` (div=10). permuter 950it on that seed, score 65 floor. seq_split still best. |

**Open angle:** get the `0x800` addend in `r0` (`mov r0, #0x800` after both sign extends) *and* keep it live until after `lsr` of `rx`, then dead so `adc r0, lr, #0` can reuse `r0`. A named `k` born early cannot sit in `r0` (sign `asr`s need `r0`). A literal dies too soon, so `adc` runs before `rx`.


andrewboudreau (commented):
# Review — PR #1447 "Bank near-miss for ov006 func_ov006_020dbe9c (div 15→7)" (lunavyqo)

**Verdict: merge.** The number is real. I reproduced it, and I reproduced every other
measurable number in the PR body, the attempt tree and the CLAIMS row. Two non-blocking
asks below.

Shape: 3 files, +7/−1, no `src/` file, no `NONMATCHING` banner removed, nothing enrolled.
I deliberately did **not** run `rombuild.py --no-rom`: there is no new/changed `src/` file,
no new `extern`, and nothing enters the link from this diff, so a 5-minute link buys nothing.

---

## Verification

Everything below is a number **I measured**, not one I read. Compiler: `mwccarm 2004/b56`
(`match.CANONICAL`), flags `M.DEFAULT_FLAGS`, via `match.compile_c` + `match.compare` on
scratch copies (never in `build/`).

### The headline claim — div 15 → 7: **REPRODUCED EXACTLY**

| record | stored | my word-diff (`match.compare`, reloc-wildcarded) | my `nearmiss_db.evaluate()` |
|---|---|---|---|
| `origin/main` (old) | 15 | **15** | 16 |
| `pr1447` (new tip) | 7 | **7** | **7** |

Both size-exact at 0xe0 = 224 bytes on both sides. So the banked "15 → 7" is true under the
metric the DB has been using for these rows.

One metric footnote, **pre-existing on main, not this PR's doing**: main's stored `15` is a
`match.compare` *word* count (the old `c_source`'s own comment says "fdiff 15/56"), while
`nearmiss_db.evaluate()` scores that same source at 16 (its difflib opcode metric). The two
metrics disagree by one on the old record. They **agree exactly at 7** on the new one, so the
new record is consistent under both and the improvement is real under either reading.

- `target_hex` on the PR branch is **byte-identical** to `origin/main`'s (not silently edited).
- It **equals the real ROM bytes** at `ov006:0x020dbe9c` for 224 bytes
  (`extracted/overlays/overlay_0006.bin`, base `0x020bfec0`), confirmed by direct slice.
- `size` 224 matches `symbols.txt` `size=0xe0`.

### Not pin-shopped

Swept all 25 installed compilers. Only four produce a size-exact 0xe0 function
(`1.2/base`, `1.2/sp2`, `1.2/sp2p3`, `2004/b56`); every other version emits 0xd4 or 0xd8.
On **all four**, old = 15 and new = 7. There is no version on which the old source wins.
The improvement holds at the canonical pin and is not a lucky-version artifact.

### Residual — exactly where the PR says it is

Full word compare of the new tip. **+0x00 through +0x7c is byte-identical** (the "expand
thru +0x7c" claim is literally true, all 32 words). The 7 diffs are:

```
        ROM                                 candidate
+0x80   lsr  r4, r6, #0xc                   adc  r0, lr, #0
+0x84   ldrb r6, [r3, #0x1be]               lsr  ip, ip, #0xc
+0x88   orr  r4, r4, r5, lsl #20            ldrb lr, [r3, #0x1be]
+0x8c   adc  r0, lr, #0                     orr  ip, ip, r0, lsl #20
+0x90   lsr  ip, ip, #0xc                   lsr  r4, r6, #0xc
+0x94   orr  ip, ip, r0, lsl #20            orr  r4, r4, r5, lsl #20
+0xbc   ldr  r0, [r0, r6, lsl #2]           ldr  r0, [r0, lr, lsl #2]
```

i.e. a pure block permutation — the candidate finishes the **sine** `>>12`/`orr` before the
**cosine** one, ROM does cosine first — plus `idx_l` colored `lr` instead of `r6`, which
cascades into the single `+0xbc` operand. That is precisely what the PR body and the
`610d9094…` attempt note describe. Stores (+0x9c…+0xb0), prologue and epilogue all match.

### Every claimed lever — independently re-measured, all confirmed

I rebuilt each lever against the shipped tip (not the old tip — the #1461 stale-lever trap):

| lever | measured |
|---|---|
| **the PR's lever**: revert split back to `((s64)v<<12)+0x800` | 7 → **15** (worth 8 words; it *is* the whole delta) |
| `#pragma opt_propagation off` removed | 7 → **21** (worth 14 words — genuinely load-bearing) |
| `idx_l` moved after `ry` | 7 → **11** (worth 4 words) |
| `split_only_cos` intermediate (claimed 11 in attempt `60d17b5d…`) | **11** ✔ |
| `unsigned k = 0x800` (claimed "ROM `>>12` order but k@r6, div=10") | **10** ✔, and the disasm confirms the order claim: `lsr r4` lands at +0x80 before `lsr ip` at +0x90, with `mov r6,#0x800` instead of `mov r0,#0x800` |

I also attacked the tip myself with the tree's own documented alternative
(`notes/mwccarm-codegen.md:184`, "`(int)(((long long)a*b + 0x800) >> 12)` is the Fix12
multiply") and four other spellings:

| my variant | div |
|---|---|
| `(int)(((s64)cosv * 0x1000 + 0x800) >> 12)`, cos first | 24 |
| same, `idx_l` read first | 24 |
| cos shift/add/`>>12` fully, then sin | 14 |
| `s64 k = 0x800` | 10 |
| `int k = 0x800` | 11 |
| `unsigned k` + `idx_l` first / after `ry` | 24 / 15 |

Nothing beats 7. **The odd-looking hand-split s64 expand is justified** — the documented
Fix12-multiply idiom is 17 words *worse* here. "seq_split still best" survives my attack.

### Bookkeeping

- **No duplicate key.** Raw rows 189 on both sides, `ledger.key_of` collision count 0.
  Exactly one record per `(ov006, 0x020dbe9c)`.
- **Nothing destroyed.** Main's old record carried **no `floor`** object, so no floor mark
  is dropped. Its lore prose *is* dropped — see "Should fix" #1. The evidence otherwise
  survives in `config/match_attempts.jsonl` under the parent chain.
- **All 5 attempt rows are schema-VALID** (`match_attempts.validate_attempt`). Parent chain
  is well-formed and closes: `84629ac6… → da62918531284769a701af70f7ec9dab`, which exists on
  main. `divergences`/`prevBestDivergences` are internally consistent (15→15→11→7→7→7). None
  of the five `attemptId`s collides with the ~34 pre-existing duplicate ids on main.
- **AI disclosure present** on every new row: `kind:"ai"`, `model:"grok-4.6"`,
  `reasoning:"xhigh"`, `harness:"grok-build"`, `author:"lunavyqo"`.
- **API claim lock genuinely released** — `claims_md.is_held("0x020dbe9c")` → `False`, both
  with the API and under `CLAIMS_NO_API=1`. (Unlike #1444/#1448/#1461's kept locks.)
- **Function identity**: `ov006:0x020dbe9c`, `size=0xe0`, sole caller
  `from:0x020de66c kind:arm_call` inside `_ZN11dScMgCoin_c6RenderEv` (0x020de63c, size 0x60).
  No word anywhere in any module equals `0x020dbe9c`, so it is **not** in a vtable — this is a
  private helper, not a virtual, and there is no vtable/typeinfo evidence to weigh against the
  name. The PR makes **no** class claim at all (it uses the raw `func_ov006_…` name
  throughout), so the #1461 wrong-class failure cannot occur here. For the record it is a
  `dScMgCoin_c::Render` helper: it builds a 2×2 rotation from `data_02082214[]` sin/cos and
  hands it to `func_ov004_020b023c`.
- **CI**: `PR validation` pass (2m55s), `submit` pass. `mergeable MERGEABLE / CLEAN`;
  `git merge-tree origin/main pr1447` is conflict-free against current main. The
  `mergeable UNKNOWN` GitHub reported was a stale computation.

### Not run, and why

- `rombuild.py --no-rom` — no `src/` file, no new externs, nothing enrolled (see above).
- `langmode_audit` / `check_references` / `eligible.py` — no source file in the diff.
- The intermediate `work/dbe9c/*.c` drafts named in `srcPath` are not committed (no `work/`
  in the tree), so those rows are ledger prose only; I verified their *numbers* by
  reconstructing the variants, which is the checkable part.

---

## Blocking

None.

---

## Should fix

**1. The record's lore comment was deleted, and it is provably free to keep.**
Main's `c_source` carried a five-line block recording the levers and the residual:

```
/* near-miss tip: pool base r4 + cos ldrsh r6 + sin ldrsh r4 match;
 * residual dual-s64 expand interleave ... fdiff 15/56 on mwccarm 2004/b56.
 * Levers: opt_propagation off (base ip->r4), named cosv (cos lr->r6),
 * idx_l between rx and ry. */
```

The new `c_source` has **no comment at all**. The DB record is the artifact the next
permuter run and the next model read as their seed, and it now ships four measured,
load-bearing levers (pragma = 14 words, split = 8, `idx_l` placement = 4, named `cosv`) with
nothing saying so. This is the same shape flagged on #1448 — lore living only in the CLAIMS
row where no future reader of the source sees it — except here a comment that *existed* was
removed, which is Pass A item 8.

I measured the fix: re-adding a lore block to the new tip leaves it **size-exact at div=7**.
Zero cost. Please carry it forward, ideally with the measured word costs, e.g.

```c
/* near-miss tip, div=7 (word diff) vs ROM on mwccarm 2004/b56, size-exact 0xe0.
 * Levers, each MEASURED -- do not tidy:
 *   #pragma opt_propagation off   14 words (removing it: 7 -> 21)
 *   split shift then += 0x800      8 words (combined form: 7 -> 15)
 *   idx_l read between rx and ry   4 words (moved after ry: 7 -> 11)
 *   named s16 cosv                 colors the cos ldrsh into r6
 * Residual: ROM does the cos >>12 block (+0x80..+0x88) before the sin one; we
 * emit sin first, and idx_l lands in lr not r6, cascading into
 * ldr r0,[r0,lr,lsl #2] at +0xbc. unsigned k=0x800 fixes the order but colors
 * k in r6 (div=10). The documented Fix12-multiply spelling
 * (int)(((s64)v*0x1000+0x800)>>12) is div=24 here -- the hand-split is correct. */
```

**2. "floor" is declared in prose but `mark-floor` was never run — say which you mean.**
Commit `2762bf9` is headlined *"(div=7 floor)"*, and both the PR body and the CLAIMS row say
"score 120 floor" / "score65 floor". But the banked record carries **no `floor` object**, and
I measured the consequence:

| | main | pr1447 |
|---|---|---|
| `export-close --max-div 8` seeds exported | **18** | **19** |
| DB `floor` count | 45 | 45 |

At div=15 the record was outside the close-seed window; at div=7 it is inside it, unfloored,
so it **newly re-enters automated permuter compute** — after this PR's own ledger records two
permuter runs on it (768 it @ score 120, 950 it @ score 65) that found nothing. That is the
#1445 pattern recurring.

Two honest readings, and only the author can pick:
- If "floor" means the DB sense (residual is compiler-internal, stop spending compute):
  run `python tools/nearmiss_db.py mark-floor --name func_ov006_020dbe9c --class ordering
  --evidence "..."` and the seed count stays 18.
- If "floor" only ever meant the *permuter score* bottoming out — which the PR's
  **"Open angle"** paragraph strongly implies, since it names a concrete next lever — then
  the re-arm is *intended* and nothing needs doing except dropping the word "floor" from the
  commit headline, where it reads as the DB sense.

I lean toward the second: this PR reads as live work, not a death certificate. Either way it
should not be ambiguous in the commit subject.

Note the class word matters here (`export-close --category` routes on it): the correct class
for this residual is **`ordering`** with a register-allocation cascade — see "Worth checking"
#1 for a tool caveat.

---

## Worth checking

**1. The category classifier mis-routes this residual — is that worth a tool issue?**
`categorize_misses.classify_entry` returns **"missing logic (ROM does more)"** for this
record. It is not: the function is size-exact and the diff is a literal block permutation of
identical instructions plus one register color. difflib splits the move into `insert`(3) /
`delete`(1) / `delete`(2) / `replace`(1), and the `sorted(cs) == sorted(ts)` "instruction
reorder" test never fires because the delete/insert land in separate opcodes. Consequence:
`export-close --category "instruction reorder,register allocation"` would **not** pick this
record up, while the plain `--max-div 8` path would. Pre-existing tool behaviour on main, not
introduced by this PR — but it is directly relevant to #2 above, and it means the effective
re-arm surface is narrower than the raw 18→19 suggests.

**2. Should the tip use named offsets from a `dScMgCoin_c` header rather than `+0x5000`?**
The tip spells `c + 0x5000`, `+0x1a8`, `+0x1ac`, `+0x1bd`, `+0x1be`, and
`c + 0x5100 + 0xb8`. For a near-miss DB draft raw offsets are entirely normal and I am **not**
asking for a change here. But when this eventually graduates to `src/`, the #1448 check
applies: grep `include/*_c.h` for whether an auto-generated `dScMgCoin_c` /
`dScMgMemory2_c.h` already names any of those fields before baking offset soup in. (`+0x5000`
on a scene object is a large embedded sub-object, and `+0x5100+0xb8` being written as two
addends is itself a recovered-structure hint.)

**3. Is the CLAIMS row deliberately not recording the improvement?**
Every other near-miss row of this author's in `CLAIMS.md` states the delta
("near-miss improved **319→228→209**", "**253→24**"). This row records only the *rematch*
("rematch from div=7 no beat"), so a reader of `CLAIMS.md` alone never learns that this PR
moved the tip 15→7. Consider "near-miss improved **15→11→7** (match.py words, size-exact
0xe0); rematch from div=7 no beat …".

---

## Mechanical

- **PR body overstates one line.** "Stores, prologue, and the call setup already match" —
  the stores and prologue do, but `ldr r0, [r0, lr, lsl #2]` at +0xbc *is* call-argument
  setup and is one of the 7 diffs. The same sentence names the `lr`/`r6` cascade, so this is
  imprecision rather than a false claim; the attempt-row note ("+ idx_l lr vs r6 cascade")
  gets it right.
- **Author column lost its `(Grok)` tag.** Every other `lunavyqo` row in `CLAIMS.md` reads
  `lunavyqo (Grok)`; this one is bare `lunavyqo`. Same drift as #1445. (The provenance itself
  is fine — `model:"grok-4.6"` is on all five attempt rows.)
- **Date skew.** CLAIMS row says `2026-08-13`; both commits are `2026-08-12`. Trivial.
- **Not findings — do not bill these.** (a) `addr` stored as an int rather than a hex string
  (`ledger.norm_addr` accepts both; 27 rows on main already do it). (b) The record's line
  moving from 49 to 33 in `db.jsonl` — `save_db` sorts by divergence, so a div=7 row sorting
  upward is expected. (c) `nearmiss/db.jsonl` and `CLAIMS.md` are `merge=union`; with four
  sibling near-miss PRs open, a replaced row can transiently duplicate after one of them
  merges, and the `update-chaos-data` workflow's `nearmiss_db.py dedupe` collapses it keeping
  the **lower** div (7). I confirmed `git merge-tree` against current main is conflict-free
  today. (d) `CLAIMS.md`'s `## Claims` table has no `|---|` separator row — pre-existing on
  main, not introduced here.

---

## Summary

This is the most honest bank in the sweep so far. Six independent numbers claimed across the
PR body, the commit messages, the CLAIMS row and five attempt rows (7, 15, 11, 10, plus the
"expand through +0x7c" and the "unsigned-k restores ROM order" structural claims) — **all six
reproduced exactly**, and the tip survived five alternative spellings I threw at it including
the tree's own documented Fix12 idiom. Merge it; the two asks are a free comment and one
ambiguous word in a commit subject.

lunavyqo:

PR #1448, opened 2026-08-12
Match func_ov006_020f5b98 byte-identical (mwccarm 2004/b56)

lunavyqo:
Byte-identical match of `func_ov006_020f5b98` (ov006, 0x020f5b98, size 0xa8) on mwccarm 2004/b56.

Walks 20 HUD records of stride 0x18 and calls `Hud_RenderSprite` when active. From the stored div=23 near-miss.

Load-bearing shape:
- `#pragma opt_propagation off`
- `do { ... i += 1; p += 0x18; } while (i < 20)`
- declare `i` first, `p` last
- inline `p + 0x51xx` field reads (no named `p + 0x5000` base)

`match.py --strict-relocs` MATCH; `linkcheck` VERIFIED (0 blind).

Attempt tree (`config/match_attempts.jsonl`, `functionId=ov006:0x020f5b98`):

| attemptId | parent | status | div | note |
|---|---|---|---|---|
| `01K5B98A…001` | null | no_progress | 23 | reconfirm included tip |
| `01K5B98A…002` | 001 | no_progress | 23 | 70+ variants + permuter, no beat |
| `01K5B98A…004` | 002 | near_miss | 14 | prop_off + do-while; i@r4; body schedule OK |
| `01K5B98A…005` | 004 | near_miss | 8 | p first / i last; i↔p swap only |
| `01K5B98A…003` | 002 | matched | 0 | i first / p last; final match |

Provenance: `kind=ai` model=`grok-4.6` reasoning=`xhigh` harness=`grok-build` author=`lunavyqo`.


andrewboudreau (commented):
# Review — PR #1448 "Match func_ov006_020f5b98 byte-identical (mwccarm 2004/b56)" (lunavyqo)

**Verdict: merge.** The match is real, the link is real, and every "lever" the author
claims in the `CLAIMS.md` row measured genuinely load-bearing. Nothing blocking.
Everything below is byte-free polish, all of it measured, none of it a reason to hold
the PR.

---

## Verification

Everything here I ran locally, on `pull/1448/head` (`7c8e04ef1`), against `origin/main`.

**Address/size derived from config, not from the PR description.**
`config/arm9/overlays/ov006/delinks.txt` adds `src/func_ov006_020f5b98.c: .text
start:0x020f5b98 end:0x020f5c40` → size `0xa8`. Cross-checked against
`config/arm9/overlays/ov006/symbols.txt:1007`:
`func_ov006_020f5b98 kind:function(arm,size=0xa8) addr:0x020f5b98`. Agrees.
No per-file override in `config/rombuild-versions.txt`, so the pin is the tree default.

**1. Byte gate at the canonical pin** — `tools/build_pin.py:verify`, verbatim:

```
pin: ('2004/b56', None)
verify(src, 'func_ov006_020f5b98', 0x020f5b98, 0xa8, 'ov006') -> (True, '2004/b56')
```

**2. Link gate** — `tools/linkcheck.py`, verbatim:

```json
{ "name": "func_ov006_020f5b98", "module": "ov006", "addr": "0x020f5b98",
  "verdict": "VERIFIED", "diffs": [], "blind": 0 }
```

`blind: 0` matters: it means every relocation resolved to a real address and the
*linked* bytes equal the ROM, so the relocation-wildcard hole (fakematch.md §3 /
Trap B) is closed here, not merely unexamined.

**3. Real `mwldarm` link** — `python tools/rombuild.py --no-rom`, verbatim tail:

```
[3/6] mwccarm: 10818 enrolled source file(s), -j20 (8 on an alternate toolchain version)
      10774 reused from cache, 44 compiled
source-built functions: 10,818  (1,945,940 / 2,211,124 code bytes, 88.01%)
  reproducing: 10,818
  mismatching: 0
ROM-build analysis: PASS
```

`rombuild.enrolled()` confirms `src/func_ov006_020f5b98.c` is in the enrolled set, so
this is a link that actually contains the new file — not a green tick that skipped it.

**4. Vacuous-match / banner checks** — all negative:

```
git diff origin/main...pr-1448 -- 'src/*' | grep '^-.*NONMATCHING'   -> no hits
grep -E '^asm void|dcd 0x|__asm' src/func_ov006_020f5b98.c            -> no hits
git diff origin/main...pr-1448 -- '*.c' '*.cpp' '*.h' | grep '^-\s*(//|/\*|\*)'  -> no hits
```

The body is 30 lines of ordinary C — one `do/while` over 20 elements — producing 42
instructions (0xa8). It is not a transcription. Only one `src/` file is touched, and it
is an addition; no existing matched file is edited or deleted.

**5. Phantom-extern sweep** — the file declares three externs, all resolve:

| extern | found |
|---|---|
| `Hud_RenderSprite` | `config/arm9/overlays/ov004/symbols.txt:59` (`size=0xe4 addr:0x020af68c`) |
| `data_ov006_0213d45c` | `config/arm9/overlays/ov006/symbols.txt:3839` (`kind:data(any)`) |
| `data_ov006_02142490` | `config/arm9/overlays/ov006/symbols.txt:4707` (`kind:bss`) |

The #1438 failure mode does not recur. The real link above independently confirms it.

**6. Ratchet** — `python tools/langmode_audit.py --check langmode-baseline.json` →
`langmode ratchet PASS`. `delaunder.find_sites(text, ("MASK","ROUNDTRIP","WIDEN"))` on
the new file returns **0 sites**. The #1444 launder regression does **not** recur: this
file contains no laundering idiom of any kind.

**7. CI** — `gh pr checks 1448`: `PR validation pass`, `ratchet pass`, `submit pass`.
`mergeable_state: clean`.

**Not checked:** whether `Hud_RenderSprite` genuinely takes five arguments and returns
`void` — the return type is invisible to the linker in either language and the result is
discarded here. The declaration used is the tree's majority spelling (16 of 26 existing
`extern` declarations of that symbol are exactly this one), so this PR is not the place
to settle it.

---

## Blocking

None.

---

## Should fix

**1. The four load-bearing levers are shipped with no comment in the file.** They are
recorded only in the `CLAIMS.md` row, which is not where a future reader of
`src/func_ov006_020f5b98.c` will look — and a later "tidying" pass will delete every one
of them. I measured each by editing and re-running `build_pin.verify` at 2004/b56:

| lever removed / rewritten | result |
|---|---|
| drop `#pragma opt_propagation off` | **COSTLY** — 27 words differ |
| `do/while` → `for (i = 0; i < 20; i++)` | **COSTLY** — size changes (999) |
| declaration order → `char *p` first | **COSTLY** — 14 words differ |
| `flag = 0; if (..) flag = 1;` → `flag = (.. != 2);` | **COSTLY** — 22 words differ |
| raw `a0 + 0x51xx` → recovered `struct HudCell *` | **COSTLY** — 36 words differ |
| raw offsets → pre-offset `p = a0 + 0x51a8` | **COSTLY** — 36 words differ |
| `data_ov006_0213d45c[kind][frame]` (2-D) | **COSTLY** — 25 words differ |

So: **every claim in the CLAIMS row is true**, and the "no named base" choice is not
laziness — `notes/mwccarm-codegen.md` documents exactly this under *"Big-offset
placement: let the compiler split, don't pre-offset"*, discovered in this same overlay
(`func_ov006_020e83bc`). The pragma + 0/1-selector pairing is documented too, at §6r
(*"`#pragma opt_propagation off` keeps a 0/1 ternary selector as a stack scalar"*) —
which is precisely the `flag` shape here. Per `fakematch.md` claim 3b, an uncommented
pragma is indistinguishable from a slot-machine pull; this one is a documented lever and
deserves to say so. Adding the comments cannot change codegen.

**2. Three byte-free type/naming recoveries, each measured FREE:**

| edit | result |
|---|---|
| `*(Fix12i *)(p + 0x51a8) >> 12` (both coordinate reads) | **FREE** |
| `extern void *data_ov006_02142490[23];` (size known from `src/__sinit_ov006_021314e4.c`) | **FREE** |
| `#include "dScMgMemory2_c.h"` + `((struct dScMgMemory2_c *)a0)->unk_53d4` | **FREE** |
| named offset `#define`s for the seven raw offsets | **FREE** |
| rename `a0`→`self`, `p`→`e`, `k`→`idx` | **FREE** |
| `i++` for `i += 1`; `0x14` for `20` | **FREE** |

The third one is worth calling out: `include/dScMgMemory2_c.h` already exists on main and
already names offset `0x53d4` as `unk_53d4`; the file spells it raw. That is the same
under-recovery pattern as #1444, in a much milder form — here it is one field of seven,
and the other six are genuinely blocked by the big-offset lever above.

I verified the whole package together (comments + `#define`s + `Fix12i` + `[23]`) still
returns `(True, '2004/b56')`, so this is one edit, not seven experiments.

---

## Worth checking

**Is this `dScMgMemory2_c::DrawCards`?** The only caller is
`src/func_ov006_020f73f4.c`, which main already annotates
`// recovered name: dScMgMemory2_c_Render`. The body loops 20 × 0x18 over an array based
at `+0x51a8`, and its siblings agree on the shape: `func_ov006_020f5c40.cpp` runs the
same `0x14` × `0x18` loop dispatching a per-element state through a pointer-to-member
table, and `func_ov006_020f5a64.cpp` compares two elements picked by `+0x53f0`/`+0x53f1`
and plays a "match"/"mismatch" sound. That reads as the memory/concentration card
minigame, 20 cards.

Supporting evidence for the field guesses, from the ROM: `data_ov006_0213d45c` dumps as

```
0000 0000 0000 0000 0000 | 0000 0001 0002 0004 0003 | 0000 0001 0002 0006 0005 | ...
```

— unmistakably rows of 5, so it is `u16 [N][5]` indexed `[cardKind][frame]`, and
`+0x51bd` (the second index) is the value `func_ov006_020f5de0` counts *down* to zero
before setting state 2. So `+0x51bd` is a flip-animation frame counter and
`data_ov006_02142490[k]` is one of 23 sprite handles filled in by
`src/__sinit_ov006_021314e4.c`.

Suggested (a comment, not a symbol rename — matching the caller's own convention, and
avoiding the langmode "renamed, never migrated" ratchet):

```c
// recovered name: dScMgMemory2_c_DrawCards
```

with field guesses `+0x51a8/+0x51ac` = `Fix12i x, y`, `+0x51b8` = card kind,
`+0x51ba` = live/visible, `+0x51bd` = flip frame, `+0x53d4` = scene state.

**Provenance:** clean. Every identifier in the file is either the tree's own
(`Hud_RenderSprite` already exists in `ov004/symbols.txt`, `Fix12i` in `include/types.h`,
`dScMgMemory2_c` in `include/dScMgMemory2_c.h`) or an address-derived `data_ov006_*`
name. Nothing here looks lifted from a vendor SDK header. `config/match_provenance.jsonl`
discloses `{"kind":"ai","model":"grok-4.6","reasoning":"xhigh","harness":"grok-build"}`.

---

## Mechanical

- **Bookkeeping is complete and schema-valid.**
  - `CLAIMS.md` row is `**done**`, not `**active**` — `claims_md.HELD_STATUS` only holds
    `active`/`partial`, so the file row will not block the scheduler.
  - `nearmiss/db.jsonl`: exactly one line removed (189 → 188), the right one; `grep -c
    020f5b98 nearmiss/db.jsonl` → 0. That row carried a `floor` marked
    `regperm+ordering` on 2026-08-06 with "100+ pure-C variants + permuter ~700it" — this
    PR breaks a recorded floor, which is worth saying out loud.
  - `config/match_attempts.jsonl`: all 1,580 rows pass `match_attempts.validate_attempt`,
    including the 12 for this function. `tools/test_match_attempts.py` 10/10 OK,
    `tools/test_match_provenance.py` 20/20 OK (2 skipped).
  - `match_provenance.lookup_provenance('ov006', 0x020f5b98)` resolves.
- **Residue (same as #1444, not this PR's fault):** the row ends "API clm_00006bfa9beb
  kept", and `python tools/claims_md.py` still lists `ov006 0x020f5b98-0x020f5c40` as one
  of 8 live **API locks** (it disappears under `CLAIMS_NO_API=1`, confirming it is the
  service side, not the file). Post-merge `is_held()` will keep treating this range as
  taken. Ask the author to release the API claim when it merges.
- Small inconsistency: the CLAIMS row says "from near-miss 23→14→0", but the attempt tree
  is `002 (div 23)` → branch A `003 matched (div 0)` and branch B `004 (14)` → `005 (8)`.
  The match came off the div-23 tip, not off the 14/8 branch. The commit title
  ("23→14→8") is the accurate one.
- No `.clang-format` in this tree, so no format gate applies. No register-named locals
  (`sp0`/`r4`/`var_r8`/`new_var`), no `& 0xFFFFFFFFu` no-ops, no dead locals, no magic
  `0x04......`/`0x05......` addresses, no `extern "C"`, no hand-spelled mangled name. The
  symbol is unmangled so plain `.c` (no `//cpp`) is correct.
- `delinks.txt` entry is in address order between `020f5a64` and `020f5c40` and uses the
  same two-line `complete` / `.text start:.. end:..` form as its neighbours; the real link
  confirms it.

lunavyqo:

PR #1450, opened 2026-08-12
Reconcile the Linux lane with cons, and make both platforms build again

tangosdev:
`port-linux-fwd-sweep` forked at `d723fc71d` and cons moved 119 commits past it. The Linux edits are all port-side and `_WIN32`-guarded so they survived, but the gap had to close before more work landed on a tree the port line no longer looks like.

## The four conflicts that were decisions

- **`func_ov085_0212e4a4` had two host copies**, one per lane. Kept cons's `Ov085_LakituBro_ClosestPlayer.cpp`, which covers that body plus three sibling LakituBro states, and dropped the Linux lane's single-body copy so only one definition links.
- **`_Z14ApproachLinearRsss`**: cons fixed this face to forward to `ApproachLinear` instead of `ApproachLinear2` (the sign turn-forever bug). That name *is* the Itanium mangling of its own callee, so taking the fix makes it self-recurse on GCC. It was only safe on Linux while it called the wrong function. Guarded `_WIN32`, the way this file already handles that class.
- **`heap_vtable` TRAP list**: took cons's shorter list. Slots 6/7/13/14/15 have real seated bodies now, so the Linux list would have re-trapped working slots.
- **`MovingMeshCollider` dtor**: cons seats a `__fastcall` shim at MSVC's folded slot 0, unusable from GCC's cdecl dispatch. Added a cdecl twin at the Itanium deleting slot, read off the ROM rather than assumed: `_ZTV18MovingMeshCollider` (`0x02099434`) relocates slot 0 to D1 `0x0203a470` and slot 1 to D0 `0x0203a444`.

## This lane had broken the Windows build

`861b25389` swapped `__declspec(align(8))` for `PORT_ALIGN(8)` in three HAL files. `PORT_ALIGN` only arrives via the CMake `-include hal/port_msvc_compat.h`, which sits inside the `if(NOT MSVC)` branch, so MSVC never saw it and the Windows port build has been broken since. The lane's "Windows byte-match INTACT" check was `git diff -- src/`, which never touches `port/`. The include is explicit in those three files now.

## Ten Windows-only features that had never met a Linux compiler

Startup-error dialog and file, ntr reservation diagnostics (`MAX_PATH`), the crash-classifier executability probe, level-boot QPC timing, `IsBadReadPtr` in two more files, `__int64`, `VK_F8`/`VK_F9` for the save-state binds, and a `RaycastGround` cast. Each is guarded or given a POSIX equivalent.

One is a deliberate capability gap worth calling out: **`port_addr_is_exec` answers 0 on Linux**, so no stack word is ever labelled a return address and Linux backtraces are unlabelled. Answering 1 would label data addresses as code, which is exactly what seeded the bogus `+0019be40` crash family. Misinformation reaching the classifier is worse than missing information. The real `/proc/self/maps` probe belongs with the rest of the deferred crash-forensics lane.

`linux_defsyms.txt` regenerated (155 to 161 flags) for the `alternatename` pragmas cons added. That staleness was the entire link failure.

## Verification

Each platform in its own build dir:

| | result |
|---|---|
| Windows | `ninja walk_window`, 4137 targets, exit 0 |
| Linux | `ninja -k 0 walk_window`, 4399 targets, links, exit 0 |
| Linux run | 300-frame headless castle grounds, `EXIT=0`, full actor census, player walks |

Merged `src/` is byte-identical to cons (`git diff 4b323dc1c -- src/` is empty), so the byte gate is unaffected and also proves nothing here. Nothing in CI builds `port/` on either platform, which is how the Windows breakage above went unnoticed.


andrewboudreau:
# Review — PR #1450 "Reconcile the Linux lane with cons, and make both platforms build again"

Author **tangosdev** (maintainer). 90 files, +4026/−87. GitHub: `CONFLICTING` / `DIRTY`.

**Verdict: merge with fixes, after resolving the conflict deliberately.** The engineering is
sound and the one ROM-evidence claim in the body verifies exactly. But the conflict is *not*
mechanical, and the naive resolution ("take HEAD") silently regresses save-state capture on
Windows in four of the six conflicted files.

---

## 0. What this PR actually is — read this before anything else

**The base branch is `port-mount-noseat-cluster`, not `main`.** This is an internal merge inside
the port lane. Two facts follow, and they reshape the whole review:

- **All 90 files are under `port/`. Zero `src/`, `include/`, `config/`, or root `tools/`.**
  Proven by tree hash, not by inspection:

  | ref | `src/` tree | `config/` tree |
  |---|---|---|
  | merge-base `4b323dc1c` | `5bef20f803…` | `797e0b8a38…` |
  | PR head `bf880bf9f` | `5bef20f803…` | `797e0b8a38…` |
  | base tip `4ed287364` | `5bef20f803…` | `797e0b8a38…` |

  Bit-identical across all three. The byte gate cannot move, and neither can the merge result's.
- `port/` is referenced by **no** `.github/workflows/*.yml` and **not** by `tools/rombuild.py`
  or `tools/eligible.py`. Nothing in CI builds this PR's output on either platform. (The body
  says so; confirmed.)

### File grouping (90 files)

| group | files | note |
|---|---|---|
| `port/hal/` | 52 | the bulk: self-collision guards, vtable/ABI fixes, section macros |
| `port/tools/` | 18 | 11 new `_lin_*.sh` WSL harness scripts + `gcc_aliases.py` (new, 353L) + `regen_linux_defsyms.sh` (new) + `gen_catalog_from_dump.py` (new) + generator edits to `hostgen/ovdata/romdata.py` |
| `port/unmatched/` | 12 | mostly one-line `#include <cstddef>` / guard additions; `Door_Behavior.cpp` +33 |
| `port/tests/` | 3 | `fault_probe.h` +146, `walk_window.cpp` +113 |
| `port/ntr/` | 2 | trivial |
| root of `port/` | 3 | `CMakeLists.txt` +183/−13, `linux_defsyms.txt` (new, 161L), `slice_gate18.txt` |

54 commits, by concern: **23** self-collision guards · **8** vtable/ABI · **8** misc ·
**5** generators/tooling · **4** build flags/CMake · **3** crash diagnostics ·
**1** WSL harness · **1** SDL2 seam · **1** the merge itself.

779 of the 4026 added lines are `#if/#else/#endif` preprocessor lines across 59 files.

---

## Verification — what I actually ran

Windows, primary worktree on `main`, scratch worktree at PR head (since removed).

- `git merge-tree` and a real trial `git merge` of `origin/port-mount-noseat-cluster` into
  PR head → **6 conflicted files, 10 hunks, 195 conflicted lines, 0 renames.**
- Tree-hash comparison of `src/` and `config/` across merge-base / PR head / base tip (above).
- Macro-coverage sweep over all of `port/`: every file using `PORT_ALIGN` / `PORT_GROUPED_DECL*` /
  `PORT_OVSEC` either includes `port_msvc_compat.h` (9 files) or self-defines the macro
  (the two generators). **0 uncovered.** The PR's central Windows fix is complete.
- `port/CMakeLists.txt:62` — `add_compile_options(-include .../hal/port_msvc_compat.h)` is
  indeed inside the `if(NOT MSVC)` block opened at line 27. The body's diagnosis is correct.
- Deleted-comment grep over the whole diff: **4 hits, all 4 blocks survive elsewhere in their
  own file** (moved past a new `#endif`). No lore lost. Clean.
- `pyflakes` on the five touched/added `port/tools/*.py`: 2 trivial hits (below).
- ROM-evidence claim in the body, checked against `config/arm9/symbols.txt`:
  `_ZTV18MovingMeshCollider` @ `0x02099434` (line 4346), `…D1Ev` @ `0x0203a470` (1468),
  `…D0Ev` @ `0x0203a444` (1467). **Exact.** Credit where due — that is the right way to
  state a vtable-slot claim.
- `/alternatename` pragma census: merge-base 1199, PR head 1199, **base tip 1220**.

### What I could NOT check

- **The Linux build and the 300-frame run.** No Linux/WSL toolchain reachable from this session.
  The body's `ninja -k 0 walk_window`, 4399 targets, `EXIT=0` claims are unverified here. Said
  plainly rather than guessed at. I reviewed the Linux side by reading.
- **The Windows `ninja walk_window` build.** Not attempted — the port build is 4137 targets and
  the lane is 582 commits behind main, so a partial failure would not be attributable to this PR.
- **I deliberately did not run `tools/rombuild.py --no-rom`.** It would build this lane's `src/`,
  which is 582 commits behind `main` — the numbers would describe the lane, not the PR, and would
  read as a huge regression against main's `106/106, 88.00%` for reasons having nothing to do
  with #1450. The tree-hash table above is *stronger* evidence than a build would be: the PR's
  diff outside `port/` is exactly zero bytes, so `eligible.py` enrolment, module fidelity and the
  match percentage are untouchable by construction. (The "unbuildable file silently drops out"
  hazard also cannot apply — no `src/` file changes.)

---

## Blocking

Nothing that makes the PR wrong. One thing that makes **merging it carelessly** wrong:

### B1. The conflict resolution must not drop `.dsstate`, and "take HEAD" does

**Diagnosis first: this is a genuine content conflict, not the phantom rename/rename pattern.**
`git merge-tree` reports zero renames. The branch forked at `4b323dc1c`, is **85 commits behind**
the base tip and **54 ahead**, and the base branch's 85 newer commits are themselves 100% `port/`.
Two live lanes edited the same HAL files. **A restack is the wrong remedy here** — it would replay
54 commits against a moved base instead of resolving 6 files once.

The substance: since `4b323dc1c`, cons introduced `port/hal/dsstate_seg.h` — one `.dsstate`
section family bracketed by `$aaa`/`$zzz` sentinels that `lk6_savestate.cpp` memcpy's wholesale.
Its header comment says this scheme exists *because* the previous hand-maintained 23-symbol list
missed ov002's `ANIM_PTRS` and that was the 0.2.0 restore crash.

The Linux lane, working from the pre-`dsstate` base, portable-ised the *old* `.{tag}$NNNN` scheme
into `PORT_OVSEC` / `PORT_ALIGN8` / `PORT_GROUPED_DECL_A`. That was correct against its base.
But it means **4 of the 6 conflicts are the same collision**, and in every one of them the `HEAD`
(Linux) side has no `.dsstate` at all:

| file | HEAD side | cons side | taking HEAD costs |
|---|---|---|---|
| `port/tools/ovdata.py` (4 hunks) | `PORT_OVSEC(".{tag}$NNNN", ".{tag}", …)`, `PORT_ALIGN8` | `pksect()` → `.dsstate$pkNNN_NNNN`, `#pragma data_seg(".dsstate$mmm")` | **every generated overlay global leaves the captured span** |
| `port/hal/actor_classes_star.cpp` | `.hvsstar$0000/0001` | `.dsstate$hvsstar0000/0001` | 32 bytes of `hvsstar` state uncaptured |
| `port/hal/fader_wipes.cpp` (hunk 2) | `extern "C" PORT_ALIGN(8) …` | `DSSTATE_BEGIN` + `__declspec(align(8))` | the `DSSTATE_BEGIN` marker |
| `port/hal/heap_vtable.cpp` | `#include <execinfo.h>` | `#include "dsstate_seg.h"` | the header itself |

None of these produce a compile or link error. The failure mode is a save-state restore that
silently keeps stale globals — i.e. exactly the bug `dsstate_seg.h` was written to kill.

**The resolution shape that keeps both intents** is to keep the Linux macro but feed it cons's
section names, e.g. in `ovdata.py`:

```python
f'PORT_OVSEC("{pksect(slot)}", ".dsstate", {rom_align(a)})'
```

with `PORT_OVSEC`'s existing `_MSC_VER` branch emitting `__pragma(section(fullsec,…)) +
__declspec(allocate(fullsec))` (→ MSVC gets `.dsstate$pkNNN`, capture preserved) and its GCC
branch emitting `__attribute__((section(basesec), aligned, used))`.

**Open design question for the GCC branch, please answer it explicitly rather than defer:**
`dsstate_seg.h`'s sentinel trick is `#pragma section`, which is MSVC-only, and `.dsstate` is not
a valid C identifier so GCC's automatic `__start_/__stop_` symbols do not apply. Does the Linux
lane have a save state at all yet? If not, say so in the resolution comment — "GCC side routes
to `.dsstate` for parity; the sentinels are MSVC-only until the Linux save state lands" is a
fine answer and stops the next reader re-deriving it.

**Post-resolution check that is one grep** (should be non-zero on both):
`grep -c 'dsstate' port/tools/ovdata.py port/hal/actor_classes_star.cpp port/hal/fader_wipes.cpp port/hal/heap_vtable.cpp`

### Exact unblock sequence

```
git fetch origin pull/1450/head:pr1450
git fetch origin port-mount-noseat-cluster
git checkout pr1450
git merge origin/port-mount-noseat-cluster        # 6 files conflict, 10 hunks
```
Resolve, per file:
- `port/tools/ovdata.py` (4 hunks) — **keep HEAD's `PORT_OVSEC`/`PORT_ALIGN8` macro bodies,
  swap the emitted section names to cons's `.dsstate$…`.** The hard one; the other three follow it.
- `port/hal/actor_classes_star.cpp` (1) — HEAD's `#include "port_msvc_compat.h"` +
  `PORT_GROUPED_DECL_A`, with cons's `.dsstate$hvsstar…` names.
- `port/hal/fader_wipes.cpp` (2) — **both sides, they are unrelated.** Hunk 1: HEAD adds an
  include, cons adds a 30-line design comment; keep both. Hunk 2: cons's `DSSTATE_BEGIN` +
  HEAD's `PORT_ALIGN(8)`.
- `port/hal/heap_vtable.cpp` (1) — both includes; cons's `dsstate_seg.h` unconditionally,
  HEAD's `<execinfo.h>` under `#else`.
- `port/hal/cxxname_bridge.cpp` (1) — HEAD's expanded `Model::LoadFile` body exists to dodge the
  extern-C self-recursion on Itanium; cons's one-line `Model::LoadFile()` call is the Windows
  path. `#ifdef _WIN32` / `#else`, the way this file already handles the class.
- `port/hal/reverse_bridges.cpp` (1) — cons's whole `ClsnResult`/`ClsnResultIntFace`
  `/alternatename` block under `#ifdef _WIN32`; see W1 below for the Linux half.
```
../tools/regen_linux_defsyms.sh          # MANDATORY, see S1 — do not skip
```
Do **not** rebase or restack. Expected outcome: one merge commit, 6 files touched by hand.

---

## Should fix

### S1. `linux_defsyms.txt` is already stale again, before the PR even lands

The body correctly identifies checked-in-generated-list staleness as "the entire link failure",
regenerates 155 → 161 flags, and ships `regen_linux_defsyms.sh`. But the input to that generator
is the `/alternatename` pragma set, and **cons has added 21 more since the merge base** (1199 →
1220). The regenerated 161-flag list was computed against 1199. It is stale on arrival.

Two asks:
1. **Re-run `regen_linux_defsyms.sh` as part of the conflict resolution** (in the sequence above),
   not as a follow-up. Otherwise the merge lands with the exact defect it set out to fix.
2. The list should not be checked in at all. It is *derivable* — `gcc_aliases.py` already takes
   `--altnames` from a `grep` of the pragmas. Deriving it at configure/link time, or adding a
   `--check` mode that fails the build when the pragma set has moved, removes a whole class of
   silent link failure. Every future cons commit that adds an `/alternatename` re-breaks Linux
   with no diagnostic until someone runs a link.

### S2. `--allow-multiple-definition` disables the diagnostic that finds this PR's own bug class

`port/CMakeLists.txt:58`. The comment is honest — "The HAL copy (linked first) is the one kept,
which is the intended one" — but that is an assertion about CMake source ordering that nothing
checks. Two consequences worth naming in the comment, or better, guarding:

- An unrelated `target_sources` reorder silently swaps a HAL host copy for a byte-locked ROM
  body, with no warning. MSVC at least emits `LNK4006` for the same situation;
  `--allow-multiple-definition` is completely silent.
- 23 of this PR's 54 commits exist to guard *accidental* symbol collisions (extern-C forwarders
  self-recursing under Itanium mangling). After this flag, the linker can no longer distinguish
  an intended duplicate from a new accidental one — the next self-collision will present as a
  runtime hang, not a link error.

Cheapest mitigation, and it composes with S1: `regen_linux_defsyms.sh` already runs `nm` over
every `walk_window` object. Have it also emit the *multiply-defined* set to a checked-in
expected list, so a new collision shows up as a diff instead of a stack overflow.

### S3. `port_msvc_compat.h`'s own header comment states the false premise that broke Windows

Lines 9–12: *"This header is FORCE-INCLUDED on the non-MSVC build only (CMake `-include`) … On
MSVC it is never included and the keywords keep their real meaning."* That is true of the
`__fastcall`/`__thiscall`/`__int64` shims (lines 22–41, inside `#ifndef _MSC_VER`) and **false**
of `PORT_ALIGN` and `PORT_GROUPED_DECL_A` (lines 43–75, outside it) — which are exactly the two
macros that broke the Windows build. Please split the comment: "the keyword shims are non-MSVC
only and arrive via `-include`; **`PORT_ALIGN`/`PORT_GROUPED_DECL*` are needed on both platforms
and every TU that uses them must include this header explicitly**." In this tree comments are
read as load-bearing lore, and this one currently invites the same mistake again.

### S4. Nothing builds `port/` in CI, which is how the Windows breakage lived

The body says this. It is the root cause of the whole PR, and it is the only finding here that
prevents recurrence. Out of scope to fix *in* #1450, but it should become an issue before this
lane grows further: even a configure-only `cmake -S port -B …` job on both platforms would have
caught the missing `-include` at line 62.

---

## Worth checking (questions, not assertions)

### W1. `reverse_bridges.cpp` — is the Linux `int` spelling an ODR problem now?

Cons changed the shadow to `struct ClsnResult { unsigned GetClsnID() const; }` so the MSVC
decoration names the newly-matched `src/_ZNK10ClsnResult9GetClsnIDEv.cpp` body (`…QBEIXZ` vs
`…QBEHXZ`), and added a `ClsnResultIntFace` + `/alternatename` for the `int`-returning callers.
On Itanium the return type is **not** in the mangling, so both spellings are the *same* symbol
`_ZNK10ClsnResult9GetClsnIDEv` — which is why the Linux lane guarded its shadow out entirely.

After the merge: on Linux the real body comes from the matched TU. Does any Linux caller still
declare it as `int`-returning? If so that is an ODR mismatch (`int` vs `u32`), harmless in
practice on i386 but worth a one-line comment saying it is deliberate — and `ClsnResultIntFace`
plus its `#pragma comment(linker, …)` are dead weight on Linux and should sit inside the
`_WIN32` guard, not just be ignored.

### W2. `port_addr_is_exec` returns 0 on Linux — but the PR ships the probe it says is deferred

`port/tests/fault_probe.h:54-66` defers the POSIX answer ("parsing `/proc/self/maps` … belongs
with the rest of that work"). The reasoning for preferring 0 over 1 is right and well argued.
But **this same PR writes that exact probe** in `port/unmatched/Door_Behavior.cpp:165-188`.

Not a simple hoist, though, and this is the part worth thinking about: `port_door_fn_in_image`
computes the **union span** of all `r-xp` ranges (`lo = min start`, `hi = max end`) and falls
open when maps is unreadable. For its purpose (reject an obviously stomped pointer) that is
fine. As `port_addr_is_exec` it would be *wrong in precisely the way that seeded the
`+0019be40` family* — a data mapping sitting between two executable ones passes the union test.
So: is it worth a per-range membership version (parse once into a sorted array, binary search)
rather than leaving Linux backtraces unlabelled? That is maybe 25 lines and closes the gap
honestly instead of deferring it.

### W3. `port/hal/host_platform_linux.{h,cpp}` (498 lines) — is the SDL2 seam finished?

The first commit calls it "phase 1, unbuilt". It *is* wired in by the end
(`CMakeLists.txt:123`, behind an option, and 6 files include the header), so the "unbuilt"
label is stale by the tip. Just confirming it is intended to be live in this PR and not a
partially-landed seam.

---

## Mechanical

- `pyflakes port/tools/gcc_aliases.py`: `:40 'sys' imported but unused`,
  `:233 local variable 'nxt' is assigned to but never used`. New file, trivial.
- `port/tools/_lin_*.sh` — 11 new scripts with leading-underscore names, which in this tree
  usually reads as personal scratch. If they are the supported Linux workflow, they deserve
  real names and a line in whatever the port's README is; if they are scratch, they probably
  should not be in the commit. (Not blocking either way — just pick one.)
- No build artifacts, `.o`, `__pycache__`, or `build/` output in the diff. Clean.
- No hand-edited `config/**` generated files (none touched at all).
- `port/slice_gate18.txt` change is a comment-only expansion documenting the
  `func_ov085_0212e4a4` host-copy decision. Good — that is the right place for it.

---

## Should this be split? — yes, but not into 8 PRs

Pass C's question is "one change or several wearing one hat", and the honest answer is *several*.
But the merge commit is genuinely atomic (it resolves four cross-lane decisions and both
platforms have to move together), so the useful split is **two PRs, not eight**:

1. **The reconciliation** — the merge commit, the four documented decisions, the three
   `port_msvc_compat.h` includes that fix Windows, the `CMakeLists.txt` MSVC/`NOT MSVC` guards,
   and the regenerated `linux_defsyms.txt`. This is the PR the title describes and it is the
   part that needs careful review. Maybe 15 files.
2. **The Linux portability sweep** — the 23 self-collision-guard commits plus the 8 vtable/ABI
   commits. 59 files, 779 of the 4026 lines are `#if/#endif`, and 14 files change ≤3 code lines.
   Highly repetitive and reviewable in bulk *precisely because* it is repetitive.

Two things that are independently separable and would be better off on their own, since neither
is needed to make either platform build:

- The **WSL harness** (11 `_lin_*.sh`, 1 commit) — pure tooling, zero build impact.
- The **SDL2 platform seam** (`host_platform_linux.{h,cpp}`, 498 lines) — a new subsystem that
  arrived under a "make the build work" title.

Given this is an internal lane merge and not a contribution to `main`, I would not insist on the
split — the review cost is already paid. But if any of it gets replayed toward `main`, land the
SDL2 seam and the harness separately; a new 498-line subsystem should not ride in on a build fix.

andrewboudreau:
Closing. This is the tip of an unlanded Linux/PC-port lane (`port-linux-fwd-sweep` -> base `port-mount-noseat-cluster` -> ...), `CONFLICTING` against a `main` that has moved 2,842 files since the branch point. It is not rebasable in any useful sense and the port build is ungated in CI, so nothing here has a byte gate behind it.

**Harvest list — 74 files at the branch tip (`bf880bf9f7473c15a4080affd3ae0f2d1ff5e361`) that this PR changes and that do not exist on `origin/main` @ `7869b817d`.** `port-linux-fwd-sweep` and `port-mount-noseat-cluster` are the only branches carrying them, and #1450 is the only open PR on the lane, so closing it leaves all 74 reachable by branch ref alone.

**23 introduced by this PR itself (absent from its own base):**

```
port/hal/host_platform_linux.cpp        port/hal/host_platform_linux.h
port/hal/pmf_devirt_shims.cpp           port/hal/port_msvc_compat.h
port/linux_defsyms.txt
port/tools/_lin_build.sh                port/tools/_lin_build_native.sh
port/tools/_lin_copy_assets.sh          port/tools/_lin_copy_extracted.sh
port/tools/_lin_gdb_crash.sh            port/tools/_lin_gdb_warp.sh
port/tools/_lin_probe_input.sh          port/tools/_lin_run_level.sh
port/tools/_lin_run_live.sh             port/tools/_lin_run_selftest.sh
port/tools/_lin_run_window.sh           port/tools/_lin_warp.sh
port/tools/gcc_aliases.py               port/tools/gen_catalog_from_dump.py
port/tools/regen_linux_defsyms.sh
port/unmatched/BabyPenguin_StateTick.cpp
port/unmatched/func_0204322c.cpp        port/unmatched/func_02043fdc.cpp
```

**51 more inherited from `port-mount-noseat-cluster` and modified here — that branch has no open PR either, so they are equally unlanded:**

```
port/hal/actor_class_faces.cpp           port/hal/actor_classes.cpp
port/hal/actor_classes_bbh.cpp           port/hal/actor_classes_bob_enemy.cpp
port/hal/actor_classes_bob_world.cpp     port/hal/actor_classes_bowserpuzzle.cpp
port/hal/actor_classes_ccm.cpp           port/hal/actor_classes_flame.cpp
port/hal/actor_classes_jrb.cpp           port/hal/actor_classes_koopa_chuckya.cpp
port/hal/actor_classes_l7.cpp            port/hal/actor_classes_ov002g200.cpp
port/hal/actor_classes_ov012.cpp         port/hal/actor_classes_ov064.cpp
port/hal/actor_classes_ov064_gate178.cpp port/hal/actor_classes_ov072.cpp
port/hal/actor_classes_ov081.cpp         port/hal/actor_classes_ov094.cpp
port/hal/actor_classes_star.cpp          port/hal/actor_classes_wf_enemy.cpp
port/hal/actor_faces_bob.cpp             port/hal/actor_registry.cpp
port/hal/bob_enemy_bridges.cpp           port/hal/bob_enemy_header_faces.cpp
port/hal/bob_enemy_shadow_faces.cpp      port/hal/camera_bridges.cpp
port/hal/cxx_aliases.cpp                 port/hal/door_ring_faces.cpp
port/hal/fader_wipes.cpp                 port/hal/level_boot.cpp
port/hal/level_change.cpp                port/hal/method_faces.cpp
port/hal/mmc_vtable.cpp                  port/hal/oam_lists.cpp
port/hal/particle_bridges.cpp            port/hal/player_bridges.cpp
port/hal/reverse_bridges.cpp             port/hal/sub_actors.cpp
port/hal/sub_screen.cpp                  port/slice_gate18.txt
port/tests/walk_window.cpp
port/unmatched/ActorDerived_Spawn.cpp    port/unmatched/BabyPenguin_StateEnter.cpp
port/unmatched/Door_Behavior.cpp         port/unmatched/Door_Render.cpp
port/unmatched/MrBlizzard_Behavior.cpp   port/unmatched/MrBlizzard_StateDispatch.cpp
port/unmatched/Player_InitResources.cpp  port/unmatched/RabbitKey_StateSeat.cpp
port/unmatched/StarMarker_D1.cpp         port/unmatched/func_0204322c.cpp
port/unmatched/func_02043fdc.cpp
```

The largest single artifacts, by line count at the tip: `port/tests/walk_window.cpp` (5,267), `port/hal/level_boot.cpp` (2,920), `port/hal/actor_classes.cpp` (2,344), `port/hal/cxx_aliases.cpp` (1,579), `port/hal/actor_classes_bob_world.cpp` (1,471), `port/hal/level_change.cpp` (1,353).

**Do not delete `port-linux-fwd-sweep` or `port-mount-noseat-cluster`.** They are the only copy.

**Follow-up should not rebase this.** Re-cut against current `main` in atomic slices: (1) `port/hal/host_platform_linux.{cpp,h}` + `port/tools/_lin_*.sh` + `regen_linux_defsyms.sh` + `port/linux_defsyms.txt` as one host-platform slice, since the defsyms file is generated by the script and the two must land together; (2) `port/hal/port_msvc_compat.h` alone; (3) `port/tools/gcc_aliases.py` / `gen_catalog_from_dump.py` / `ovdata.py` as a tooling slice; (4) each `port/hal/actor_classes_*.cpp` group behind its own `port/slice_gate*.txt`, checked with `port_refcheck.py` — which is blind to field names, so each slice needs the sub-object vptr re-stamp verified by hand. `port/unmatched/*.cpp` is behavioural-oracle material, not matched source; harvest it as reference, do not land it in `src/`.

PR #1451, opened 2026-08-12
Log no_progress floor reconfirm for ov006 func_ov006_020f3e68 (div=23)

lunavyqo:
Floor reconfirm for `func_ov006_020f3e68` @ `0x020f3e68` (ov006, size `0xa8`) on **mwccarm 2004/b56**.

`validate` is not expected to change: `src/` is untouched. The included near-miss still compiles size-exact with **match.py words=23**. Residual is welded:

- loop `i` in `sb` vs ROM `r4` (then table / `-1` / `self+0x5000` cascade)
- tip emits early `moveq`/`movne`/`str`; ROM delays `movne` until after `mla`/`ldrh`

This session (attempt tree in `config/match_attempts.jsonl`):

- `c96975d3a03e49c180a1798776ad3b66` — reconfirm tip at words=23 (parent `dbbf12df…` near-miss draft)
- `6042224db22447eeb26cf1adc9e61d12` — permuter ~350it score 1110→930 but match.py **25** (worse; `i=*(p+0x51b8)` abuse)
- `20c0c12816854783af7a0d7fd1385c63` — fresh retry: m2c override+do-while; sibling `020f3f10` C++ method; `020f5388` pad-struct; IV/schedule pragmas; goto loops. Probes: `i` lands in `r4` with only `data_ov006_0214236c` live; adding `data_ov006_0213d168` + the walking pointer locks `i` in `sb`.

Claims API `clm_06c43da80343` / `clm_0e9bc9f7bfbd` **released**. Tip remains in `nearmiss/db.jsonl` (div 24 / words 23). No `src/` change.


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1451 — ov006 `func_ov006_020f3e68` (div=23)

### The one finding that matters

**The #1445 floor defect, straight.** The PR title says "floor reconfirm", the CLAIMS row says
"retry no_progress FLOOR **words=23**", and the attempt note says "floor reconfirm match.py
words=23" — and the `nearmiss/db.jsonl` record for `ov006:0x020f3e68` has **no `floor` object**,
and this PR does not add one (the PR does not touch `nearmiss/db.jsonl` at all). The floor is
asserted in three places in prose and recorded nowhere a scheduler reads.

This is the **third** session on this function (main already carries three CLAIMS rows dated
08-06 / 08-06 / 08-07). Seed impact at `--max-div 8` is nil (the record is div=24), but the
entry is unfloored at *every* threshold ≥24, so `export-close`/`refine_wl` at a wider tier will
keep offering it as live compute. The fix is one command and the evidence text is already
written in the row:

```
python tools/nearmiss_db.py mark-floor --name func_ov006_020f3e68 --class <regalloc> \
  --evidence "i@sb vs ROM r4 (+ table/neg/fixed cascade) + early moveq/movne/str vs ROM delayed movne after mla/ldrh; 6k/6y/6ar/6aq/6c/6at, decl 24-perms, named hoists, array/struct views, pragmas, -opt noschedule, permuter ~350it"
```

### Verification

`match.compare` words = **23** exactly (the number CLAIMS and all three rows use), size-exact
`0xa8`. `evaluate()` = 26; the DB record stores 24 and the first row's `base.divergences:24`
correctly cites it. No internal contradiction — the row uses the words metric and says so.
3 rows schema-valid, parented onto main's `dbbf12dfb20c`. `target_hex` == ROM.
CLAIMS `**released**`, `is_held` False.

### Mechanical

- Row `6042224db224` has `srcPath:"/tmp/perm_ov006_020f3e68_batch3/output-930-1/source.c"` —
  an absolute machine-local path; the sibling rows use repo-relative `work/…`.

**Verdict: merge with one fix** — run `mark-floor`, or drop the word "FLOOR" from the row.

---

---

   recomputing all three.
   rows.

PR #1452, opened 2026-08-12
Log no_progress floor reconfirm for ov006 func_ov006_0211e72c (div=26)

lunavyqo:
Log the matching try for `func_ov006_0211e72c` (ov006 `0x0211e72c`, size `0xac`, mwccarm 2004/b56).

Not a byte match. Tip stays at **div=26**. No `src/` file.

What was tried:
- Near-miss draft with hoisted `0 / 1 / -1 / 0x1000` and scalar `OAM::Render` — size-exact, still 26 (table in `r4` not `fp`; `-1` in `sl` not `r4`).
- Ghidra-shaped rewrite (`data[e*7+d]`, no named table pointer): table lands in `fp` at the ROM pool-load slot, frame `0x1c` matches, **30 words**.
- 6at pointer-class `1` and `0x1000` put those values in `sb` / `r8`. Residual weld is `base` / `i` / `0` / `-1` coloring.
- C89 rank, 6y booster/scope, pragma pair, struct view, permuter ~900 it — no word-div beat of 26.

Ledger: `config/match_attempts.jsonl` (attempt `05d30e71fc614dd099be509a09fda4e1`). Claim kept active on `clm_ad8cbf0d146d`.


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1452 — ov006 `func_ov006_0211e72c` (div=26)

- Reproduces: `evaluate()` = **26** (= stored), `match.compare` words = 27, size-exact `0xac`.
  2 rows schema-valid, `target_hex` == ROM.
- **No floor recorded and none needed** — the CLAIMS row is honestly `**active**` and the body
  says "Claim kept active on `clm_ad8cbf0d146d`". `is_held` **True**, which is correct and
  intentional here. This is live work, not a floor claim.
- **The PR title is wrong**: "Log no_progress floor reconfirm" contradicts both its own body and
  its own CLAIMS row. Title-only defect; the ledger is right. A squash-merge would land the
  wrong summary on main (the #1459 stale-title failure mode, in a milder form).
- div=26 > 8, so no `export-close --max-div 8` impact either way.

**Verdict: merge, retitle.**

---

---

   recomputing all three.
   rows.

lunavyqo:

PR #1453, opened 2026-08-12
Bank near-miss for Player::St_SwingPlayer_Main (div 8→4)

lunavyqo:
Improve the stored near-miss for `_ZN6Player19St_SwingPlayer_MainEv` (ov002 `0x020d9fec`, size `0x3c4`) on mwccarm **2004/b56**.

**8 → 4** match.py words, size-exact. Lever: dropped 4th argument on `ApproachLinear2` via a function-pointer cast on the `step==0` path only. That reserves `r3` for the live signed-diff web so `mAngleYSpeed` loads into `ip` (ROM). Residual is the shift destination (`ip` vs ROM `r3`) on four words.

Ledger + tip: `config/match_attempts.jsonl`, `nearmiss/db.jsonl`. No `src/` file (not a byte match).


andrewboudreau (commented):
# Review — PR #1453 "Bank near-miss for Player::St_SwingPlayer_Main (div 8→4)" (lunavyqo)

**Verdict: merge.** div=4 is a number I **reproduced**, not one I read — and so is the 8 it
replaces, the 4-word residual, the exact location of every residual word, and the lever's
worth. Bookkeeping is clean: no duplicate key, `target_hex` untouched and ROM-identical,
schema-valid attempt chain, AI model disclosed, API claim lock genuinely released.

One finding that matters and is **not** a blocker for a near-miss bank but must be recorded
in the record itself: the lever is a **4-argument call to a 3-argument function**. It is a
register-allocator hint that emits no instruction, not recovered source, and nothing
containing it can ever graduate to `src/`. Nothing in the PR says so.

Shape: 3 files, +5/−1. No `src/` file, no `NONMATCHING` banner removed, nothing enrolled.
I deliberately did **not** run `rombuild.py --no-rom`: no source file, no new extern, nothing
enters the link from this diff, so a 5-minute link buys exactly nothing.

---

## Verification

Compiler: `mwccarm 2004/b56` (`match.CANONICAL`), `M.DEFAULT_FLAGS`, via `match.compile_c` /
`match.extract_func` on scratch copies in `tempfile` (never in `build/`). Records pulled from
`git show origin/main:nearmiss/db.jsonl` vs `HEAD`.

### The headline claim — **REPRODUCED**

| record | stored `divergences` | my `match.compare` words | my `nearmiss_db.evaluate()` | size |
|---|---|---|---|---|
| `origin/main` | **11** | **8** | **11** | 0x3c4 exact |
| PR tip | **4** | **4** | **4** | 0x3c4 exact |

So the PR title's "8 → 4" is a `match.py` **word** claim and it is exactly right. The DB
*field* moves 11 → 4 because main's stored value was the `nearmiss_db.evaluate()` difflib
number — the same two-metric drift documented on #1447, here pre-existing on main, not
introduced. Both metrics agree at 4 on the new record, so the tip is consistent under either
reading and strictly better under both. See Mechanical #1.

- Byte-identical to `origin/main`'s (**not** silently edited).
- Equals the **real ROM bytes** at `ov002:0x020d9fec` for 964 bytes
  (`extracted/overlays/overlay_0002.bin`, base `0x020ad660`) — direct slice compare.
- `size` 964 = 0x3c4 matches `symbols.txt` `size=0x3c4`.
- Exactly **one** record per `(ov002, 0x020d9fec)`; 189 rows both sides, `ledger.key_of`
  collision count **0**.

### Not pin-shopped — swept all 25 installed compilers

Only four produce a size-exact 0x3c4 function at all (`1.2/base`, `1.2/sp2`, `1.2/sp2p3`,
`2004/b56`); every 2.0/* emits 0x39c and every dsi/* 0x364–0x368. On **all four**,
old = 8 and new = 4. There is no version on which the old source wins. This is the same
shape as #1447 and it settles the question.

### Residual — exactly where the CLAIMS row says

New tip, 4 differing words, **all four are the same web colored `ip` instead of ROM `r3`**:

```
        ROM                          candidate
+0x150  asrlt r3, r0, #0x10          asrlt ip, r0, #0x10
+0x158  asrge r3, r0, #0x10          asrge ip, r0, #0x10
+0x15c  cmp   r3, #0                 cmp   ip, #0
+0x174  add   r0, r0, r3             add   r0, r0, ip
```

That is literally "shift dest/cmp/add use ip (step) vs ROM r3" — two shift destinations, the
`cmp`, and the `+=` add. The row is precise. Mechanically: ROM coalesces `step` into `diff`'s
r3 (diff dead after `lsl r0,r3,#9` at +0x154); the candidate coalesces it into `av`'s ip
(av dead after `cmp ip,#0x800` at +0x148). Both are legal; mwcc just ranks the webs
differently.

For contrast, main's record diverged at **+0x128, 0x134, 0x138, 0x13c, 0x144, 0x148, 0x14c,
0x154** — the *first* web (`ldrsh r3` vs ROM `ldrsh ip` for `av`). The PR fixes that web
outright and leaves only the second. The improvement is structural, not a lucky shuffle.

### The lever — re-measured, and it is worth exactly what is claimed

| variant (mutating the **shipped** tip) | div |
|---|---|
| **PR tip** `((AL4)ApproachLinear2)(p, 0, 0x20, diff)` | **4** |
| revert to plain 3-arg call | **8** |
| 4th arg = `av` instead of `diff` | 12 |
| 4th arg = `step` | size 0x3cc |
| 4th arg = `0` | size 0x3c8 |
| `s32 step` → `int step` | 4 (free) |
| `s32 step` → `s16 step` | size 0x3c8 |
| decl `step` before `av` | 4 (free) |
| `*(s16*)x += step` → `*q = *q + step` (the `St_WallJump_Init` compound-assign idiom) | 4 (no change) |
| same, re-reading the address instead of a temp | 15 |

So the lever is real, load-bearing, and **specific**: it must be `diff` in the 4th slot.
I tried the tree's documented `x += k` vs `x = x + k` lore against the `+0x174 add`
— it is **neutral** here (still 4), and the naive respelling that re-reads the address costs
11 words. Worth recording that it was tried.

### Identity — `Player::St_SwingPlayer_Main` at ov002:0x020d9fec

The PR asserts nothing main does not already say: the name is verbatim
`config/arm9/overlays/ov002/symbols.txt:865`
(`_ZN6Player19St_SwingPlayer_MainEv kind:function(arm,size=0x3c4) addr:0x020d9fec`), unchanged
by this diff. Corroboration I gathered independently:

- **It is a `Player` state member function.** Its *sole* reference anywhere is
  `relocs.txt: from:0x0210a494 kind:load to:0x020d9fec`, a word inside the dense
  pointer-to-member pool at `0x0210a4xx` (strict `(0, ptr)` 8-byte pairs).
- **It is a Main, not an Init or Cleanup.** 0x3c4 bytes, a 4-way `switch` on a step counter at
  `c+0x6e3`, `Player::FinishedAnim` / `SetAnim` / `ChangeState`, `CylinderClsn::Clear`+`Update`
  every frame. `St_SwingPlayer_Cleanup` is 0x28 and `St_SwingPlayer_Init` is 0x6c.
- **Contiguity.** `St_SwingPlayer_Cleanup` (0x020d9fc4, size 0x28) ends *exactly* at
  0x020d9fec, and `St_SwingPlayer_Init` (0x020da3b0) begins *exactly* at 0x020d9fec + 0x3c4.
  The triple is contiguous in `.text`, and its other two members are already matched in
  `src/` — this is the last of the three.
- I **could not** confirm which state from the pointer pool, and I checked rather than
  assumed: only **2 of 24** complete `St_*` triples in ov002 group inside a 0x18 window, so
  that region is a flat PMF constant pool, **not** an array of `Player::State` structs (the
  real `State` object, `data_ov002_0211013c`, is **bss**). The two apparent groupings are
  coincidence. **No #1461-style wrong-name signal — there is nothing here to disagree with.**

### What the replacement destroys — the floor, and the evidence survives

Main's record carried a real `floor` object; the PR drops it:

```json
{"class": "register allocation (two-web rank tie)",
 "evidence": "2026-08-01 deep re-crack (grok-build): match-words=8 pure rank-tie av->ip vs r3
  at +0x128. Schedule 241/241. Tried: sibling St_GrabBowserTail if/else+r3 naming; 100+ pure-C
  ...; permuter 750it score40 floor; 1.2/sp2p3+2004/b56 identical. Residual IR-invariant:
  mwccarm always ranks memory-loaded av above computed signed-diff for r3.", "date": "2026-08-01"}
```

Dropping it is **documented-correct** on a strictly-better ingest, and here it is also
*substantively* correct: the floor's own conclusion ("IR-invariant … always ranks memory-loaded
av above computed signed-diff for r3") is precisely what this PR disproves, at exactly the
+0x128 word it names. The evidence is not lost — it survives near-verbatim on **main** in
`config/match_attempts.jsonl` (rows `298185fc…` "deep re-crack: sibling GrabBowserTail …
permuter 750it base-score=40 floor … IR-invariant under 1.2/sp2p3 and b56", `bc69717b…`, and
`906970e6…`). **Nothing is destroyed that isn't recoverable.**

Neither the old nor the new `c_source` has a lore comment, so unlike #1447 no comment prose was
deleted. (See Should fix #1 — one should be added anyway.)

### The floor re-arm — measured, and here it is intended

| | main | pr1453 |
|---|---|---|
| DB `floor` count | 45 | **44** |
| `export-close --max-div 8` seeds (unfloored, div ≤ 8) | 18 | **19** |
| records with div ≤ 8 including floored | 33 | 34 |

So the record does newly enter automated compute. Unlike #1445 this reads as **intended live
work**, and the PR earns that reading: the attempt row is `status:"near_miss"` with
`improvedNearMiss:true` (not `no_progress`), and the CLAIMS row names an open trade-off
("One-web mutate-diff restores step=r3 but loses av=ip"). Re-arming a record whose floor you
just disproved is the correct outcome. The only residue is naming — see Mechanical #2.

### Bookkeeping

- **3 attempt rows, all schema-VALID** (`match_attempts.validate_attempt` → no errors).
- **Parent chain well-formed**: `9adbe20e…` (root, `base.kind:"near_miss_draft"`) →
  `d8420eb7…` → `b42c0659…`, each parent resolving. `divergences`/`prevBestDivergences`
  internally consistent (8→8→8→4). A `null` root is the tree norm for this function (main's
  three prior rows are all roots too). No `attemptId` collides with the ~34 pre-existing
  duplicate ids on main; no duplicates within the PR.
- **AI disclosure on every row**: `kind:"ai"`, `model:"grok-4.6"`, `reasoning:"xhigh"`,
  `harness:"grok-build"`, `author:"lunavyqo"`.
- **API claim lock genuinely released** — `claims_md.is_held("0x020d9fec")` → `False` both with
  the API and under `CLAIMS_NO_API=1`. Same clean behaviour as #1445/#1447; **not** the kept
  lock of #1444/#1448/#1461.
- **No `src/` draft** exists for this function, so the #1445 banner-lag issue cannot apply.
- **CI**: `PR validation` pass (8m15s), `submit` pass. `mergeable MERGEABLE / CLEAN`, and
  `git merge-tree --write-tree origin/main HEAD` is **conflict-free** against current main —
  the `mergeable UNKNOWN` GitHub reported was a stale computation, confirmed not assumed.

### Not run, and why

- `rombuild.py --no-rom` — no `src/` file, no new extern, nothing enrolled. A link buys nothing.
- `langmode_audit` / `check_references` / `eligible.py` / `linkcheck.py` — no source file in
  the diff and the DB `c_source` is not compiled by any gate.
- The `work/swing_tip.cpp` / `work/swing_div4.cpp` paths in `srcPath` are not committed (no
  `work/` in the tree), so those rows are ledger prose; I verified their *numbers* by
  reconstructing the variants, which is the checkable part.
- **One claimed number I could NOT reproduce** — see Worth checking #2.

---

## Blocking

None.

---

## Should fix

### 1. The lever is a 4-arg call to a 3-arg function, it emits **no instruction**, and the record does not say so

This is the finding that matters, and it is new for this author — every previous lever
(#1447's s64 split, #1448's pragma and decl order, #1444's volatile) was legal C that could
ship. This one cannot.

```c
typedef void (*AL4)(s16*, s16, s16, s16);
...
((AL4)_Z15ApproachLinear2Rsss)((s16*)(((int)c + 0x69c) & 0xFFFF...LL), 0, 0x20, diff);
```

`_Z15ApproachLinear2Rsss` demangles to `ApproachLinear2(short&, short, short)` — three
parameters — and it is **not a guess**: the function is matched and in-tree at
`src/_Z15ApproachLinear2Rsss.cpp`, `symbols.txt` `addr:0x0203aed8 size=0x5c`. And the ROM
proves it at the call site, which I disassembled:

```
+0x1a4  ldr r0, [pc, #0x204]
+0x1a8  mov r1, #0
+0x1ac  add r0, r4, r0
+0x1b0  mov r2, #0x20
+0x1b4  bl  0x203aed8            <- reloc; r0/r1/r2 only, NO r3 setup
```

The ROM passes three arguments. The candidate is **byte-identical here** — the 4th argument
costs zero instructions, because `diff` is already in r3. Comparing the old and new candidates
directly, they differ in **12 words, none of them at the call site**: the cast is a pure
register-allocator hint that pins `diff`'s web to r3 so `av` colors to `ip` like the ROM.

That is legitimate *information* — "the ROM has diff in r3 here" is a real discovery worth
banking — but the *mechanism* is a fiction, and calling through a wrong-arity function pointer
is UB. For a near-miss DB entry that is acceptable; the DB is a seed store, not source. The
problem is that this `c_source` is what the next permuter run and the next model read as their
starting point, and nothing in it warns that this line is a dead end. Please add a comment.
**I measured it: adding a lore block leaves the tip size-exact at div=4. Zero cost.** e.g.

```c
/* near-miss tip, div=4 (match.py words) vs ROM on mwccarm 2004/b56, size-exact 0x3c4.
 * Lever, MEASURED: the 4-arg fnptr cast on the step==0 ApproachLinear2 call is worth 4
 * words (plain 3-arg call: 4 -> 8). It emits NO instruction -- diff is already in r3 --
 * it only pins diff's web to r3 so av colors to ip like the ROM.
 * It is a CODEGEN HINT, NOT RECOVERED SOURCE: _Z15ApproachLinear2Rsss really takes 3 args
 * (src/_Z15ApproachLinear2Rsss.cpp, 0x0203aed8) and the ROM sets only r0/r1/r2 at
 * +0x1a4..+0x1b4. It must be replaced before this can graduate to src/.
 * Residual 4 words, all `step` colored ip vs ROM r3:
 *   +0x150 asrlt  +0x158 asrge  +0x15c cmp  +0x174 add
 * ROM coalesces step into diff's r3; we coalesce it into av's ip.
 * The 2026-08-01 floor ("IR-invariant two-web rank tie") is DISPROVED by this tip. */
```

Same shape as #1447/#1448 — load-bearing lore living only in the CLAIMS row, where no future
reader of the source sees it — except this time the lore is also a *warning*.

### 2. `Fix12` was silently respelled to `s32` in two extern declarations, and reverting is free

The diff changes

```c
-extern void _ZN6Player7SetAnimEji5Fix12IiEj(char* c, u32 anim, int a, Fix12 b, u32 d);
+extern void _ZN6Player7SetAnimEji5Fix12IiEj(char* c, u32 anim, int a, s32 b, u32 d);
-extern u32 _ZN8Particle...(u32 h, u32 id, Fix12 x, Fix12 y, Fix12 z, void* v, void* cb);
+extern u32 _ZN8Particle...(u32 h, u32 id, s32 x, s32 y, s32 z, void* v, void* cb);
```

The mangled names still spell `5Fix12IiE` = `Fix12<int>`, so the declarations now contradict
the symbols they declare, and `typedef s32 Fix12;` on line 7 is left dead. This is a type-lore
regression of the kind noted on #1438/#1444. **I measured it: restoring both `Fix12` spellings
is div=4, size-exact — free.** Please revert it (or, if the typedef really is meant to go,
delete line 7 too — that is also free at 4).

---

## Worth checking

### 1. Is there a spelling of the r3 pin that is *not* an ABI lie? — the graduation blocker

Not a request to change this PR; a question about what the next attempt should look for. The
banked fact is "ROM keeps `diff` in r3 across the `step==0` call". The real source must achieve
that some other way — `diff` genuinely live past the call, a differently-shaped `if/else`, or
`ApproachLinear2` being reached through something else. I did not find one (revert = 8), and
the author's own sweep row lists ~40 shapes that did not. But since the tip *cannot* graduate
while the cast is in it, it seems worth stating in the record as the explicit open angle — the
way #1447's PR body did. Would you add an "Open angle" line?

### 2. The one number I could not reproduce: "One-web mutate-diff restores step=r3 but loses av=ip (back to 8)"

Both CLAIMS and the attempt note make this claim. My two spellings of "reuse `diff` as `step`"
— `diff = (s16)(...)` and via an `s32 step` temp assigned back — both **size-regress to 0x3c8**
rather than landing at a size-exact 8, so I could not confirm it. That is almost certainly my
spelling being wrong rather than the claim being wrong (every other number in the PR
reproduced exactly), but it is the one assertion I am taking on trust. Could you paste the
exact mutate-diff variant, or drop the parenthetical? Distinguishing "not reproduced locally"
from "wrong" matters here and I cannot do it from the row alone.

### 3. Should the residual's floor class be recorded now, or is it deliberately open?

If a follow-up concludes div=4 is itself the floor, the right class is `register allocation`
(coalescing preference between two dead-at-the-same-point webs) — the same class word main's
dropped floor used, and still accurate for the surviving half. Note the #1447 caveat:
`categorize_misses.classify_entry` mis-routes shapes like this, so `export-close --category`
and the plain `--max-div` path will disagree about whether this record is armed.

---

## Mechanical

1. **The PR title's metric is not the DB field's metric.** Title/body say "8 → 4"; the
   `divergences` field goes 11 → 4. Both are true (8 is `match.compare` words, 11 is
   `nearmiss_db.evaluate()`), and the body does say "match.py words" — but a reviewer diffing
   the JSONL sees 11→4 and a reviewer reading the title sees 8→4. Consider "8→4 match.py words
   (DB field 11→4, evaluate metric)". Pre-existing tree drift, not this PR's doing.
2. **The branch name and first commit subject assert a floor the PR itself refutes.** Branch
   `bank/ov002-020d9fec-div8-floor`; commit 1 is *"Log no_progress floor reconfirm … (div=8)"*;
   commit 2 disproves it. That is honest history and I would not squash it away, but the CLAIMS
   row should probably note that the 2026-08-01 floor was **dropped because disproved**, since
   the row currently just states the new residual and a reader cannot tell a floor ever existed.
3. **"dropped 4th argument on `ApproachLinear2`"** (PR body, and "6ab dropped 4th arg" in the
   attempt note) reads as *removing* an argument; the change **adds** one. Please reword —
   this is the single most important sentence in the PR and it currently says the opposite of
   what the code does.
4. **Author column lost its `(Grok)` tag** — this row is bare `lunavyqo`. Same drift as
   #1445/#1447. (Softer than it sounds: 49 of ~393 rows on main are already bare. Provenance
   itself is fine, `model:"grok-4.6"` on all three attempt rows.)
5. **Date skew.** CLAIMS row says `2026-08-13`; both commits are `2026-08-12`. Trivial.
6. **Not findings — do not bill these.** (a) `addr` stored as an int (`ledger.norm_addr`
   accepts both; 27 rows on main do it). (b) `source` field changing from
   `"lunavyqo-batch5-swing"` to `"log_attempt"` — that is just which tool wrote the row.
   (c) `nearmiss/db.jsonl` and `CLAIMS.md` are `merge=union`, so a replaced row can transiently
   duplicate after a sibling near-miss PR merges; `update-chaos-data`'s `nearmiss_db.py dedupe`
   collapses it keeping the lower div (4). I confirmed `merge-tree` is clean today.
   (d) `CLAIMS.md`'s `## Claims` table has no `|---|` separator row — pre-existing on main.

---

## Summary

Every checkable number in this PR reproduced: 8, 4, the 0x3c4 size on both sides, the lever's
exact worth (4 → 8 on revert), all four residual words at the offsets and in the roles the
CLAIMS row names, `target_hex` identical to both main and the ROM, and 8→4 holding on all four
compilers that can build the function at all. The floor drop is not merely permitted here — the
floor said this residual was IR-invariant and the PR disproves it at the precise word it named.
Bookkeeping is the cleanest of the author's bank PRs so far: no duplicate key, valid chain,
disclosed model, claim lock released.

The one thing to fix before or after merge is a **free comment**: the lever is a 4-argument call
to a 3-argument function that emits no code, and the record — which is a seed for the next
automated run — must say that, or someone will spend a cycle trying to graduate it.

lunavyqo:

PR #1454, opened 2026-08-12
Log no_progress floor reconfirm for ov006 func_ov006_020ee994 (div=6)

lunavyqo:
Log the matching try for `func_ov006_020ee994` (ov006 `0x020ee994`, size `0x168`, mwccarm 2004/b56).

Not a byte match. Tip stays at **div=6** (fdiff align replace=6, size `0x170` vs ROM `0x168`). No `src/` file.

What was tried:
- Included near-miss tip (irreducible second loop entry): rematerializes seed / `#0xa` / last-ptr with `{r4,r5,lr}`, extra `cmp`/`blo` (+8B). Residual `mul r0` vs ROM `mul r2` + `ldr last` between mul and asr.
- Natural do-while / m2c rewrite: hoists those three into `r4/r7/r8`, size `0x164`.
- Common + obscure pragmas, O0 mix, launders, volatiles, Duff/switch, named temps, C++ `dScMgJump_c` members — no beat of the tip.
- Permuter ~465it on the do-while: low scores pulled `RandomIntInternal` out of the loop (semantically broken). Faithful perms stay hoisted.

Ledger: `config/match_attempts.jsonl` (attempts `36beb730…` / `06e49626…` / `37ffdee5…` / `4337326d…` / `16e4efbd…`). Floor mark `const-hoist` on the existing tip in `nearmiss/db.jsonl`. Claim released.


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1454 — ov006 `func_ov006_020ee994` (div=6) — **the exemplar of the five**

- **This is the only one of the five that actually records a floor.** It adds
  `floor:{class:"const-hoist", evidence:"…", date:"2026-08-13"}` to the DB record. DB floor
  count **45 → 46**; unfloored `export-close --max-div 8` seeds **18 → 17**. This is precisely
  the `mark-floor` step every prior review in this sweep found missing. Nothing else on the
  record changed — `c_source`, `divergences`, `target_hex` all untouched, so no lore drop.
- Numbers reproduce, and the metric is a **third** one. Stored 6 is the **fdiff `--align`
  replace count**, not `evaluate()` (19) and not `match.compare` words (999, size differs).
  `python tools/fdiff.py --align` returns
  `counts={'equal': 84, 'replace': 6, 'insert': 2, 'delete': 1}`, size `0x170` vs ROM `0x168` —
  verbatim the attempt-row note "fdiff align eq=84 rep=6 +2-1". The PR body discloses it up
  front. Honest; just be aware you cannot compare this 6 to 1452's 26 or 1451's 23.
- 5 attempt rows, all schema-valid, well-parented tree with four honest `div=999` dead ends.
- `target_hex` == ROM. CLAIMS row `**released**`, `is_held` False.
- Mechanical: row `16e4efbd…` has no `srcPath` while its four siblings do.

**Verdict: merge.** Hold this one up as the shape the others should copy.

---

---

   recomputing all three.
   rows.

PR #1457, opened 2026-08-13
Log no_progress floor reconfirm for ov009 func_ov009_0211145c (div=3)

lunavyqo:
Reverified `func_ov009_0211145c` (ov009, 0x0211145c, size 0x17c) from the stored near-miss tip under mwccarm **2004/b56**.

`tools/match.py` still reports **div=3**, size-exact. The whole function is byte-identical except the sound-call tail:

- ROM: `ldr r0,=0xff06a000; add r1,sl,#0x74; str r0,[sl,#0x168]; mov r0,#0x6a; bl func_0201267c`
- ours: `ldr r3,=0xff06a000; add r1,sl,#0x74; mov r0,#0x6a; str r3,[sl,#0x168]; bl`

Outgoing arg `0x6a` always precolors r0; the rematerializable pool const stays in leftover r3, so mov/str freely reorder (no anti-dep). Same residual after comma/equal-arm/6y/inlined store-then-return-0x6a/C++ helpers/pragma matrix/`-opt noschedule`/`-O2`/`-O3`. Sibling `func_ov009_021116ec` matches the same stores without the call (s colors r0).

Did not touch `src/` (existing NONMATCHING asm hatch is already on main; C still does not reproduce). Tip remains in `nearmiss/db.jsonl`. Attempt logged; claim released.


lunavyqo:
Opened against a stale leftover remote branch of the same name, not this session's commit. Replacing with a correct-head PR.

PR #1458, opened 2026-08-13
Log no_progress floor reconfirm for ov009 func_ov009_0211145c (div=3)

lunavyqo:
Reverified `func_ov009_0211145c` (ov009, 0x0211145c, size 0x17c) from the stored near-miss tip under mwccarm **2004/b56**.

`tools/match.py` still reports **div=3**, size-exact. The whole function is byte-identical except the sound-call tail:

- ROM: `ldr r0,=0xff06a000; add r1,sl,#0x74; str r0,[sl,#0x168]; mov r0,#0x6a; bl func_0201267c`
- ours: `ldr r3,=0xff06a000; add r1,sl,#0x74; mov r0,#0x6a; str r3,[sl,#0x168]; bl`

Outgoing arg `0x6a` always precolors r0; the rematerializable pool const stays in leftover r3, so mov/str freely reorder (no anti-dep). Same residual after comma/equal-arm/6y/inlined store-then-return-0x6a/C++ helpers/pragma matrix/`-opt noschedule`/`-O2`/`-O3`. Sibling `func_ov009_021116ec` matches the same stores without the call (s colors r0).

Did not touch `src/` (existing NONMATCHING asm hatch is already on main; C still does not reproduce). Tip remains in `nearmiss/db.jsonl`. Attempt logged; claim released.


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1458 — ov009 `func_ov009_0211145c` (div=3)

- Reproduces: `evaluate()` = `match.compare` words = **3**, size-exact `0x17c`.
  2 rows schema-valid, parented onto main's `0e6fd99a77d8`. `target_hex` == ROM.
- **The floor is already recorded on main** (`class:"ordering"`, dated 2026-07-30), so the entry
  is already excluded from `export-close`. Not the #1445 defect — nothing is re-armed and
  nothing is destroyed.
- **But this is the fifth reconfirm of the same div=3.** Main already carries **six** CLAIMS rows
  for `0x0211145c`: four prior `**released**` floor reconfirms (07-30 ×3, 07-31), a stale
  `**active**`, and a 07-13 `**done**`. Plus the DB floor since 07-30. The mechanism designed to
  stop this compute has been in place for two weeks and four sessions ran anyway.
- **The `**released**` in this row is ineffective.** `is_held(ov009:0x0211145c)` is **True** even
  under `CLAIMS_NO_API=1`, because main's stale 2026-07-30 `**active**` row (`CLAIMS.md:167`)
  still holds it. Pre-existing, not caused by this PR — but this PR is the natural place to flip
  that one word.

**Verdict: merge**, with the two asks above (they are about the workflow, not the diff).

---

---

   recomputing all three.
   rows.

PR #4, opened 2026-06-27
Match _ZN6Camera16CleanupResourcesEv

Moundistz:
Byte-exact match for `_ZN6Camera16CleanupResourcesEv` verified with mwccarm 1.2/sp2p3.

**Oracle verification (tools/match.py):**
```
TARGET _ZN6Camera16CleanupResourcesEv @ 0x0200d9d0 size 0x30  bytes: 00402de904d04de2480190e5000050e30200000a000050e30000000a76bc00eb0100a0e304d08de20040bde81eff2fe1
  1.2/sp2p3: MATCH

========================================
MATCHING VERSIONS: 1.2/sp2p3
```

- Module: arm9
- Address: 0x0200d9d0
- Size: 0x30

PR #3, opened 2026-06-27
Match _ZN6Camera11ChangeStateEPNS_5StateE

Moundistz:
Byte-exact match for `_ZN6Camera11ChangeStateEPNS_5StateE` verified with mwccarm 1.2/sp2p3.

**Oracle verification (tools/match.py):**
```
TARGET _ZN6Camera11ChangeStateEPNS_5StateE @ 0x0200cb70 size 0x70  bytes: 30402de904d04de20050a0e1540195e50140a0e1100010e204d08d120000a0133040bd181eff2f11381195e5010054e10800000a30009fe5000051e10200001aac7300eb0500a0e18eefffeb384185e50000a0e3a601c5e50500a0e1c4ffffeb04d08de23040bde81eff2fe1c8b00902
  1.2/sp2p3: MATCH

========================================
MATCHING VERSIONS: 1.2/sp2p3
```

- Module: arm9
- Address: 0x0200cb70
- Size: 0x70

PR #1459, opened 2026-08-13
Bank near-miss for Wiggler::Behavior (div 143→122)

lunavyqo:
Floor reconfirm of `_ZN7Wiggler8BehaviorEv` (ov034 @ 0x02112b5c, mwccarm 2004/b56).

The tip is still **div=122**, size-exact 0x6e0 / stack 0x74. Variants that did not beat the tip:

- `//cpp` + `Vector3` stack objects: +8 bytes
- flattening `Frame` into loose locals: size 0x684
- 702 pool/local respells: +4 bytes
- camera reorder: 126
- pointer decl-order grid: 124–126

Residual: first-loop setup hoist, 702 add-order, 6f8 regperm, Fix12 sincos interleave, camera r0/r3.

Tip remains in `nearmiss/db.jsonl`. No `src/` file (not a match).


andrewboudreau (commented):
# Review — PR #1459 "Bank near-miss for Wiggler::Behavior (div 143→141)" (lunavyqo)

**Verdict: merge — but retitle first.**

**The 122-vs-141 question resolves to (a), cleanly.** The branch has two commits from two work
sessions (`2d6619b1` 143→141, `a799b608` 141→122). The single replaced `nearmiss/db.jsonl`
record holds **122**, and I reproduced it. Row A's 122 is **real and backed**; row B's 143→141
is a **superseded intermediate** that was true of the first commit's record and is now history;
the **PR title and body are stale**, describing a state the branch does not ship. It is not (b)
— the DB record *was* updated to 122 — and it is not (c) — nothing here is aspirational.

Everything else is the cleanest bank PR of this author's five. I reproduced **every rung of the
ladder**: 143 → 141 → 137 → 122, size-exact 0x6e0 at every step, by peeling the named levers off
the shipped tip one at a time at the canonical pin. `target_hex` is untouched and ROM-identical,
identity is settled from the vtable, no floor was destroyed, and both API claim locks are
released.

Shape: 3 files, +8/−1. No `src/` file, no `NONMATCHING` banner removed, nothing enrolled,
`config/match_provenance.jsonl` untouched. I deliberately did **not** run
`tools/rombuild.py --no-rom`: nothing in this diff enters the link.

---

## Verification

Compiler `mwccarm 2004/b56` (`match.CANONICAL`), `M.DEFAULT_FLAGS`, via `match.compile_c` /
`match.extract_func` on scratch copies in `tempfile` — never in `build/`. Records pulled from
`git show origin/main:nearmiss/db.jsonl` vs `pr1459`.

### The headline numbers — REPRODUCED, and both metrics reported

| record | stored `divergences` | my `match.compare` words | my `nearmiss_db.evaluate()` | size |
|---|---|---|---|---|
| `origin/main` | **143** | **143** | 176 | 0x6e0 exact |
| PR tip | **122** | **122** | 119 | 0x6e0 exact |

**On this PR the stored field is the `match.compare` word count on *both* sides**, so the
143 → 122 delta is apples-to-apples and every number in the CLAIMS rows, the attempt notes and
the `c_source` banner is in that same metric. The metric drift documented on #1447/#1453 is
present in the tree but **not inside this PR** — the two metrics simply disagree by +33 / −3,
in opposite directions, and neither stored value is the `evaluate()` number.

One consequence worth knowing (Mechanical #5): `nearmiss_db.py ingest` stores `evaluate()`'s
number, so a re-ingest of this same `c_source` would rewrite the field to **119**, not 122.

### The whole ladder reproduces — I peeled the levers off the shipped tip

Not one number in this PR is taken on trust. Starting from the shipped `c_source` and reverting
the named levers in reverse order:

| variant | div | size |
|---|---|---|
| **PR tip as shipped** | **122** | 0x6e0 exact |
| − the `new_var` hoist of `*(u32*)(c+0x8e4)` | **137** | 0x6e0 exact |
| − also the `(c+i)+0x702` / `0x6f8` reshape (back to `static off702`) | **141** | 0x6e0 exact |
| − also the `r7,r6,r5,r4` decl order (back to `r4,r6,r5,r7`) | **143** | 0x6e0 exact |

That last line lands exactly on main's stored record. **Every claimed rung — 143, 141, 137,
122 — is confirmed, size-exact, at the canonical pin.** The `141` in the PR title is therefore
not wrong, it is simply two commits out of date.

Lever worth, measured against the **shipped** tip (the #1461 stale-lever check):

| mutation of the shipped tip | div |
|---|---|
| revert decl order only | 124 (lever still live, worth 2 — same 2 it was worth at 143) |
| revert `new_var` hoist only | 137 (worth **15**) |
| `new_var` hoisted *after* the cos→sin pair instead of between | **148** — worse than not hoisting |
| pure rename `r7/r6/r5/r4` → role names | 122 (free) |
| remove `#pragma opt_strength_reduction off` | size 0x6ec — **load-bearing** |
| remove `#pragma opt_common_subs off` | 122 — **inert** |
| drop all 5 `+ (zero & 0)` no-ops and their 3 `s32 zero = 0;` locals | 122 (free) |

The decl-order lever is **not** stale (contrast #1461): it is still worth its 2 words at the
tip. The `new_var` placement is exquisitely specific — one statement later and it is *worse
than not doing it at all* — which is a fact the record should carry (Should fix #3).

- Byte-identical to `origin/main`'s record (not silently edited).
- Equals the **real ROM** at `ov034:0x02112b5c`: found at file offset `0x19bc` in
  `extracted/overlays/overlay_0034.bin`, implying base `0x021111a0`, which is exactly the
  lowest symbol in `config/arm9/overlays/ov034/symbols.txt` (`_ZN7WigglerD1Ev`). **Unique
  occurrence** in the overlay.
- `size` 1760 = 0x6e0 = `symbols.txt` `size=0x6e0`.
- **Exactly one** record for `(ov034, 0x02112b5c)`; 189 rows both sides; `ledger.key_of`
  collisions **0** on both.

### Not pin-shopped — swept all 25 installed compilers

| | main's record | PR tip |
|---|---|---|
| `2004/b56` (canonical) | div=143 size-exact | **div=122 size-exact** |
| `1.2/base`, `1.2/sp2`, `1.2/sp2p3` | div=143 size-exact | size 0x6d4 (short) |
| every other 1.2 / 2.0 / dsi | size mismatch | size mismatch |

Confirmed at the canonical pin, which is the pin the tree builds with — so the number stands.
Worth stating plainly though: **the new tip is size-exact only at 2004/b56**, where main's was
size-exact on four versions. The cross-version corroboration that settled #1447 and #1453 is
not available for this record, so the claim rests on the canonical pin alone. That is the
correct pin and I am not treating it as a defect — but it is a change in the record's character
and the banner already says `2004/b56`, which is right.

### Identity — `Wiggler::Behavior` is CORRECT (no #1461 recurrence)

The PR asserts nothing main does not already say (`symbols.txt:33`
`_ZN7Wiggler8BehaviorEv kind:function(arm,size=0x6e0) addr:0x02112b5c`), but I settled it from
the ROM rather than from prose:

- `_ZTV7Wiggler` @ `0x021144bc`. **Slot 6 holds `0x02112b5c`.** Slots 16/17 are
  `_ZN7WigglerD1Ev` / `_ZN7WigglerD0Ev` — the destructor pair in its canonical place, which
  fixes the slot indexing.
- **Slot 6 is universally the `Behavior` slot**: I scanned every overlay for classes having
  both a `_ZTV<C>` and a `_ZN<C>8BehaviorEv`, and all 40+ (`Player`, `Coin`, `Bullet`,
  `InvisiblePole`, `daObjMarioCap_c`, `dScMgBase_c`, …) put `Behavior` at slot 6.
- Slot −1 typeinfo → `0x0211445c` = `_ZTI12daHanachan_c`, whose `_ZTS` string reads
  **`12daHanachan_c`**. Hanachan is Wiggler; the tree's English name and the ROM's RTTI name
  agree about the same object.
- Corroborating: factory `Wiggler_Spawn` @ `0x021136a4`, and six sibling `_ZN7Wiggler*` files
  already matched in `src/` (`InitResources`, `CleanupResources`, `Render`, `D0`, `D1`,
  `Wiggler_Spawn`). `Behavior` is the last big one in the class.

### Residual — verified cluster by cluster, and the c_source's numbers are one step stale

The 122 differing words fall in seven clusters. Mapping them to the categories the rows name:

| cluster | n | category named | verdict |
|---|---|---|---|
| `+0x198..+0x1bc` | 10 | "setup i/0x1000-before-ptrs (10)" | ✅ exact — `mov r8,#0`/`mov fp,#1`/`mov r0,#0x1000` vs the `add r6,sb,#0x160` pointer movs |
| `+0x214..+0x218` | 2 | "702 add-order (2)" | ✅ exact — `add r0,sb,r0; add r0,r0,r8` vs `add r1,sb,r8; add r0,r1,r0` |
| `+0x22c..+0x248` | 7 | "6f8 regperm (7)" | ✅ exact — pure r1/r2/r3/sl permutation around the `#0x600` / `#0xf8` split |
| `+0x2bc..+0x2d0` | 6 | "inc i++ before ptrs (6)" | ✅ exact — `add r8,r8,#1; cmp r8,#5` hoisted above the four pointer adds |
| `+0x354..+0x52c` | **87** | "sincos/part (101); idx (1)" | ❌ **stale — it is 87 now, not 102** |
| `+0x598..+0x5c0` | 10 | "cam (10)" | ✅ exact — `mov r3,#0`/`mov r0,#0x1f4000` vs `mov r0,#0`/`mov r3,#0x1f4000`, a pure r0↔r3 swap |

Five of the six numbers are exactly right. The sixth is the giveaway: the banner's list sums to
**10+2+7+6+101+1+10 = 137**, while its own header says 122 — see Should fix #2. The 15-word gap
is precisely the `new_var` lever's win, and it lands entirely inside the sincos/particle region,
which is internally consistent.

For contrast, main's 143 clusters are 10 / 15 / 6 / 101 / 1 / 10 = 143 — main's stored number is
also exactly its residual. Neither record is fudged.

The residual is overwhelmingly **pure register permutation** in the allocator's global basin
(ROM r3/r4/r1 where the candidate uses r0/r3), which matches the CLAIMS prose and the older
attempt notes about an "r6/r7 vs cyl r5/r4/r6 global allocator tradeoff".

### The floor — nothing destroyed, and the #1445 defect does NOT recur

- **Neither record carries a `floor` object.** DB floor count **45 → 45**.
- `export-close --max-div 8` seeds: **18 → 18** unfloored (33 → 33 including floored).
  Also unchanged at 16, 32 and 64. At `--max-div 128` the record newly enters (109 → 110), but
  nothing in the tree exports at 128; the documented permuter seed command is `--max-div 8`.
  **There is no scheduler impact.**
- Neither the new CLAIMS rows nor the new attempt rows declare a floor. Main's *older* CLAIMS
  row (2026-07-25) declares a prose floor ("global r6/r7↔cyl r5/r4/r6 tradeoff + static off702
  + sincos/cam regperm"); this PR partly disproves it (the static is gone, the decl-order
  basin moved) and correctly does not re-assert one.
- The attempt rows are `status:"near_miss"` with `improvedNearMiss:true` for every improving
  step, and `no_progress` only for the two genuine dead ends. Per the #1453 rule this is
  correct live work.

### Bookkeeping

- **5 new attempt rows, all schema-VALID** (`match_attempts.validate_attempt` → `None` on each).
  `tools/test_match_attempts.py` 10 passed, `tools/test_match_provenance.py` 18 passed.
- **Parent chain resolves and is well-formed** — it is a *tree*, correctly:
  `52e36e4b…` (pre-existing on main) → `60684dfb…` (no_progress 143) → `23cd8656…` (141) →
  branches to `189b1fd3…` (no_progress 141, the dead end) **and** `7c1a9e2b…` (137) →
  `04466cd2…` (122). Every `parentAttemptId` resolves; `divergences`/`prevBestDivergences`
  consistent at every edge; no `attemptId` duplicated.
- **AI disclosure on all five rows**: `kind:"ai"`, `model:"grok-4.6"`, `reasoning:"xhigh"`,
  `harness:"grok-build"`, `author:"lunavyqo"`.
- **Both API claim locks genuinely released.** `claims_md.held_targets()` hit the live API
  (reports `8 API locks`) and **none of the 8 ranges is in ov034**;
  `is_held(addr="0x02112b5c")` and `is_held(name="_ZN7Wiggler8BehaviorEv")` are **False both
  with the API and under `CLAIMS_NO_API=1`**. Same clean behaviour as #1445/#1447/#1453, not
  the kept-lock pattern of #1444/#1448.
- **No `src/` draft** exists for this function, so the #1445 banner-lag issue cannot apply.
- **CI**: `PR validation` pass (3m37s), `submit` pass. `mergeable MERGEABLE / CLEAN`, and
  `git merge-tree --write-tree origin/main pr1459` is **conflict-free** against today's main —
  confirmed, not assumed.

### Not run, and why

- `rombuild.py --no-rom` — per brief #11: no `src/` file, no new extern, nothing enrolled.
- `langmode_audit` / `check_references` / `eligible.py` / `linkcheck.py` — no source file in
  the diff; a DB `c_source` is compiled by no gate.
- The `srcPath` values (`scratch/w_tip.c`, `scratch/_ZN7Wiggler8BehaviorEv.c`) are not committed
  — same ledger-prose convention as #1453's `work/` paths. I verified the *numbers* they claim,
  which is the checkable part.
- **Nothing in this PR failed to reproduce.** Unlike #1453 there is no number I am taking on
  trust.

---

## Blocking

None.

---

## Should fix

### 1. The title and PR body describe a state the branch does not ship

Title: *"Bank near-miss for Wiggler::Behavior (div 143→141)"*. Body: *"Word diffs **143→141**"*,
plus a "Residual clusters" paragraph that is the **141** breakdown (`sincos/particle coloring
(101)`, `static off702 extra load + 0x6f8 addressing (13)`) and a "Lever:" paragraph naming only
the decl-order lever. The merged record is **122**, produced by two further levers.

This is a squash-merge title, so it becomes main's commit message. Please retitle to
**"div 143→122"** and rewrite the body around the shipped tip:

- Levers: decl order `r7,r6,r5,r4` + ROM pointer roles (2 words); `(c+i)+0x702` replacing the
  `static off702` plus the `s32 val`/`uval>>4` respelling (4 words); permuter hoist of
  `*(u32*)(c+0x8e4)` into a function-scope temp between the particle cos and sin lookups
  (15 words).
- Residual: 10 / 2 / 7 / 6 / 87 / 10 (see §Residual above).

Nothing is wrong with the *history* — two honest commits, and I would not squash the second
away. Only the summary is stale.

### 2. The `c_source` banner's residual list sums to 137 while its own header says 122

```
// NONMATCHING: size=0x6e0 stack=0x74 OK; div=122 word diffs (permuter new_var hoist) (2004/b56).
...
// Residual: setup i/0x1000-before-ptrs (10); 702 add-order (2); 6f8 regperm (7);
// inc i++ before ptrs (6); sincos/part (101); idx (1); cam (10).
```

10+2+7+6+101+1+10 = **137**. The list was carried over from the 137 tip and the last lever's
win was never subtracted. Measured: the sincos/particle region is now **87** words
(`+0x354..+0x388` 9, `+0x3a8..+0x3c4` 5, `+0x3e8..+0x52c` 73) and the separate `idx (1)` word
is gone — it was absorbed. So `sincos/part (101); idx (1)` should read `sincos/part (87)`, and
then the list sums to 122. Every other number in it is exactly right, which is why this one
stands out. Free to fix.

### 3. Deleted lore — the #1447 pattern recurs, and three of the four dropped items are still live

Main's `c_source` carried **two** stacked banner blocks; the PR replaces both with one. The
second block is dropped entirely:

```c
// NONMATCHING: size=0x6e0 (exact) stack=0x84 (want 0x74) ~224 non-bl diffs.
// Wall: (1) 0x10 spill before frame shifts sp offsets; (2) Fix12 sincos head schedule;
// (3) loop setup reg birth (fp=one, r8=i, r4=0x40c). Permuter 8365->7030.
```

Item (1) is genuinely obsolete (the stack is 0x74-exact now). But:

- **(2) "Fix12 sincos head schedule" is still the largest live cluster** — 87 of the 122 words.
- **(3) "loop setup reg birth (fp=one, r8=i, r4=0x40c)" is still the `+0x198` cluster** — the
  10 words the new banner calls "setup i/0x1000-before-ptrs".
- **"Permuter 8365->7030" survives nowhere in the tree.** I grepped `config/match_attempts.jsonl`
  on main: the only hit for `8365` is a coincidental substring of an `attemptId`
  (`9a52af383656`). `sincos head`, `reg birth` and `0x40c` return **zero** rows. Unlike #1453,
  this evidence is *not* preserved elsewhere — deleting it destroys it.

Also dropped from the first block: *"static s32 off702 frees fp for one=1 (frame spo=0xc)"*.
That one **does** survive, in main's attempt row `52e36e4b…`.

**I measured it: re-adding a carry-forward block leaves the tip at div=122, size-exact. Free.**
Suggested text:

```c
// Carried forward from the pre-0x74-frame era (still live, do not delete):
//   Wall (2) Fix12 sincos head schedule -- still the largest residual cluster
//     (+0x354..+0x52c, 87 words).
//   Wall (3) loop setup reg birth (fp=one, r8=i, r4=0x40c) -- still the +0x198 cluster (10).
//   Wall (1) 0x10 spill before frame shifted sp offsets -- SOLVED, stack is 0x74 exact.
//   static s32 off702 freed fp for one=1 (frame spo=0xc); dropped at 141->137 in favour of
//     DecIfAbove0_Byte((c+i)+0x702). Permuter score history 8365->7030.
// new_var placement is exact: hoisting it AFTER the cos->sin pair instead scores 148,
//   i.e. worse than not hoisting at all (137). Measured 2026-08-13 on 2004/b56.
```

### 4. Two CLAIMS rows for one function, one of them superseded inside the same PR

Multi-row is **convention, not a defect** — 56 addresses on `main` already carry more than one
row (max 6), and `.gitattributes` explicitly documents duplicate CLAIMS rows as harmless under
`merge=union`. So this is not a schema problem. It is a *readability* problem in a file whose
stated purpose is human coordination: a reader now sees two `**released**` rows for
`0x02112b5c` dated the same day, one claiming 141 with a residual breakdown that is no longer
true of anything in the tree, and one claiming 122. Row A happens to sort above row B, so
top-down reading gets the right answer by luck.

Either collapse them into one row (`143→141→137→122`, one claim id, one residual list), or
leave row B in place with `— superseded by the 122 row above` appended. Row B's claim id
`clm_ad6d0da6ac20` should stay recorded either way; both are released.

---

## Worth checking

### 1. `#pragma opt_common_subs off` is inert — is it doing anything at all?

Measured on **both** records:

| | `-opt_common_subs` | `-opt_strength_reduction` |
|---|---|---|
| main's 143 record | 143 (inert) | size 0x6e8 — load-bearing |
| PR's 122 tip | 122 (inert) | size 0x6ec — load-bearing |

So `opt_strength_reduction off` is real (and the `189b1fd3…` attempt note's *"SR-off
load-bearing"* is confirmed), while `opt_common_subs off` buys nothing at either end of the
ladder. It is **pre-existing on main**, so I am not billing this PR for introducing it — but the
PR is rewriting the banner anyway, and in a one-function-per-file tree a pragma is a claim about
the whole TU. Drop it, or add `// measured inert at 143 and 122, kept for X`?

### 2. Five `+ (zero & 0)` no-ops and three `s32 zero = 0;` locals are free to delete

Measured: removing all five sites *and* the three now-dead locals leaves div=122, size-exact.
This is Ghidra/permuter residue of the #1438 kind, pre-existing on main (5 sites there too).
Not a blocker for a DB seed — but the record cannot graduate to `src/` carrying it, and it is
free to shed now. Any reason to keep them?

### 3. One of the three named 141→137 sub-levers measures inert

CLAIMS row A lists the 141→137 win as *"(c+i)+0x702 + 6f8 split + s32>>4 asr"*. From the 137
state, respelling `*(u16*)((p2 + 0x600) + 0xf8)` back to `*(u16*)(p2 + 0x6f8)` leaves div at
**137** — the `0x600 + 0xf8` split is doing nothing *in that combination*. The 4 words come from
dropping the static and the `s32 val` / `uval >> 4` respelling. Is the split load-bearing in
some other combination (it was clearly load-bearing at 141, where "static off702 + 0x6f8" was a
13-word cluster), or can it be dropped from the lever list?

### 4. `new_var` and `char *r7/r6/r5/r4` — free to rename, but the names *are* the lever

Both are the classic fakematch tells (#1438: `new_var`; Pass A #11: register-named locals), and
I confirmed a pure rename to role names (`pBody`/`pAnim`/`pHgt`/`pApr`) is **free** — still 122,
size-exact. But in this record those names are load-bearing *documentation*: the whole lever is
"make these four pointers live in ROM's r7/r6/r5/r4", and the decl order is what achieves it.

For a DB seed I would not insist. If you do rename, the register mapping has to move into the
comment or the next reader loses the lever's meaning. Would you rather rename + comment, or keep
the names and add `// names encode the ROM register roles; the DECL ORDER is the lever (2 words)`?

### 5. Should the record note that the tip is now 2004/b56-only?

Main's `c_source` was size-exact on four compiler versions; this tip is size-exact only at the
canonical pin (the 1.2 trio now emits 0x6d4). Not a problem — 2004/b56 is the tree's pin, and
the banner correctly says so. But it is a fact a future re-cracker would want, and it is one
free line.

---

## Mechanical

1. **Non-ASCII `→` in `c_source` breaks `nearmiss_db.evaluate()` on a default Windows Python.**
   `nearmiss_db.py:81` does `pathlib.Path(tmp).write_text(src)` with no `encoding=`, so under
   cp1252 this record raises `UnicodeEncodeError: 'charmap' codec can't encode character
   '→'`. I had to run everything under `python -X utf8`. **Pre-existing** (5 of 189 records
   on main have non-ASCII `c_source`, including this one), but this PR takes this record from 4
   arrows to 8. Cheapest fix here: write `->`. Real fix: `write_text(src, encoding="utf-8")` in
   the tool — worth a separate one-liner PR.
2. **The improvement chain names a `138` that no attempt row records.** Banner reads
   `277→214→197→153→143→141→138→137→122`; the ledger goes 141 → 137 with no 138. Harmless, but
   either add the row or drop the step.
3. **`(Grok)` tag missing from the author column** on both new rows — same drift noted on
   #1445/#1447/#1453, and ~49 of ~393 rows on main are already bare. Provenance itself is fine:
   `model:"grok-4.6"` on all five attempt rows.
4. **`srcPath` points at uncommitted `scratch/` files** — same convention as #1453's `work/`.
   Not a finding, noted so the next reviewer does not chase it.
5. **A re-ingest would rewrite the field to 119.** `nearmiss_db.py ingest` stores
   `evaluate()`'s difflib count; both this record and main's store the `match.compare` word
   count. The DB genuinely mixes the two metrics across rows (confirmed on #1447 and #1453 as
   well). Tree-wide pre-existing drift, not this PR's doing — but it means `export-close
   --max-div N` is filtering on inconsistent units. Worth a tracked issue, not a PR comment.
6. **Not findings — do not bill these.** (a) `addr` stored as a hex string here (both accepted
   by `ledger.norm_addr`). (b) Two `no_progress` rows among the five — dead ends are supposed to
   be logged. (c) `nearmiss/db.jsonl` and `CLAIMS.md` are `merge=union`; a sibling bank PR
   merging first leaves a transient duplicate row, and `update-chaos-data`'s `nearmiss_db.py
   dedupe` collapses it keeping the **lowest** divergence — which is this PR's 122. Correct
   outcome by construction. (d) `CLAIMS.md`'s `## Claims` table still has no `|---|` separator
   row — pre-existing on main.

---

## The maintainer judgment call (brief item 10)

**Is banking incremental near-miss progress on a nowhere-near-matching function worth its own
PR?** Yes, on the merits, and no, in this shape.

On the merits: 0x6e0 is a 440-word function and 122 words differ, so 72% of it is already right
and the residual is now cleanly classified into six named clusters, five of them tiny. That is
not noise — it is a 15% reduction plus a *structural* result (the first loop body is
byte-identical now, and the two remaining big clusters are both pure allocator basin). The DB is
the seed store for the automated permuter, and a strictly-better seed with a reproducible lever
ladder is exactly what it is for. The alternative — sitting on it until the function matches —
loses the ladder if the author stops.

On the shape: this is the author's **fifth** open bank PR (#1445, #1447, #1453, #1459, #1460),
each 3 files, each +5/−1 to the same three union-merged ledgers, each needing a reviewer to
compile two `c_source` blobs to say anything at all. They cannot conflict with each other by
construction. **These should be one PR per batch, not one per function** — "Bank 5 near-miss
improvements (ov002/ov006/ov034/…)" with one CLAIMS row per function and one attempt subtree
each would cost the same review effort *once* instead of five times, and would remove the
temptation to let a title go stale across sessions, which is the one real defect here.

**Two same-function rows in an "Active claims" file**: acceptable as a mechanism (documented
harmless), wrong as a record. A coordination table should say what is true *now*; history
belongs in `match_attempts.jsonl`, which already carries all five steps properly parented. One
row, one claim id, the full ladder in the note. See Should fix #4.

---

## Summary

The number in the title is stale; the number in the database is real. `122` reproduced
size-exact at the canonical pin, and so did `137`, `141` and `143` — I rebuilt the entire ladder
by peeling the levers off the shipped source, and every rung landed on its claimed value.
`target_hex` is untouched and byte-identical to the ROM, the identity is confirmed from the
vtable (slot 6, the universal `Behavior` slot, in the vtable whose D1/D0 are `_ZN7WigglerD1Ev`/
`_ZN7WigglerD0Ev` and whose RTTI reads `daHanachan_c`), no floor existed or was destroyed, there
is no scheduler impact, the attempt tree is schema-valid and correctly parented, the model is
disclosed, and both claim locks are released.

Three free fixes before merge: **retitle to 143→122**, **correct the banner's residual list
(it sums to 137)**, and **carry forward the deleted `Wall (2)/(3)` lore, which describes 97 of
the 122 words that are still wrong and survives nowhere else in the tree**.

lunavyqo:

PR #1460, opened 2026-08-13
Bank near-miss for ov060 func_ov060_021140c0 (div 17→11)

lunavyqo:
Improve the stored near-miss for `func_ov060_021140c0` (ov060, 0x021140c0, size 0x1f4) on mwccarm 2004/b56.

**Not a byte match.** `src/` is unchanged. Tip in `nearmiss/db.jsonl`.

- Started from the included near-miss (DB 17 / match.py words 18).
- **17→11**: sibling-style `s32 rnd; u32 hi; u32 m = hi % 10u` fixes the r3/r5 modulo residual; `u16 dir[3]` restores ROM `ldrh` (shared `Vector3_16` is signed).
- Residual 11 words: `pos.x` in r2 vs ROM r3, `0xe8` in r3 vs ROM ip, `0xa000` hoisted to ip vs late `mov r3,#0xa000` + `&dir` in r2 (cascade). Extra register pressure can put `0xe8` in ip but then hoists `0xa000` to lr (div 19).
- ~30 C variants + permuter ~960 iterations, floor 460 (no word-div beat).
- Attempts logged; claim released (`clm_8b6d5e5d9fb6`).


andrewboudreau (commented):

### What I ran

  `CLAIMS_NO_API=1`.
  pairs**.

---

## PR 1460 — Bank near-miss ov060 `func_ov060_021140c0` (div 17→11)

### Verification

Both records compiled at **2004/b56**, size-exact `0x1f4` both sides:

| record | stored `divergences` | `evaluate()` | `match.compare` words | size |
|---|---|---|---|---|
| main | 17 | **17** | **18** | 0x1f4 exact |
| PR   | 11 | **10** | **11** | 0x1f4 exact |

Both metrics improve. The PR body's "Started from the included near-miss (DB 17 / match.py
words 18)" is exactly right. The **stored field switched metric**: main's 17 was the
`evaluate()` number, the PR's 11 is the `match.compare` word count (`evaluate()` says 10).
That is the same 1-off metric drift documented on #1447/#1453, not dishonesty — but see the
cross-cutting note below, `divergences` now carries three different meanings in this tree.

`target_hex` byte-identical to main's record **and** to the ROM at `ov060:0x021140c0`
(500 bytes). One record for the key. 4 new attempt rows, all `validate_attempt()` → OK,
chain parents onto main's `e6119d42cf5c`.

### Blocking

None.

### Should fix

- **The record ships four unexplained levers and zero lore.** The new `c_source` introduces
  the `s32 rnd / u32 hi / u32 m = hi % 10u` modulo split, `u16 dir[3]` in place of the shared
  `Vector3_16` (because the shared one is signed and emits `ldrsh`), and the `tbl` / `scale` /
  `grav` named hoists — all load-bearing per the PR body, none explained inside the record.
  Nothing was *dropped* (main's `c_source` had no comment either, and neither record carries a
  `floor`), so this is not the #1447/#1459 deleted-lore pattern — it is the same lore-lives-only-
  in-the-PR-body pattern. A comment in a DB `c_source` is free by construction. Ask for it.

### Worth checking

- One `& 0xFFFFFFFFFFFFFFFFLL` launder survives in the record. DB-only, so no langmode ratchet
  exposure today — but it will need measuring if this record ever graduates to `src/`.

### Not findings / credit

- **The floor question resolves in the PR's favour.** Main's record had no `floor`; the PR adds
  none. Commit 2's headline says "(div=11 floor)", but the CLAIMS row names an explicit open
  lever ("Open: pin 0x1000 in lr so tbl takes r8") and the winning row is
  `status:"near_miss" + improvedNearMiss:true`. By the #1453 grading rule that is *correct live
  work*. Measured: unfloored `export-close --max-div 8` seeds **18 → 18** (both 17 and 11 are
  >8); at `--max-div 16` the entry newly **enters**, 34 → 35. The PR arms the entry — the exact
  opposite of the #1445 defect.
- Real improvements, do not re-bill: moved off local `struct Vector3` / `Vector3_16` shadows to
  `#include "common.h"` + `decl_Actor.h` + `decl_Animation.h`; dropped two no-op
  `& 0xFFFFFFFF` launders; renamed `func_ov060_02115a30` → `Bowser_IsAnimAtLastFrame`, which is
  a **real symbol** (`config/arm9/overlays/ov060/symbols.txt:64`) already spelled that way by
  two sibling `src/` files. No phantom externs.
- CLAIMS row `**released**`; `is_held(ov060:0x021140c0)` **False** with and without the API.
- `addr` stored as int rather than hex string — fine, `ledger.norm_addr` accepts both.

**Verdict: merge** (one free should-fix).

---

---

   recomputing all three.
   rows.

lunavyqo:

PR #1461, opened 2026-08-13
Match 3 functions byte-identical (mwccarm 2004/b56)

lunavyqo:
Matched three functions byte-identical on mwccarm 2004/b56:

- `func_ov062_02117724`
- `func_ov075_021190a4`
- `func_ov075_0211867c` (0x0211867c, size 0x238) — linkcheck VERIFIED. Levers: `#pragma opt_strength_reduction off` + `opt_common_subs off`, goto-merged 0-select tail, `(base+(i<<2))[1]`.


andrewboudreau (commented):
# Review — PR #1461 "Match func_ov062_02117724 byte-identical (mwccarm 2004/b56)" (lunavyqo)

**Verdict: merge.** The match is real and fully verified at link level. Nothing blocking.
Everything below is "should fix" / "worth checking" — including one measured, *false*
compiler-lore claim in the `CLAIMS.md` row and PR body, and a class misattribution.

---

## Verification

Address/size derived from config, never from the PR text:

- `config/arm9/overlays/ov062/delinks.txt:166-168` — `src/func_ov062_02117724.c:` `.text start:0x02117724 end:0x02117994` (0x270)
- `config/arm9/overlays/ov062/symbols.txt:41` — `func_ov062_02117724 kind:function(arm,size=0x270) addr:0x02117724`
  (pre-existing; the PR does **not** touch `symbols.txt`, only adds the delink entry.)

All four gates run by me, on the PR branch, at the canonical pin:

| check | command | result (verbatim) |
|---|---|---|
| compile+bytes | `build_pin.verify('src/func_ov062_02117724.c','func_ov062_02117724',0x02117724,0x270,'ov062')` | `True 2004/b56` (pin resolved by `compiler_for` → `('2004/b56', None)`) |
| relocation destinations | `tools/linkcheck.py --name func_ov062_02117724 --addr 0x02117724 --size 0x270 --module ov062` | `"verdict": "VERIFIED", "diffs": [], "blind": 0` |
| real `mwldarm` link | `python tools/rombuild.py --no-rom` | `source-built functions: 10,818 (1,946,396 / 2,211,124 code bytes, 88.03%)` / `reproducing: 10,818` / `mismatching: 0` / `module fidelity: 106/106 exact, 100.000000% of compared bytes` / `ROM-build analysis: PASS` |
| ratchet | `python tools/langmode_audit.py --check langmode-baseline.json` | `langmode ratchet PASS` |

**`blind: 0`** — the reloc-wildcard hole is *closed*, not merely unexamined: every relocation
resolved and the linked bytes equal the ROM. The rombuild log shows `10817 reused from cache,
1 compiled`, and `rombuild.enrolled()` contains `src/func_ov062_02117724.c` (10,818 entries) —
so the green tick is not one that skipped the new file.

Fakematch sweep — all clean:

- No `NONMATCHING` banner removed anywhere: `git show 413318921 | grep '^-.*NONMATCHING'` → empty.
- Not a vacuous match: no `asm void`, no `dcd`, no `__asm`. Real C, 63 lines, 5 externs, one function.
- No register-named locals (`sp\d+`/`r\d`/`var_r\d`/`new_var`) — grep empty.
- No magic `0x04......`/`0x05......` hardware literals.
- **Phantom-extern sweep: all 5 externs resolve** in `config/arm9/symbols.txt` —
  `func_0201267c` (:377), `_ZN8Particle6System3NewEjj5Fix12IiES2_S2_PK11Vector3_16fPNS_8CallbackE`
  (:951), `_ZN8Particle19SetSelfDestructFlagEj` (:931), `_ZN8Particle6System12FromUniqueIDEj` (:952),
  `data_02082214` (:3173). The #1438 defect did not recur.
- Launder ratchet: `delaunder.find_sites(text, ("MASK","ROUNDTRIP","WIDEN"))` on the new file → **0 sites**.
  The #1444 blocker did not recur.
- Arity check on the mangled callee: `PK11Vector3_16f` is `const Vector3_16f*` (11-char type name),
  so `New` really is 7 parameters, and the 7-arg local declaration is right, not a truncation.

CI: `gh pr checks 1461` → `PR validation pass 9m29s`, `ratchet pass`, `submit pass`. `mergeStateStatus CLEAN`.

**What I could not check:** nothing material. I did not audit the two pre-existing `no_progress`
attempt rows from earlier sessions, only that they exist and validate.

---

## Blocking

None.

---

## Should fix

### 1. One of the three claimed levers is not load-bearing — measured

`CLAIMS.md`, the PR body and the commit message all name three levers. I measured each by
editing a scratch copy (same file stem, so `compiler_for` still resolves the pin) and re-running
`build_pin.verify`:

| lever | edit | result |
|---|---|---|
| `#pragma opt_propagation off` | removed | **COSTLY** — `999 word(s) differ` (i.e. the function *size* changes) |
| non-volatile `Vector3` + `*(volatile int *)&pos.y = ...` | plain `pos.y = pos.y + (k<<12)` | **COSTLY** — `999 word(s) differ` |
| same, vs the landed sibling's shape | `volatile Vector3 pos` (as `src/func_ov062_021175c0.c` does) | **COSTLY** — `7 word(s) differ` |
| **"first-arm load/sub interleave"** | both arms written plainly and symmetrically | **FREE — still `True 2004/b56`** |

So levers 1 and 2 are real and well-earned. Lever 3 is not: given the volatile-y-store, the
interleave carries nothing. I think this is a conflation — the interleave probably *was* the
div=5→0 delta relative to the **near-miss tip**, which used `volatile V3 pos` (I read the
retired `nearmiss/db.jsonl` line). It is not a lever in the *shipped* file. Please drop it from
the `CLAIMS.md` row and the PR body, or restate it as "was needed under the old volatile-struct
shape".

The whole `if/else` block can then be written as, byte-identically:

```c
        if (h <= a2) {
            int idx = (unsigned short)ang >> 4;
            pos.x = pos.x - k * data_02082214[idx * 2];
            pos.z = pos.z - k * data_02082214[idx * 2 + 1];
        } else {
            int idx = (unsigned short)ang >> 4;
            pos.x = k * data_02082214[idx * 2] + pos.x;
            pos.z = k * data_02082214[idx * 2 + 1] + pos.z;
        }
```

— 8 fewer lines, no `i1`/`i2`/`v`/`w`/`ev`/`ew` scaffolding, `idx` scoped where it is used.
I verified this exact variant end-to-end: `build_pin.verify` → `True 2004/b56`, and
`linkcheck` → `"verdict": "VERIFIED", "diffs": [], "blind": 0`.

(For completeness, these were **COSTLY** and should stay as written: hoisting `idx` above the
`if` (`999`), reusing `iscb` instead of recomputing `iscb2` (`999`), dropping the `(short)`
truncation on the `ps+0x50` fixup (`999`).)

### 2. The two load-bearing levers ship with no comment in the `.c`

Recurring ask, third PR running (#1444, #1448). The lore lives only in the `CLAIMS.md` row,
where no future reader of `src/func_ov062_02117724.c` will ever see it. Two lines, byte-free:

```c
/* 6r: opt_propagation off -- removing it changes the function's size. */
#pragma opt_propagation off
...
        /* The Vector3 must stay non-volatile and only this one store volatile:
           a wholly-volatile Vector3 (as src/func_ov062_021175c0.c uses) costs 7 words. */
        *(volatile int *)&pos.y = pos.y + (k << 12);
```

`opt_propagation off` is documented at `notes/mwccarm-codegen.md:933` (§6r) and used by 49 other
files, so citing the section is cheap and correct.

### 3. The function is a **Koopa** helper, not Chuckya

Title, body, commit message and the `CLAIMS.md` row all say "Chuckya particle spawn". The
evidence says otherwise, and this is the kind of label that later misleads TU-slicing work:

- Chuckya's block in `ov062/symbols.txt` runs `_ZN7ChuckyaD1Ev` 0x02115ee0 → `Chuckya_Spawn`
  0x02117478+0x54 = **ends 0x021174cc**. Koopa's begins there (`_ZN5KoopaD1Ev` 0x021174cc) and
  runs to `Koopa_Spawn` 0x0211970c. 0x02117724 is inside **Koopa**.
- All three callers in `ov062/relocs.txt` (0x02118748, 0x02118b78, 0x02118d84) are in the Koopa
  block, and the chain walks up to `_ZN5Koopa8BehaviorEv` (0x021192ac / 0x021192b8). Nothing in
  the Chuckya range references it.
- The offsets corroborate: `Koopa_Spawn`/`KoopaSmall_Spawn` both `new` 976 = 0x3d0 bytes, which is
  exactly `include/Koopa.h`'s `Koopa_size_must_be_0x3d0`, and `t + 0x358` lands on
  `mModelAnim` (0x300) `+ 0x58` = `ModelAnim::currFrame` (`include/ModelAnim.h:108`).

The `*(u16*)(t+0xc) == 0xcb` test is then almost certainly the big-vs-`KoopaSmall` variant check
(both spawners build `_ZTV5Koopa`), which is also what the `k = 0xf : 0xa` offset distance says.

### 4. Release the API claim after merge

`python tools/claims_md.py` still lists `ov062 0x02117724-0x02117994`; `CLAIMS_NO_API=1 python
tools/claims_md.py` does not. So `clm_732122bd2dde` is a **live API lock**, not a file row, and
`is_held()` will keep blocking the range post-merge. Same residue as #1444 and #1448 — the row
even says "API clm_... kept". One-line mechanical fix.

---

## Worth checking

- **Is `h` the integer animation frame?** `(unsigned int)(*(int*)(t+0x358) << 4) >> 16` is bit-for-bit
  `(unsigned short)(*(int*)(t+0x358) >> 12)` — and 0x358 is `mModelAnim.currFrame`, which is 20.12
  fixed point. Written as `>> 12` it reads as what it is (integer frame number) and matches the
  tree's fixed-point idiom. I measured the `>> 12` spelling: **FREE**, `True 2004/b56`, and
  `linkcheck VERIFIED / blind: 0` in combination with the §1 cleanup. Would you take it?
- **`unsigned short a4` is real — worth saying so.** Widening the 5th parameter to `unsigned int`
  costs `1 word`, so the asymmetric signature `(u32, u32, u32, u16)` is a genuine recovered type,
  not a leftover. That is a nice result and deserves a line in the PR body rather than looking
  like an oversight.
- **`include/Koopa.h` can gain two fields from this function.** It exists on main and already names
  the 0x110–0x3d0 span, but declares `u8 pad_398[0x4]` and `u8 pad_3cc[0x2]`. This function reads
  `*(u8*)(t+0x398) == 1` and writes 0/1 to `*(u8*)(t+0x3cd)` — so both are real `u8` fields, and
  0x3cd sits next to the already-named `unk_3ce` that the sibling `func_ov062_021175c0.c` uses as
  the same kind of once-per-window latch. Separate header PR, not this one.
- **Name the function?** Behavior is unambiguous: during either of two animation-frame windows,
  once per window (latch at 0x3cd), play SFX 0xe4 at `&mCamSpacePosX` (0x74) and spawn particle
  effect 0xf9 at the Koopa's position offset `k` units along `mAngleY + 0x4000`. Something like
  `Koopa_SpawnStepDust(Koopa *self, u32 startA, u32 endA, u32 startB, u16 endB)` follows the
  existing `Chuckya_ChangeState` / `KoopaSmall_Spawn` free-function convention in this overlay.
  A rename needs the `symbols.txt` + `delinks.txt` + `rombuild-versions.txt` pin moved together,
  so it is a follow-up, not this PR.
- **Return type of `New`.** This file declares it `int`; the tree's dominant spelling is `u32`
  (33 sites) or `void*` (8), and the immediate sibling `func_ov062_021175c0.c` uses `u32`.
  `uid` is only tested `== 0`, so it is byte-free either way — matching the sibling would just
  reduce the tree's 8-way divergence on one symbol. Pre-existing debt, not this PR's defect.
- No identifier in the file looks lifted from a vendor SDK header — the only names are ROM
  symbols (`func_0201267c`, the three `Particle::*` manglings, `data_02082214`) and locals.

---

## Mechanical

- Bookkeeping is **complete**, and better than #1438's per-function gaps:
  - `CLAIMS.md` row is `**done**` (row-logging is established convention here — not a finding).
  - `nearmiss/db.jsonl` entry for 0x02117724 retired (`-1` line); 0 remaining references.
  - `config/match_attempts.jsonl` +2 rows, chained `parentAttemptId` div=5 → `status:"matched"`
    div=0; `config/match_provenance.jsonl` +1 row disclosing
    `{"kind":"ai","model":"grok-4.6","reasoning":"xhigh","harness":"grok-build"}`.
  - All rows validate: `MA.validate_attempt` / `MP.validate_match_provenance` clean over the whole
    files (1,577 attempt rows, 0 bad); `pytest tools/test_match_attempts.py
    tools/test_match_provenance.py` → `28 passed, 2 skipped`.
  - **The "only 2 attempt rows" concern is unfounded**: main already carries 2 rows for
    `ov062:0x02117724` from the earlier near-miss sessions, so the branch has 4 and the trail is
    continuous. The row count differs from #1444/#1448 because the earlier legwork was logged in
    earlier PRs, not because it is missing.
- `delinks.txt` entry is a clean insertion in address order between 0x021175c0 and 0x02117994; the
  declared range is exactly the `symbols.txt` size. No `symbols.txt`/`relocs.txt` edits.
- No `src/` file deleted or rewritten (the #1438 "lands a .c that deletes a better .cpp" pattern
  did not recur). One credit where it is due: the file uses `common.h`'s real `Vector3`, where
  both the near-miss tip and the landed sibling `func_ov062_021175c0.c` declare a local shadow
  `typedef struct { int x, y, z; }`. That is a strict improvement on the shadow-declaration debt.
- Formatting matches the tree (4-space, K&R). The repo deliberately has no `.clang-format`, so no
  format gate applies.

lunavyqo:

PR #1473, opened 2026-08-13
Bank near-miss for ov071 func_ov071_021211e0 (div=7)

lunavyqo:
Banked a size-exact 0x314 tip for `func_ov071_021211e0` (mwccarm 2004/b56).

What improved: yoff-first first/third vectors, `Vector3_16` copy for the angle triple, materialized spawn bool, and y-then-x-then-z stores on the case-7 target.

Residual: 7-word r0/r1 swap on the case-7 player pointer (pointer/x vs y). Permuter ~1300 iterations stayed at score 50. Tip is in `nearmiss/db.jsonl`.

PR #1472, opened 2026-08-13
Match func_ov071_02120d30 byte-identical (mwccarm 2004/b56)

lunavyqo:
Match `func_ov071_02120d30` (ov071, 0x02120d30, size 0x3dc) byte-identical on mwccarm 2004/b56.

Mr. I state machine: ApproachLinear spin wrap, particle pair, four-way state switch. Final lever was the yadj/zcopy Vector3 build for Actor::Spawn so the y/z/x load and x/z/y store order match the ROM.

`match.py` MATCH; `linkcheck.py` VERIFIED.

PR #1471, opened 2026-08-13
Match _ZN9TowerStep8BehaviorEv byte-identical (mwccarm 2004/b56)

lunavyqo:
Match `TowerStep::Behavior` (`_ZN9TowerStep8BehaviorEv`, ov015 `0x021124d0`, size `0x1cc`) byte-identical on mwccarm **2004/b56**.

Levers:
- inverted clamp ternaries `(y<lim)?lim:y` / `(y>lim)?lim:y` for the `movge`/`movle` dest-in-r2 form
- named `flag = (y >= bound)` / `(y <= bound)` to materialize the `movge`/`movlt` pair

Replaces the NONMATCHING C stub. `match.py` MATCH + `prepush_linkcheck` VERIFIED.

The validate bot table is the review.

PR #1470, opened 2026-08-13
Match KnockDownPlank::InitResources byte-identical (mwccarm 2004/b56)

lunavyqo:
Match `_ZN14KnockDownPlank13InitResourcesEv` (ov015 `0x02111960`, size `0x208`) byte-identical on mwccarm 2004/b56.

Loads the plank model/KCL, flips yaw by `0x8000`, raycasts the two rotated offsets, and returns 0 when star 1 is missing in level 7 above the Y threshold.

Levers: `#pragma opt_propagation off`, u64-mask on the yaw increment, named `zero` for the Vector3 stores. `linkcheck` VERIFIED (0 blind).

PR #1474, opened 2026-08-13
Match N functions byte-identical (mwccarm 2004/b56)

lunavyqo:
Single squash of 20 open lunavyqo PRs so they are not sitting as a pile of tiny reviews.

**Matches (byte-identical, mwccarm 2004/b56):**
- `Flamethrower::Behavior`
- `TowerStep::Behavior` (`.c` → `.cpp`)
- `KnockDownPlank::InitResources` (`.c` → `.cpp`)
- `func_ov071_02120d30`
- `func_ov006_020f5b98`
- `func_ov005_020c16e4`
- `func_ov062_02117724`
- `func_ov075_0211867c`
- `func_ov075_021190a4`
- `func_ov080_02125460`
- `func_ov091_021339fc`
- `func_ov102_0214b53c` (replaces NONMATCHING `.cpp`)

**Near-miss / floor logs** (ledger + `nearmiss/db.jsonl` only, except the existing NONMATCHING `func_ov005_020c0378.cpp` tip update):
ov071 `021211e0`, ov005 `020c0378`, ov075 `02119dc4`/`0211afb0`, ov002 `020bfa74`/`020d3b9c`/`020d9fec`, ov060 `021140c0`, Wiggler::Behavior, ov006 `0211e72c`/`020dbe9c`, ov003 `020af8a0`.

Supersedes #1438 #1444 #1445 #1447 #1448 #1452 #1453 #1459 #1460 #1461 #1462 #1463 #1464 #1467 #1468 #1469 #1470 #1471 #1472 #1473.


andrewboudreau:
Reviewed the consolidation end to end. **The matching work is real — 12/12 verified.** Two things I'd like fixed before it lands, both small, plus one mechanical step to make it mergeable.

## Verification

Everything below was rebuilt in a worktree holding **the actual merge of this PR into current `main`**, not the PR in isolation — its base is two PRs stale (#1465, #1466), and I wanted the numbers to reflect what would really land.

All 12 match claims rebuilt at the canonical pin `mwccarm 2004/b56` via `build_pin.verify`, **with the reloc-destination check supplied** — not a byte-only compare:

| Function | Module | Size | Verdict |
|---|---|---|---|
| `_ZN12Flamethrower8BehaviorEv` | ov095 | 0x470 | **PASS** |
| `_ZN9TowerStep8BehaviorEv` | ov015 | 0x1cc | **PASS** |
| `_ZN14KnockDownPlank13InitResourcesEv` | ov015 | 0x208 | **PASS** |
| `func_ov071_02120d30` | ov071 | 0x3dc | **PASS** |
| `func_ov006_020f5b98` | ov006 | 0x0a8 | **PASS** |
| `func_ov005_020c16e4` | ov005 | 0x33c | **PASS** |
| `func_ov062_02117724` | ov062 | 0x270 | **PASS** |
| `func_ov075_0211867c` | ov075 | 0x238 | **PASS** |
| `func_ov075_021190a4` | ov075 | 0x0f8 | **PASS** |
| `func_ov080_02125460` | ov080 | 0x0fc | **PASS** |
| `func_ov091_021339fc` | ov091 | 0x270 | **PASS** |
| `func_ov102_0214b53c` | ov102 | 0x44c | **PASS** |

No file carries a per-file pin override and `config/rombuild-versions.txt` is untouched, so the default pin is the right one for all 12.

**Banner accounting.** 10 removed `NONMATCHING` lines, 1 added, all reconciled: 3 were the stacked duplicate banners in `func_ov005_020c0378.cpp` (still `NONMATCHING`, div 77→44, banner at byte 7 — inside the 200-byte window, so still correctly *not* a match claim). The other 7 are the files above that became matches. Every one rebuilt PASS.

**No transcriptions** — zero `asm void` / `dcd` hits, `asm_policy` exits 0. Worth stating explicitly for `func_ov091_021339fc`: it *was* the asm hatch in this diff, and it has been **replaced by real C that byte-matches**. `inline_asm` drops 14 → 13.

Other gates: `port_refcheck` **393 references, 0 stale**. `affected_src` names 5 consumers of `Flamethrower.h`; I rebuilt the 3 unchanged ones (`InitResources`, `D0`, `D1`) — all PASS, ripple nil. `eligible.classify` — 15/15 enrolled files eligible including rule 5. Object-symbol dump on the three `//cpp` files — no double-mangling. AI provenance is disclosed on all 12.

**Not run:** the full `eligible.py` sweep (and therefore `check_references.py` / `check_data_definitions.py`), and the full `rombuild.py` link. I substituted targeted `eligible.classify` runs covering rules 1–5 for every touched file, but the tree-wide bracket is unmeasured.

**Attribution is settled** — the 4 changed-credit entries are lunavyqo finishing drafts, which is what the finisher rule is for. I've applied the `attribution-override` label; no action needed from you.

## Needs doing before merge

**1. The branch needs `git merge origin/main`.** GitHub reports this conflicting, but only on `CLAIMS.md`, `config/match_attempts.jsonl` and `nearmiss/db.jsonl` — nothing in `src/`. Those three are `merge=union` in `.gitattributes`, and **that driver runs locally but not on github.com**, so a merge that is clean on your machine still shows dirty there. Merging `main` in locally resolves all three silently. (The duplicate ledger rows this produces are harmless — CI's `nearmiss_db dedupe` ranks `(div, -floored)`, so improved tips and floors both win.)

**2. `include/private/ov080_mat.h` ships a load-bearing claim that is false at the current pin.**

```c
/* The union wrapper on `param` is load-bearing for mwccarm 2004/b56
   TEXIMAGE_PARAM coloring (plain `u32 param` misses the r2/r3 rank). */
```

I tested this: replacing `union { u32 w; } param;` with a plain `u32 param;` (and `param.w` → `param`) still reproduces `func_ov080_02125460` byte-for-byte at 2004/b56 with reloc destinations checked. Control run with the union intact also matches, so the comparison is clean. It was most likely true at the retired `1.2/sp2p3` pin and didn't get re-tested at the flip. Either drop the union or correct the comment — this is going into a shared header, and a comment that lies costs more than no comment, because the convention here is to never rewrite a comment you didn't author.

**3. Two files are `//cpp` but stop halfway — and their classes already exist.** `_ZN9TowerStep8BehaviorEv.cpp` and `_ZN14KnockDownPlank13InitResourcesEv.cpp` compile as C++ but still spell `extern "C" int _ZN9TowerStep8BehaviorEv(char *self)` and poke `self + 0x37c`, `self + 0x390`, `self + 0x384`…

Meanwhile `include/TowerStep.h` and `include/KnockDownPlank.h` already model both classes on the real `Platform` chain, already declare `int CleanupResources(); int Render(); virtual ~TowerStep();`, and already **name every offset those two files hand-poke** — `unk_378 / unk_37c / unk_380 / unk_384 / unk_388 / unk_38c / unk_390 / unk_391 / unk_392` for TowerStep, `unk_378 … unk_397` for KnockDownPlank. The sibling `_ZN9TowerStep6RenderEv.cpp` is already `int TowerStep::Render()` against that header. And this PR does exactly the right thing one file over — it adds `int Behavior();` to `include/Flamethrower.h` and writes `int Flamethrower::Behavior()`.

Asking as a verify-cycle question rather than a patch: **does `int TowerStep::Behavior()` against `TowerStep.h` still match?** Two hazards to bracket — including `Platform.h` can pull in a synthesised constructor that mwcc inlines into holders, and the class carries a `virtual ~TowerStep()`; `build_pin.verify` on the file plus its `affected_src` closure settles both. A `TODO:` naming the class is an acceptable answer if it doesn't come free.

This is also the whole langmode story below, so fixing it clears that too.

## The langmode green is borrowed, not earned

`langmode_audit --check langmode-baseline.json` **passes**, and that is the file CI prefers, so the gate is green as configured. Commit 2 did its job and then some:

| metric | main | merged | delta |
|---|---|---|---|
| `launder_or_forced` | 282 | **281** | **−1** |
| `inline_asm` | 14 | **13** | **−1** |
| `nonmatching_drafts` | 9 | 7 | −2 |
| `mangled_c` | 550 | 548 | −2 |
| `cpp_still_handspelled` | 144 | **146** | **+2** |

But against the **authoritative chaos-data copy** the merge *fails*: `cpp_still_handspelled 144 → 146`, plus two `shadow_decls` +1s. `main` passes that same check, so the rise is this PR's. It's masked only because the stale root `langmode-baseline.json` (last written by #1359) is looser on every field — and that workflow's own inline warning says the file was meant to be deleted in a follow-up. Not your doing, and I'll clean it up separately.

The +2 is exactly item 3 — I diffed the `--list cpp-handspelled` name lists to confirm. Fixing those two files clears the rise and the chaos-data failure with it.

## Should fix

**4. Eight steering pragmas, none tagged.** `opt_strength_reduction off` ×2, `opt_propagation off` ×4, `opt_common_subs off`, `opt_loop_invariants off` across `_ZN12Flamethrower8BehaviorEv.cpp:6`, `_ZN14KnockDownPlank13InitResourcesEv.cpp:31`, `func_ov005_020c0378.cpp:33`, `func_ov006_020f5b98.c:2`, `func_ov062_02117724.c:9`, `func_ov075_0211867c.c:23-24`, `func_ov080_02125460.c:4`. One line each naming the documented lever (`// 6e: EBB-local CSE`, per `notes/mwccarm-codegen.md` §6e–6f). An uncommented pragma is indistinguishable from a slot-machine pull, and mwccarm silently accepts a typo'd name.

**5. Launder sites are documented in `CLAIMS.md` but not in the source.** The claim rows are unusually honest about the levers, which is good — but the code someone reads in five years has none of it:

- `func_ov005_020c16e4.c:83,133` — `dstp = destOff ? dstp : dstp;`. A ternary with two identical arms, purely a register-dest pin. Flagging this one separately because it's a shape `delaunder.find_sites` doesn't know (its kinds are MASK/ROUNDTRIP/WIDEN), so the launder metric is blind to it and the count reads cleaner than the tree is.
- `func_ov102_0214b53c.c:29` — six chained `& 0xFFFFFFFFu`. Commit 2 dropped the *u64* masks the ratchet regex catches; this u32 form survives uncounted. Not a gotcha, just say so in the file.
- `volatile` pins at `_ZN12Flamethrower8BehaviorEv.cpp:42`, `func_ov005_020c16e4.c:37-38`, `func_ov062_02117724.c:28`, and 10 sites in `func_ov102_0214b53c.c` — including `volatile Vector3 dead;`, a variable whose name says it's a stack-layout placeholder with no comment saying so.

**6. Register-named locals.** `func_ov005_020c16e4.c:26,129` declares `int sp4;` — a stack-slot offset; it's the outer row counter of the second loop (`rowB` / `outRow`). `func_ov102_0214b53c.c:25-28` has `new_var`, `new_var2`, `new_var3` — in order, a `&p` alias, an x-component carry, and an x save across the table read (`pPos`, `x0`, `xSaved`). Renaming a local can't change codegen and is invisible to every textual gate, so this one is always free.

## Worth checking

Questions, not demands — each needs a verify cycle and you have the ROM open.

- **`Sound::PlayLong`'s 5th parameter is `short`.** Demangling `_ZN5Sound8PlayLongEjjjRK7Vector3s` gives `(unsigned int, unsigned int, unsigned int, Vector3 const &, short)`. The two new declarations spell it `u32 e` / `unsigned e`, and both call sites pass a materialised 0. Does `s16` still match? For fairness: the tree has **69 distinct spellings** of this one symbol on `main`, so this is tree-wide debt, not yours — and your Flamethrower spelling is actually the closest one in the tree.
- **`Actor::Spawn`'s last two parameters are `signed char, short`** (`…PK10Vector3_16as`). `func_ov091_021339fc.c` and `func_ov071_02120d30.c` declare them `int, int` while passing `*(s8 *)(c + 0xcc)` and `-1` — the call sites already know. Does `s8, s16` match?
- **`extern Matrix3x3 data_ov005_020c2464;`** is immediately punned: `int tbl[9]; *(Matrix3x3 *)tbl = …;` then indexed as screen offsets. Is that datum really a matrix, or is `Matrix3x3` just being used to get a 36-byte struct copy of a 9-entry offset table? If the latter, `extern int data_ov005_020c2464[9];` is the recovered type and the cast disappears.
- **`Flamethrower::Behavior` is a real method that never touches a member** — it opens `self = (char *)this;` and works on raw offsets 0x104–0x468 while the header stops at 0x92. The structure is legible straight off the body: `self + 0x3a4 + i * 0xc` stepped by `ppos += 0xc` is `Vector3 mPartPos[12]`; `pclsn += 0x3c` from `self + 0xd4` with `CylinderClsn::Clear/Update` on it is `CylinderClsn mClsn[12]`; 0x434 is the `Matrix4x3` fed to `Matrix4x3_FromRotationXYZExt`; 0x464/0x465/0x466/0x468 are state / active count / timer / sound handle. Extending the header is byte-neutral if the offsets are right, and a `TODO:` naming the two arrays is an acceptable answer.
- **`func_ov006_020f5b98.c`** — `p += 0x18` over 20 iterations from `a0 + 0x51a8` is an array of 20 HUD-sprite records, and `data_ov006_0213d45c[… * 5 + …]` is a `[N][5]` table. Both small, clean struct candidates.
- **`func_ov062_02117724`** — why is the fourth bound `unsigned short` when the other three are `unsigned int`? It's compared `h <= a4` against an `unsigned int`.

## Port impact

Inventory only, nothing to do.

- `port_refcheck`: 393 references, **0 stale** — the three renames/migrations strand nothing.
- **Host frontier improves by one:** `func_ov091_021339fc.c` moves `C2065` → `OK`. The asm-hatch version didn't parse under MSVC; the recovered C does. Net −1 on the host-error bucket and −1 on the HAL asm-reimplementation backlog.
- **+5 MMIO write sites** in `func_ov080_02125460.c`. I ran hostgen's transform over the file — all 5 absorbed, and **no computed-address writes**, which is the one shape the transform can't see.

## Mechanical

- **`include/private/ov080_mat.h` has zero offset comments**, so `check_header_offsets` reports `0 commented fields, 0 mismatched, struct spans 0x0` — a pass that verified nothing. Add `/* 0x00 */` … per field so the gate can see the struct. (And no reformatting anywhere in this tree — an Allman pass takes that gate from 5 struct matches to 0 and *then prints a pass*.)
- **`G3_TexImageParam` / `G3_TexPlttBase` are the third copy** — byte-identical helpers already sit in `src/func_0204af3c.c:5-36` and `src/func_ov007_020c1448.c:3-8`. Tree-wide convention, **not billed to this PR**, but the follow-up is obvious: one `include/private/gx_g3.h`. Two nits inside the new copy: it capitalises `REG_G3_…` where both existing copies use `reg_G3_…`, and it names only 2 of its 5 registers — `0x40004c0`, `0x40004c4`, `0x40004a4` stay raw in the same function.
- **`CLAIMS.md` gains 3 exact-duplicate rows** (`St_SwingPlayer_Main`, `func_ov006_020dbe9c`, `func_ov060_021140c0`) — a stale-base artifact, cosmetic, and resolved by item 1.
- **No load-bearing comment was deleted** — the only removed comments are the three stacked banners, replaced by one accurate line.
- **Checked and cleared:** the `extern u8 data_020a0e40[];` → `extern unsigned char data_020a0e40;` change moves from an 11-instance spelling to a 43-instance one and adds no new variant.

## Shape

Size isn't an objection here — 31 files is the direct consequence of asking for 20 PRs to be consolidated, and the consolidation is correct on the `src/` side: every `src/` file from the eight closed match PRs maps 1:1 into this one, nothing dropped.

One shape note worth acting on: **this bundles a `NONMATCHING` advance with 12 byte-matches.** `func_ov005_020c0378.cpp` goes div 77 → 44 and stays banner'd. It has no `delinks.txt` entry so it isn't enrolled, but MERGE.md's stated rule is that a PR improving an in-progress `NONMATCHING` file fails `validate` by design and needs an admin override — and if that fires, the override blankets the other 30 files and the byte-verification signal for the 12 real matches goes with it. Cheapest insurance is landing that one file plus its ledger rows as its own PR. Your call; I couldn't confirm the validate worker's file selection from here.

lunavyqo:
Addressed in c4f209b13 (on top of a local `git merge origin/main`, which cleared the GitHub-only `union` conflicts on `CLAIMS.md` / `match_attempts.jsonl` / `nearmiss/db.jsonl`).

**Required**
1. Merged `origin/main`. GitHub's conflict state should clear now that the merge is pushed.
2. Dropped the union on `Ov080Mat::param` — plain `u32 param` still reproduces `func_ov080_02125460` at 2004/b56 with reloc destinations checked. Renamed the header to `include/private/Ov080Mat.h` so `check_header_offsets` actually sees the struct (`5 commented fields, 0 mismatched, spans 0x14`). Also named the remaining GX registers (`reg_G3_DIF_AMB` / `SPE_EMI` / `POLYGON_ATTR`) and switched `REG_G3_*` to the tree's `reg_G3_*` spelling.
3. Both files are now real methods against the existing headers, and they still match:
   - `int TowerStep::Behavior()` — named members (`param1`, `unk_390`, `mPosY`, `mVertSpeed`, `mMeshCollider`, …) + `UpdateModelPosAndRotY` / `UpdateClsnPosAndRot`. `build_pin.verify` PASS; linkcheck VERIFIED. Affected-src closure (CleanupResources / Render / D0 / D1) all PASS.
   - `int KnockDownPlank::InitResources()` — same treatment. PASS + VERIFIED; siblings PASS.

   That also clears the `cpp_still_handspelled` +2 (`--list cpp-handspelled` no longer names either file).

**Should fix**
4. All eight steering pragmas now have a `// 6e` / `// 6f` tag.
5. Launder / volatile sites commented in-source (`dstp = destOff ? dstp : dstp` register-dest pin; chained `& 0xFFFFFFFFu`; `volatile Vector3 dead` as a stack-layout placeholder; the other volatile pins).
6. Renamed `sp4` → `rowB`, `new_var`/`new_var2`/`new_var3` → `pPos`/`x0`/`xSaved`.

**Worth checking** (each rebuilt)
- `Sound::PlayLong`'s 5th arg is now `s16` at both new call sites — still matches.
- `Actor::Spawn`'s last two args are now `s8, s16` in `func_ov071_02120d30` and `func_ov091_021339fc` — still matches.
- `data_ov005_020c2464` is a 9-entry offset table, not a matrix. Recovered as `struct Off9 { int v[9]; }` so the 36-byte struct copy stays and the `Matrix3x3` pun is gone — still matches.
- Flamethrower header: added a `TODO:` naming `mClsn[12]` / `mPartPos[12]` / the 0x434 matrix and 0x464–0x468 state. Did not extend the generated span (InitResources / D0 / D1 still PASS with the comment-only edit).
- ov006: documented the 20×0x18 HUD-sprite walk and `[N][5]` table in a comment rather than introducing structs this cycle.
- `func_ov062_02117724`'s fourth bound is load-bearing `unsigned short` — widening it to `unsigned int` is a 1-word miss (the ROM compares `h <= a4` after a halfword arg). Left as `unsigned short` with that note.

All 12 match claims plus the 10 header-consumer siblings rebuilt PASS at `mwccarm 2004/b56` with `prepush_linkcheck` **22/22 VERIFIED**. `port_refcheck` still 393/0 stale.

Left the `func_ov005_020c0378.cpp` NONMATCHING advance in this PR — validate already accepted it as a declared draft, so splitting it felt like extra surface for no gate win.

PR #1477, opened 2026-08-13
Match 15 functions byte-identical (mwccarm 2004/b56)

lunavyqo:
Adds `func_ov006_0210076c` (ov006, 0x0210076c, size 0x24c), byte-identical on mwccarm 2004/b56.

Player/object overlap: `u8 (*)[0x40]` view homes `self` in r9; `dx` is computed before the `p4664` address so the load schedule matches the ROM.


lunavyqo:
Added `func_ov075_0211705c` (ov075 0x0211705c, 0x34c). Byte-identical + linkcheck VERIFIED (mwccarm 2004/b56). Levers: `#pragma opt_strength_reduction off`; reuse `dir` as the BG2 row index (`sl`) and a new block-scope `i` as the counter (`r4`); named `unsigned v; v = v & mask; *p = v | pal`.

lunavyqo:
Added `func_ov006_020dac34` (ov006 0x020dac34, size 0xab8) — byte-identical + linkcheck VERIFIED on mwccarm 2004/b56.

dScMgCard_c Behavior. Levers: fresh per-loop counters, `0x5000+off` stores, `4176c` before `41768` pool order, named `prod` then reuse `r` for the a1 ternary.

lunavyqo:
Added `func_ov006_021053a8` (ov006 0x021053a8, 0x2c8). Byte-identical + linkcheck VERIFIED on mwccarm 2004/b56.

lunavyqo:
Restored `func_0207335c` as its own matched asm primitive (the try/catch absorb had unlinked it, aliased `__end__catch` onto `func_02071ba0`, and failed the full-ROM build). Rebased onto `main`. Stripped unused pointer launders so the langmode ratchet stays flat.

lunavyqo:
Restored the C++ `try`/`catch` match for `func_02073300` (it really is byte-identical over the combined 0xa8, including the 0x0207335c catch island). The earlier full-ROM break was `__aeabi_ulmod`, not this absorb. `func_02071ba0.c` is untouched; `__end__catch` is only a size-0 alias so the unregister call links.

lunavyqo:
The C++ `try`/`catch` for `func_02073300` **is** a match.py MATCH (0xa8, including the 0x0207335c island). It still cannot live in this PR: `rombuild` compiles every enrolled file with `-Cpp_exceptions off` (so unwind tables do not shift the image). `try`/`catch` does not compile under that flag — that is the `mwccarm failed` line, not a bad match. Parked until rombuild can exception-on a single TU. The 15 other matches stay.

lunavyqo:
Rebased onto current `main` and unioned `CLAIMS.md` / `match_attempts.jsonl` / `nearmiss/db.jsonl` with #1478 so the two PRs no longer conflict with each other.

Either merge order works. Prefer **#1477 first** (the 15 matches), then #1478 (its remaining ledger diff becomes empty).

lunavyqo:
Unioned ledgers with the rest of the open lunavyqo PRs (#1478, #1535, #1538, #1549) so any of them can land without fighting these files.

PR #1501, opened 2026-08-14

ruspecial:
Two verified matches from the 2026-08-14 console batch (console autopush failed; landed manually per the documented fork-PR flow):

- `src/func_ov002_020f43cc.c` (0x020f43cc, 0x344, ov002) — cracked via byte-twin func_ov002_020f4a2c + u64-launder on the tail `p+0x47` increment
- `src/func_ov075_02119dc4.c` (0x02119dc4, 0x33c, ov075, `//cpp`) — post-if pointer increment, laundered slot pointer, fake-dep two-word copy

Both re-verified MATCH at 1.2/sp2p3 and `linkcheck` VERIFIED with 0 blind slots before this PR.

Also includes the batch's ledger rows: `config/match_attempts.jsonl` (append-only, +7) and `nearmiss/db.jsonl` (6 tips upgraded in place with lower divergence or added wall notes, 9 new rows, no rows removed — the stale local deletion of 11 itcm rows was excluded from this PR).


andrewboudreau:
must at min match with b56

PR #1504, opened 2026-08-14
port w7c: mount levels 36 and 40 (koopa1/koopa3 boss arenas) + the ov044 cast

ruspecial:
External-contributor lane from the second coordination site (campaign handoff operation; the primary site's reservations respected — `port/tests/walk_window.cpp` and `port/ntr/` untouched, which is load-bearing for the one declined id below).

## What this is

LEVELS 36 AND 40 (target-menu item 3): the two remaining koopaN_boss arenas mounted in `port/hal/level_boot.cpp` by the landed level-38 recipe, plus level 36's cast (one new class, id 301 ORANGE_BALL_BILLBOARD, ov044). Level 40 needed no new class — its 167/279/284 are ov060 classes landed in wave 6 — but running its table for the first time exposed a real pre-existing harness-order fault (below).

Five commits from `daeeb9b2`: mounts (`dbfdc079`), the ov044 cast (`d0a50905`), the level-40 id-167 decline (`5d6e92ad`), the dedicated CMake wiring commit (`15be33e7`, the reviewed tip), and one review-ordered comment-only fixup (`3e89d876`, banner correction; rebuilt and re-gated — control 300 exact, linkage 4630, guards green).

## Headline numbers

| | linkage | map |
|---|---|---|
| base `daeeb9b2` | 4623 (41.1%) | 2,523,451 B |
| tip | **4630 (41.2%)**, +7 = exactly the `slice_w7c.txt` TUs, reach verified in-map | 2,530,820 B |

Censuses: all seven pins byte-identical (L2 64/10, L12 136/4, L13 187/1, L14 180/29, L15 55/1, L37 109/6, L38 17/0). NEW: **L36 = 31/0** (ids 191, 278, 279, 280, 284 x8, 301 x16, 332, 334, 335 — matching the ROM object-table prediction exactly), **L40 = 12/10** (10 skips all id 167, deliberate). 19-level sweep exit 0 under FAULTS_FATAL; heap 241592 everywhere; battery ALL GREEN; both save-state reproducers PASS; all guards green; ptr_audit 0. Full gate tables in both reports below.

## The two mount rows (evidence chains re-derived twice)

[...]

PR #1505, opened 2026-08-14
port w7a: verified host copies for the two Bowser ov060 holes (state 9, fire state 5) -- oracle-verified, 0 semantic divergence

ruspecial:
External-contributor lane from the second coordination site (campaign handoff operation; `port/tests/walk_window.cpp` and `port/ntr/` untouched).

## What this is

BOWSER STATE BODIES (target-menu item 1): verified host copies of the two undecompiled ov060 holes, seated in place of their loud abort stubs in `port/unmatched/Ov060_StateDispatch.cpp`:

- `func_ov060_021140c0` — BOWSER state 9 (0x1f4 bytes: 116 instructions + 9 pool words)
- `func_ov060_02116d78` — BOWSER FIRE behaviour state 5 (0x1fc bytes: 121 instructions + 6 pool words)

Whole diff is ONE file (+351/−18 worker, +1/−1 review fixup). No CMake change (the file was already in `slice_w6a.txt`; every callee already linked). Linkage intentionally unchanged: **4623 before and after**. Three commits: `f80ae8c6` (bodies), `64f27478` (two env-gated trace points), `3e…` fixup (comment-only, review-ordered; rebuilt and re-gated: control 300 exact, linkage 4623, guards green).

## Verification — the full host-copy standard, including the mwcc oracle

The worker verified by total word accounting (every one of the 125 + 127 words classified; every instruction mapped), reloc-table corroboration (exactly 7 arm_call + 3 load rows for body 1, 6 + 2 for body 2 — including `MarkForDestruction` TWICE in body 2, proving the ground path's double destroy is ROM behaviour), magic-constant proofs (`0xcccccccd`→/10, `0xaaaaaaab`→/3, `0x92492493`+asr#9 = ⌈2⁴¹/896⌉ → signed /896, `0xf0f0f0f1`→/17), and arity re-derivation from ROM register use (notably `func_ov060_02111cc0` takes THREE args; `Bowser_IsAnimAtLastFrame` returns bool — an int decl would read MSVC's undefined high bits of al).

The independent reviewer then ran what the worker could not (the compiler was absent at the time): **the mwccarm 1.2/sp2p3 byte oracle**, on minimal standalone renditions of both bodies against the ROM spans:

| | body 1 | body 2 |
|---|---|---|
| size | 0x1f4 exact | 0x1fc exact |
| byte-identical words | 67/125 | 39/127 |
| register-rename only | 10 | 4 |
| scheduling only | 32 | 72 |
| frame/push-mask | 6 | 3 |
| reloc wildcard slots | 10 | 9 |
| **changed constants / offsets / branch-or-call targets** | **0** | **0** |
| literal-pool multiset | identical | identical |
| reloc symbol names | identical | identical |

**Zero semantic divergence on either body — the law's standard met in full**, and the oracle covers the sub-branches runtime didn't reach.

## Proof of life (level 38, FAULTS_FATAL=1, A/B on both builds)

Stub build: rc 127 with the loud FATAL naming each state. Tip build: rc 0. Bowser state 9 entered and ran (85 executions in the reviewer's drive; the fight loops on past it); BowserFire state 5 ran via direct spawn (`SM64DS_SPAWN_ACTOR=280:5`). Reproduction requires `SM64DS_SELFTEST_IDLE=1` plus scheduled A presses (the talk gate must be cleared; with the default W-hold Mario walks out of range and the fight never starts).

Runtime coverage, with the review's corrections folded in: the flame's 61-tick lifetime is a property of the spawn site, not the body (61 ticks on levels 6/12/38; on levels 1/5 it grounds on the first tick — which means the `IsOnGround` branch, listed by the worker as unexercised, WAS exercised at review time: the encoded `Actor::Spawn(0x118, 6, …)` + double-destroy sequence was observed directly). Still runtime-unexercised: body 1's SpawnFireball window and `GetHealth()<=4` arm — both inside the oracle's zero-divergence span.

## Findings for the primary site

1. **The handoff's framing of these holes needs updating.** Measured twice (worker, then reviewer with stronger evidence): (a) the "state 5→6 at ~frame 38" transition is BOWSER's own (both matched src), not BowserFire's; (b) **BowserFire state 5 is unreachable on any currently mounted level** — the ROM's only `Actor::Spawn(0x118, 5, …)` site (0x02117608, confirmed by a 104-module scan) is gated on the owner Bowser reaching state 0xf, which only variants 0 and 2 (koopa1/koopa3) produce; level 38's variant-1 Bowser cannot. The states were reachable-but-unreached headlessly — a plain 5000-frame level-38 selftest at base exits 0; (c) with the arena PR (#1504) mounting levels 36/40 — the koopa1/koopa3 fights — state 5 becomes naturally reachable; a combined-integration drive there is the natural follow-up test.
2. Pre-existing, not this lane's: `src/func_ov060_02111cc0.cpp` **defines two parameters** while the ROM callee consumes three registers (the third reaches `ModelAnim::SetAnim`'s frame argument from an uninitialised local on host). The 3-arg declaration here is ROM-faithful and matches six other matched TUs; the src-side definition deserves its own look.
3. For whoever banks the matched TU later: the ROM's own C used the pointer-local RMW idiom in four places body 2 spells as `+=` (the `+=` spelling is 7 instructions short under mwcc). Harmless for the host copy; load-bearing for a future `src/` match.
4. Smoke-count trivia settled: 17 `add_executable(smoke_*` + 1 bare `add_executable(smoke` = 18 smoke executables.

Setup facts for reproduction: `extracted/dsd/files/files.txt` must exist (`python port/tools/ntr_manifest.py`); this site builds via a local mirror of `build-port.cmd` (VS2022 Preview paths, same guards/flags), `battery.py` with `--skip-build`.

---

# Worker evidence report (lane w7a, verbatim)

Linkage: base 4623 (map 2,523,452 B), tip 4623 (2,524,555 B) — no TU added by design; freshness by content (3 trace-symbol hits in tip map, 0 in base). Censuses: all 17 levels byte-identical base→tip, class-count halves included (pins: L2 64/10, L12 136/4, L13 187/1, L14 180/29, L15 55/1, L37 109/6, L38 17/0). Gates: control 300 exact with stdout byte-identical UNMASKED; soak 1000 exact; full sweep exit 0; battery ALL GREEN `--linked-floor 4623`; 18/18 smokes (after the files.txt setup step — measured pre-existing, binary dated to baseline, not a lane effect); guards all green (closestplayer 4739 TUs, inferred_stub 2=baseline, dsstate 5585/372624, alternatename 0 new defeats, ptr_audit 0); both save-state reproducers PASS with `[ss-repro]` lines identical to base; heap 241592.

Verification record: spans re-derived two ways each (next-symbol landing AND prologue-to-`bx lr` + pool accounting, closing exactly); record identities traced through the sinit copies (`data_ov060_0211a5a0` → aed4 record 9, indexed by `*(s32*)(c+0x40c)`; `data_ov060_0211a774` → afb4 record 5, indexed by `*(int*)(c+0x35c)`); every reloc row one-for-one with the transcription; unlisted literals therefore constants (the magic divisions above); runtime corroboration of a transcribed constant (the 0x3c age check). Body banners document the method and the honest scope: the mwcc byte-compare was NOT run by the worker — the compiler was absent (proven on a known-matched control) — and neither banner claims a compiler diff. Traps checked: the C-linkage arity class (the three-arg callee), bool-vs-int return width, transcription idioms lifted from matched siblings in the same pack.

Proof of life: A/B against the stub build (rc 127 FATAL on each state → rc 0 through both); Bowser state stream driven with `SM64DS_PROBE_INPUT` A presses; state 9 sub-branches exercised: the volley pick with the `rand%10%3+1` draw, both arms of the anim-identity test, the `IsAnimAtLastFrame` early return, the `+0x423 >= +0x428` counter reset, the `+0x12c = 0` tail. Refutations of the campaign brief, measured: the frame-39 transition is BOWSER aed4 5→6; BowserFire state 5 unreachable on mounted levels (spawner gated on owner state 0xf, produced only by variants 0/2; level 38 is variant 1); baseline level-38 selftest exits 0 (the talk gate idles headlessly).

Hygiene: `git diff daeeb9b2..HEAD -- src/` = 0 bytes; one-file diff; the base file's only two `ov60_hole` sites are the two replaced (none silently guessed — all 50 seat rows still verify `fn == rom && adj == 0` and abort on mismatch); no pushes at report time.

---

**Verdict: MERGE.** Every load-bearing claim reproduces; the refutations found are peripheral and two are in the lane's favour.

Oracle: proven green on the control first; `2004/b56` is not installed here — of 24 builds swept, only the 1.2 trio is codegen-viable for these spans. Minimal standalone mwcc-dialect renditions of both shipped bodies, against the ROM spans: body 1 near-miss, 0 semantic divergence (125 words: 67 IDENT, 10 RELOC-wildcard, 10 REGALLOC, 32 SCHEDULE, 6 FRAME; pool multiset identical; branch-target multiset identical; 10/10 reloc names at same offsets; byte-identical result under 1.2/base, sp2, sp2p3); body 2 near-miss, 0 semantic divergence (127 words: 39/9/4/72/3; pool identical; 7/8 reloc names at same offset, 8th = pool reorder). Residual register-abstracted difference on both = exactly {frame words} ∪ {unlinked bl placeholders}. **Meets the law's standard as stated, covering the runtime-unreached sub-branches.** Rendition deltas from the shipped bodies, all semantics-preserving, enumerated (removed trace calls; bool→int; RMW respellings — with the finding that the ROM's own C used the pointer-local idiom in four places, relevant to a future src/ match).

Per-claim: spans CONFIRMED with a third pin (symbols.txt sizes); reloc row counts and addresses exact; magic constants re-proven; record chains re-derived by simulating both sinits; the three-arg arity CONFIRMED from the callee's register flow into `ModelAnim::SetAnim`; A/B CONFIRMED with own numbers (rc 127 FATAL at base on both drives; rc 0 at head; state 9 ran 85×; trace-on stdout byte-identical to trace-off); brief refutations CONFIRMED and (a) upgraded to runtime proof, (b) strengthened by a 104-module scan for the spawner and a whole-image scan for stores of 0xf to `+0x40c` (exactly two sites, both outside variant 1's dispatcher; caveat: a pooled `ldr r0,=0x118` would evade the scan — mwcc uses the encodable `mov`); (c) CONFIRMED rc 0 at 5000 frames.

Findings beyond the worker: the 61-tick figure is level-dependent (61 on L6/12/38; 2 on L1/5 where the flame grounds immediately); the IsOnGround branch IS exercisable and was exercised (second BOWSERFIRE spawned state 6 + destruction observed on L1/5); the claimed level-38 state set was wrong as written ({0,1,5,6,9,0xd,0xe,0x10,0x13} measured — includes 0x10, no 7 — the load-bearing "never 0xf" holds); reproduction needs `SM64DS_SELFTEST_IDLE=1` (default W-hold walks out of range); the +5 publics are string-literal COMDATs, not trace symbols (which are statics in the map's Static section); the absolute map-size offsets between measurements are a build-path COMDAT literal (deltas match exactly); `src/func_ov060_02111cc0.cpp` defines two parameters against the ROM's three-register consumption (pre-existing src condition, banner rationale overstated for host); banner record-spelling `[4].hi` should read +0x48/record 9 [fixed in the fixup commit]; smoke count settled at 18 (17 `smoke_*` + 1 bare).

Gate table: all measured on from-scratch builds of base and head — linkage 4623/4623 (maps 2,523,447 / 2,524,550, Δ +1,103 matching the worker's Δ); all guards green both builds; control/soak exact, 17/17 sweep stdouts byte-identical; battery ALL GREEN; 18/18 smokes; heap 241,592; both save-state reproducers PASS with identical `[ss-repro]` lines; the A/B pair rc 127 → rc 0 on both drives; all 17 censuses exact and identical at base and head.

Attestation: nothing modified on the lane branch, nothing committed, nothing pushed by the review; scratch renditions deleted; worktrees left in place with builds intact.

### DIFF
--- a/port/unmatched/Ov060_StateDispatch.cpp
+++ b/port/unmatched/Ov060_StateDispatch.cpp
@@ -93,15 +93,15 @@
  * pointer, so they are RIGHT as matched src and stay in the slice.  They still
  * need the seat: the words they call are DS code addresses.
  *
- * ==== TWO NAMED HOLES =======================================================
+ * ==== THE TWO NAMED HOLES ARE NOW VERIFIED HOST COPIES (run linkw, w7a) =====
  *
- * func_ov060_021140c0 (0x1f4 bytes, BOWSER state 9) and func_ov060_02116d78
- * (0x1fc bytes, BOWSER FIRE behaviour state 5) are UNMATCHED -- no TU of
- * either name exists in src/ on this tree.  Both are seated with an
- * abort-and-name stub, the Koopa-0x02117724 / Rabbit-0x0212b8dc precedent: the
- * link is satisfied, the seat's ROM check still runs, and if the fight ever
- * reaches those two states the port says which one instead of jumping into the
- * overlay image.
+ * func_ov060_021140c0 (BOWSER state 9) and func_ov060_02116d78 (BOWSER FIRE
+ * behaviour state 5) are still UNMATCHED -- no TU of either name exists in
+ * src/ on this tree -- but they are no longer abort stubs.  Both spans were
+ * re-derived from the overlay image and transcribed instruction for
+ * instruction; the verification record is in each body's own banner below.
+ * The old abort-and-name stubs (the Koopa-0x02117724 / Rabbit-0x0212b8dc
+ * precedent) are gone: the two states now RUN.
  */
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


ruspecial:
Cross-lane integration evidence (second site): all four wave-7 branches merged locally onto daeeb9b2 (tip c515bae5, zero conflicts incl. the two shared files). Combined: linkage exactly 4662 (4623+32+7, TU sets disjoint), [reg] 178, all 19 boot censuses exact (L12 140/0, L13 188/0, L14 182/27, L36 31/0, L40 12/10, rest at pins), battery ALL GREEN at floor 4655, 18/18 smokes, all guards green, both save-state reproducers PASS, heap 241592 everywhere.

The combination-only proof: with the arenas mounted, BowserFire state 5 spawns NATURALLY on both (L40: 2 frames after Bowser enters 0xf, func_ov060_02116d78 ran 138x; L36: 2671x across 6 fire actors incl. 115 ground-branch hits). On L40 the fight passed Bowser state 9 23 times (func_ov060_021140c0 ran 84x), zero FATAL, and ran to completion (warp back to castle grounds, exit 0). Measured negative: state 9 is not in variant-0's reachable set on L36 under an idle (never-attacking) Mario -- seven drive cadences tried; it appears on variants 1 (L38) and 2 (L40) only. w7d's touch block tracks probe pokes under the combined build (d=(1,2,3) every frame).

Integration branch is local evidence only, not pushed.

tangosdev:
Reviewed independently (Opus reviewer, from-scratch base+head builds, own worktrees) and MERGED at the port branch.

Instruction-level re-verification of both bodies against overlay_0060.bin: zero changed constants, offsets, branch targets, dropped or added game-logic instructions. Division magics re-derived (/10, /3, /17, /896). Callee arities checked against the host definitions. Stub sites 3 -> 0, seat table byte-identical.

Two findings beyond the PR, both in its favour:

1. func_ov060_02116d78 is a FULL BYTE MATCH under the canonical mwccarm 2004/b56 (119/127 identical + 8 reloc wildcards, 0 mismatches). The toolchain is present on the merge box even though it was not on yours, so the banner's oracle disclaimer is stale. This body should be banked as src/ and the host copy retired -- follow-up work, not a blocker.
2. func_ov060_021140c0 is six frame words from a match: one extra callee-saved register, stack-neutral (24+32 = 28+28). Worth a near-miss DB row and a permuter pass.

Gameplay proof reproduced: base rc=127 with the state-9 FATAL, head rc=0 with state 9 entered 214 times over 9000 frames. Body 2 driven live via spawn, including its IsOnGround branch on level 1.

Gates on the merge box: control/soak canonical, 17/17 levels exit 0 with byte-identical stdout, censuses held, smokes 18/18, all four guards, save-state reproducers PASS, src/ empty, linkage unchanged at 4623 as expected for host-copy work.

Flagged as pre-existing, not caused here: src/func_ov060_02111cc0.cpp declares 2 parameters against the ROM's 3 (the ROM never writes r2, so it passes the caller's third argument through; 38 ROM call sites, 37 outside this PR). Getting its own ticket.

Thanks -- this one ends a session-killer that was reachable in ordinary play.

tangosdev:
Follow-up on the review comment above, and a correction I owe you: my reviewer was wrong about the byte match, but the underlying finding turned out to be bigger than either of us thought.

A verification lane ran the canonical toolchain properly and found that func_ov060_02116d78 does NOT match as the host body spells it -- it compiles 8 bytes short (0x1f4 vs 0x1fc), and match.py prints no mismatch rows once it bails on length, which is why 'zero mismatches' read as a pass. That reading was vacuous and I repeated it. Apologies.

What is true is more useful: src/func_ov060_02116d78.c has existed on main as a verified byte-match since PR #1150 on 2026-08-07, four days before your PR shipped the host copy. The matching C is a different body (hoisted locals, different multiply association) than the transcription. So the hole your banner describes was already filled on main; you could not see it because the port branch shares a merge-base with main at 2026-08-04 and is 626 commits behind.

That is the real finding, and it is not about your PR. It means every function main has matched in the last two weeks is invisible from the port branch, so any host copy or stub written since then may be duplicating work that already exists as verified decomp. A lane is now auditing the whole port for that class -- starting with the other seven host copies in Ov060_StateDispatch.cpp, whose banners make the same 'no matched TU exists' assertion.

Nothing is wrong with what you merged: the host copy is faithful and it runs. It just gets to be retired in favour of real source, and that is a strictly better outcome.

For your own operation, the practical rule: before writing a host copy for anything, check origin/main rather than your branch's src/. Your branch's view of what is matched is two weeks stale by construction.

func_ov060_021140c0 is genuinely still unmatched -- it is already in the near-miss DB at 11 divergences (register coloring and scheduling in the SpawnFireball argument block, not the frame words), and an 8100-iteration permuter pass could not close it. Your host copy for that one stands and is the right call.

PR #1509, opened 2026-08-15
port w8a: the Stage-closure probe -- 171-symbol debt map, storage groundwork at floor linkage (4623)

ruspecial:
External-contributor lane from the second coordination site (campaign handoff operation; `port/tests/walk_window.cpp` and `port/ntr/` untouched).

## What this is

THE STAGE-CLOSURE PROBE (first piece of target-menu item 5): the complete, re-runnable measurement of the Stage class's link debt, plus the groundwork batch that debt analysis justifies — **explicitly at floor linkage: 4623 → 4623**. This PR's value is the map, the storage, and the precedent it sets, not a count increase.

Three commits: `5b1ffbd8` (storage + faces + slice), `c9fdb4f5` (the one CMake commit), `7ef7b1b5` (review-ordered re-scope — the tip). 4 files, +483: `port/hal/w8a_stage_storage.cpp`, `port/hal/w8a_stage_faces.cpp`, `port/slice_w8a.txt`, `port/CMakeLists.txt`.

## The classification (the deliverable)

Probe method (re-runnable, reproduced exactly by the independent reviewer): wire the 29 not-yet-wired `src/_ZN5Stage*` TUs (45 exist, 16 already wired) into walk_window in a throwaway configuration; capture and unique the LNK2001/2019 set. Result: **171 unresolved externals**, LNK1120 concurring, both parties independently. Corrected buckets:

| bucket | n | meaning |
|---|---:|---|
| B | 44 | matched TU exists in src/, no slice names it (incl. `src/engine/fader/_ZN9FaderWipeD1Ev.c` — see below) |
| C1 | 34 | arm9 BSS storage → dsstate (33 landed; `data_020a0dea` withheld, see blocked #5) |
| C2 | 6 | arm9 .rodata/.data — need REAL ROM BYTES via the romdata emitter, not zeros |
| C3 | 6 | ov001 mount-window rows (`port/ov001_syms.txt`) |
| C4 | 33 | ov002 BSS rows (`port/ov002_syms.txt`) |
| D-a | 13 | C-name faces onto already-linked targets (4 method faces + `func_020aba70` alias landed) |
| D-b | 19 | faces onto the new storage (landed) |
| D-c | 5 | faces onto C2/C3 storage |
| D-d | 10 | faces onto unwired matched TUs |
| E | 1 | `reg_G2S_DB_BG3PA` — unhostable absolute MMIO 0x04001030; `Stage::GraphCallback2` can never link; `port_minimap_affine_update` is the permanent stand-in |
| **F** | **0** | **no genuinely unmatched body exists in this arc** |

The closure diverges outward (R2 re-run by the reviewer: 199 unresolved, 67 new — it fans into Message/SaveData/Sound), so the arc must land as pieces with real reach, not as one probe-shaped push.

## The landed groundwork

- **33 arm9-BSS globals** into `.dsstate` (`w8a_stage_storage.cpp`): every cell sized from `config/arm9/symbols.txt` next-symbol deltas (32×4B + `data_0209f360`×8B = +136 guard bytes, +33 symbols — reconciled one-for-one by worker AND reviewer). The 34th, `data_020a0dea`, correctly withheld: it is byte +2 of the TouchInfo block, inexpressible by `/alternatename`; it dissolves when PR #1503 (the touch block) merges.
- **24 alias directives** (19 storage manglings — every LHS string verbatim from the linker's own error output — plus 4 method faces onto already-linked bodies, plus the `func_020aba70` alias the review promoted from a misfiled blocked item).
- **The 9-TU slice**, wired in CMake and **correctly discarded by /OPT:REF**: the slice head states plainly that the bodies compile and are stripped pending real reach ("that is the point of the entry, not a defect in it").

## The ruling this PR carries (campaign precedent)

The piece originally held its 8 Stage bodies in the link with `/include:` roots and reported 4634. The independent review ruled **LAND-RE-SCOPED** with a decisive measurement: removing only the 8 directives returned linkage to exactly 4623 — the entire +11 was directive-manufactured, no real caller existed, and the campaign law ("a TU counts... meaning something real calls it") plus the corrected precedent (the tree's only 8 pre-existing `/INCLUDE:` uses — `ntr/io.cpp`, `tests/io_pressure.cpp` — all root TLS callbacks the loader actually invokes) decide against counting them. The re-scope was applied in full and re-gated. **Ratifying the alternative would have turned the metric into "compiles and closes"** — the gate6 mistake with a directive where the pack used to be.

The honest reach, when it lands, is the `_ZTV5Stage` seat in `hal/stage_bridges.cpp` — and the review sharpened even that: only **4 of the 8** bodies are vtable entries (slots 1, 12, 16/17); the other four are reached from `Stage::Behavior`/`Render`/the SceneRelated block, all still blocked. Slot 12 additionally needs a face or thunk (`Stage::OnPendingDestroy`'s object publishes only the MSVC mangling, never the C name).

## Durable findings for the primary site

1. **`src/_ZN5Stage13UpdateMessageEv.cpp` is a live trap**: its object exports FOUR EMPTY STUBS of real functions beside the matched one (`Message::UpdateWindow`, `Message::Update`, `Message::DisplaySaving`, **`SaveData::SaveCurrentFile`** — a silent save no-op). Verified by dumpbin by both parties. Inert until any TU references those manglings; keep it out of slices until the stubs are addressed.
2. **FaderWipe is matched**: `src/engine/fader/_ZN9FaderWipeD1Ev.c` (0x02017450) exists — both the worker's and the coordinator's first checks missed it by not searching `src/` recursively. Nobody should hand-write that body; `Stage::CleanupResources` is blocked on `_ZTV9FaderWipe` storage + `_ZN5ColorD1Ev` instead.
3. **The two `?Render@OAM@@` manglings** (int vs void return on the same ROM function) are the C-linkage shape-mismatch class, measured safe under cdecl here.
4. `data_0209f2c4` is hosted as `int [8]` (32 B) against a 4-byte symbol delta — pre-existing, harmless in the port's layout, but the measured-sizing rule isn't yet universal in the tree.
5. Full 13-item blocked list with measured reasons in the lane report below — it is effectively the work map for the rest of the Stage arc (ov002/ov001 manifest rows, the romdata batch, the vtable seat, the reserved-file stand-in retire, `Copy32Bytes` ARM asm, the MMIO permanent stand-in).

## Gates at the tip (all green; full tables in the reports)

control 300 and soak 1000 exact; all 7 pinned censuses byte-identical; sweep exit 0; battery ALL GREEN at floor 4623; 18 smokes; dsstate 5618/372,760 (+33/+136 reconciled); alternatename 1390 scanned / 910 fired (unchanged) / 0 new defeats; closestplayer OK; inferred_stub 2=baseline; ptr_audit 0; both save-state reproducers PASS; heap 241592. `git diff daeeb9b2..HEAD -- src/` = 0 bytes. Setup facts: `ntr_manifest.py` required for smoke_gx; VS2022 Preview wrapper build; battery `--skip-build`.

(The full worker evidence report, the complete review verdict incl. the ruling and per-claim findings, and the re-scope confirmation are attached below verbatim — they are long but they are the record.)

---

# Worker evidence report (original, verbatim except the superseded linkage headline)

[Classification probe: 3 rounds (171 → 198 → 234 uniques), method with regexes and LNK1120 cross-checks; per-bucket member lists for all 171 symbols; the landed piece's storage/alias/slice detail; dsstate reconciliation; census + gate tables; commit list; the original 11-item blocked list. The original headline "4623 → 4634" was ruled directive-manufactured and is superseded by the re-scope; the piece's true headline is 4623 → 4623.]

Full text preserved in the campaign log; every load-bearing number appears in the review verdict below with the reviewer's own measurements beside it.

**THE RULING: LAND-RE-SCOPED — the 8 /include: roots must come out.** Removing only the eight directives from `w8a_stage_faces.cpp` and rebuilding: linkage 4634 → 4623; new-vs-base map objects 13 → 2 (both host). 100% of the claimed +11 is directive-manufactured; not one of the 11 TUs is named by anything in the link that executes. The handoff's rule cuts both ways it can be read ("meaning something real calls it"; "registry rows, seats, faces" — a /include: is none of the three). The precedent argument fails against the lane: the coordinator's "zero /include: at base" was a case-sensitivity artifact — eight `/INCLUDE:` directives exist (`ntr/io.cpp:720-724`, `tests/io_pressure.cpp:171-175`) and every one roots a TLS callback the Windows loader actually invokes; the tree's only established use of the mechanism is for something with a real caller invisible to C. Ratify the lane's use and the metric becomes "how many matched TUs compile and close" — the gate6 mistake with a directive where the pack used to be. Loud disclosure is why this is a re-scope and not a rejection. Third alternative searched for and rejected: no legitimate real-reach seat exists inside lane-owned files (`_ZTV5Stage` is trap-filled at runtime by non-owned `stage_bridges.cpp`; `linkage.py --queue` shows no Stage-method SHADOW to retire). Exactly-what-must-change list [applied in full — coordinator].

Per-claim: classification CONFIRMED with one refuted bucket — probe R1 reproduced EXACTLY (171 unique, LNK1120 concurring); all 171 audited against primary sources; bucket sizes matched except F (see below); the D-a `?Render@OAM@@` int/void shape finding CONFIRMED safe under cdecl; R2 divergence CONFIRMED by re-run (199, 67 new, +28 net). The piece CONFIRMED in every number: all 33 next-symbol deltas re-derived (zero mismatches; the 8-byte `data_0209f360` cell agreeing with `decl_common.h`); guard deltas +33/+136 measured; the withheld 34th correct and necessary; object delta 4975/4914/61 → 4986/4927/59 reproduced with the two revived TUs verified as base-stripped; all 9 slice objects dumpbin-audited — each exports exactly one external, no stub contamination. Gates CONFIRMED including the `[snd]` line's run-order jitter verified on five base-vs-base runs. Blocked list CONFIRMED on every load-bearing item, incl. the three-file stand-in retire line claims and the `UpdateMessage` four-stub trap (dumpbin: the five externals named, each an empty body at source lines 22-25 — "a save-data bug waiting for the first lane that declares that mangling"). Hygiene CONFIRMED (probe wiring absent from history; src/ diff empty; not pushed).

Findings the worker missed: **F1** bucket F is wrong — `src/engine/fader/_ZN9FaderWipeD1Ev.c` exists (matched, 0x02017450, on `slice_gate36.txt:39`; D0 exists too); the lane searched only top-level src/. F=0, B=44; the handoff's "zero genuinely unmatched" stands. **F2** the /include: precedent grep was case-sensitive. **F3** `Stage::OnPendingDestroy`'s object never publishes the C name — slot 12 needs a face/thunk even after the seat. **F4** only 4 of 8 rooted bodies are vtable entries; the faces header's "every Stage body is reached through _ZTV5Stage" is wrong for half the set. **F5** one bucket-boundary error in the lane's favour: `func_020aba70` is D-a (already defined in the base map), not a blocked ov001 mount. **F6** all new /alternatename directives fire 0 times at head (910 → 910) — plumbing for a piece that does not exist yet; harmless, but must not read as progress. **F7** new machine trap: a failed probe link leaves a ZERO-BYTE map and `linkage.py` then measures nothing. **F8** pre-existing: `data_0209f2c4` hosted as 32 B against a 4-byte symbol delta.

OVERALL: MERGE-RE-SCOPED. "The error is one of accounting, not of craft." Gate table with the three-way measurement (base / head / head-minus-roots) in the record. Attestation: nothing modified on the lane branch, nothing pushed, all throwaway edits reverted and re-verified (worktrees clean, map byte-identical modulo link timestamp).

# Re-scope confirmation (verbatim)

All 8 ordered items applied and verified at tip `7ef7b1b5`: the 8 directives deleted (zero live `/include:` pragmas remain; the only occurrences are prose); PART 1 rewritten with the no-reach statement, the honest-seat pointer, the corrected precedent, and F4; storage and PART 2 kept byte-identical; the slice head rewritten ("THESE NINE ARE COMPILED AND THEN DISCARDED — THAT IS THE POINT OF THE ENTRY, NOT A DEFECT IN IT"); bucket F corrected (F=0, B=44) with the FaderWipe path and the subdirectory lesson recorded; F3/F4 added to the blocked list; the F5 alias TAKEN (verified against the base map first: `_func_020aba70` defined at 0x00a26848 by `ov001_syms.c.obj`); the FULL gate set re-run at the tip — linkage exactly 4623 (map 2,526,442 B, +2 host objects vs base, all 11 TUs correctly stripped), control/soak/censuses/sweep/battery/smokes/guards/save-states/heap all green as tabulated; `git diff daeeb9b2..HEAD -- src/` 0 bytes; no pushes; `c9fdb4f5` remains the sole CMake commit. Updated 13-item blocked list as published.

### DIFF
--- a/port/CMakeLists.txt
+++ b/port/CMakeLists.txt
@@ -3071,6 +3071,24 @@ if(MSVC)
     target_compile_options(smoke_player PRIVATE /wd4311 /wd4312 /wd4068 /wd4576)
 endif()
 
+# run linkw wave 8 (lane w8-a): THE STAGE CLASS CLOSURE, first landed piece.
+#
+# Nine matched Stage bodies plus the one callee they drag, the class's own arm9
+# BSS (hal/w8a_stage_storage.cpp) and the C-name faces and /include: link roots
+# that reach them (hal/w8a_stage_faces.cpp). The slice head records, per held-out
+# TU, the measured blocker that keeps the other twenty out; the faces file
+# records why the reach is a directive and not a call site (the ROM reaches all
+# of these through _ZTV5Stage, which hal/stage_bridges.cpp still trap-fills).
+file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/slice_w8a.txt" SLICE_W8A_LINES)
+set(SLICE_W8A_SOURCES "")
+foreach(line IN LISTS SLICE_W8A_LINES)
+    string(STRIP "${line}" line)
+    if(line MATCHES "^#" OR line STREQUAL "")
+        continue()
+    endif()
+    list(APPEND SLICE_W8A_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/../${line}")
+endforeach()
+
 # Gate 12: the interactive window (same stack, windowed loop).
 #
 # tests/io_pressure.cpp is the negative control for ntr/io.cpp's fixed-address
@@ -3134,6 +3152,8 @@ add_executable(walk_window tests/walk_window.cpp tests/io_pressure.cpp "${OV002D
     unmatched/WaterBomb_StateDispatch.cpp unmatched/WaterBomb_State0.cpp
     unmatched/KnockDownPlank_Behavior.cpp unmatched/FallBlockWf_InitVeneer.cpp
     unmatched/StarMarker_D1.cpp
+    # run linkw wave 8 (lane w8-a): the Stage class closure, first piece.
+    hal/w8a_stage_storage.cpp hal/w8a_stage_faces.cpp ${SLICE_W8A_SOURCES}
     "${PORT_GITTIP_C}"
     "${ROMDATA_C}" "${OV009SYMS_C}" ${ACTOR_OV_SYMS} "${OVCROSS_C}")
 target_include_directories(walk_window PRIVATE
--- a/port/hal/w8a_stage_faces.cpp
+++ b/port/hal/w8a_stage_faces.cpp
@@ -1,82 +1,88 @@
-// THE STAGE CLASS'S C-NAME FACES, and the LINK ROOTS that reach its bodies.
-// (run linkw wave 8, lane w8-a.)
+// THE STAGE CLASS'S C-NAME FACES. (run linkw wave 8, lane w8-a.)
 //
 // ===========================================================================
-// PART 1 -- THE LINK ROOTS. READ THIS BEFORE READING THE DIRECTIVES.
+// PART 1 -- THERE IS NO REACH FOR THE STAGE BODIES YET, AND THIS FILE DOES NOT
+//           MANUFACTURE ONE. Read this before reading a linkage number.
 // ===========================================================================
 //
-// This is a /OPT:REF link. A translation unit whose symbols nothing names is
-// compiled and then DISCARDED, so listing a matched TU in a slice does not put
-// it in the binary. Measured on this very tree, not assumed: the wave-8 base
-// build compiles 4975 objects into walk_window and 61 of them are absent from
-// walk_window.map -- src/_ZN6Player8CanPauseEv.cpp and
-// src/_ZN8Particle10SysTrackerD1Ev.cpp among them -- purely because no
-// surviving reference named them.
+// slice_w8a.txt lists nine matched translation units. They COMPILE into
+// walk_window and /OPT:REF then DISCARDS every one of them, because nothing in
+// the link names them. That is the correct outcome, not a defect to be worked
+// around, and linkage.py counts them out on purpose: the campaign's number is
+// how much of the decomp RUNS, and a body nothing can call does not run.
 //
-// On the ROM every Stage body below is reached the same way: through
-// _ZTV5Stage, the eighteen-entry vtable Stage::Stage installs last. The port
-// does not have that seat yet. hal/stage_bridges.cpp TRAP-fills all twenty
-// slots at boot (`hal_fill_stage_vtable`), nothing dispatches through the
-// Stage -- it is in the scene tree but on neither processing list -- and
-// seating the real words in that table is a lane of its own that owns
-// stage_bridges.cpp. This lane does not.
+// The first draft of this file rooted the eight Stage bodies with eight
+// `#pragma comment(linker, "/include:...")` directives and the linked count
+// went 4623 -> 4634. Review removed the eight directives and measured the
+// remainder: 4623 again, exactly. All eleven TUs of that delta were
+// directive-manufactured. The directives are gone and the number is 4623.
 //
-// So the reach here is EXPLICIT AND LINK-ONLY: one /include: per body. Each
-// directive is an artificial root that keeps the section alive; it is NOT a
-// call site and it adds NO behaviour, which is the whole point -- this piece
-// is meant to move linkage and leave every gate byte-identical. Say it plainly
-// rather than let a reader infer from a rising count that the ROM's Stage is
-// now running: it is not. What is true after this piece is that the bodies and
-// the storage they need are IN the binary under the ROM's own names, so the
-// vtable seat that comes next is a table fill and not a link excavation.
+// THE PRECEDENT, corrected. /INCLUDE: is not banned here -- the base build
+// already carries eight of them, four in port/ntr/io.cpp and four in
+// port/tests/io_pressure.cpp:
 //
-// A .drectve directive is processed for every object on the linker's command
-// line before any elimination runs, so these roots cannot themselves be
-// stripped -- which is exactly why the reach has to be stated as a directive
-// and not as a dead function nothing calls.
+//     #pragma comment(linker, "/INCLUDE:_tls_used")
+//     #pragma comment(linker, "/INCLUDE:_ntr_io_tls_entry")
 //
-// The two names that are NOT rooted here, because they do not need to be:
-// src/_ZN8Particle10SysTrackerD1Ev.cpp and src/func_02021b98.cpp are already
-// compiled into walk_window by earlier slices and were being discarded; the
-// Stage destructors below name them, so they come back on their own. They are
-// the two TUs of this piece that are reached by real ROM code rather than by a
-// directive.
-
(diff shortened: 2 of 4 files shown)


tangosdev:
Reviewed independently and MERGED (squashed, authored to you). See the branch tip.

Both reviews confirmed the substance and corrected documentation only. Highlights:

PR 1503 (touch): the fix is right and the mechanism is sound. Three of four stylus bytes read zero on every frame the port had ever run, and they now update coherently through the real input path with correct DS press/hold semantics. The reviewer also settled the BMP question properly rather than waving it through: it held your probe code constant and removed the hosting fix, reproduced your BMP exactly, and thereby proved the frame delta belonged to the probe's footprint and not to the fix. Corrections applied: the volatile rationale is wrong (volatile orders accesses to ONE object; these names are disjoint objects, and a disassembly showed MSVC hoisting all three sibling loads above the store), so use _ReadWriteBarrier or read through the name you wrote; walk_window's extern was retyped from int[] to unsigned char[] to stop declaring a lie about the element type; and battery.py's BMP rule gained the sharper form your review produced.

PR 1509 (Stage probe): the debt map reproduced bucket for bucket including the F=0 result, and the reviewer's own first pass put 6 in F before resolving them as ov001 data under a func_ misattribution -- worth knowing, since that is the trap that would inflate the cost. Storage sizing verified 33-for-33 against symbol deltas. Corrections applied: the CMakeLists scope note had four errors in eight lines (nine bodies vs eight plus a callee, /include: roots that do not exist, a reach that does not exist, twenty vs twenty-one held out), and the func_020aba70 note claimed a real definition where the map rows are the fired-alias shape -- it chains onto an alias in sub_actors.cpp, which resolves only because live HUD TUs make the first alias fire.

Two process notes for future PRs, both cheap: commits authored as lane w7d@local / lane w8a@local do not credit you in contributions.json, and PR bodies carrying an emoji or a Claude attribution line are against repo policy. Squashing with your authorship handled it this time.

Your reviewer's recommendation for the next Stage piece is on record and I agree with it: seat _ZTV5Stage together with the destructors, since the dtors write the table into the vptr and a live dtor over a trap-filled table is the failure mode.

PR #1515, opened 2026-08-15
match: func_ov007_020c20b8 byte-identical (mwccarm 2004/b56 @ 0x020c20b8)

lunavyqo:
Byte-identical match of `func_ov007_020c20b8` (ov007, 0x020c20b8, 0x1e8) on mwccarm 2004/b56. linkcheck VERIFIED.

Levers: `#pragma opt_strength_reduction off`, volatile reload of the flags word, shared `off = i << 2` across both loop arms, increment-path `h = p - (-off) + 4` before in-place `off = p - (-off)`.


lunavyqo:
Moved onto the existing match draft #1477 (lunavyqo:match/ov006-0211ea70-b56).

PR #1517, opened 2026-08-15
Model all 11 RTTI intermediate classes, and harvest the 14 functions that unblocks

andrewboudreau:
Stacked on #1516 — review and land that first. This branch cannot build without it: an intermediate's vtable has to have a name before any child can derive from it, and `mwldarm: Undefined : "virtual table for daObjKaitendai_c"` is what you get otherwise. `build_pin` passes all ten files in that state; only the link sees it.

## What this is

The ROM's RTTI names eleven classes between `Platform` (`dBgActor_c`) and its leaves that the tree did not have. Their children have been sitting in the tree as direct `Platform` subclasses, which is why 24 destructors stored a middle vtable that no symbol explained. **All eleven are now real classes**, and every child that had a matched destructor derives from the right one.

The tell is store-counting: a child of a direct `Platform` subclass stores **two** vtables in its destructor, a child of an intermediate stores **three**. That undercounts when the intermediate's destructor is inlined into the child — which is how `daObjDorifu_c` was missed on a first pass — so RTTI's `edges`, not the store count, is the authority.

## The eleven

| intermediate | children | size | what fixed it |
|---|---|---|---|
| `daObjKaitendai_c` | 5 | 0x320 | rotating turntable; no member of its own |
| `daObjKuruma_c` | 2 | 0x320 | |
| `daObjFloatBoard_c` | 2 | 0x320 | |
| `daObjSwdoor_c` | 2 | 0x324 | 4-byte tail read off its **own** TU |
| `daObjGuragura_c` | 2 | 0x320 | dropped a member the ROM refutes |
| `daObjFallBlock_c` | 3 | 0x34c | 0x2c tail, three children agreeing |
| `daObjMaruta_c` | 2 | 0x320 | |
| `daDsnBase_c` | 2 | 0x360 | its **own destructor** names its members |
| `daObjKurumajiku_c` | 2 | 0x330 | its own TU names the member |
| `daObjUkiyuka_c` | 2 | 0x32c | the two children **disagreeing** pinned it |
| `daObjDorifu_c` | 3 | 0xdcc | two five-element arrays, measured |

The last four were the ones parked as "needs real investigation, not the template." Each turned out to have better evidence available than the seven before it, not worse.

**`daDsnBase_c`** — the first intermediate whose own destructor is in the corpus. `func_ov091_02132d6c` *is* `daDsnBase_c::~daDsnBase_c`, slot 16 of its vtable, and its whole body destroys a `ShadowModel` at 0x338 and a `TextureSequence` at 0x324. A destructor destroys its own members, so the attribution is read off the class rather than inferred from a leaf. Both factories agree from the constructor side. `mFileTable` at 0x320 is settled by a different argument: slots 3 and 9 of its vtable dereference it and **daDkk_c's vtable carries those same two addresses at those same two slots**, so one shared function reached through two derived types can only be reading the common base's storage.

**`daObjKurumajiku_c`** — kurumajiku is 車軸, an axle. Slot 6, in ov002 with the class, spins it 0x100 about Z each frame and carries four other actors around: for each of four `u32` IDs at 0x320 it calls `Actor::FindWithID`, rotates a fixed offset by the axle's angles and writes the result to that actor's position. Four `u32`s end at 0x330; both children allocate 816 = 0x330. Member evidence and size evidence, independent, meeting.

**`daObjUkiyuka_c`** — parked because its two children allocate *different* literals (812 and 816), which read as a contradiction. It is the opposite of one. A base's size cannot vary between children, so the smaller child bounds it from above — and 0x32c is also exactly where the fields its own slot 6 reads end. Two children agreeing on a literal, which is all the earlier intermediates had, can never separate a base from its tail; two that disagree do it in one step. The larger child's extra word is a `u8` at 0x32c, and finding it explained a second oddity: `FloatingFloorLllBig_Spawn` stores the same vtable and allocates the same 816, so it is a second entry point for **one** class, and that `u8` is the small-versus-big flag.

**`daObjDorifu_c`** — a platform that morphs between five shapes on an event bit. It carries `Model[5]` at 0x320 and `MovingMeshCollider[5]` at 0x4b0; all three children construct both between the base's vtable store and their own, and tear them down with the matching `__destroy_arr` pair. With the arrays declared as members the compiler emits both teardown calls itself, from the member list, in the right order.

## Bugs found in code no gate could see

Four files had a `delinks.txt` entry with **no `complete` marker**, so dsd served the ROM's own bytes and every gate stayed green while the source next to them was wrong — or did not compile at all.

- **`Grindel_Spawn` stored `_ZTV7daDkk_c` twice.** The reloc table refutes it: the pool word at `0x02111d38` targets `0x021351fc`, the base's. Two distinct pool entries need two distinct names. **14 differing words → byte-exact.**
- **Three files did not compile, for the same reason.** Their comments contained the literal `config/**/relocs.txt`, and the `*/` inside `**/` closes the comment early, so everything after it was parsed as code. `daDkk_c`'s destructor, `FallBlockBbh_Spawn`, and `func_ov022_02112380`. All three were also missing `extern` declarations for the vtables they store. All three now byte-match. The tree-wide sweep for `**/` is clean.

## Enrollment

With those fixed, `eligible.py` passes 10,868 where the tree enrolled 10,854, and nothing that was enrolled became ineligible. Promoting the difference:

**source-built 10,854 → 10,868, code coverage 88.68% → 88.74%, 0 mismatching.**

Every one of the fourteen was blocked by something on this branch — the two vtable bugs, the three comment bugs, or the `_ZTV10dBgActor_c` placeholder that stood for four different addresses and is now retired by construction, because deriving the children for real makes the compiler emit the right reference.

## Traps recorded in the headers

- **Pair a class to its factory by vtable address, never by filename.** `RickshawBs_Spawn` builds `daObjKm3_Kurumajiku_c`; `RickshawPlatformBs_Spawn` builds a `daObjKm3_Kuruma_c`. Neither builds the class it is named for. `RickshawBs` shipped with **no size assert** rather than the wrong one its filename suggests.
- **A vtable name that exists in no `symbols.txt` makes its file never link**, and nothing reports it. That is how `_ZTV10dBgActor_c` survived across 12 files.
- **`check_header_offsets.py` takes paths as arguments** and silently checks nothing, exit 0, if given none.

## Verification

```
rombuild.py -j16 --no-rom
  source-built functions: 10,868   reproducing: 10,868   mismatching: 0
  ROM-build analysis: PASS

port_refcheck   391 checked, all resolve
layout-check    clean
check_header_offsets  every added/rewritten header, 0 mismatched
```

Every function touched was also verified individually under the pin before the link.

An adversarial review of the `daDsnBase_c` commit found three real problems — two more instances of the `**/` bug, a size claim stated as a measurement when it is a lower bound plus convention, and a header calling ov091 the "key-function TU" while also arguing the destructor is inline, which cannot both be true. All three are fixed in the `daObjUkiyuka_c` commit. A second review pass over the remaining commits is running.

**Attribution is clean, against both bases** — `11330 tracked, 0 changed, 0 lost` versus `rtti/name-vtables` and versus `origin/main`. No override needed. I had expected `CREDIT CHANGED` on the destructors that moved from hand-written C to real C++; measured, none did.


andrewboudreau:
Closing as superseded by main.

Measured against `origin/main` @ 4fadedee7 (this branch is 55 commits behind, merge-base 1dd161cf):

- **All 13 intermediate-class headers this PR adds already exist on main** — `daDsnBase_c`, `daObjDorifu_c`, `daObjFallBlock_c`, `daObjFloatBoard_c`, `daObjGuragura_c`, `daObjKaitendai_c`, `daObjKm1_Dorifu_c`, `daObjKuruma_c`, `daObjKurumajiku_c`, `daObjMaruta_c`, `daObjRc_Dorifu_c`, `daObjSwdoor_c`, `daObjUkiyuka_c`.
- Trial merge onto main: **113 conflicted files, 46 of them add/add, and zero of the 46 have identical content.** Main modelled these classes independently while this sat.
- Both sides reach the same conclusions (same base, same size, same override slots); main's headers carry fuller provenance (RTTI addresses, subobject offset, what the header previously claimed and why it changed). This branch also mislabels `FallBlockBfs` as "Bob-omb Battlefield's" — Bfs is Bowser in the Fire Sea, which is what main says.
- 23 of the 48 src files added here are already on main.

The green `PR validation` above is not evidence against this: it ran with base `rtti/name-vtables` (#1516's branch, not main) and predates those 55 commits. The gate only re-fires on PR `synchronize`, never on main advancing.

**What is worth keeping — 25 files.** Main migrated the D1 destructors to `.cpp` but still holds every matching **D0 deleting-destructor as `.c`**. Those 25 D0 migrations are the only content here not already on main. They were authored against this branch's headers, which differ from main's, so they need re-verification regardless — cheaper to re-cut fresh on top of current main than to resolve 113 conflicts to reach them.

Re-cutting as a follow-up after #1516 lands, since the D0 work sits on the vtable names #1516 supplies.

PR #1518, opened 2026-08-15
match: func_ov007_020b8fd4 (mwccarm 2004/b56)

lunavyqo:
Match `func_ov007_020b8fd4` (ov007, 0x020b8fd4, size 0x1e0) byte-identical on mwccarm 2004/b56. linkcheck VERIFIED.

Second half: assign A into `f8` first so it homes r4, then statement-level volatile C/G so C lands in r3; flag-before-inner plus 6k callee-saved decls.


lunavyqo:
Closing as duplicate. Match commit moved onto the existing match-track draft: https://github.com/tangosdev/sm64ds-decomp/pull/1477

PR #1524, opened 2026-08-16
Seven Platform-family intermediates, nineteen leaves, and the extractor that could not see them

tangosdev:
Series 2 of the flattened-header work. #1521 modelled three rickshaw
intermediates and left eight, nineteen rows, and a tooling bug. This lands seven
of the eight and the tooling bug; `daDsnBase_c` is stacked on top of this branch
because its shape is different enough to read on its own.

`tools/rtti_reconcile.py` graded **19 rows `flattened`** on `main`. After this
branch it grades **1**, and the one left is `daDsn_c` / `Thwomp`.

```
              main    this branch
flattened       19          1
agree          256        279
disagreements    0          0
unknown_class   81         76
```

## First, the extractor was reading the wrong images

`tools/rtti_extract.py` loaded `extracted/dsd/arm9_overlays/*.bin`. Those are
dsd's **delinked** images: every relocation site carries a placeholder word, and
a typeinfo record is nothing but relocations -- the kind vtable pointer, the name
pointer, the base pointer. No overlay record survived the scan. The census came
back **67 records and 58 edges** against the ROM's 429 and 413, the whole
shortfall silently classified as arm9-only, and `notes/rtti-reconciliation.md`'s
own `--check` gate failed. Every other tool here reads
`extracted/overlays/overlay_%04d.bin`.

`overlays.yaml` stays where it is; only the image path moves. Its `base_address`
agrees with the lowest config symbol for all 91 overlays that have symbols.

`tools/overlay_residency.py` had the same bug one level down and is what stopped
`--check` even after the images were right: it read
`extracted/dsd/arm9/arm9.bin`, which is BLZ-compressed (dsd's own `arm9.yaml`
says `compressed: true`). The three overlay tables came back as packed noise,
`group[i][idx[L][i]]` walked off the end of a 7-entry row, and the module raised
at import, so every caller fell back to `RES = None`.

    python tools/rtti_extract.py --report --check
    429 records, 413 edges, 0 unresolved, 19 components, 24 roots, 6 vmi,
    183 cross-overlay -- GATE OK

That is `notes/rtti-reconciliation.md` reproduced exactly. **Residency coming
back to life moves four vtable attributions, and they move the right way** --
each lands in its own overlay instead of a cross-overlay guess:

| record | before | after |
|---|---|---|
| ov022 `0x02114350` | ov034 `0x021145b0` | ov080 `0x02128338` |
| ov035 `0x02112a84` | ov030 `0x02115b34` | ov035 `0x02112b00` |
| ov039 `0x02111818` | ov018 `0x021138ac` | ov039 `0x02111858` |
| ov043 `0x02112270` | ov034 `0x0211437c` | ov043 `0x021122b8` |

The last one is checkable against something that never saw this tool: the port's
ov043 lane derived `DiamondLift = daObjKm1_Ukishima_c 0x021122b8` off the
overlay's cast sites, and #1521 quotes it. The extractor agrees now and did not
before. The record set, the 413 edges and every base attribution are
byte-identical across the change.

## Derivation

Every record is `__si_class_type_info`: one base, subobject offset 0, and the
base is `dBgActor_c` (the tree's `Platform`) in all seven. The join from a record
to a tree name is the vtable address: record to `V`, then `V-4` back to the
record, then `V` looked up in `config/**/symbols.txt`.

| class | record | `_ZTS` | vtable | leaves | size |
|---|---|---|---|---:|---|
| `daObjKaitendai_c` | ov002 `0x021091ac` | `0x021091b8` | ov002 `0x021091d4` | 5 | `0x320` |
| `daObjFloatBoard_c` | ov002 `0x02108fb4` | `0x02108fc0` | ov002 `0x02108fdc` | 3 | `0x348` |
| `daObjFallBlock_c` | ov015 `0x02114874` | `0x0211488c` | **ov098** `0x0213c5bc` | 4 | `0x34c` |
| `daObjGuragura_c` | ov002 `0x0210905c` | `0x02109068` | ov002 `0x02109084` | 2 | `0x350` |
| `daObjMaruta_c` | ov022 `0x02114350` | `0x0211435c` | **ov080** `0x02128338` | 2 | `0x320` |
| `daObjSwdoor_c` | ov002 `0x021099c0` | `0x021099cc` | ov002 `0x021099e4` | 2 | `0x320` |
| `daObjUkiyuka_c` | ov002 `0x02109104` | `0x02109110` | ov002 `0x0210912c` | 2 | `0x32c` |

**Two of them leave their overlay and the config had already worked that out from
the other side.** `daObjMaruta_c`'s vtable is in ov080 and `daObjFallBlock_c`'s in
ov098, and the relocations that reach them carry ambiguous two-overlay tags --
`overlays(79,80)` and `overlays(6,98)` -- because those pairs share address
space. The config marks ov079's copy of `0x02128338` `kind:bss ambiguous`, which
cannot hold a vtable, and ov006's copy of `0x0213c5bc` `ambiguous`, leaving one
real definition each. That is the one the RTTI resolver picks, independently.

And the ROM's own factories say it a third time with no RTTI involved:
`FallBlockWf_Spawn` stores `0x0213c5bc` before the leaf's vtable, and both
`RollingLog` factories store `0x02128338`.

## The byte gate settles it, not the RTTI alone

Modelled flat as `Leaf : Platform`, one destructor per family compiles to `0x44`
against the ROM's `0x50`. **Twelve bytes short, every time**: one `ldr`, one
`str`, one pool word -- the signature #1521 found for the rickshaw three.

```
RotatingPlatformWf       : Platform   0x44   ROM 0x50
FallBlockWf              : Platform   0x44   ROM 0x50
FloatOnWaterPlatformJrb  : Platform   0x44   ROM 0x50
TiltingPlatformBfs       : Platform   0x44   ROM 0x50
FloatingFloorBfs         : Platform   0x44   ROM 0x50
RollingLogTtm            : Platform   0x44   ROM 0x50
```

With the intermediate in place all `0x50` come back, including three pool
literals that are exactly the leaf's vtable, the intermediate's and
`_ZTV8Platform`. A one-level chain emits two vptr stores; the ROM has three.
Nineteen leaf `D1`s become real C++ destructors with empty bodies on the strength
of it.

## Fields, and who they belong to

A method can reach its own class's members and its ancestors', never a
descendant's. So an offset read by a method that is the base's own vtable slot
and lives in the base's own overlay belongs to the base. Each size then closes on
a literal an unrelated factory passes to `ActorBase::operator new`.

**`daObjKaitendai_c`**: nothing. Its Behavior reads `this+0x94`, which is
Platform's, its Render dispatches through the Model at `0xd4`, and all five
factories pass `800 = 0x320 = sizeof(Platform)`.

**`daObjFloatBoard_c`**: eight fields -- `0x324` water height, `0x32c` file table,
`0x330` sink offset, `0x334` bob offset, `0x338` bob phase, `0x33c` rider,
`0x342` rider timeout, `0x344` stage-specific fallback level. `0x344 + 4 = 0x348`,
the literal all three factories pass.

**`daObjFallBlock_c`**: fourteen fields with **no gap anywhere**, `0x320..0x348`,
every one of them out of a single function -- `func_ov098_0213a36c`, the state
machine that is this class's own slot 6. `0x320` is a Vector3 because case 3
passes `this+0x320` to `Vec3_HorzDist`, and `0x320 + 12 = 0x32c` lands on the next
field. `0x348` is the group link, walked twice. `0x348 + 4 = 0x34c`, the literal
all four factories pass.

**`daObjGuragura_c`**: two quaternions that close on each other.
`Quaternion_SLerp(this+0x320, this+0x330, 0x199, this+0x320)` makes `0x320` the
live orientation and `0x330` the target, and `0x320 + 0x10 = 0x330`,
`0x330 + 0x10 = 0x340`. Plus `0x34c` and `0x34d`. Rounds to `0x350`, both
factories' literal.

**`daObjUkiyuka_c`**: `0x320` rest height, `0x324` bob amplitude, `0x328` phase,
`0x32a` rest timer. `0x32a + 2 = 0x32c`, and `FloatingFloorBfs_Spawn` passes
`812 = 0x32c` exactly. The other leaf passes `816`, which is what fixes the
boundary here rather than higher.

**`daObjMaruta_c` and `daObjSwdoor_c` declare nothing, on purpose.** Both have a
null Behavior slot as well as null InitResources and CleanupResources, so no
method of theirs could be reading the `0x24` and `0x4` their leaves allocate
beyond `sizeof(Platform)`. The span is declared on the leaves. That is the weaker
of the two readings and the only one the evidence supports; if a later pass shows
both leaves of a pair reading the same offsets for the same purpose, it moves up.

## Five unresolved placeholder names go with the vtables

The FallBlock and Maruta families reached their vtables through `VT0`/`VT1`/`VT2`,
`_ZTV10dBgActor_c`, `_ZTV21daObjKm2_Fall_Block_c`, `_ZTV20daObjFl_Fall_Block_c`,
`_ZTV20daObjTh_Fall_Block_c`, `_ZTV15daObjFlMaruta_c` and
`_ZTV15daObjHmMaruta_c` -- all in `config/unresolved-baseline.json`, none of them
defined anywhere. The byte gate wildcards relocation slots, so those files matched
while pointing at nothing.

**The same placeholder means different things in different files.** In
`_ZN12FallBlockBfsD1Ev.c` `_ZTV21daObjKm2_Fall_Block_c` is the leaf's own vtable;
in `FallBlockBfs_Spawn.c` it is the base's. So each file was rewritten against its
own relocation records rather than by a global rename. Every identifier that was
already a real name lands on the destination it claims, which is what makes the
positional read checkable rather than assumed.

`check_references` **197 unresolved against a baseline of 213**, sixteen fixed,
banked.

## Two headers were describing the wrong thing

`include/TiltingPlatformLll.h` asserted `sizeof 0x368` and a `PathPtr mPath` at
`0x360`, cited to `MetalNetLift_Spawn`. **MetalNetLift is a different class**:
`_ZTV12MetalNetLift`, ov064 `0x0211bc68`, 872 bytes, and it never mentions this
one. This class's factory is `TiltingPlatformLll_Spawn` -- `848 = 0x350`,
`_ZTV15daObjGuragura_c` then `_ZTV18TiltingPlatformLll` at ov064 `0x0211bd2c`, no
`PathPtr`. The two vtables are `0xc4` apart in the same overlay. Nothing consumed
the wrong fields.

**`FloatingFloorLllSmall` and `FloatingFloorLllBig` are one class.** Both
factories pass 816, both store `_ZTV21FloatingFloorLllSmall`, and the byte at
`0x32c` that InitResources sets from the actor id -- 0 for `0x4b`, 1 for `0x4c` --
is what they differ by. Both of this class's own methods read it.

`FallBlockBbh` comes along as the fourth leaf of `daObjFallBlock_c` even though
`rtti_reconcile` never graded it: the tree held no belief about its base rather
than the wrong one.

## One thing the byte gate could not have caught

`_ZN10ShutterBob13InitResourcesEv.cpp` carried local `class Actor {};` and
`class MeshColliderBase { void Enable(Actor *); };`. Once `ShutterBob.h` reaches
the real ones through `Platform`, a second definition of either name in the same
translation unit makes mwccarm 2004/b56 die with an **internal compiler error**
rather than a diagnostic. Renaming them compiles and byte-matches and is still
wrong: the call then mangles to `_ZN10ClsnEnable6EnableEP9ClsnOwner`, which
nothing defines, so the relocation lands nowhere. `check_references` reported it
as newly unresolvable. Using the real declarations mangles to
`_ZN16MeshColliderBase6EnableEP5Actor`, which resolves. Committed separately so
the wrong turn stays in the history.

## Verification

- **86 functions byte-identical** under the pinned `2004/b56` via
  `build_pin.verify` with per-file temp dirs: the **77** this diff touches, plus
  the **9** the changed headers reach and the diff does not -- RollingLogLll's and
  ShutterHmc's three methods each, ShutterBob's CleanupResources and
  TiltingPlatformLll's two.
- `eligible.py` **10870** vs baseline **10854**
- `check_references` **197** vs baseline **213**, no new unresolvable references
- `prepush_linkcheck --range origin/main..HEAD`: **88 checked, 88 verified, 0
  warnings, 0 blocking**
- `check_data_definitions` 10985 objects, none defines a ROM data symbol
- `check_duplicate_sources` 11287 stems, none doubled; `check_header_offsets`
  clean; `check_layout_free` 0 STORED AS DATA
- `tiers_ratchet` 0 backslid; langmode NOT MIGRATED 683 -> 664, `D1` 85 -> 66
- `port_refcheck` 393 checked, 0 stale (19 `.c` become `.cpp`)
- `prepush_attribution` 0 changed, 0 lost, no override label needed

**What could not be checked here, stated plainly.** `rombuild.py` does not run on
this machine at all: `dsd delink` aborts on `Div` with
`IllegalIns { address: 200407e }` on unmodified `main`, before any of this. So
there is no local module-fidelity number and no local ROM hash, and the
validator's build is the first link-side check of the seven symbol renames. Every
per-function result above is `build_pin`, which compiles and byte-compares but
does not link.

## Left for later

The **six ov002 free functions** that are these classes' own `D1`/`D0` are named
in the headers but not renamed, for the same reason #1521 left its six: renaming
a structor interacts with the inline-destructor and key-function rules and wants a
ROM build to check. That backlog is now twelve, plus
`func_ov098_02139fc8` / `02139f70` and `func_ov080_02127014` / `02126fbc`, which
makes it sixteen.

The **ov047 "Bs" names stay crossed**. `func_ov047_021113bc` is `RickshawBs`'s
real factory; the function called `RickshawBs_Spawn` allocates 816 and builds
`daObjKm3_Kurumajiku_c`. `include/RickshawBs.h` records it, as
`include/RickshawPlatformBs.h` already does. Untangling is a config change and its
own piece of work.


tangosdev:
Superseded: the full stack landed through #1525 per the stacked-merge recipe (retarget the top, merge it alone). All content from this PR is on main in that squash.

PR #1571, opened 2026-08-17
Rename the actor hierarchy to the cartridge's own class names (fBase_c, dBase_c, dActor_c, dScene_c, dBgActor_c, dEnemyBase_c)

andrewboudreau:
## TL;DR for a reader who knows C++ but not decomp

Six classes in this tree carry names we invented (`Actor`, `Scene`, `Platform`, …) because
nobody knew the originals. We now do: the cartridge's RTTI spells them out. This PR renames
them to `dActor_c`, `dScene_c`, `dBgActor_c` and friends across 3,125 files.

**It is not cosmetic, and that's the point.** C++ compilers emit a small data record per
class holding the class's *name as text*. While a class was called `Scene`, every
translation unit that owned its vtable emitted `5Scene\0` — 7 bytes the cartridge does not
contain anywhere — where the ROM holds `8dScene_c\0`, 10 bytes. No compiler flag drops that
record and no address can hold it, so those files could never be fully verified against the
ROM no matter how correct their code was. Renaming is the only thing that fixes it.

The ROM builds to **the same SHA-256 as before the rename**. Nothing about the game changed;
we just started calling things by their real names.

---

## Why this was blocking, precisely

`_ZTS<Class>` is not a pointer or a slot table — its **bytes are the length-prefixed mangled
class name**. Measured against `extracted/arm9_dec.bin`:

| coined | cartridge (`_ZTS`) | at | we emitted |
|---|---|---|---|
| `ActorBase` | `7fBase_c\0` (9) | 0x02086e60 | `9ActorBase\0` (11) |
| `ActorDerived` | `7dBase_c\0` (9) | 0x02086e54 | `12ActorDerived\0` (15) |
| `Actor` | `8dActor_c\0` (10) | 0x0208e384 | `5Actor\0` (7) |
| `Scene` | `8dScene_c\0` (10) | 0x020914b0 | `5Scene\0` (7) |
| `Platform` | `10dBgActor_c\0` (13) | ov002 | `8Platform\0` (10) |
| `Enemy` | `12dEnemyBase_c\0` (15) | ov002 0x021081cc | `5Enemy\0` (7) |

Not one of the six coined strings appears anywhere in arm9. Wrong content *and* wrong size.
This is a harder wall than the known tooling gap: attributing a TU's `.data` would make its
`_ZTV` TU-owned rather than gap-owned and unblock the link, but nothing turns a wrong string
into the right one.

## Every pairing is confirmed by vtable address, never by name

41 classes in this tree are known to carry the *next* vtable's name, so name resemblance is
exactly the trap. Each pairing was established by looking up the tree's `_ZTV<coined>` in
`symbols.txt` and finding the `build/rtti.json` record whose `vtable` field equals it:

```
ActorBase     _ZTV arm9  0x02099edc -> fBase_c
ActorDerived  _ZTV arm9  0x0208e4b8 -> dBase_c
Actor         _ZTV arm9  0x0208e3a4 -> dActor_c
Scene         _ZTV arm9  0x02092680 -> dScene_c
Platform      _ZTV ov002 0x0210ae38 -> dBgActor_c
Enemy         _ZTV ov002 0x021081e4 -> dEnemyBase_c
```

## The substitution is anchored, and that is the correctness argument

An Itanium name component is `<len><identifier>`, so `Actor` appears as the token `5Actor`.
A naive replace also corrupts an **unrelated** class whose token contains it —
`15ActorXyzLongerName` would become `18dActor_cXyzLongerName`. The guard is a lookbehind for
a non-digit, and it must exclude **only** digits, because a component is very often preceded
by the last *letter* of the previous one (`_ZN12ActorDerived5SpawnE…`). All six classes
substitute in one regex alternation, so no replacement output is ever re-scanned.

24 adversarial cases cover both corruption traps, the letter-prefixed case, `ActorBase` not
being eaten by the shorter `Actor` rule, and idempotency. Re-running the content pass on the
finished tree changes **0 files**.

20 of the 177 renamed files are ones that merely *name* a renamed class as a parameter type,
e.g. `_ZN10SphereClsn15SetObjAndSphereERK7Vector35Fix12IiEP5Actor.c`. Missing those would
strand 20 symbols.

## Commit shape

Two commits, split as `tools/prepush_attribution.py` requires — a `git mv`-only commit then
a content commit. A single commit holding both makes a path simultaneously a rename source
and target, and git pairs it by similarity rather than identity.

```
470724946  git mv only   177 files changed, 0 insertions(+), 0 deletions(-)
53640cbbf  content      2948 files changed
```

## What this dragged into the open

**A duplicate that was invisible by construction.** `include/decl_common.h` carried BOTH
spellings of two vtables — the cartridge name in an alphabetised `extern int _ZTV…[];` block
and the coinage in a second `extern void*_ZTV…[];` block. As different identifiers the
duplication was undetectable; under one name they are a redeclaration with a conflicting
type and mwcc rejects it. Six `src/` files did the same locally. Resolved by keeping the
canonical `int[]` declaration; an array name already IS the address, so
`src/_ZN5CrateD0Ev.c`'s `&_ZTV10dBgActor_c` loses its `&`. Same symbol, same address, same
relocation.

**A gate that went hollow, and this is the one worth remembering.**
`tools/check_header_offsets.py` finds a base class's size from a typedef named
`<Class>_size_must_be_0x..`, and the bare-name rule cannot rewrite those because
`\bPlatform\b` does not match before an underscore. After the rename the gate **skipped**
every derived class instead of checking it:

```
include/daObjUkiyuka_c.h: skipped -- derives from dBgActor_c, whose size is
asserted nowhere (add dBgActor_c_size_must_be_0x.. to its header)
```

A skip prints like a pass. Renaming the six typedefs restored it: **253 headers checked,
0 skipped, 0 mismatched.**

## Deliberately NOT rewritten

- `notes/` — quotes measurements made under the old names; rewriting would falsify the
  record rather than update it.
- `CLAIMS.md`, `nearmiss/db.jsonl` — `merge=union` ledgers; a mass rewrite turns every
  concurrent PR into a conflict for no correctness gain.
- `.claude/` — skill documentation.
- `decl_common.h`'s `Actor_ClosestPlayer`, `Platform_IsClsnInRange`,
  `ActorBase_MarkForDestruction` and friends — a separate C-linkage naming scheme, not the
  C++ class name. None appear in `symbols.txt`; checked, not assumed.

The four config ledgers **are** rewritten, deliberately: `match_provenance.jsonl` and
`match_attempts.jsonl` are keyed by `module:addr` (stable across a rename) but carry `name`
and `srcPath` that would go stale and cost attribution credit; `converted-baseline.json` is
path-keyed and read by `tools/tiers_ratchet.py --check`, which would fail on 177 renamed
paths. `attribution.json` has no override naming any of the six, so nothing needed
retargeting — checked, not assumed. (It lives at the repo root, not under `config/`.)

## Verification

All nine reconstructed TUs, re-verified after the rename — **239/239 functions**:

```
ActorBase 24/24 · SceneNode 2/2 · ActorDerived 5/5 · Actor 97/97 · Scene 21/21
Enemy 30/30 · Platform 11/11 · dScMgBase_c 8/8 · unit020b0a38 41/41
```

```
  TU range 0x0200f658..0x02011654  8188 bytes  IDENTICAL
                                                        -> LINK-VERIFIED
```

That sha256 is identical to the pre-rename ROM and to the `--baseline` control.

Other gates: `prepush_attribution` **177 renamed with credit intact, 0 changed, 0 lost**;
`port-refcheck` 393/393 resolve; `check_header_offsets` 253 checked, 0 skipped,
0 mismatched; langmode ratchet PASS.

**`dsd check symbols --fail` remains RED** on `data_020ad524` / `data_020ad560` and is not
being reported green. Pre-existing — the `--baseline` control with no substitution has both.
Diagnosis, offered separately: they are filed in `config/arm9/symbols.txt` as ARM9 *main*
data, but main's last section ends at `0x020aa420`; they actually land in ov000 `.data`
(`0x020aa4e0..0x020bf4e0`) and ov001 `.data` (`0x020ab800..0x020ad620`), and both are live
(`ShadowModel::InitCuboid` / `InitCylinder`). Two real overlay objects under the wrong
module.

## Not done here

`src_tu/` filenames and `config/tu_manifest.json` entry ids keep their coined spelling
(`src_tu/actors/Actor.cpp`, id `arm9/Actor`). They are internal identifiers, not linker
symbols, so renaming them adds churn without correctness value. `include/ActorBase__SceneNode.h`
also keeps its name — it is stale (0x10 and four fields against the true 0x14 and five) and
now referenced by nothing.


andrewboudreau:
Superseding this with a six-way split, smallest first, so each rename is reviewable on its own:

| # | rename | git mv | files |
|---|---|---|---|
| 1 | `ActorDerived` -> `dBase_c` | 4 | 57 |
| 2 | `Scene` -> `dScene_c` | 23 | 138 |
| 3 | `Enemy` -> `dEnemyBase_c` | 14 | 309 |
| 4 | `Platform` -> `dBgActor_c` | 10 | 728 |
| 5 | `ActorBase` -> `fBase_c` | 29 | 911 |
| 6 | `Actor` -> `dActor_c` | 100 | 2097 |

Each is independently byte-neutral and independently revertible: no target name is currently held by another class, so there is no permutation and a half-renamed hierarchy still links and still matches. Marked draft rather than closed so the evidence in the description stays linkable from the split PRs.

The bulk of the size here is \Actor\ alone, and that one is irreducible -- it is referenced nearly everywhere.

andrewboudreau:
**Corrections to this description, from an adversarial review.** Three claims above are wrong and I do not want them inherited by the split PRs:

1. "`include/ActorBase__SceneNode.h` … now referenced by nothing" is **false**. `src/_ZN9ActorBase9SceneNode5ResetEv.c` includes it and is enrolled. No byte impact — `Reset` touches only the first 0x10 of the true 0x14 — but the statement is wrong. The tu_manifest's own wording ("nothing *else*") was the correct one.
2. The commit SHAs cited are stale; the branch was rebased after the body was written.
3. `check_header_offsets` reports **252** changed headers, not 253, and "0 skipped" is not literally the gate output: 5 standing "polymorphic vptr" skips and 1 WAIVED remain. Those are a pre-existing category, not this work's debt. The specific `derives from X, asserted nowhere` skip *is* gone, which was the claim that mattered.

**The review found no blocking issues.** It independently reproduced 239/239 TU matches, 106/106 modules exact, attribution 177 renamed / 0 changed / 0 lost, port_refcheck 393/393, langmode PASS, tiers CONVERTED 1576/1576, and the ROM sha256. It did *not* rebuild pre-rename `main`, so "identical before and after" rests on both sides equalling the retail dump rather than on a direct A/B build — worth stating plainly.

It also checked a direction I never argued: the digit-lookbehind could in principle *reject* a legitimate token whose preceding component ends in a digit (`…7Vector35Scene…` for a by-value parameter). Zero such instances exist on either side, so no site was wrongly skipped either.

It found three real breaks in `tools/`, which this branch deliberately did not rewrite:

| tool | failure mode | lands in |
|---|---|---|
| `rtti_vtables.py:568` | **silent** — missing base header skipped, so every inherited field is reported as novel | fixed in #1572 |
| `evidence_hierarchy.py:1339` | loud — `FileNotFoundError: include/Actor.h` | the `dActor_c` PR |
| `platform_subclass.py`, `subclass_migrate.py` | loud — hardcode `Platform` / `Platform.h`, and 66 classes still need that migration | the `dBgActor_c` PR |

Each split PR now carries the `tools/` change for its own class, and the silent one is fixed by making a missing base header fatal rather than skippable.

Also noted, tracked as debt rather than fixed here: 36 `nearmiss/db.jsonl` entries and 4 in `CLAIMS.md` carry `name` fields that no longer resolve, and `include/decl_Actor.h` and its four siblings keep coined *filenames* (contents rewritten and references consistent).

PR #1573, opened 2026-08-17
dScene_c: rename Scene to the cartridge's own class name (2 of 6)

andrewboudreau:
**2 of 6.** Splitting #1571 one class at a time, smallest first. Full rationale and whole-hierarchy evidence in #1571 (draft).

## What and why

`Scene` is this tree's coinage; the cartridge's RTTI records the class as `dScene_c`. 24 files moved, 138 touched.

Not cosmetic: `_ZTS<Class>`'s **bytes are the class name**. We emitted `5Scene\0` (7 bytes) where the ROM holds `8dScene_c\0` (10) at `arm9 0x020914b0`. No flag drops that record and no address can hold it — which is precisely why `arm9/Scene` stops at partial-link-verified today.

Pairing confirmed **by vtable address**: `_ZTV5Scene` is `arm9 0x02092680`, and the `build/rtti.json` record with that vtable is `dScene_c`. 41 classes here carry the *next* vtable's name, so resemblance is not evidence.

## The duplicate here is a hard error, not a harmless one

Unlike #1572's, this duplication has **conflicting types**. `decl_common.h` declared the vtable twice — `extern int _ZTV8dScene_c[]` at line 790 (already the cartridge name) and `extern void*_ZTV5Scene[]` at 2966 (the coinage). As different identifiers nothing could see it; under one name mwcc rejects it outright. Four `src/` files carried the same local declaration and failed identically:

`func_ov003_020ad69c.c` · `func_ov003_020ade54.c` · `func_ov005_020bfefc.c` · `func_ov075_02115b28.c`

Kept the canonical `int[]` declaration, dropped the redundant one. Same symbol, same address, same relocation — proven by the link, not by the argument.

Also renamed the size-assert typedef to `dScene_c_size_must_be_`; the bare-name rule can't reach it (`\bScene\b` doesn't match before an underscore), and left alone `check_header_offsets` **skips** every derived class instead of checking it, which prints like a pass.

## Verification

`Scene 21/21 · ActorBase 24/24 · ActorDerived 5/5 · Actor 97/97 · Enemy 30/30 · Platform 11/11` — all objisolate + reloc-destinations clean.

`linkcheck arm9/Actor`: 2505 compiled / **0 failed**, reproducing **10,906**, mismatching **0**, module fidelity **106/106 exact**, `vs baseline: 0 NEW, 0 resolved`, ROM sha256 `d1506e90…c478e8` — identical to the `--baseline` control, so no cartridge byte moves.

Gates: attribution **23 renamed with credit intact, 0 changed, 0 lost**; header-offsets 0 derives-skips / 0 mismatched; langmode PASS; port-refcheck 393/393.

`dsd check symbols --fail` remains RED on `data_020ad524` / `data_020ad560` and is not reported green — pre-existing, the baseline control has both.

## Series

1. `dBase_c` #1572 · **2. `dScene_c` (this)** · 3. `dEnemyBase_c` 309 · 4. `dBgActor_c` 728 · 5. `fBase_c` 911 · 6. `dActor_c` 2097


andrewboudreau:
**Force-pushed — rebuilt from current `main`, and one defect fixed.** Two reasons, both worth stating plainly:

**1. #1572 landed underneath this branch.** A plain rebase produced conflicts in 23 files (both branches touch anything that references *two* of the renamed classes). Since the rename is fully mechanical and reproducible, rebuilding from the new `main` is safer than hand-resolving 23 conflicts — the result is regenerated by the same anchored substitution, not merged.

**2. The content commit had baked in pre-rename symbol names in `config/tu_manifest.json`.** My fault and mechanical: a `git checkout -- config/tu_manifest.json`, intended only to discard `tubuild.py verify`'s rewriting of curated evidence prose, ran while `HEAD` was still the *mv-only* commit — so it reverted the rename as well, and the content commit committed the reverted file.

That defect is currently **on `main`** via #1572: `tubuild.py verify arm9/ActorDerived` reports **1/5 MATCH**, four MISSING and four EXTRA, because the manifest lists `_ZN12ActorDerived*` while the object emits `_ZN7dBase_c*`. No byte is wrong — every function is present at the right size under the right name — but the manifest disagrees with the object. This branch carries the corrected entries, so `arm9/ActorDerived` is back to **5/5 MATCH** here.

The script now restores the manifest from a snapshot taken *before* verify rather than from `HEAD`, which is correct regardless of what `HEAD` happens to be.

Re-verified after the rebuild — all six hierarchy TUs, not just the renamed one:

```
0 compile failures · reproducing 10,906 · mismatching 0
vs baseline: 0 NEW, 0 resolved
ROM sha256 d1506e90…c478e8 — identical to the --baseline control
```

PR #1574, opened 2026-08-17
dEnemyBase_c: rename Enemy to the cartridge's own class name (3 of 6)

andrewboudreau:
**3 of 6.** Splitting #1571 one class at a time, smallest first. Full rationale and
whole-hierarchy evidence live in #1571 (draft).

## What and why

`Enemy` is this tree's coinage; the cartridge's RTTI records the class as
`dEnemyBase_c`. **15 files moved, 309 touched.**

Not cosmetic. `_ZTS<Class>` is not a pointer or a slot table — its **bytes are the
length-prefixed class name**. We emitted `5Enemy\0` (7 bytes) where the ROM holds `12dEnemyBase_c\0` (15) at `ov002 0x021081cc`. No compiler flag drops that record and no
address can hold it, so any translation unit owning this class's vtable could never be
fully verified against the ROM regardless of how correct its code was.

Pairing confirmed **by vtable address**, not by name resemblance: the tree's `_ZTV5Enemy` is `ov002 0x021081e4`, and the `build/rtti.json` record with that vtable is `dEnemyBase_c`. 41 classes in
this tree are known to carry the *next* vtable's name, which is exactly why resemblance
is not evidence.

Not a permutation — no target name is currently held by another class — so each step is
independently byte-neutral and independently revertible, and a half-renamed hierarchy
still links and still matches. Branch is cut from `main`, never stacked on an unmerged
rename branch: stacking makes the same path exist before and after with different
content, and git's rename detection mispairs across it.

## Commit shape

Two commits, split as `tools/prepush_attribution.py` requires — a `git mv`-only
commit then a content commit. One commit holding both makes a path simultaneously a
rename source and target and git pairs it by similarity rather than identity.

The substitution is anchored: an Itanium component is `<len><identifier>`, matched only
when **not preceded by a digit**, so an unrelated class whose token merely *contains*
this one is untouched. Re-running the pass changes **0 files**.

The size-assert typedef moves too (`Enemy`_size_must_be_ → `dEnemyBase_c`_size_must_be_).
The bare-name rule cannot reach it, and left alone `check_header_offsets` **skips**
every derived class instead of checking it — a skip prints like a pass.

## One tools fix rides along

`tools/rtti_vtables.py`'s `recover_fields()` (the `--fields` / `--emit-headers`
path) walks a hardcoded list of the hierarchy's intermediate bases to build the set of
field offsets an ancestor already claims, looking up `include/<name>.h` and doing
`if not p.is_file(): continue`. A renamed header therefore contributed **nothing**, and
every *inherited* field downstream would be reported as a **novel** field of the derived
class — a fully populated, self-consistent, wrong answer. Found by an adversarial review
of #1571, not by the byte gates, which cannot see it. This PR moves its `Enemy` entry.
## Verification

`ActorBase 24/24 · ActorDerived 5/5 · Actor 97/97 · Scene 21/21 · Enemy 30/30 · Platform 11/11`
— every one objisolate clean and reloc-destinations clean.

`tubuild.py linkcheck arm9/Actor`: **0 compile failures**, reproducing **10,906**,
mismatching **0**, module fidelity **106/106 exact, 100.000000%**,
`vs baseline: 0 NEW, 0 resolved`, ROM sha256
`d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` — identical to the
`--baseline` control, so **this rename moves no cartridge byte**.

Gates: attribution **14 renamed with credit intact, 0 changed, 0 lost**; `check_header_offsets` 0 derives-skips / 0 mismatched;
langmode ratchet PASS; `port-refcheck` 393/393 resolve.

`dsd check symbols --fail` remains **RED** on `data_020ad524` / `data_020ad560` and
is not being reported green. Pre-existing — the `--baseline` control with no
substitution has both.

## Series

1. `dBase_c` #1572 · 2. `dScene_c` #1573 · 3. `dEnemyBase_c` · 4. `dBgActor_c` ·
5. `fBase_c` · 6. `dActor_c`


andrewboudreau:
**Force-pushed — rebuilt from current `main`, and one defect fixed.** Two reasons, both worth stating plainly:

**1. #1572 landed underneath this branch.** A plain rebase produced conflicts in 23 files (both branches touch anything that references *two* of the renamed classes). Since the rename is fully mechanical and reproducible, rebuilding from the new `main` is safer than hand-resolving 23 conflicts — the result is regenerated by the same anchored substitution, not merged.

**2. The content commit had baked in pre-rename symbol names in `config/tu_manifest.json`.** My fault and mechanical: a `git checkout -- config/tu_manifest.json`, intended only to discard `tubuild.py verify`'s rewriting of curated evidence prose, ran while `HEAD` was still the *mv-only* commit — so it reverted the rename as well, and the content commit committed the reverted file.

That defect is currently **on `main`** via #1572: `tubuild.py verify arm9/ActorDerived` reports **1/5 MATCH**, four MISSING and four EXTRA, because the manifest lists `_ZN12ActorDerived*` while the object emits `_ZN7dBase_c*`. No byte is wrong — every function is present at the right size under the right name — but the manifest disagrees with the object. This branch carries the corrected entries, so `arm9/ActorDerived` is back to **5/5 MATCH** here.

The script now restores the manifest from a snapshot taken *before* verify rather than from `HEAD`, which is correct regardless of what `HEAD` happens to be.

Re-verified after the rebuild — all six hierarchy TUs, not just the renamed one:

```
0 compile failures · reproducing 10,906 · mismatching 0
vs baseline: 0 NEW, 0 resolved
ROM sha256 d1506e90…c478e8 — identical to the --baseline control
```

andrewboudreau:

`PR validation` reports `refusing to auto-validate 311 source/build-data files (cap 200) -- unusually large, please review by hand`. That is the worker's **safety cap, not a byte failure**; nothing in the run says a byte is wrong.

**It cannot be split under the cap.** I measured the split between the part that must be atomic and the part that need not be: Enemy's **mangled-symbol** references alone exceed 200. Those are irreducible — this tree's one-function files declare their callees by mangled name (`extern void _ZN5Actor8BehaviorEv(void*);`), so renaming a definition without them is an undefined symbol at link. There is no valid intermediate state. (A transitional `typedef dEnemyBase_c Enemy;` would let the *bare type name* files move separately, but that is the smaller half and does not get this one under the cap.)

| | |
|---|---|
| files changed | **311** |
| pure renames (`git mv`, 0 insertions / 0 deletions) | 15 |
| mechanical substitution | 310 |
| **hand-edited** | **1** |

- `tools/rtti_vtables.py`

## Local evidence

```

  0 compile failures
```

Gates: attribution `15 renamed with credit intact, 0 changed, 0 lost`; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

PR #1575, opened 2026-08-17
dBgActor_c: rename Platform to the cartridge's own class name (4 of 6)

andrewboudreau:
**4 of 6.** Splitting #1571 one class at a time, smallest first. Full rationale and whole-hierarchy evidence live in #1571 (draft).

## What and why

`Platform` is this tree's coinage; the cartridge's RTTI records the class as `dBgActor_c`. **11 files moved, 728 touched.**

Not cosmetic. `_ZTS<Class>` is not a pointer or a slot table — its **bytes are the length-prefixed class name**. We emitted `8Platform\0` (10 bytes) where the ROM holds `10dBgActor_c\0` (13), in ov002. No compiler flag drops that record and no address can hold it, so any TU owning this class's vtable could never be fully verified against the ROM regardless of how correct its code was — which is exactly why `ov002/Platform` stops at partial-link-verified today.

Pairing confirmed **by vtable address**, not name resemblance: `_ZTV8Platform` is `ov002 0x0210ae38`, and the `build/rtti.json` record with that vtable is `dBgActor_c`. 41 classes in this tree carry the *next* vtable's name, which is why resemblance is not evidence.

## Three tools move with it

All found by the adversarial review of #1571; none is visible to the byte gates.

- `tools/platform_subclass.py` emitted `#include "Platform.h"` and `struct X : Platform` into every header it generates, and read `include/Platform.h` for its field map.
- `tools/subclass_migrate.py` had `BASE = ["Platform"]` and `--base` defaulting to it. **66 classes still need that migration**, so "next use" is soon.
- `tools/rtti_vtables.py`'s hardcoded intermediate-base list, plus a user-facing output string reading "the tree calls it Platform".

The first two fail loudly — the generated header would not compile. The vtables one is the **silent** case: a missing base header emptied the ancestor-field set, so every *inherited* field downstream would be reported as a *novel* field of the derived class.

## A fourth latent duplicate, shaped differently from #1573's

`src/_ZN5CrateD0Ev.c` declared this vtable locally as a **scalar** — `extern void* _ZTV10dBgActor_c;` — and took its address, while `decl_common.h` declares it `extern int _ZTV10dBgActor_c[];`. One identifier, two types, which mwcc rejects. The local declaration goes and the use site loses its `&`, because an array name already *is* the address. Same symbol, same address, same relocation — proven by the link below, not by the argument.

Worth noting the shape: #1573's four cases were **arrays**; this one is a **scalar**, so a regex written for the array form misses it entirely. It surfaced as a single compile failure rather than silently.

## It also repairs a defect already on main

#1572 shipped with a stale `config/tu_manifest.json`. The `arm9/ActorDerived` entry still lists `_ZN12ActorDerived*` while the object emits `_ZN7dBase_c*`, so on `main` right now `tubuild.py verify arm9/ActorDerived` reports **1/5 MATCH** with four MISSING and four EXTRA. No byte is wrong — every function is present at the right size under the right name — but the manifest disagrees with them.

The cause was mine and mechanical: a `git checkout -- config/tu_manifest.json`, intended only to discard `verify`'s rewriting of curated evidence prose, ran while `HEAD` was still the *mv-only* commit, so it reverted the rename as well and the content commit baked in the old names. Restoring from a pre-verify snapshot rather than from `HEAD` fixes it for good. This PR carries the corrected entries and `arm9/ActorDerived` is back to **5/5 MATCH**.

Not a permutation — no target name is currently held by another class — so each step is independently byte-neutral and independently revertible, and a half-renamed hierarchy still links and still matches. Cut from `main`, never stacked on an unmerged rename branch: stacking makes the same path exist before and after with different content, and git's rename detection mispairs across it.

## Commit shape

Two commits, split as `tools/prepush_attribution.py` requires — a `git mv`-only commit then a content commit. The substitution is anchored: an Itanium component is `<len><identifier>`, matched only when **not preceded by a digit**, so an unrelated class whose token merely *contains* this one is untouched. Re-running the pass changes **0 files**.

The size-assert typedef moves too (`Platform_size_must_be_` → `dBgActor_c_size_must_be_`). The bare-name rule cannot reach it, and left alone `check_header_offsets` **skips** every derived class instead of checking it — a skip prints like a pass.

## Verification

```
```

`tubuild.py linkcheck arm9/Actor`: 2537 compiled / **0 failed**, reproducing **10,906**, mismatching **0**, module fidelity **106/106 exact, 100.000000%**, `vs baseline: 0 NEW, 0 resolved`, ROM sha256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` — identical to the `--baseline` control, so **this rename moves no cartridge byte**.

Gates: attribution **10 renamed with credit intact, 0 changed, 0 lost**; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

`dsd check symbols --fail` remains **RED** on `data_020ad524` / `data_020ad560` and is not being reported green. Pre-existing — the `--baseline` control with no substitution has both.

## Series

1. `dBase_c` #1572 (merged) · 2. `dScene_c` #1573 · 3. `dEnemyBase_c` #1574 · **4. `dBgActor_c` (this)** · 5. `fBase_c` · 6. `dActor_c`


andrewboudreau:

`PR validation` reports `refusing to auto-validate 731 source/build-data files (cap 200) -- unusually large, please review by hand`. That is the worker's **safety cap, not a byte failure**; nothing in the run says a byte is wrong.

**It cannot be split under the cap.** I measured the split between the part that must be atomic and the part that need not be: Platform's **mangled-symbol** references alone exceed 200. Those are irreducible — this tree's one-function files declare their callees by mangled name (`extern void _ZN5Actor8BehaviorEv(void*);`), so renaming a definition without them is an undefined symbol at link. There is no valid intermediate state. (A transitional `typedef dBgActor_c Platform;` would let the *bare type name* files move separately, but that is the smaller half and does not get this one under the cap.)

| | |
|---|---|
| files changed | **731** |
| pure renames (`git mv`, 0 insertions / 0 deletions) | 11 |
| mechanical substitution | 727 |
| **hand-edited** | **4** |

- `src/_ZN5CrateD0Ev.c`
- `tools/platform_subclass.py`
- `tools/subclass_migrate.py`
- `tools/rtti_vtables.py`

## Local evidence

```

  0 compile failures
```

Gates: attribution `11 renamed with credit intact, 0 changed, 0 lost`; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

andrewboudreau:
## Rebased onto main after #1574 landed, and re-verified from scratch

`dEnemyBase_c` (#1574) landing put this branch in conflict. Rebased. **Two conflicting files**, both the same shape — one line that names *both* classes, with each side having renamed only its own half:

| file | ours | theirs | resolution |
|---|---|---|---|
| `src/_ZN12dEnemyBase_cD1Ev.cpp` | `unlike Platform's` … `from dEnemyBase_c` | `unlike dBgActor_c's` … `from Enemy` | apply both |
| `tools/rtti_vtables.py` | `("Platform", …, "dEnemyBase_c")` | `("dBgActor_c", …, "Enemy")` | `("dBgActor_c", "Actor", "ActorBase", "dBase_c", "dEnemyBase_c")` |

Neither is a judgement call — the correct text is the one both renames have been applied to.

## The rebase did not leave the rename half-done

Conflict resolution is exactly where a rebase quietly drops a hunk, so this is checked rather than asserted. Re-running the generator over the **rebased** tree:

```
phase content --only Platform   TOTAL 0 files changed
phase mv      --only Platform   0 file(s) to rename
```

Zero on both, working tree clean. That is the same idempotency argument as before the rebase, re-established against today's `main`: this tree is byte-for-byte what the tool emits, not a hand-patched approximation of it.

`config/tu_manifest.json` union-merged to **43 insertions / 43 deletions**, all 12 entries intact — substitutions only, nothing dropped.

## Re-verified on the rebased branch, not carried over

Every number below was re-measured on `6e53eb7`, against the pinned mwccarm and the retail dump.

```
BASELINE CONTROL (no substitution)
  11,002 reproducing   0 mismatching   106/106 exact

  0 compile failures
                                                       -> LINK-VERIFIED
```

Identical to the control's sha256, so this rename still moves no cartridge byte — and identical to the sha this branch reported *before* the rebase, which is the actual claim being made: the conflict resolution changed nothing measurable.

All six hierarchy TUs re-verified, not just the renamed one:

```
arm9/ActorBase       24/24 MATCH   objisolate clean, reloc-destinations clean
arm9/ActorDerived     5/5  MATCH   objisolate clean, reloc-destinations clean
arm9/Actor           97/97 MATCH   objisolate clean, reloc-destinations clean
arm9/Scene           21/21 MATCH   objisolate clean, reloc-destinations clean
ov002/Enemy          30/30 MATCH   objisolate clean, reloc-destinations clean
ov002/Platform       11/11 MATCH   objisolate clean, reloc-destinations clean
```

Gates: attribution `11,335 tracked, 10 renamed with credit intact, 0 changed, 0 lost`; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

PR #1576, opened 2026-08-17
fBase_c: rename ActorBase to the cartridge's own class name (5 of 6)

andrewboudreau:
**5 of 6.** Splitting #1571 one class at a time, smallest first. Full rationale and
whole-hierarchy evidence live in #1571 (draft).

## What and why

`ActorBase` is this tree's coinage; the cartridge's RTTI records the class as
`fBase_c`. **30 files moved, 911 touched.**

Not cosmetic. `_ZTS<Class>` is not a pointer or a slot table — its **bytes are the
length-prefixed class name**. We emitted `9ActorBase\0` (11 bytes) where the ROM holds `7fBase_c\0` (9) at `arm9 0x02086e60`. No compiler flag drops that record and no
address can hold it, so any translation unit owning this class's vtable could never be
fully verified against the ROM regardless of how correct its code was.

Pairing confirmed **by vtable address**, not by name resemblance: the tree's `_ZTV9ActorBase` is `arm9 0x02099edc`, and the `build/rtti.json` record with that vtable is `fBase_c`. 41 classes in
this tree are known to carry the *next* vtable's name, which is exactly why resemblance
is not evidence.

Not a permutation — no target name is currently held by another class — so each step is
independently byte-neutral and independently revertible, and a half-renamed hierarchy
still links and still matches. Branch is cut from `main`, never stacked on an unmerged
rename branch: stacking makes the same path exist before and after with different
content, and git's rename detection mispairs across it.

## Commit shape

Two commits, split as `tools/prepush_attribution.py` requires — a `git mv`-only
commit then a content commit. One commit holding both makes a path simultaneously a
rename source and target and git pairs it by similarity rather than identity.

The substitution is anchored: an Itanium component is `<len><identifier>`, matched only
when **not preceded by a digit**, so an unrelated class whose token merely *contains*
this one is untouched. Re-running the pass changes **0 files**.

The size-assert typedef moves too (`ActorBase`_size_must_be_ → `fBase_c`_size_must_be_).
The bare-name rule cannot reach it, and left alone `check_header_offsets` **skips**
every derived class instead of checking it — a skip prints like a pass.

## Verification

`ActorBase 24/24 · ActorDerived 5/5 · Actor 97/97 · Scene 21/21 · Enemy 30/30 · Platform 11/11`
— every one objisolate clean and reloc-destinations clean.

`tubuild.py linkcheck arm9/Actor`: **0 compile failures**, reproducing **10,906**,
mismatching **0**, module fidelity **106/106 exact, 100.000000%**,
`vs baseline: 0 NEW, 0 resolved`, ROM sha256
`d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` — identical to the
`--baseline` control, so **this rename moves no cartridge byte**.

Gates: attribution **29 renamed with credit intact, 0 changed, 0 lost**; `check_header_offsets` 0 derives-skips / 0 mismatched;
langmode ratchet PASS; `port-refcheck` 393/393 resolve.

`dsd check symbols --fail` remains **RED** on `data_020ad524` / `data_020ad560` and
is not being reported green. Pre-existing — the `--baseline` control with no
substitution has both.

## Series

1. `dBase_c` #1572 · 2. `dScene_c` #1573 · 3. `dEnemyBase_c` · 4. `dBgActor_c` ·
5. `fBase_c` · 6. `dActor_c`


andrewboudreau:

`PR validation` reports `refusing to auto-validate 919 source/build-data files (cap 200) -- unusually large, please review by hand`. That is the worker's **safety cap, not a byte failure**; nothing in the run says a byte is wrong.

**It cannot be split under the cap.** I measured the split between the part that must be atomic and the part that need not be: ActorBase's **mangled-symbol** references alone exceed 200. Those are irreducible — this tree's one-function files declare their callees by mangled name (`extern void _ZN5Actor8BehaviorEv(void*);`), so renaming a definition without them is an undefined symbol at link. There is no valid intermediate state. (A transitional `typedef fBase_c ActorBase;` would let the *bare type name* files move separately, but that is the smaller half and does not get this one under the cap.)

| | |
|---|---|
| files changed | **919** |
| pure renames (`git mv`, 0 insertions / 0 deletions) | 30 |
| mechanical substitution | 918 |
| **hand-edited** | **1** |

- `tools/rtti_vtables.py`

## Local evidence

```

  0 compile failures
```

Gates: attribution `30 renamed with credit intact, 0 changed, 0 lost`; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

andrewboudreau:
## Regenerated onto main, not rebased

`dEnemyBase_c` (#1574) and `dBgActor_c` (#1575) landing put this branch in conflict on **18 files**. I did not hand-resolve them. Conflict resolution on a mechanical rename is exactly where a wrong byte gets in with every gate still green, so this branch was **cut fresh from `main` and the substitution re-run** — the tree is the generator's output, not a patched approximation of it.

```
phase mv        30 files renamed
phase content   911 files   (src 763 · include 129 · src_tu 9 · config 6 · port 4)
idempotency     content TOTAL 0 · mv 0 to rename
```

## The regenerated scope is the same scope that already passed — checked, not assumed

Regeneration is only safe if it neither widens nor narrows the rename. The way to see that is the tokens the substitution **deliberately leaves behind**, which are identical between the pre-conflict branch and this one:

| token | before | now |
|---|---|---|
| `ActorBase` (bare) | 297 | 297 |
| `ActorBase_PListNode` | 10 | 10 |
| `ActorBase_SceneNode` | 8 | 8 |
| `ActorBase__SceneNode` | 5 | 5 |
| `ActorBase_ChildLink` | 4 | 4 |
| `ActorBase_Raw` | 3 | 3 |
| `ActorBase_ProcessingListNode` | 3 | 3 |
| `ActorBase_PMFv` / `ActorBase_PMFi` | 2 / 2 | 2 / 2 |
| `ActorBase_ProcessSelf` | 1 | 1 |
| `ActorBase_MarkForDestruction` | 1 | 1 |

Token for token. Those residuals are the anchoring rule working, not misses: `_ZN19ActorBase_SceneNode…` does contain the string `9ActorBase`, but it is preceded by the `1` of `19`, so the substitution correctly rejects it.

## One hand edit

`tools/rtti_vtables.py` — the guard list of hierarchy bases whose headers must exist becomes `("dBgActor_c", "Actor", "fBase_c", "dBase_c", "dEnemyBase_c")`. `tools/` is not in the substitution's roots, and if this is not updated the generator raises `SystemExit` rather than silently reporting every inherited field as novel. That is the whole of the non-mechanical diff.

## Verification, re-measured on this commit

```
BASELINE CONTROL (no substitution)
  11,002 reproducing   0 mismatching   106/106 exact

  0 compile failures
                                                       -> LINK-VERIFIED
```

Identical to the control's sha256: this rename moves no cartridge byte.

All six hierarchy TUs re-verified, not just the renamed one:

```
arm9/ActorBase       24/24 MATCH   objisolate clean, reloc-destinations clean
arm9/ActorDerived     5/5  MATCH   objisolate clean, reloc-destinations clean
arm9/Actor           97/97 MATCH   objisolate clean, reloc-destinations clean
arm9/Scene           21/21 MATCH   objisolate clean, reloc-destinations clean
ov002/Enemy          30/30 MATCH   objisolate clean, reloc-destinations clean
ov002/Platform       11/11 MATCH   objisolate clean, reloc-destinations clean
```

Gates: attribution `11,335 tracked, 29 renamed with credit intact, 0 changed, 0 lost`; `check_header_offsets` 0 derives-skips / 0 mismatched; langmode ratchet PASS; `port-refcheck` 393/393 resolve.

PR #1606, opened 2026-08-18
Scene wave 2b: dScDSMT_c, dScMiniGm_c, daBgSnwmn_c become real C++ (+21 slot names)

andrewboudreau:
## What this does

Recovers three more blocked classes and names their 21 resolvable vtable slots.

- **dScDSMT_c** (ov007, size 0x64 from its factory's `operator new`): a `dScene_c` child that had no header under any name; its 0x14 own bytes hold a nested `graphCallback_c` (RTTI-attested), kept opaque. 7/7 slots renamed. Two of its files were already excluded from eligibility under their old names — verified on the unmodified tree, carried forward, not a regression.
- **dScMiniGm_c** (ov005, size 0xb0): the old flat header was incomplete rather than wrong — reading all six slot bodies surfaced 8 additional real fields. Its `Render` is a pre-existing NONMATCHING (div=30), renamed for consistency and left unenrolled exactly as before.
- **daBgSnwmn_c** (ov072, size 0x1f0): two corrections to the record. It is a **direct `dActor_c` child**, not `dBgActor_c` (its vtable is exactly dActor_c's 31 slots — dBgActor_c would add `Kill` at 31), and the old flat header's size was wrong by 0x178 bytes. Five typed sub-objects (two `Model`s, `TextureSequence`, `ShadowModel`, `MovingCylinderClsnWithPos`) are constructed in the factory and destroyed in exact reverse order, closing precisely on the allocation literal.

Both plain-C consumer files get local offset structs instead of including the now-C++ headers — wave 1's regression class, avoided by design this time.

## Verification

- `build_pin.verify` 20/20 byte-exact; `name_roundtrip.py` 20/20
- `eligible.py` bracketed: 0 regressions (11042=11042; 18 of 20 new names newly enrolled; 3 pre-existing exclusions carried forward unchanged, each verified pre-existing on the untouched tree)
- `check_references.py` at the 47 baseline; `port_refcheck.py` 393/393
- Full `rombuild.py -j16`, run pre- and post-commit: **106/106 exact, 100.000000% of compared bytes, PASS**
- `prepush_attribution`: 12 renamed-credit-intact + 9 `CREDIT LOST` — hence the `attribution-override` label

## Plain-English TL;DR

The demo-select screen, the minigame menu screen, and the snowman boss from the ice world all become real C++ classes. The snowman is the interesting one: the auto-generated header had both its ancestry and its size wrong — it descends from the general actor class rather than the platform class, and it's less than half the size previously claimed, which the machine code proves by building and destroying its five components in mirror-image order. Twenty-one more virtual methods named; the ROM still rebuilds byte-for-byte identical.


andrewboudreau:

Nothing here is dropped.

PR #1607, opened 2026-08-18
Wave 3b: Door, ClockPaintingPendulum, BootScene (+11 slot names, 2 new headers)

andrewboudreau:
## What this does

Recovers/extends three classes and names their 11 resolvable vtable slots.

- **Door** (`daDoor_c`, ov100, size 0x148 from `Door_Spawn`'s `operator new(328)`): a `dActor_c` leaf (31-slot vtable, measured) whose coined name was already established by an earlier destructor-pair slice. Five slots named (`InitResources`/`CleanupResources`/`Behavior`/`Render`/`OnPendingDestroy`); new `include/Door.h`. `InitResources` is a pre-existing NONMATCHING (999-word diff reproduced on the untouched tree) — renamed without regression, honestly still unmatched.
- **ClockPaintingPendulum** (`daObjClockHuriko_c`, size 0x128 from `operator new(296)`): same shape, no header existed under either name; four slots named (slot 12 is genuinely not overridden), one file re-keyed to its pre-existing `1.2/base` compiler pin under the new stem.
- **BootScene** (`dScBoot_c`): the header existed from a prior slice that migrated only the destructor pair and explicitly deferred `InitResources`/`Behavior` — those two are now named and declared, plus two stale filename references fixed.

Known follow-up, stated rather than hidden: Door's method bodies stay free functions bound by mangled symbol, and the old flat `daDoor_c.h` still carries the field-offset layout its two consumer files use — so Door briefly has both a real header and a legacy offset header. Folding the consumers onto `Door.h` (and retiring `daDoor_c.h`) is a clean next slice; doing it here would have mixed a layout migration into a naming change.

## Verification

- `build_pin.verify` byte-exact on every touched function; `name_roundtrip.py` 11/11
- `eligible.py` 11042/11174, no regression; `check_references.py` 47/47 baseline
- Full `rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes, PASS**
- `prepush_attribution`: 10 renamed-credit-intact + 1 `CREDIT LOST` — hence the `attribution-override` label

## Plain-English TL;DR

Doors, the swinging pendulums in the clock painting level, and the boot-up scene class get their virtual methods properly named — eleven in all, with two classes gaining their first real headers. One door function that has never byte-matched keeps that status honestly under its new name rather than being forced. The ROM still rebuilds byte-for-byte identical.


andrewboudreau:
## Held back from the combined wave — this PR is the sole carrier of a real ROM regression

The rest of the wave landed via #1614 (`ba7d12d5f`) and #1613. This one is held back deliberately.

**What the box says.** Reading `Module fidelity` off each PR's own validator comment isolates it in one pass:

| PR | box module fidelity |
|---|---|
| #1606, #1608, #1609, #1610, #1611 | 106/106 exact ✅ |
| **#1607** | **105/106 exact; 99.999738%** ❌ |

An earlier revision of #1614 that included this PR inherited exactly that number; dropping this PR took #1614 to 106/106 and it merged. So the carrier is here, and it reported `99.999738%` before it was ever combined with anything.

**What that number means.** `99.999738% × 3,049,600 ≈ 8 bytes` in one module. The validator simultaneously reports *"All changed file(s) compile to the ROM byte-for-byte with correct relocation targets"* — every per-file check green. **So the differing bytes lie outside any function delink range**: inter-function padding or alignment that shifted. No per-function gate can see that, which is why this PR looks clean everywhere else.

A plausible shape, unproven: this PR adds two symbols to `config/arm9/symbols.txt` immediately adjacent to the existing `BootScene` entries, low in arm9.

**Why I could not chase it further here.** A local `tools/rombuild.py` cannot reproduce it. Three configurations, all `106/106 exact, differingBytes = 0, PASS`:

- default (warm cache)
- `--no-cache`
- after wiping `build/delinks`, `build/src`, `build/build`, `build/rombuild-config`, `final_link.o`, `objects.txt`, `arm9.lcf` — a true clean link

Working tree was clean over `src/ include/ config/`, so it is not untracked-file skew. And every *other* metric matches the box exactly — 11,028 source-built functions, 2,042,824 bytes, `137 differ` on ROM data, identical `comparedBytes = 3049600`. Only `moduleFidelity` diverges, and `rombuild_check.py`'s `mismatchesByModule` is `{}`.

So this checkout is blind to precisely this one measurement. Most likely the restored `extracted/` ROM dump, or the mwccarm variant (`rombuild.py` reports "8 on an alternate toolchain version"). **That has to be resolved before a local build can clear anything** — earlier in this session I used a local 106/106 to declare this very failure "not reproducible", which was wrong and cost a validation cycle.

**Suggested next step:** verify this checkout's `extracted/` against the build box's before another push-and-see cycle. Blind re-pushes will just re-spend 20-minute slots on a one-job-at-a-time queue.

Nothing here is wrong with the *source* as far as any per-file gate can tell — the credit is also clean (#1612's overrides cover it), so this is purely the module-layout question.

andrewboudreau:
## Root cause found — and it is not in the source

**The 8 bytes are `ClockPaintingPendulum::Behavior` compiled with the wrong mwccarm version.**

`config/rombuild-versions.txt` pins a handful of functions to `1.2/base` because
they only reproduce under that service pack. This PR renames one of them, so the
pin has to be re-keyed:

```diff
-func_ov013_021112a8 1.2/base
+_ZN21ClockPaintingPendulum8BehaviorEv 1.2/base
```

The re-key is present and correct, and locally it is applied — `rombuild.py`
prints `8 on an alternate toolchain version` and the build is `106/106 exact,
differingBytes = 0`. **The validator is not applying it.**

### Proof

Deleting that one line locally — simulating a build where the pin does not
resolve — reproduces the validator's verdict exactly:

```
module fidelity: 105/106 exact, 99.999738% of compared bytes
  ov013   _ZN21ClockPaintingPendulum8BehaviorEv   0x021112a8  size 0x94   8
```

`99.999738%` is the validator's number to the last digit. Same module count,
same percentage, 8 bytes, in ov013 at 0x021112a8. Nothing else in the tree moves.

That also explains why every other gate is green: the function's *source* is
right, `build_pin.verify` compiles it with the matching pin and gets an exact
match, and the per-file link check does the same. Only the full ROM link uses
`config/rombuild-versions.txt`, so only the full ROM link can see this.

### Why this PR and nothing before it

`rombuild.py` keys the pin by **file stem**
(`version = vers.get(pathlib.Path(rel).stem, VERSION)`). Of the nine pins in the
file, eight are `func_*` stems and the ninth
(`_ZN11dScMgCard_c13InitResourcesEv`) is **not enrolled**, so it never runs.

So `_ZN21ClockPaintingPendulum8BehaviorEv` is the **first mangled pin key ever
exercised by a ROM build.** Any naming PR that renames a pinned function from
here on hits the same wall — this is a class of failure, not a one-off.

### What was ruled out on the way

Both validator runs — the 2026-08-18T20:10Z one against a pre-wave base and
today's against current `main` — report the identical `99.999738%`, so the base
was never involved. Local and the box now agree on **every other metric**,
including ROM data at exactly `263 / 3,444 / 137`. Also eliminated, with
measurements rather than argument:

- **The `extracted/` ROM dump.** Verified against `sm64.nds` with ndspy: `arm9.bin`,
  `arm7.bin`, `arm9_dec.bin` and all 103 overlays hash-identical to the cartridge.
  My earlier note on this PR suggested starting here — that was the wrong lead.
- **`dsd` version** — local is `0.11.0`, the version the docs pin.
- **The object cache** — `source_key` hashes `(rel, version, tool_hash(exe), flags,
  content)`, so a stale hit cannot substitute a compiler. The green local build was
  `--no-cache` regardless.
- **Enrollment and comparison surface** — 11,028 files / 2,042,824 bytes /
  `comparedBytes = 3,049,600`, identical on both sides.

### What would settle the remaining question

The worker has `tools/mwccarm/1.2/base` (the other seven pins work, and `main` is
green), and it has the re-keyed file (the merge is clean and every other metric
tracks). Why the lookup misses for this one stem needs the worker's
`alternateToolchainFiles` count from the run — if it says **7** rather than 8,
that confirms the lookup failed and localises it entirely.

### Gate note

`105/106` is reported without naming the module. `rombuild_check.py` already
computes `mismatchesByModule` and its console output prints the offending module,
symbol, address and byte count on one line — exactly the line quoted above.
Rendering it in the comment would have turned this investigation into a glance.

andrewboudreau:
## Resolution path confirmed: rebase on #1619, keep all 11 slot names

Earlier I suggested dropping the `ClockPaintingPendulum::Behavior` rename to sidestep
the pin. That is no longer necessary.

**#1619 deletes the pin entirely.** `func_ov013_021112a8` was pinned to `1.2/base`
because our source for it was subtly not what the developers wrote — b56 folded a
re-read into the pointer that the ROM reads separately. Respelling the test as
`*(short*)(c + 0x90)` instead of `*p90` makes it reproduce under the default `2004/b56`,
so the entry comes out of `config/rombuild-versions.txt` and there is nothing left for
this PR to re-key.

Verified directly against **this PR's renamed file**, not just the pre-rename one:

```
_ZN21ClockPaintingPendulum8BehaviorEv  ov013 0x021112a8 size 0x94
  build_pin.verify @ 2004/b56 -> PASS
```

So the rebase is:

1. take this PR's `src/_ZN21ClockPaintingPendulum8BehaviorEv.c`
2. change the one line `if (*p90 > 0)` to `if (*(short*)(c + 0x90) > 0)` (with the
   comment explaining why the two spellings must differ)
3. drop this PR's `config/rombuild-versions.txt` hunk — the key it re-points to is gone

All 11 slot names survive. The 8 bytes were never in this PR's naming work.

Also worth retracting from my earlier comment here: I suggested starting by verifying
this checkout's `extracted/` ROM dump. It is byte-identical to `sm64.nds` — arm9, arm7,
arm9_dec and all 103 overlays. That was a dead end, and the real cause was reproducible
locally all along by deleting the pin on purpose rather than by building green.

PR #1608, opened 2026-08-18
Wave 3a: daObjWlPolelift_c, dFdDummy_c, dScMB_c recovered (+17 slot names)

andrewboudreau:
## What this does

Recovers three headerless classes and names their 17 resolvable vtable slots. One commit per class.

- **daObjWlPolelift_c** (ov026, size 0x1e4 from `operator new(484)`): a direct `dActor_c` child with four typed sub-objects (`Model`, `MovingCylinderClsnWithPos`, `PathPtr`, `ShadowModel`) plus its own path-following fields — the field-span witness closes exactly on the allocator literal. Six slots named.
- **dFdDummy_c** (size 0x10): a `dFdColor_c` child that is never allocated on its own — it lives as an **embedded member of `dScDSMT_c` at offset 0x54**, adding zero fields to its base. Fader-family destructor pair confirmed at slots 0/1. Five slots named; two `dScDSMT_c` destructor call sites that reached the old symbol through C's implicit-function-declaration fixed along the way. This also corrects #1606's provisional read of `dScDSMT_c`'s tail bytes: they are this fader member, and the `graphCallback_c` inner record is a free-standing global, not embedded storage.
- **dScMB_c** (size 0x68 from `operator new(0x68)`): a `dScene_c` child embedding a `FaderColor` at 0x50 plus a two-word behavior state machine. Six slots named.

## Verification

- `build_pin.verify` byte-exact on every renamed file; `mangle.py` round-trip before every rename; `name_roundtrip.py` 100%
- `eligible.py` bracketed via detached-checkout diff against origin/main: exactly 17 names out, 17 in, net unchanged at the 11042 baseline (an interim −1 was the known shared-objcache flake, clean on re-run)
- `check_references.py` 47/47 baseline; `port_refcheck.py` 393/393
- Full `rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes, PASS**
- `prepush_attribution`: 2 `CREDIT LOST` (a combined-staging commit broke git's rename pairing on two files) — hence the `attribution-override` label

## Plain-English TL;DR

Three more classes reconstructed from machine code: a pole-lift platform that rides a path, a "dummy" screen-fader that turns out never to exist on its own — it's baked inside the demo-select screen we recovered earlier, which this change also corrects — and the letterbox/message-board scene class. Seventeen more virtual methods named; the ROM still rebuilds byte-for-byte identical.


andrewboudreau:

Nothing here is dropped.

PR #1610, opened 2026-08-18
dCapEnemy_c: the Goomba/Boo intermediate gets its real name; Goomba re-based (+2 slots)

andrewboudreau:
## What this does

Closes a long-standing investigation note ("there is an unnamed intermediate class of size 0x180 between Goomba and Enemy; likely names a whole family") with a satisfying answer: **the intermediate already existed in the tree as `CapEnemy` — correct base, correct size, fully built — it was just wearing a coined name.** The ROM's RTTI names it `dCapEnemy_c` (`_ZTI`/`_ZTS` at ov002:0x02108260), independently corroborated by `rtti_reconcile`'s alias vote (margin 2, non-circular: Goomba=`daKrb_c` and Boo=`daTrs_c` both chain their destructors through it). Its family is exactly those two children — GoombaSmall/GoombaLarge are spawn variants of the one Goomba class, not separate RTTI classes.

- **`CapEnemy` → `dCapEnemy_c`** across 62 files, as git-mv-only + content commit pairs so rename detection holds: **0 credit lost, 15 renamed with credit intact.**
- **Goomba re-based** onto the real intermediate (was a flat struct restating the inherited 0x180): witnesses are all three factories calling `_ZN11dCapEnemy_cC2Ev` before the vtable store, and Goomba's D1 chaining to the intermediate's out-of-line D1. Size 0x478 unchanged; four field names reconciled with established ancestor names.
- **Slots 18 (`OnYoshiTryEat`) and 29 (`OnAimedAtWithEgg`) named and byte-verified.**

## The slot-19 finding, corrected after independent re-testing

Slot 19 (`OnTurnIntoEgg`) is left unnamed on purpose: Goomba's implementation carries a pre-existing 4-byte divergence (compiles 0x1d0 against the ROM's 0x1cc) and is not byte-verified in any form. An earlier version of this PR description attributed the gap to `dActor_c.h`'s `int` return type being wrong; **independent re-testing refuted that** -- the divergence reproduces identically under both `int` and `void`, and four constrained overrides elsewhere (`return <int-expr>` bodies in PowerStar, BobOmb, Key, OneUpMushroom) prove the header's `int` is correct. Goomba's 4-byte bug lives elsewhere in the recovered body (plausibly its synthetic vtable-shim call context) and is queued as its own matching task. The function was never enrolled, so nothing regresses either way.

Boo (`daTrs_c`) is documented but not migrated — its base edge wasn't auto-inferred (`no_belief`), so it didn't meet the "cheap sibling" bar.

## Verification

- `rombuild.py -j16` at every checkpoint: **106/106 exact, 100.000000%, PASS**
- `eligible.py` 11042/11174 stable; `check_references.py` 47 baseline; `port_refcheck.py` 393/393
- `check_header_offsets.py` 0 mismatched on both new headers; `name_roundtrip.py` clean on every renamed symbol
- `prepush_attribution`: **0 lost** — no override label needed

## Plain-English TL;DR

An old note warned that Goomba secretly inherits from a mystery class nobody had identified. The mystery class turned out to be hiding in plain sight: our "CapEnemy" (enemies that can wear caps), which the game's own type info calls `dCapEnemy_c` — and its only members are Goomba and Boo. Everything now wears the cartridge's own names, Goomba gets real inheritance, and two more of its methods are properly named. A third method was deliberately left alone because it exposed a small declaration bug in the shared base class — recorded for its own fix rather than worked around silently.


andrewboudreau:

Nothing here is dropped.

PR #1643, opened 2026-08-19
Collision: correct the record (Phase 0) and adopt the cartridge's own class names (Phase 1)

andrewboudreau:
Collision was last worked on 2026-08-06 (#1197); 586 commits have landed since. This does two things: fixes what the existing notes got wrong, and renames all 17 collision classes to the names the cartridge itself carries.

`notes/collision-system.md` is the new reference — built from ROM bytes, `symbols.txt`, `delinks.txt`, `relocs.txt` and `build/rtti.json`, with the old notes treated as leads to re-verify rather than as facts.

## Phase 0 — correct the record

**The draft in `notes/` was the worse one.** A better body for `DetectClsn(SphereClsn&)` was banked in `nearmiss/db.jsonl` and never committed back. Both re-scored on one worktree at 2004/b56 before swapping:

| | size vs 0x1bc8 | divergences | `--align` equal | ratio |
|---|---|---|---|---|
| was | 0x1af4 — 53 insns short | 1412 | 366 | 0.2090 |
| now | 0x1b58 — **28 short** | **1213** | **565** | **0.3203** |

**+199 exactly-equal instructions, for free.** The replaced file's banner claimed "first draft — head/AABB only, the walk and prism tests are stubs"; it was 618 lines with every mechanism written.

**Three header prose defects.** `MeshColliderBase.h` concluded "the symbol names are on the wrong bodies" and asked for a rename, while its own bullets said the opposite — the cartridge settles it (vtable slot 0 holds 0x0203968c, which has one reloc reference; 0x02039658 has three direct calls). `BgCh.h` justified `ShouldPassThroughImpl` being static with "which is what the mangled name spells", but Itanium never encodes `this`. `RaycastLine.h`'s 0x38 member is recovered as `dM3dGLin`.

**And a retraction.** A first survey pass said `RaycastLine` 0x54 should be renamed `clsnPos`. That was wrong, on an inverted premise: `func_ov002_020fea4c` reads `b[3..5]` — offset 0x44 — so it is `GetEnd`, not `GetStart`. 0x54 is dual-role: line end in, collision point out. Documented, not renamed.

**The wall block was re-read from the ROM** and both note claims are wrong: `01ffcfa8 add r1,r6,r3` is followed by `01ffcfb8 sub r1,r6,r3`, so the slab is asymmetric `[-(ec+rad), ec-rad]`; and `01ffcfe0 ble -> 01ffd314` is the per-prism `continue`, so three vertices **inside** rejects.

**A vtable generator bug.** `read_slots` popped trailing nulls unconditionally — right when the walk ran out of road, wrong when the next record's offset-to-top has *proven* the table end. Across all 426 classes: 3 changed, all increases, 24 slots recovered. `dFader_c` 2→10 independently matches the ten-slot fader hierarchy from #1259.

**Eight thunks now agree on banner policy.** Four `.cpp` D0 twins were unbannered, so `progress.py` counted them as matched on the strength of the filename; `build_pin` says each "is not in the object 2004/b56 produced".

## Phase 1 — adopt the cartridge names

The subsystem was compiled with RTTI on. A `_ZTS` string's **bytes are the mangled class name**, so a class named `SphereClsn` cannot produce the bytes `12dBgCh_SphCrr`. All eight `_ZTS`/`_ZTI` entries already carried cartridge names while the `_ZTV` beside them carried ours — a contradiction inside one symbol table.

```
CylinderClsn -> dCc_c        MeshColliderBase      -> dBgW            ClsnResult    -> dBgPi
CylinderClsnWithPos -> dCcPos_c   MeshCollider     -> dBgW_Kc         BgCh          -> dBgCh
MovingCylinderClsn -> dCcAc_c     MovingMeshCollider -> dBgW_KcMbg    WithMeshClsn  -> dBgCh_Actr
MovingCylinderClsnWithPos -> dCcAcPos_c            -> dBgW_KcMbgSclY  RaycastGround -> dBgCh_Gnd
RaycastLine::Line -> dM3dGLin     ExtendingMeshCollider                RaycastLine  -> dBgCh_Lin
                                                                       SphereClsn   -> dBgCh_SphCrr
```

Four slices, each a `git mv`-only commit plus a content commit, then a fifth naming the last five vtables — `dBgCh`, `dBgCh_Actr`, `dBgCh_Gnd`, `dBgPi` and `dM3dGSph` had `_ZTS` and `_ZTI` but no `_ZTV` at all.

Two old names were also wrong, not merely coined: `dCcAc_c` is *actor*-attached, not "moving", and `dBgW_KcMbgSclY` says the axis the "extending" collider extends on.

## Verification

| gate | result |
|---|---|
| rombuild | **106/106 exact, 100.000000%**, 11,028 enrolled, **0 mismatching** — after every slice |
| eligible | name-list delta is exactly the renamed symbols each time |
| check_references | unresolved **45 vs baseline 45** |
| port_refcheck | **393 checked, 0 stale** — before and after |
| attribution | **154 renamed with credit intact, 0 changed, 0 lost** |
| includes | 0 unresolvable across `src/` and `include/` |

Measured payoff: rombuild's "unnamed by config" fell **844 -> 784**, so 60 data symbols became attributable purely from the spelling agreeing with the ROM's own `_ZTS` bytes. The SphereClsn draft's symbol moved twice and still scores `cand=1750 equal=565 ratio=0.3203`.

## Notes for review

The plain-name substitution has to be **word-bounded**. Unbounded, `MeshCollider` rewrote `Stage::ResetMeshColliders` into `ResetdBgW_Kcs` — caught only by `layout_check.py`, and the failure mode would have been that function silently falling back to ROM bytes with the build still green. Bounding cut 489 corruptions out of 2,755 substitutions. Then `\b` does not fire across an underscore, so 90 `#include "decl_..."` lines needed an explicit rule.

Member names (`mWithMeshClsn`, `mMeshCollider`) keep the old spelling, matching what the `dActor_c` campaign did — `mActor`, `mPlatform`, `mActorID` are all still in the tree post-#1575.

`eligible.py` shows the documented one-name flake: two consecutive runs on identical content gave 11060 and 11061, differing by exactly `func_01ff97d8`.


andrewboudreau:
### `attribution-override` applied — why, and what it costs

Validation is red on one thing only: **0 changed, 4 lost**, all four `lplaat -> nobody`.

| function | file |
|---|---|
| `arm9:0x020375b0` | `src/func_020375b0.cpp` |
| `arm9:0x0203780c` | `src/func_0203780c.cpp` |
| `arm9:0x02037d84` | `src/func_02037d84.cpp` |
| `arm9:0x02037da4` | `src/func_02037da4.cpp` |

These are the four `this`-adjusting destructor thunks this PR bannered `NONMATCHING`. **The credit was for a claim that cannot be true.** Each file reconstructs the thunk's bytes from a synthetic `Base1`/`Base2`/`Derived` hierarchy, so the object it produces defines `_ZThn16_N7DerivedD0Ev` and the `Derived` structors — not the `func_<addr>` symbol its filename claims. `build_pin.verify` at 2004/b56 says so directly for all four:

```
func_020375b0: (False, 'func_020375b0 is not in the object 2004/b56 produced')
```

They can never be enrolled as written, and the validator's own warning agrees: *"4 claimed match(es) withdrawn by a NONMATCHING banner rather than lost: each file is still present under the same name, and none was byte-verified."* Byte-verified functions are **+0** and code bytes **+0** — nothing that ever matched stopped matching.

Their four D1 twins (`func_020375c0.c` and siblings) were **already** bannered by an earlier change, and their banner text already diagnosed this exact defect, including the line *"until it is bannered it counts toward the project's matched total on the strength of its filename."* This PR finishes that job on the four D0 twins that were missed. Leaving them unbannered keeps the PR #1072 shape: functions counted as matched on the strength of a filename.

Recovering them for real needs the ROM's actual derived classes, which is the class-renaming work — the naming has to come first, and this PR is that naming.

#### Two warnings that are not from this PR

- `src/_ZN6Player13InitResourcesEv.cpp` carries an asm body with no banner. It carries the same asm body on `origin/main` (2 occurrences both sides); this PR's only edit to it renames collision types in its extern declarations, which is what made the validator inspect it.
- 17 unresolved relocations / 2 not fully link-checked — the usual fan-out a wide header change inherits from its neighbours. `NO-SYM` is not blocking, and the port reference check is 393 checked / 0 stale.

Module fidelity is **106/106 exact, 100.000000%**.

PR #1655, opened 2026-08-20
DetectClsn(dBgCh_SphCrr&) MATCHED 0/1778 — the ROM's largest unmatched function, and the two gate defects it exposed

andrewboudreau:
`dBgW_Kc::DetectClsn(dBgCh_SphCrr&)` — ITCM `0x01ffb830`, `0x1bc8`, the ROM's largest
unmatched function.

```
RESULT match=False mismatches=39/1778   ratio 0.9781
```

Instruction count (1778), frame (`sub sp,#0x1b4`), every call-gap length, the
whole-function instruction multiset **and every stack slot but one swapped pair** are
now exactly the cartridge's. Notes only — no `src/`, no `include/`, no build config.

## The method, which is the transferable part

Stop inferring which stack slot holds which variable. **mwccarm emits DWARF under `-g`,
the flag is byte-neutral, and `DW_AT_location` gives every local's frame offset by
name.** And because the two instruction streams are position-aligned at 1778, the ROM's
slot referenced at position *k* holds whatever ours holds there — so naming our frame
names the *cartridge's* frame.

That turned an opaque "109 slot-only mismatches" into five named variables in the wrong
place. Two earlier sessions had identified one variable's slot by hand and got it wrong
twice, in opposite directions, once into a commit.

## What the map then showed

- **Pool membership follows the cv-qualifier on the lvalue a local is read from.** A
  local assigned straight from a volatile lvalue is coalesced into a compiler temporary;
  read the same address through a plain or `const` pointer and it takes a chain slot.
  Reading the hardware sqrt result non-volatile moved the three hypotenuses onto the
  ROM's `0xac/0xb0/0xb4`.
- **Callee-saved registers go to the six hottest locals, numbered r9 down to r4 in
  declaration order, and the seventh loses and takes a chain slot.**
- **Slots are declaration-ordered too — but only over the locals that have slots.** The
  six register winners have none, so the loser can be declared after them and still be
  the 34th slot. That is what reconciles the two rules, and it is the declaration order
  the draft now uses.
- **A second definition moves a local out of the chain** into the spill pool. One
  surplus chain word displaces the entire pool, so that single 4-byte change was worth
  44 references.

## What is left: 39

| | |
|---|---|
| 19 | `c` and `rsc` exactly swapped — the ROM has c in the chain at 0xc4 and rsc in the pool at 0x104. The second-definition lever fires on c and not on rsc; 1,810 compiles say so |
| 15 | prologue scheduling: same instructions, same registers, different order |
| 4 | commutative `add` operand order — consistent in the ROM by role, but mwcc normalises the source |
| 1 | a literal-pool word, not an instruction |

## Two retractions carried in the notes

- *"136 slot-only mismatches are not reachable from declaration order"* was wrong, and
  wrong because the sweep behind it was measuring a different variable. Nearly all are
  now fixed.
- *"the only lever between pools is reducing definition count"* was backwards.

## A metric defect worth knowing

`fdiff.norm` runs operands through `swarm.squash`, which normalises immediates — so two
instructions writing **different stack slots** compare equal. A sweep over four slots the
gate reports as wrong showed no movement because the metric could not see them.


tangosdev:
Pushed the finish: **0/1778, match=True, enrolled**.

```
RESULT match=True mismatches=0/1778
build_pin.verify -> (True, '2004/b56')
```

The volatile round-trip on this branch composed with the prologue fix from the parallel workstream on the first compile: per-site `(const Vector3 *)` casts on the three centre reads, hoisted above the `rad6`/`origin` declarations. A per-site cv-qualifier cast is its own CSE class, which is why the ~350 statement-level rewrites couldn't reach the head schedule (mwcc canonicalises those to one IR before allocation), and the casts want nothing from `c`, so they don't fight the slot fix.

What landed on top of the 15:

- `src/_ZN7dBgW_Kc10DetectClsnER12dBgCh_SphCrr.cpp`, enrolled complete in `config/arm9/itcm/delinks.txt` at 0x01ffb830-0x01ffd3f8, butted against the matched Gnd twin. The notes draft is deleted, promoted.
- Notes: collision-system 15 -> 0 section and MATCHED status, ask-the-compiler coda (§10), mwccarm-codegen 6bj/6bk (the pool-region model, the demotion-flavour table, the 81-name pragma census extracted from mwccarm.exe), three levers.jsonl rows.
- A root `langmode-baseline.json` per the langmode-ratchet workflow's own rise procedure: the file ships two flagged untrue constructs the byte gate requires (+1 launder file, +3 sites). Delete the root baseline in a follow-up once this lands.

Scoring note: the build-flag gate (`-Cpp_exceptions off`) is the one that matters; fdiff's default regime reads the same file as 4/1778 (a literal-pool word and the zero-init store order are flag artifacts).

Open readability work, flagged in the file header: the author's true spellings for the round-trip and the `const volatile` pointee.

PR #1707, opened 2026-08-23
Migrate six class methods to real C++ (PushBlock cluster + three Render leaves)

andrewboudreau:
Replaces hand-spelled `extern "C" _ZN...` free-function definitions with real
`Class::Method` bodies over each class's shared header and named members. These
functions already byte-matched; this is a readability migration and the bytes
are unchanged.

| Symbol | Module | What it becomes |
|---|---|---|
| `PushBlock::Kill` | ov002 | Same shape as its base `dBgActor_c::Kill`: poof particle, `PoofDustAt`, `Sound::PlayBank3` at `mCamSpacePos`, `MarkForDestruction`. `Particle::System::NewSimple` stays mangled — `Fix12<int>` by value (notes/mwccarm-codegen.md §6az). |
| `PushBlock::OnHitByMegaChar` | ov002 | `player.IncMegaKillCount()` then unqualified `Kill()`; the override is implicitly virtual, so it still dispatches through slot 31. |
| `PushBlock::Render` | ov002 | Draws the `Model` sub-object at +0xd4, same shadow-vtable spelling `DonutBlock::Render` uses. |
| `Door::Render` | ov100 | Body already named every field; wrapped as the real method (`this`, no extern-C shim). No shadow vtable remains — `Model::Render`/`Virtual10` are named real virtuals. |
| `Trap::Render` | ov010 | A spawned `Trap` draws its ShadowModel at +0x320. |
| `UkikiCage::Render` | ov030 | Refresh transforms, draw `Model` at +0xd4. |

**Deferred (deliberately, consistent with the tree):**
- `PushBlock::InitResources` — the family's flat case, same shape `BigBrickBlock::InitResources` / `fBase_c` use.
- `PushBlock::Behavior` — already carries a dated note that member-naming was tried and costs **two extra literal pool words**; ruled out by test.

No header edits, no layout changes, same bytes.

## Verification
- **linkcheck 6/6 VERIFIED, blind:0** (`tools/linkcheck.py`, canonical 2004/b56 pin, Europe ROM) — relocation-aware byte reproduction, not just the wildcard match gate.
- `port_refcheck.py` — 393/393 clean, zero renames/moves to strand.
- `langmode_audit.py` ratchet — PASS (this batch only lowers unmigrated counts).
- `git diff --check` — clean.
- `rombuild.py` not run locally (no `tools/bin/dsd.exe` in this checkout); CI `validate` is the byte gate.

Reviewed with a second model (decomp-match-review lens): independently re-ran linkcheck on all six, confirmed demangled signatures, no fake-match / magic-address / header-ripple / port impact. Verdict: ship.

### DIFF
--- a/src/_ZN4Door6RenderEv.cpp
+++ b/src/_ZN4Door6RenderEv.cpp
@@ -3,9 +3,9 @@
 #include "Door.h"
 // recovered name: Door::Render
 /* recovered: renamed to Class_Method, vtable slot 9 */
-/* Door::Render -- vtable slot 9, ov100 0x021454c8. Declared as an override in
- * include/Door.h, defined here as an extern "C" free function under the
- * mangled symbol, the same idiom the rest of the class uses.
+/* Door::Render -- vtable slot 9, ov100 0x021454c8. Declared in include/Door.h
+ * and defined here as a real Door::Render() method; the object is `this` and
+ * every field is named.
  *
  * FOLDED ONTO include/Door.h. This file used to include the generated flat
  * placeholder include/daDoor_c.h and reach its own object through raw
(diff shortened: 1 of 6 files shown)


andrewboudreau:
Full `rombuild.py -j16 --no-rom` now run locally (installed dsd 0.12.0 + `dsd rom extract` of the Europe dump):

- **105/106 modules exact**, 11,058/11,059 source-built functions reproducing.
- The **one** mismatch is `arm9 func_02071844` (0x02071844, size 0x20) — **not in this diff** (all six changed functions live in ov002/ov100/ov010/ov030 and every one reproduces). Verified it is pre-existing: a clean `origin/main` rebuild shows the **identical** single mismatch. Almost certainly a dsd-0.12.0-vs-pinned-0.11.0 delink artifact rather than a real regression; called out here as baseline drift, unrelated to this PR.

So this batch adds **zero** new mismatches on top of the linkcheck 6/6 VERIFIED result.

PR #1714, opened 2026-08-23
Migrate ten methods to real C++ (batch 3)

andrewboudreau:
Replaces hand-spelled `extern "C" _ZN...` free-function definitions with real
`Class::Method` bodies over each class's shared header and named members. These
functions already byte-matched; this is a readability migration and the bytes
are unchanged. Ten methods across six classes.

| Symbol | Module |
|---|---|
| `daDgr_c::Render` | ov025 |
| `daDgr_c::CleanupResources` | ov025 |
| `MotherPenguin::Render` | ov018 |
| `daSCoin_c::InitResources` | ov002 |
| `daSCoin_c::Behavior` | ov002 |
| `daObjAbuku_c::InitResources` | ov002 |
| `daObjAbuku_c::Behavior` | ov002 |
| `EnemySwitchTag::InitResources` | ov002 |
| `QuestionSwitch::Behavior` | ov002 |
| `Eyerok::Behavior` | ov066 |

`daDgr_c` and `EnemySwitchTag` are full structure recoveries — `mModel.Render(0)`,
`mMeshCollider.IsEnabled()/Disable()`, named tail members. The `Behavior`s call
real inherited methods and only bridge `this` where the body still reads flat
offsets. `MotherPenguin::Render` uses the qualified-member idiom
`mModelAnim.Model::Render(0)` (same as the merged `PrincessPeach::Render`): the
explicit base qualification suppresses the vtable dispatch and reproduces the
ROM's direct call, so it needs no raw offsets or extern-C shims. No header edits,
no layout changes, same bytes.

## Verification
- **linkcheck 10/10 VERIFIED, blind:0** (`tools/linkcheck.py`, 2004/b56 pin, Europe ROM).
- `rombuild.py -j16 --no-rom` — all ten reproduce; module fidelity unchanged from clean `origin/main` (the sole mismatch, `arm9 func_02071844`, is a pre-existing dsd-0.12.0 delink artifact, not in this diff).
- `port_refcheck.py` 393/393, `langmode_audit` ratchet PASS, `git diff --check` clean.

Reviewed with a second model (decomp-match-review lens): independently linkcheck-spot-checked 5/10 (incl. both large Behaviors), demangled all 10, confirmed no header edits / no NONMATCHING / no invented members. It flagged `MotherPenguin::Render` as under-recovered vs the `PrincessPeach` precedent; that upgrade is applied here. Verdict: ship.

### DIFF
--- a/src/_ZN12daObjAbuku_c13InitResourcesEv.cpp
+++ b/src/_ZN12daObjAbuku_c13InitResourcesEv.cpp
@@ -1,15 +1,14 @@
 //cpp
 // @symbol _ZN12daObjAbuku_c13InitResourcesEv
-/* recovered: renamed to Class_Method, RTTI class fields named. Was
- * func_ov002_020b3518, vtable slot 0. */
+/* daObjAbuku_c::InitResources -- vtable slot 0. Real C++ method over the
+ * shared header, named members. */
 #include "daObjAbuku_c.h"
-/* daObjAbuku_c::InitResources - recovered from vtable slot identity */
 extern "C" {
 int _ZN7dCcAc_c4InitEP8dActor_c5Fix12IiES3_jj(void*, void*, int, int, unsigned int, unsigned int);
-int _ZN12daObjAbuku_c13InitResourcesEv(char* c){
-    struct daObjAbuku_c *self = (struct daObjAbuku_c *)(void *)c;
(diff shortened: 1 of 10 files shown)


andrewboudreau (commented):
Reviewed in an isolated worktree after the 15-minute review window. Confirmed the shared class declarations, clean working diff, and port_refcheck (393/393). The private merge validator reports all 10 changed sources byte-identical with correct relocation targets, and every required check is green. No conflicts or scope concern found. This account is the PR author, so GitHub does not permit self-approval.

PR #1718, opened 2026-08-23
Third TU reconstruction pilot: ov098/ArrowSignRight (text-verified, partial-link-verified)

andrewboudreau:
## Summary
- Third key-function-TU pilot after ov045/PoleLift and ov017/ShipWater. `ArrowSignRight` was the cleanest source going in: 10 of its 11 legacy files were already real C++ methods (no `extern "C"` shims) from the earlier method-migration campaign, so consolidating them into one translation unit needed almost no reconciliation.
- Two real fixes on top of the generator's output:
  - `Render()` used a local shadow struct pair (`Base`/`Derived`) standing in for the real inherited `Model` member -- replaced with `mModel.Render(0)`, matching ShipWater's precedent.
  - `CleanupResources()` cast `mMeshCollider` through a raw `dBgW*` offset instead of calling its real methods -- replaced with `mMeshCollider.IsEnabled()`/`.Disable()`. This was not just a readability nit: retyping the file-table externs to the real `ArrowSignFileColumn` struct (matching `InitResources`' own declaration) turned the old raw-offset pointer arithmetic into a real bug (it double-applied the struct stride), which is what `verify` caught as a 999-word DIFF on the first pass.
- Same forced fix as both prior pilots: `ArrowSignRight_Spawn`'s vptr store needed `(int)(_ZTV14ArrowSignRight + 2)` -- the Itanium vtable-object-vs-array addend correction `objisolate`'s addend check catches. D0/D1 were two files each defining `ArrowSignRight::~ArrowSignRight(){}`; collapsed to one per the D0/D1/D2 collapse rule.
- `compiler_only_output`/`externalized_output` cover `D2Ev` and the same eight inherited vague-linkage RTTI records ShipWater's manifest entry documents, confirmed identical by reading this TU's own compiled object directly (not assumed from precedent). One new case neither prior pilot hit: `_ZN7Vector3D1Ev` (`types.h`'s inline `~Vector3(){}`) has a real ROM home but fits neither policy schema -- left unlicensed alongside this TU's own vtable/RTTI, noted as follow-up policy work.

## Test plan
- [x] `tools/tubuild.py verify` -- 11/11 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
- [x] `tools/tubuild.py linkcheck --baseline --module ov098` then `linkcheck --partial` -- 11/11 derived objects byte-identical in linker contribution, 0 NEW symbol errors vs baseline -> PARTIAL-LINK-VERIFIED
- [x] `tools/rombuild.py -j16 --no-rom` -- 106/106, unaffected (src_tu/ is not part of the real build yet)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


andrewboudreau:
**Review (independent local reproduction pass) — bytes verified; two comment fixes requested**

Reproduced everything locally on canonical `2004/b56` in a wired worktree:

- `tubuild.py verify ov098/ArrowSignRight`: **11/11 MATCH**, objisolate clean, reloc-destinations clean → TEXT-VERIFIED; the 13 unlicensed symbols match the manifest's documented set exactly
- `linkcheck --baseline` + `linkcheck --partial`: **PARTIAL-LINK-VERIFIED** — TU `.text` range `0x02137be0..0x02138008` (1064 bytes) IDENTICAL, ROM sha256 identical to the manifest's record
- langmode ratchet PASS; `port_refcheck` 393 references resolve

Nothing blocking. Before/at merge:

1. **OnAttacked1 provenance comment is wrong about the tree** (`src_tu/actors/ArrowSignRight.cpp:194-200`): it claims `other`'s actorID "at offset 0xc has no public C++ name", but `include/fBase_c.h:94` declares `u16 actorID; /* 0x0c */` on `fBase_c`, inherited by `dActor_c`. Please correct the comment (or use `other.actorID` — same `ldrh [r0,#0xc]`; re-verify at promotion time).
2. Duplicated generator comment lines: "recovered: vtable identified…" appears twice back-to-back (lines 86–87 and 163–164). Trivial, byte-free.
3. Series-wide promotion-shape question: this TU's out-of-line dtor emits D0 *before* D1 while the ROM has D1 first (manifest honestly records `functions_occur_in_expected_order: PARTIAL [(0,1)]`). Invisible under per-function substitution; whole-range linking at promotion will want the in-class-body destructor form or documented ordering handling. Worth stating that plan durably rather than per-PR.

Housekeeping: the branch predates #1709, which rewired enroll/objisolate/rombuild for exactly this subsystem — worth a rebase + one gate re-run so the manifest entry lands against current tooling.

andrewboudreau:

PR #1719, opened 2026-08-23
Fourth TU reconstruction pilot: ov100/daObjPathLift_c (text-verified, partial-link-verified)

andrewboudreau:
## Summary
- Fourth TU reconstruction pilot (after PoleLift, ShipWater, ArrowSignRight): `ov100/daObjPathLift_c`, 8 functions, taken from shadow through text-verified and partial-link-verified.
- No forced vtable `+2` addend fix needed this time -- this class's destructor stays a declared-but-never-defined virtual (D1/D0 reproduce `PathLift`'s own out-of-line cleanup inline under the mangled names), so the TU never emits its own `_ZTV`/`_ZTI`/`_ZTS` and needed no `externalized_output` entries. The one EXTRA compiler symbol, `_ZN7Vector3D1Ev`, is the same policy-schema gap first documented on ArrowSignRight (a compiler-emitted duplicate with a real ROM home, fitting neither `compiler_only_output` nor `externalized_output`) -- left deliberately unlicensed and documented in the manifest notes.
- Reconciled against real project headers, per the standing "readable idiomatic C++, not extern-laden shadow structs" direction: `Render()`'s `Obj` virtual-dispatch shadow struct replaced with `mModel.Render(0)` (third repeat of the ShipWater/ArrowSignRight idiom); `Behavior()` and `InitResources()` converted from `extern "C"` free functions taking `char* c` into real `daObjPathLift_c::Behavior()`/`::InitResources()` methods, with raw `dBgActor_c*`/`PathLift*`/`dBgW*` casts replaced by direct calls through real inherited members.
- `mMeshCollider.SetFile()` deliberately stays `extern "C"`: its real signature takes a `Fix12<int> scale` with no converting constructor from an int literal, and materialising the call's fixed-point argument would cost stack traffic the ROM's own codegen does not have.
- `data_ov100_02148a54`/`data_ov100_02148a5c` retyped to the real `SharedFilePtr` (matching a sibling file's declaration); `decl_common.h` dropped from this TU entirely since it declared those two symbols at a less complete type, with the handful of other symbols it supplied restated locally.

## Verification
- `tools/tubuild.py verify`: 8/8 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED (re-confirmed after every real-member reconciliation swap, one at a time).
- `tools/tubuild.py linkcheck --partial`: 8/8 derived objects contribution-equivalent, TU range `0x02146d7c..0x02147328` IDENTICAL, module ov100 106/106, 0 new symbol-check errors vs baseline, scratch ROM byte-identical to stock (`sha256 d1506e90...`).
- `tools/rombuild.py -j16 --no-rom`: 106/106 module fidelity, unaffected (`src_tu/` is not wired into the real build).

## Test plan
- [x] `tools/tubuild.py verify ov100/daObjPathLift_c` -- 8/8 MATCH
- [x] `tools/tubuild.py linkcheck ov100/daObjPathLift_c --partial` -- PARTIAL-LINK-VERIFIED
- [x] `tools/rombuild.py -j16 --no-rom` -- 106/106


andrewboudreau:
**Review (independent local reproduction pass) — bytes verified; local renames requested**

- `tubuild.py verify ov100/daObjPathLift_c`: **8/8 MATCH**, objisolate clean, reloc-destinations clean → TEXT-VERIFIED (plus the single documented unlicensed `_ZN7Vector3D1Ev`)
- `linkcheck --partial` against a baseline control: **PARTIAL-LINK-VERIFIED** — TU `.text` range `0x02146d7c..0x02147328` (1452 bytes) IDENTICAL; ROM sha256 matches baseline/prior pilots
- langmode ratchet PASS; `port_refcheck` clean
- `demangle.py` spot checks agree with the shadow declarations, including the signed-short fifth param story in `Behavior()`'s comment

One request before/at merge:

1. **Register-named locals in `func_ov100_02146e70`**: `r5`, `r6`, and the decompiler branch-label `belse`. Carried verbatim from the legacy `.c`, so it's re-published debt rather than new debt — but a brand-new file is exactly where to pay it, and renaming a local is codegen-free/gate-invisible. Suggestions from reading the body: `belse` → `dropDist` (it's `(currentY − targetY) + 0x64000`, the fallback DropShadow Y), `r5`/`r6` → the abs-valued sin/scale temporaries they hold.

Worth confirming (not blocking): the hand-written extern-"C" D0/D1 pair is well justified here (a real out-of-line `~daObjPathLift_c()` would emit a call to `_ZN8PathLiftD2Ev`, which has no ROM home) — but does that stay the end state at promotion, or does promotion also want PathLift's dtor made inline-on-purpose so a real subclass dtor becomes byte-reachable? Worth a line somewhere durable either way.

Housekeeping: predates #1709 (enroll/objisolate/rombuild rewiring in this subsystem); consider rebase + gate re-run.

andrewboudreau:

PR #1722, opened 2026-08-23
Fifth TU reconstruction pilot: ov002/InvisiblePole (text-verified, partial-link-verified)

andrewboudreau:
## Summary
- Fifth TU reconstruction pilot (after PoleLift, ShipWater, ArrowSignRight, daObjPathLift_c): `ov002/InvisiblePole`, 8 functions, taken from shadow through text-verified and partial-link-verified.
- The legacy source was already unusually clean: five of the eight functions were already real C++ methods using real member access (`mClsn.Clear()`/`.Update()`, `&mClsn`), including a real out-of-line `InvisiblePole::~InvisiblePole(){}` (D1).
- Hit a new (to this series) mwcc internal compiler error: keeping that real D1 destructor alongside the legacy D0's hand-written mangled-name free function in the same TU crashes the compiler (`ELFgen.c:483`). Fix: delete the D0 body outright -- the real destructor makes the compiler emit D0 itself, byte-identical to the ROM. The forced `+2` vtable-store addend on `InvisiblePole_Spawn` was still needed (fourth confirmed occurrence).
- Populated `compiler_only_output` (the compiler-generated D2, deadstrip) and `externalized_output` (the six inherited `fBase_c`/`dBase_c`/`dActor_c` vague-linkage RTTI symbols, relocation shape independently re-read from this TU's own compiled object).
- Left this class's own `_ZTI13InvisiblePole`/`_ZTS13InvisiblePole` out of `data` and documented why: the ROM's RTTI embeds this class under a different, coined name (`daBar_c`, per the class's own header comment), so the cartridge's actual 7-character type-name string cannot hold the compiler's 13-character `"InvisiblePole"` string -- a genuine byte-length incompatibility, not a naming mismatch, deferred to Phase F same as other coined-name cases.

## Verification
- `tools/tubuild.py verify`: 8/8 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED.
- `tools/tubuild.py linkcheck --partial`: 8/8 derived objects contribution-equivalent, TU range `0x020b05d0..0x020b0748` IDENTICAL, module ov002 106/106, 0 new symbol-check errors vs baseline, scratch ROM byte-identical to stock (`sha256 d1506e90...`).
- `tools/rombuild.py -j16 --no-rom`: 106/106 module fidelity, unaffected.

## Test plan
- [x] `tools/tubuild.py verify ov002/InvisiblePole` -- 8/8 MATCH
- [x] `tools/tubuild.py linkcheck ov002/InvisiblePole --partial` -- PARTIAL-LINK-VERIFIED
- [x] `tools/rombuild.py -j16 --no-rom` -- 106/106


andrewboudreau:
**Review (independent local reproduction pass) — bytes verified; record the ICE durably**

- `tubuild.py verify ov002/InvisiblePole`: **8/8 MATCH**, objisolate clean, reloc-destinations clean → TEXT-VERIFIED; 10 unlicensed symbols = the manifest's documented FAIL-BY-DESIGN set
- `linkcheck --partial`: **PARTIAL-LINK-VERIFIED** — TU `.text` range `0x020b05d0..0x020b0748` (376 bytes) IDENTICAL; module fidelity 106/106
- langmode ratchet PASS; demangle spot checks (`dActor_c::SetRanges(Fix12<int>×4)`, `dCcAc_c::Init(…)`) agree with declarations and call sites

Cleanest pilot of the series so far — every method is already a real member, the two hand-spelled externs carry stated reasons, no register-named locals. Two asks:

1. **The D0/ICE workaround's comment says "see memory"** — that reads as a reference to something outside the repo. The fact itself is valuable and expensive to rediscover: a hand-written mangled D0 body plus a real out-of-line D1 in one TU = mwcc internal compiler error (ELFgen.c:483). Please put it in `notes/mwccarm-codegen.md` (or the pilot report) so it survives independent of chat history.
2. The `daBar_c` vs `InvisiblePole` RTTI string-length mismatch means this class can never fully promote without an alias mechanism or a rename decision — honestly deferred in the manifest, but worth an explicit line in the Phase F plan if it isn't there already.

Housekeeping: predates #1709; consider rebase + gate re-run.

andrewboudreau:
Restacked onto current `main` (was DIRTY on `config/tu_manifest.json`).

**Scope changed — please re-read before merging.** #1715 ("Build verified partitioned TUs in rombuild (M2)", landed 2026-08-23 15:52, ~5h after this branch's commit) already brought `src_tu/actors/InvisiblePole.cpp` and an `ov002/InvisiblePole` manifest entry onto main. This PR is therefore no longer "add a fifth pilot" — it is now an **upgrade of the entry main already has**:

- `src_tu/actors/InvisiblePole.cpp`: **+9 lines, comment only.** The ROM-ordinal-1 block documenting why there is no hand-written `_ZN13InvisiblePoleD0Ev` body (mwcc ICE at `ELFgen.c:483` when a hand-written mangled D0 shares a TU with a real out-of-line D1; the real `~InvisiblePole(){}` makes the compiler emit D0 itself and it still MATCHes). No code change — main's and this branch's sources are otherwise byte-identical.
- `config/tu_manifest.json`: the `ov002/InvisiblePole` entry is **replaced**, not added. Main's entry came from the older `tubuild.py partial` round; this one carries the stronger `tubuild.py linkcheck --partial` record (delink+lcf+whole-tree link+module byte compare+full ROM build), plus `compiler_only_output` (`_ZN13InvisiblePoleD2Ev` deadstrip) and `externalized_output` (the eight inherited vague-linkage RTTI records), neither of which main's entry has. 15 of main's 56 entries already use the `linkcheck --partial` schema, so this is a migration toward the current shape, not a divergence.
- Its `substitutedObjectPaths` still name the legacy per-function `src/_ZN13InvisiblePole*` files, which are all still present and still named by `config/arm9/overlays/ov002/delinks.txt` on main — the recorded link topology has not gone stale.

Conflict resolution was a three-way merge on `entries` keyed by `id`; nothing else in the manifest moved.

If the reviewer would rather not carry a whole-entry replacement, the alternative is to close this and re-land just the comment block — but the verification record is the substantive half.

PR #1723, opened 2026-08-23
Migrate two path lift methods to real C++ (batch 5)

andrewboudreau:
Migrates daObjPathLift_c::InitResources and daObjPathLift_c::Behavior from hand-spelled symbol bodies to real C++ methods over the recovered PathLift hierarchy. Uses named model, shadow, collision, position, and sound state while retaining only the verified Fix12 scalar ABI shims and explicit dBgCh_Gnd lifecycle. Validation: strict match and strict relocations at mwccarm 2004/b56; linkcheck VERIFIED with blind: 0 for both; eligibility unchanged at 11065/11187; full build 11059/11059 reproducing with 106/106 exact modules.


andrewboudreau:
**Review (independent local reproduction pass) — bytes verified; one deleted lore comment needs a durable home**

Reproduced locally on canonical `2004/b56`:

- `rombuild.py -j8 --no-rom`: **module fidelity 106/106 exact**, 11,059/11,059 reproducing, ROM-build analysis PASS
- Both changed functions verified individually with strict relocations: `_ZN15daObjPathLift_c13InitResourcesEv @0x021471e0` → **MATCH**, `_ZN15daObjPathLift_c8BehaviorEv @0x021470f4` → **MATCH**
- langmode ratchet PASS; every newly named field traced to a real declaration at the offset the replaced raw access used (`mPosX` 0x5c, `mModel` 0xd4, `mShadowModel` 0x450, …)

One request before merge:

1. **The deleted `Platform2`/`Platform3` phantom-class comment needs a durable home again.** Of the three paragraphs of mangling lore in the old Behavior.cpp: the Sound signedness note survives at its definition and the IsClsnInRange Fix12 note survives in its header — but the Platform2/Platform3 explanation ("phantom classes… names only exist after mangling… `_ZN10dBgActor_c19UpdateClsnPosAndRotEv` read as `_ZN9Platform2` + `19UpdateClsn…`; two different shadow structs grew out of one class that way") now survives **nowhere** in `src/`, `include/`, or `notes/symbol-name-provenance.md`. That trap already cost the tree two shadow structs and can't be regenerated from code — please restore a condensed version, ideally in `notes/symbol-name-provenance.md`, or as a one-liner at the `UpdateClsnPosAndRot()` call site.

Minor, non-blocking: the `(BMD_File *)Model::LoadFile(...)` cast is redundant (`void*` converts implicitly in C++); drop it whenever convenient, not worth churn by itself.

PR #1724, opened 2026-08-23
Sixth TU reconstruction pilot: ov002/CameraTag (text-verified, partial-link-verified)

andrewboudreau:
## Summary
- `src_tu/actors/CameraTag.cpp` reconstructs ov002/CameraTag (8 functions, 0x020b0748..0x020b07f8) from its 8 legacy one-function files, byte-verified against the ROM.
- `include/CameraTag.h` was extended to a real inheriting C++ class: `struct CameraTag : dActor_c` with a real `virtual ~CameraTag();` and the five real fBase_c-slot overrides, replacing the old flat `u8 pad_000[0xd4]` shell with no destructor or inheritance at all. The old header's claimed size story (0x108, a `dCcAc_c` at 0xd4) was InvisiblePole's layout, copied by factory adjacency — the same class-to-factory-by-name defect PR #1556 had to retract two findings for. `tools/opnew_sizes.py`'s own independent ROM literal (CameraTag_Spawn's `operator new(212 == 0xd4)`, `inline_store` confidence) says CameraTag is `dActor_c` (0xd0) plus one unknown 4-byte field, not a copy of a different class's layout.
- Two now-familiar forced fixes: deleting the hand-written mangled-name D0 free function (a real, already out-of-line D1 destructor coexisting with a hand-written D0 in one TU is an mwcc internal compiler error, `ELFgen.c:483` — third confirmed occurrence, same shape as InvisiblePole/#1722) and the `+2` vtable addend on the factory's own vtable store (fifth confirmed occurrence of this pattern across the pilot series).
- `config/tu_manifest.json`: new `ov002/CameraTag` entry, `text-verified`, with `compiler_only_output` (the D2 deadstrip) and `externalized_output` (six inherited base-chain RTTI objects, independently re-confirmed via direct ELF read of this TU's own compiled object). CameraTag's own `_ZTI9CameraTag`/`_ZTS9CameraTag` are deliberately left unlicensed: the ROM's RTTI name is `daCamTag_c` (10 chars) vs. the decomp's `CameraTag` (9 chars) — a different string length, not fixable by relocation or aliasing, deferred to Phase F/attribution work.

## Test plan
- [x] `tools/tubuild.py verify ov002/CameraTag` — 8/8 MATCH, objisolate clean, reloc-destinations clean → TEXT-VERIFIED
- [x] `tools/tubuild.py linkcheck ov002/CameraTag --partial` — 8/8 derived objects byte-identical, module + TU range reproduce, full ROM builds (same 7 pre-existing baseline errors, 0 NEW/0 resolved) → PARTIAL-LINK-VERIFIED
- [x] `tools/rombuild.py -j16 --no-rom` — 106/106 exact, 100% of compared bytes, unaffected


andrewboudreau:
**Review (independent local reproduction pass) — header edit verified byte-neutral; two comment fixes + one lore promotion**

- `rombuild.py -j8 --no-rom`: **module fidelity 106/106 exact** — the load-bearing check for this series' first shared-header edit: `affected_src.py include/CameraTag.h` reports exactly 6 consumers, all rebuilt, all still byte-exact
- `tubuild.py verify ov002/CameraTag`: **8/8 MATCH** → TEXT-VERIFIED; `linkcheck --partial`: **PARTIAL-LINK-VERIFIED** (TU `.text` range 176 bytes IDENTICAL, 0 new symbol-check errors vs baseline control)
- langmode ratchet PASS; `demangle.py` confirms the single real out-of-line `~CameraTag(){}` emitting both D1/D0

The −17 lines were examined directly: all are inside the header rewrite, and the deleted prose was *factually wrong* (dActor_c is 0xd0 per its compile-time guard, not 0xd4) — the replacement cites independent ROM evidence instead. That's a correction with a guard behind it, not tidying. Good change.

Before/at merge:

1. Copy-paste species error at `src_tu/actors/CameraTag.cpp:72` and `:87` — "the pole has no model…" / "The pole never moves and never thinks": this is a camera tag, not the pole (wording carried from InvisiblePole's stubs). Byte-free rewording.
2. This is now the third confirmed occurrence of the ELFgen.c:483 ICE (real D1 + hand-written D0 in one TU) and fifth of the `+2` vtable-addend fix, cited in PR bodies each time. Please promote both facts to `notes/mwccarm-codegen.md` or the pilot report so they stop living in commit messages.
3. With `daCamTag_c` (10 chars) vs `CameraTag` (9), the RTTI string-length mismatch blocking a class's own `_ZTI/_ZTS` licensing has two data points (after daBar_c/InvisiblePole). That's a pattern Phase F needs to design for rather than a curiosity — worth adding to the plan list.

Housekeeping: predates #1709; consider rebase + gate re-run.

andrewboudreau:

PR #1758, opened 2026-08-25
Land the dBgCh destructors, the collision-family D1s, and the ModelAnim family as real C++

andrewboudreau:
ModelAnim's D1 and D0, BlendModelAnim's D1 and D0, dBgCh's D2/D1/D0, and the dBgCh_Actr/dBgCh_Gnd/dBgCh_SphCrr D1s all move from hand-written .c to real C++.

**The dBgCh rename** is the load-bearing commit: `func_020354d0`, `func_020354e0` and `func_02035504` are dBgCh's D2, D0 and D1, and until they existed as named symbols the MI children's destructors had nothing to link against. include/dBgCh.h gains the family's inline `operator delete` so the D0 calls `Memory::operator_delete2` rather than the compiler's default `_ZdlPv`.

**The ModelAnim family** is the first MI destructor with a real body: BlendModelAnim conditionally frees `unk_6c` through `func_0203cbc0` (the 0xc-byte veneer the ROM uses) before the member teardown. ModelAnim's D1 is an empty body over the declared virtual destructor; the D0s are the same definitions bound to the deleting-destructor ranges.

**dBgCh_Lin stays .c** — its ROM body calls a member destructor at +0x64 before the base steps, and the empty body does not reproduce it.

**Attribution:** the renames moved credit off tangosdev because git recorded delete+add, not a rename; `attribution.json` pins both spellings for all three so the composite lineage holds.

**Gates:** every file passes `build_pin.verify` plain and strict; `rombuild -j16` → **106/106 exact, 11060/11060 reproducing**; `prepush_attribution` → **0 changed, 0 lost**.


andrewboudreau:
Superseded by #1762 — same tree, regenerated onto current `main`: 28 files / +210 −194 instead of 91 / +1983 −1500, and it merges clean.

The conflict here is not in the work. Eleven of this branch's sixteen commits are byte-identical patch-id duplicates of commits already on `main` (including its tip `ec472ac0b`, #1751), among them four "Refresh progress" bot commits — and those four are the *entire* reason `README.md` and `contributions.json` conflict. No commit of this branch's own work touches either file.

Nothing was force-pushed; this branch is intact at `59ea060b5` if anything needs recovering from it.

Closing in favour of #1762 when you're happy with it.

PR #1760, opened 2026-08-26
Merge seven one-function batches back into their translation units

andrewboudreau:
Seven TUs recovered from `src/`'s one-function files, each compiled as a single `.cpp` and byte-compared against the cartridge.

| TU | Members | Result |
|---|---|---|
| `ov023/Squasher` | 9 | 9/9 TEXT-VERIFIED |
| `ov036/daObjRcCarpet_c` | 8 | 8/8 TEXT-VERIFIED + **MODULE-LINK-VERIFIED** |
| `ov006/MgPicturePoker` | 3 | 3/3 TEXT-VERIFIED |
| `ov006/MgPairAGoneAndOn` | 2 | 2/2 TEXT-VERIFIED |
| `ov006/MgMushroomRoulette` | 2 | 2/2 TEXT-VERIFIED |
| `ov006/MgBingoBallSlotsShot` | 2 | 2/2 TEXT-VERIFIED |
| `ov006/MgWhichWiggler` | 2 | 2/2 TEXT-VERIFIED |

TEXT-VERIFIED is the full three-part line: byte MATCH + `objisolate` clean + reloc-destinations clean. `ov036/daObjRcCarpet_c` additionally reproduces its whole `.text` range `0x02112158..0x02112538` from one object with all 106 modules byte-exact.

**Every failure met along the way was a declaration-reconcile edit. None was a boundary error** — `tu_map`'s intervals held in all seven.

## The vptr store has a precondition, and both byte gates are blind to it

A merged TU stores its class's vtable by hand:

```c
p[0] = (int)(_ZTV8Squasher + 2);      // the TU DEFINES _ZTV      -> addend 8
p[0] = (int)_ZTV15daObjRcCarpet_c;    // UNDEF in the object      -> addend 0
p[0] = (int)_ZTV8Squasher + 8;        // always wrong -- an ADD instruction, not an addend
```

`+ 2` is correct **only when mwcc emits the vtable** — when the TU carries a real class definition and owns its key function, so the symbol addresses the vtable object's start, two Itanium preamble words below slot 0. Where the TU is still hand-written `extern int _ZTV<C>[];`, the object leaves the symbol `SHN_UNDEF` and `config/.../symbols.txt` already binds it to the slot array, so any addend lands past it.

I got this wrong on `daObjRcCarpet_c` first, and neither byte gate caught it:

- `match.compare` wildcards every relocated word, so `verify` printed an identical `8/8 MATCH` before and after the fix.
- `objisolate.plan()` **deliberately whitelists** an UNDEF `_ZTV` with addend exactly 8 (`tools/objisolate.py` ~L221) — the compiler generates precisely that for an inlined base destructor, and it cannot tell that from a hand-written mistake. `_ZTV10dBgActor_c` does it legitimately in this very file.

Only `tubuild.py linkcheck` caught it, as two literal-pool words 8 bytes high. **The check before writing `+ 2` is one line: is the symbol `SHN_UNDEF` in the `.o`?**

## `tools/tu_create.py`

Normalizes one shape `tubuild.py create` refuses: a function **definition** sitting inside an `extern "C" { }` block, which the splitter consumes as a declaration so `body_start` is never set. **64 of the 173 safe-pool TUs hit this.** The shim closes the block before the definition and gives that definition its own `extern "C"` prefix; assemble and manifest upsert delegate back to tubuild unchanged, so the shadow file and manifest schema stay the tool's own.

Documented limitation, and it is tubuild's not the shim's: a definition returning `struct X*` / `enum X` / `class X*` is misparsed the same way, because the splitter's first-word test sees `struct` in `_DECL_KEYWORDS`. Those need hand-assembly.

## Known, and deliberately not hidden

- **`ov023/Squasher` is TEXT-VERIFIED but not promotion-ready.** It emits 12 unlicensed symbols — `_ZTV8Squasher`, the `_ZTI`/`_ZTS` set, and `_ZN8SquasherD2Ev` — none claimed in the manifest. That is the correct consequence of a real C++ class owning its key function, but promotion stays refused until those data symbols are enrolled.
- **`check_src_tu_compiles` reports 3 failures**, in `src_tu/actors/Actor.cpp` and `src_tu/actors/HeaveHo.cpp`. Both are already on `main` and are untouched here.
- **`dsd check symbols --fail` is red**, on 7 errors. The `--baseline` control run with no TU substituted at all reproduces all 7, so 0 are attributable to this PR — but the check is not green and I am not reporting it as such.

## About the notes

The three `notes/` files carry a loud banner because they earn it: they were measured on a branch **212 commits behind `main`**, before `config/tu_manifest.json` was split into `config/tu_manifest.d/`, before the collision classes were renamed to the cartridge's own names (#1643), and before ~70 more `src_tu` TUs landed — two of which the queue still lists as candidates. On `main` the generator chain yields 522 TUs, not the 524 quoted.

The generated data they cite is **deliberately not committed**: it is derivable from `rtti_extract.py` → `rtti_vtables.py` → `tu_map.py`, and it was stale the moment `main` moved. Worth repeating that `tu_map.py` writes a self-consistent **wrong** map with exit 0 if that chain runs out of order — I reproduced that here by accident before linking `extracted/`.

The reasoning, the recipes and the failure modes are what is worth keeping; the numbers are not.

https://claude.ai/code/session_01AgdM9iZHjBQCmrKSSCMr5g


andrewboudreau:
## Reconciled against the real headers

Fair criticism that the merged TUs were "C++ with a bunch of externs" — they were. `create` emits a *shadow*: every callee hand-declared `extern "C"` under its mangled name, local `struct Base { virtual void v0(); ... }` stand-ins for real vtables, and in `daObjRcCarpet_c` **no include of the class's own header at all**, even though `include/daObjRcCarpet_c.h` already carried the full `daObjRcCarpet_c : PathLift` declaration with named members and all four methods. That byte-matches and buys nothing the `.c` files did not already have.

Two commits fix it. Hand-spelled mangled names in code: **37 → 17**, and every survivor now says why at its own line. All seven TUs still reproduce exactly, and `eligible.py` is unchanged at 11067/11188.

`Squasher` and `daObjRcCarpet_c` now have real `Behavior` / `InitResources` / `CleanupResources` / `Render` methods over named members, calling through `Model`, `ModelAnim`, `Animation`, `ShadowModel`, `dBgW`, `dBgW_Kc`, `dBgActor_c` and `Sound`. Every shadow struct is gone — `b->m(0)` turned out to be `Model::Render(const Vector3 *)`, primary-vtable slot 5.

### Two defects this surfaced

**A cross-overlay symbol collision.** `Squasher::InitResources` loaded its collision file through `&_ZN19RotatingPlatformWdwD1Ev` — **a function in ov029**. The intended symbol is `Squasher_ClsnFile`, **a bss variable in ov023**. Both sit at `0x02112080` because the two overlays load at the same address, so every byte gate was satisfied and the wrong name survived. It decodes correctly only for as long as those overlays keep colliding. Fixed.

**`AddVec3` and `Matrix4x3_FromRotationY` have no header.** Ten `src_tu/` files re-declare them locally with contradictory signatures — `int` vs `short` angle, `Vec3 *` vs `Vector3 *` vs `void *`. Filed, not fixed: a `decl_common.h` edit needs an `eligible.py` bracket of its own.

### A correction to the plan note

Shape 2 said a merged TU's vptr store "needs `+ 2` (≈88 Tier-1 TUs)". It needs `+2` **only when the object DEFINES that `_ZTV`**; an UNDEF one takes addend 0, and neither byte gate can tell them apart (`match.compare` wildcards the relocated word, `objisolate` whitelists UNDEF at addend 8). That over-generalisation was about to be applied to ~88 more TUs.

### What the remaining externs are

Mostly one wall, and it is worth recording: **by-value `Fix12<int>` at a CALL SITE**, not just at a definition. `Fix12<T>` is a constructor-less aggregate, so a literal argument needs a named temporary or a compound literal — and mwcc gives *either* one a stack slot the ROM does not have. Both spellings were measured against the bytes on two functions. The rest are `*_Spawn` factories, which hand-roll construction over raw storage: `_ZN8Particle10SysTrackerC1Ev(p + 0x471c)` has no real C++ spelling because there is no object there yet. Those files now say so instead of looking unreviewed.

### Gates

| TU | verify |
|---|---|
| ov006/MgBingoBallSlotsShot | 2/2 TEXT-VERIFIED |
| ov006/MgMushroomRoulette | 2/2 TEXT-VERIFIED |
| ov006/MgPairAGoneAndOn | 2/2 TEXT-VERIFIED |
| ov006/MgPicturePoker | 3/3 TEXT-VERIFIED |
| ov006/MgWhichWiggler | 2/2 TEXT-VERIFIED |
| ov023/Squasher | 9/9 TEXT-VERIFIED |
| ov036/daObjRcCarpet_c | 8/8 TEXT-VERIFIED + **MODULE-LINK-VERIFIED** (0 new link errors) |

All seven objisolate clean and reloc-destinations clean. `daObjRcCarpet_c` links byte-exact across its whole `.text` range.

### One thing I could not finish

`Squasher` still stops at TEXT-VERIFIED, and after this pass the reason is down to one symbol. `include/types.h` gives `Vector3` an inline in-class `~Vector3() {}`, so every TU with a `Vector3` local emits a vague-linkage `_ZN7Vector3D1Ev`. The licensing audit refuses to waive it because it has a configured ROM home — correctly, since that guard is what stops a policy hiding a genuine duplicate definition. But it does not model vague linkage, where having a ROM home *and* being emitted by every touching TU are both true and harmless. That needs a `disposition` beyond `deadstrip`; I did not attempt it, because the validator restores `tools/` from base so no PR can test its own tool change. Written up in `notes/plan-tu-merge-queue.md` §7.6.

The one policy that *is* valid was added — `_ZN8SquasherD2Ev`, compiler-generated, no ROM home, no inbound relocation. It was the blocker before `_ZN7Vector3D1Ev` became one.

https://claude.ai/code/session_01AgdM9iZHjBQCmrKSSCMr5g

PR #1794, opened 2026-08-26
Migrate the last four Tree symbols to real class form

andrewboudreau:
## What migrated

- `~Tree()` complete-object destructor (`_ZN4TreeD1Ev`)
- `~Tree()` deleting destructor (`_ZN4TreeD0Ev`)
- `Tree::InitResources()` (`_ZN4Tree13InitResourcesEv`) and `Tree::Behavior()` (`_ZN4Tree8BehaviorEv`) -- both leave hand-written `extern "C"` definitions for real class-form methods

These are the four files the header's #1719-era note called "no conversion benefit"; that note predates `mModel` becoming a real typed member array in the shared header.

## Why the compiler now owns the ABI

With `Tree : dActor_c` already real in the header (unchanged here), an empty `~Tree()` compiles to the cartridge's destructor pair byte-for-byte: vptr store, `__destroy_arr(mModel, 5, 0x50, Model::D1)`, `dActor_c::D2`; the D0 adds dActor_c's inline `operator delete` heap tail on its own. The two methods keep their exact bodies -- only the definition form changed. `InitResources` keeps the deliberately-wrong `dCcPos_c::Init` call spelling per notes/mwccarm-codegen.md 6az (true by-value Fix12<int> signature passes differently at the call site) and now addresses its model as `mModel[idx]` at the same 0xd4 + idx*0x50 offset the old shadow cast computed.

## Evidence

- Isolated bytes: 4/4 MATCH under mwccarm 2004/b56; destructors first attempt
- Strict relocations: `prepush_linkcheck` 4/4 VERIFIED, 0 blind words
- Shared header unchanged this time; no consumers affected
- Full link: `rombuild.py -j16 --no-rom` -> mismatching 0, module fidelity 106/106 exact, PASS
- port_refcheck 407 refs OK; check_dead_references 0 new; prepush_attribution 0 lost; langmode ratchet PASS vs origin/chaos-data baseline
- Candidates rejected by the full link: none


andrewboudreau:
Resolved against `main`. The Tree migration itself already landed in #1771 — `Tree` is a real class, `InitResources` uses `mPosX`/`mPosY`/`mPosZ` and `mModel[idx]`, and both destructors are an empty `~Tree()`. This branch predates that and its version of those two files would have reverted the named members to raw `(char*)this + 0x5c` casts, so every conflict resolved to main's side.

What survives is the prose, which is better than what main carries: `@symbol` banners on `InitResources` and `Behavior`, and destructor comments that say what the compiler actually emits (class vptr store, `__destroy_arr` over `mModel[5]`, the `dActor_c` base step, and for D0 the heap tail through `dActor_c`'s inline `operator delete`). Comment-only diff, no codegen change.

PR #1798, opened 2026-08-26
Update Heap paths in C++ migration guide

andrewboudreau:
Updates the Redis acquire/renew/release examples in the AI C++ migration guide after the merged Heap destructor migration changed both files from .c to .cpp.

Verification: tools/check_dead_references.py reports zero unresolved prose paths; git diff --check passes.


andrewboudreau:
Superseded by #1799, which merged the same Heap .c-to-.cpp guide-path updates into main.

PR #1816, opened 2026-08-26
Migrate BrickBlock to real C++ class form

andrewboudreau:
Reconstructs the coherent BrickBlock TU at ov002 0x020b415c-0x020b46a0: genuine dActor_c inheritance and 0xdc layout, real compiler-spelled lifecycle and PMF action methods, exact 31-slot vtable ownership, and corrected symbols/delink/rename metadata. The four tag factories truthfully remain C because the ROM calls the fBase_c actor allocator rather than global new. Validation: 14/14 linkcheck VERIFIED with blind 0, attribution preserved through five R100 moves, eligibility 11069/11189, and full ROM build 106/106 exact with zero source mismatches.


andrewboudreau:
Resolved against `main`. Only `symbols/actor_renames.tsv` conflicted, and it was an append collision — main gained nine `SnowmanBreath`/`SnowmanBreathParticle` rows from #1802 where this branch appends five `BrickBlock` helper rows. Resolved as the **union**; both row sets kept, no line dropped.

Checked the one substantive edit rather than taking it on faith. `_ZTV10BrickBlock` alloc `0x330` → `0xdc` is correct: the old value came from `BrickBlock_Spawn`, which is **not** this class's factory — it allocates 816 and stores `_ZTV13BigBrickBlock`. The four factories that actually store `_ZTV10BrickBlock` each allocate 220 = `0xdc` and call `dActor_c`'s constructor, which independently corroborates the header's `: dActor_c` base and the `0xdc` size assert:

```
GreenShellBlockTag_Spawn / MegaMushroomBlockTag_Spawn
OneUpMushroomBlockTag_Spawn / SilverStarBlockTag_Spawn
    p = _ZN7fBase_cnwEj(220);  _ZN8dActor_cC2Ev(p);  p[0] = (int)_ZTV10BrickBlock;
```

Four independent witnesses. Re-running CI on the merge.

PR #1842, opened 2026-08-27
Fix promoted source coverage in src_tu compile gate

andrewboudreau:
Teach the src_tu coverage pass to validate each manifest source at its repo-relative path, so promoted production TUs under src/ are not reported as missing. Also distinguish the legacy 0x78-byte raycast stack storage in Actor and HeaveHo from the real 0x84-byte dBgCh_Lin class. Verified with all 15 compile-gate regression tests, 88/88 TU compiles, 1,431 resolved TU symbol references, and 407/407 port references.

### DIFF
--- a/notes/ctor-migration.md
+++ b/notes/ctor-migration.md
@@ -397,7 +397,7 @@ destructor and derives from the non-polymorphic dBgPc. With no dynamic primary
 base, the compiler places dBgPi's own vptr at +0x00 and its dBgPc base at +0x04.
 The generated C2 sequence therefore constructs dBgPc at +0x04 before storing
 dBgPi's vptr. C1 now lives in `src/_ZN5dBgPiC1Ev.cpp` as real C++; the separately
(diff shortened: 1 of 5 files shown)


andrewboudreau:
Closing as superseded — #1838 already landed this exact fix on `main`:

- `dBgCh_Lin` shadow-struct rename in `src_tu/actors/Actor.cpp`/`HeaveHo.cpp` (named `dBgCh_LinPad` there vs `RaycastLineStorage` here — same fix)
- The `check_src_tu_compiles.py` promoted-TU coverage fix (ask the disk instead of the scan root — functionally identical to this PR's change)
- The `notes/ctor-migration.md` stale `.c`→`.cpp` reference (main's version is a superset)
- The corresponding test in `test_check_src_tu_compiles.py`

`main` already runs `check_src_tu_compiles.py` at 88/88 clean, so there's nothing left here to merge.

PR #1845, opened 2026-08-27
Generate View lifecycle and Render from real C++

andrewboudreau:
Stacked on #1844.

Reconstructs the RTTI/vtable-proven `View : dBase_c` base class and generates all three remaining `View` functions from real period C++:

- D1 and D0 come from an empty `View::~View()`; the compiler emits the complete `View -> dBase_c -> fBase_c` teardown and inherited heap deletion
- virtual `Render()` uses the typed `Matrix4x3` member at offset `0x50`
- introduces the shared, asserted `0x80`-byte `View` layout needed for the next `Camera` migration
- none of the three source files needs an `extern "C"` declaration

Verification:

- 3/3 focused and expanded link checks `VERIFIED`, each with `blind: 0`
- `View.h`: 1 measured field, 0 mismatches, exact `0x80` span
- 11,064/11,064 source-built functions reproducing; 106/106 modules exact
- 88/88 promoted translation units compile
- 407/407 `port/` references resolve
- 11,072 eligible functions; 44 unresolved references, with no regression


andrewboudreau:
The "ROM data verified from source fell from 2656 to 2655 symbol(s)" failure is a false positive, not a real regression. This ratchet counts raw per-object data-symbol *records*, not unique symbols — the same vtable/typeinfo symbol gets one record per object file that emits it (vague linkage), so consolidating duplicate D0/D1/D2 destructor files (this merge took `origin/main`'s already-landed versions for several classes) legitimately drops the raw record count without losing any verified data.

Verified locally with `tools/romdata_check.py` run independently against `origin/main` and this branch's merged tree, deduped to (module, symbol) → best verdict:

- unique symbols: 1190 on both sides
- missing from this branch: 0
- added in this branch: 0
- verdict regressed on any symbol: 0
- unique VERIFIED symbols: 387 (main) → 388 (this branch) — a gain, not a loss

Merging past this one check with that evidence attached rather than reworking the shared ratchet under this PR. Flagging for a follow-up: `romdata_check.summarize()` should count unique symbols at best verdict instead of raw per-object records (raw counts move with file topology, which is exactly what this migration changes on purpose).

PR #1859, opened 2026-08-27
Five more D0s become forcing helpers

andrewboudreau:
Five deleting destructors stop being hand-spelt mangled free functions and
become one `delete` expression each.

| class | module |
|---|---|
| `daObjFallBlock_c` | ov098 |
| `daObjMaruta_c` | ov080 |
| `daOts_c` | ov064 |
| `SignPost` | ov002 |
| `daDsnBase_c` | ov091 |

All five classes already define their destructor **inline in the class body**
— deliberately, so every descendant inlines it — and all five already have a
D1 file that forces the out-of-line copy the ROM's vtable slot 0 needs. Only
the D0 halves were still transcribing the vptr stores, the member teardown
chain and the tail call to the deallocator by hand.

A `delete` expression asks for the one thing the inline copy cannot satisfy,
so mwcc emits D0 and `objisolate` keeps it while dropping the forcing
function's own `.text` — the arrangement `notes/dtor-migration.md` section 3
describes and the D1 siblings next to each of these files already use.

**No header changed.** Every one of these headers already carried the inline
destructor; nothing needed adding. All five matched first try.

## Gates

- `rombuild -j16 --no-rom`: **106/106 exact**, 0 mismatching
- every symbol through `match.py --strict-relocs`
- `prepush_attribution`: `0 changed, 0 lost` — rewrite and `git mv` split, R100
- `check_dead_references`, `port_refcheck`, `layout_check`,
  `check_src_tu_compiles`: clean
- no header edit, so no eligibility bracket was needed

`ROM data from source` moves 2,583 → 2,568 verified and `differs` 97 → 95.
That is bookkeeping, not fidelity: the counts are per source file, and these
five files stop being wide TUs that pulled in several headers' worth of
vtable and RTTI rows. Module fidelity is 106/106 either way and the `differs`
side goes down, not up.


andrewboudreau:
Closing in favour of #1862, which is this PR reduced to the one file that was actually a migration (`daDsnBase_c`'s D0).

The other four files were already real C++ on `main`; this PR only swapped how they force the destructor, from "define the key function" to a `delete p;` helper. That is a re-styling, and the validator measured its cost: **ROM data verified from source fell from 2583 to 2568** -- fifteen symbols, three per file, each class's `_ZTV`/`_ZTI`/`_ZTS`. The key-function TU emits the ROM's real vtable and it verifies; the forcing helper emits a shadow one and it does not.

Local `rombuild` does not see this (it read 2,584 against main's 2,583), so the validator's ratchet was the only gate that caught it.

PR #1880, opened 2026-08-27
tu: let a promoted TU own the vtable its key function emits

andrewboudreau:
Stacked on #1878 (which is stacked on #1876).

## The wall

Multi-function isolation refused **any** non-text content, so a reconstructed TU could be promoted only while it emitted no data. Under the Itanium ABI that is exactly backwards: a class's vtable and RTTI go to the TU that defines its **key function** — which, for almost every actor here, is the destructor the TU exists to hold. The moment a TU is real enough to place its own structors, mwcc hands it `_ZTV`/`_ZTI`/`_ZTS`, and isolation threw the object out:

```
src/actors/ArrowLift.cpp: isolate: unlicensed content in text-only multi-symbol
object: section[5] .data size 0x8 defines ['_ZTI7fBase_c']; … 11 more
```

Measured over the 25 entries that were otherwise ready: **23 of them**, 9–13 data symbols each, **258 in total and every one `_ZTV`/`_ZTI`/`_ZTS`**. The manifest notes call this "Phase F" and defer it in five separate places.

## What changed

`plan_many` now does what `plan` has always done for singletons, generalised from one kept section to N:

- drop the non-text content
- **externalise its symbols**, so a weak gap-object import can never bind to our address 0 instead of the cartridge's carved-out address — the two steps are one operation, exactly as in the one-function path
- take the vtable address-point preamble back off the surviving relocations

Every RTTI relocation check `plan` makes, this makes. `_apply` was already written in terms of a *set* of kept sections; it just had no caller that passed more than one.

Two things stay refusals, deliberately:

| refused | why |
|---|---|
| an unlicensed `.text` body | the spanning delinks entry means dsd no longer fills that range from the ROM, so a stray function would land inside it |
| non-RTTI data | an ordinary owned definition — a table, a string, a global the source declares — is not carved out of the cartridge by name, so externalising it would invent an import nothing can satisfy. Better a sentence here than an undefined symbol at the link |

A genuinely text-only object still takes no surgery and comes back byte-identical.

## Promotions

`arm9/Scene` (21 functions) and `ov029/ArrowLift` (9). ArrowLift is a key-function TU with 11 data symbols and is the proof this works: it promotes **with the image unchanged**.

`ov002/Tree` was ready and is deliberately **not** here — its TU is still the un-migrated C form (`extern void* _ZTV4Tree[]` with hand-written vptr stores), so it imports the vtable rather than defining it, and consolidating would have cost `_ZTV4Tree`'s data verification. Caught by diffing the per-record dump; it is a real loss, not a counting artifact.

## `tools/tu_promote.py`

The mechanical half of a promotion: delinks entries in and out, the `src_tu` → `src` rename, the legacy `git rm`, the manifest status, and one `attribution.json` override per absorbed symbol. It compiles nothing — the proof of a promotion is `rombuild` reporting 106/106 afterwards, and running that once over a batch beats once per entry. `--dry-run` reports what each would touch.

## Verification

- `rombuild -j16 --no-rom`: **106/106 exact**, `mismatching: 0`, PASS
- ROM data **376 verified / 56 differ** — byte-for-byte the base's numbers
- `prepush_attribution`: **35 consolidated with credit intact, 0 changed, 0 lost**
- `layout_check` clean · `port_refcheck` 407/407 · `check_dead_references` no new (two prose references followed the absorbed `dScene_c` files to `src/actors/Scene.cpp`) · `check_src_tu_compiles` all
- 67 unit tests, including a new key-function-TU case asserting every externalised symbol is `SHN_UNDEF`, every dropped section is zero-sized, and the object never changes length

## What is left

The other 22 ready entries are blocked on one thing, and it is not tooling: their TUs define the destructor **out of line**, so mwcc emits `D2, D0, D1` while the ROM has `D1` then `D0`. The fix is the source form the class-form notes already describe — an inline in-class destructor plus a real instantiation — which also removes the homeless `D2` each of them carries. That is source work, one class at a time, and it is the same change that makes these files period-accurate C++.


andrewboudreau:
Triage: **leaving this open.** The working assumption was that #1993 (`tu/restack-inline-dtors`) supersedes this whole chain. It does not supersede this PR.

#1993 promotes 26 TUs. This PR promotes 2, and neither is in #1993's set:

- `src_tu/actors/ArrowLift.cpp` -> `src/actors/ArrowLift.cpp` (`config/tu_manifest.d/ov029/ArrowLift.json`)
- `src_tu/scene/Scene.cpp` -> `src/actors/Scene.cpp` (`config/tu_manifest.d/arm9/Scene.json`)

Both files are still under `src_tu/` on `origin/main` @ `7869b817d`, neither has a counterpart under `src/actors/` there, and no other open PR touches either — the only open PRs with `src_tu/` changes are #1995 (`ActorDerived`), #2002 (`daKrpa_c`, `daKpFr_c`) and #2005 (`dPathLiftActor_c`, `daObjRcCarpet_c`, both deliberately left shadow-only).

Unlike the four TUs #1993 documents as deliberately excluded, and unlike `PoleLift` (dropped in `d3cb31f4c` because it is `tools/test_tubuild.py`'s pilot fixture), there is no recorded rationale for `ArrowLift` and `Scene` being absent from #1993. They appear to have been lost in the regenerate, not rejected.

This branch is nonetheless stale: its base is `tu/pathlift-promote`, and the whole chain predates the `tools/` split that landed as #1990. **Do not rebase it.** Once #1993 lands, re-cut both promotions from `main` with `tools/tu_promote.py` and gate on a per-symbol `rombuild --data-json` diff against `main` — the headline object-record count drops on every consolidation and will not catch a lost vtable. If the re-cut lands, close this. Until then it is the only carrier.

andrewboudreau:
**Gatekeeper review — keep open. This is NOT superseded by #1993.**

I checked the coverage rather than assuming it. The #1880 → #1882 → #1884 chain promotes **66 TUs** against #1993's 26, and five promotions are uncovered by anything else in the queue: `ov029/ArrowLift` and `arm9/Scene` (#1880), `ov045/PoleLift` (#1882), `ov002/Platform` and `ov002/Warp` (#1884). Closing the chain would silently drop those five, so it stays open.

1. **Merge current `origin/main` (`bfd908b32`).** Two things moved under you: #2010 fixed a dead-reference red caused by a `src/actors/X.cpp` metavariable left in prose, and #2011 re-banked the two `daObjAbuku_c` paths that #1996 absorbed. If you resolved against an older main, redo it — do not carry a stale count forward.

2. **Resolve `config/converted-baseline.json` by REGENERATION, not key union.** Union is correct for `attribution.json` and wrong here: `score()` returns `None` for an unreadable path, so a single banked path that a promotion absorbed and deleted exits 1 and turns the ratchet red on main.

3. **A correction to guidance I issued earlier, which may be steering you wrong right now.** I told agents "the floor is 2565, never go below it." That was wrong. `tiers_ratchet.py` is a **set** ratchet: `--check` computes `missing = banked - current` and fails only when a path *leaves* the set. The printed count is informational — there is no floor to defend, and a promotion batch legitimately produces a much smaller count. The real gate is exactly two conditions:
   - every removed path is classified by the tool itself as `MOVED -- absorbed into <file> by TU promotion (<tu_id>)`, i.e. zero genuine backslides; and
   - every named absorbing file exists on the branch and is tracked by git.
   If both hold, regenerate with `--update --reason` and push. If either fails, stop and report rather than working around it.

4. **Confirm no delinks range is orphaned.** Every `promoted` entry from `python tools/tu_manifest.py list` must have an existing, tracked `promoted_source` and exactly one `complete` `.text` delinks entry naming it. This is the one failure mode the pipeline cannot self-detect: a `.text` range that stops being named `complete`, or names a path with no file there, is refilled by dsd with **retail cartridge bytes** — the ROM still byte-matches, module fidelity stays 106/106, romdata stays green, every gate passes. Only a drop in bytes-built-from-source reveals it. `tools/layout_check.py` is the dedicated gate.

Do not run `rombuild.py` or `eligible.py` — `build/` is shared across all worktrees and concurrent runs produce fake failures. Read the TU manifest only through `python tools/tu_manifest.py`, never `config/tu_manifest.d/**` directly.

If any premise above is wrong, say so on the PR with evidence rather than routing around it — I have been corrected three times today and each correction was right.

andrewboudreau:

```
```

andrewboudreau:
Closing as **superseded**. The body work here was real, but every mechanism that made it legal
has since landed on main by a different — and deliberately opposite — design.

**The blocker is `plan_many`.** At `17e1e7e36:tools/objisolate.py:342-349,366`:

```python
RTTI = ("_ZTV", "_ZTI", "_ZTS")
for index, sec in extra:
    if sec.name.startswith(".text"): continue
    for name in occupants_of(index):
        if not name.startswith(RTTI):
            owned.append(...)          # only NON-RTTI is refused
...
drop = [index for index, _sec in extra]   # every RTTI section dropped
```

So the title — *"let a promoted TU own the vtable its key function emits"* — is the inverse of
what the code does. The TU is permitted to promote **despite** emitting a vtable, by discarding
it uncompared. Nothing records which symbols were dropped, and nothing gates it
(`tools/rombuild.py:315-317` on this branch: *"Measurement only: check_object swallows its own
exceptions, and nothing here can fail the build"*).

**Main considered this and refused it.** The `plan_many` docstring that this PR *deletes* is in
your own merge-base:

> "silently dropping an unexpected helper or data section would create a hole inside a range dsd
> no longer fills from the ROM… Success therefore needs no ELF surgery: the exact input object
> is preserved."

Main solved the same goal strictly in **#1987, "Verify class data emitted by promoted C++ TUs"**:
keep `plan_many` refusing, and license each emitted record with a `deadstrip-data` row that
`romdata_check` compares to the cartridge word-for-word with relocations applied.

| | main | this branch |
|---|---|---|
| `deadstrip-data` disposition | yes | **no** (`DISPOSITIONS` has 2 entries) |
| verified `_ZTV`/`_ZTI`/`_ZTS` records | **50** across 5 TUs | **0** |
| `ov100/daObjPathLift_c.json` rows | 14 (13 `deadstrip-data`) | **1** |

**Every `tools/` change here is already on main**, mostly byte-identical: `tu_order_check.py`
(#1990, identical), `tu_promote.py` (#1990, main a superset), `rombuild.py` (#1876/#1977/#1987),
`romdata_check.py` (#1876), `objisolate.py` (#1977/#1987/#1989). Both commits unique to the
stack root landed too: `711db61b4` → `3b771dcca`, `e2691b2a9` → `0d4d3f898`. Main has since
promoted five TUs the strict way.

Also, for the record: the base of this PR (#1878, `tu/pathlift-promote`) is itself **closed**,
and the branch is 411 commits behind main with 11 conflicting files — including
`ov100/daObjPathLift_c.json`, where **all 13 verified rows are on main's side and the branch
side is empty**. A `--theirs` resolution there would destroy them.

Nothing to salvage from this one — `plan_many` is the only piece with no main equivalent, and
that is precisely because main rejected it. Re-cut promotions in small batches from current
main using the landed `deadstrip-data` policy, one `compiler_only_output` row per emitted
record with `canonical_module`/`canonical_address`, exactly as `ov100/daObjPathLift_c.json`
does today.

PR #1882, opened 2026-08-27
TU: inline the destructors, then promote twenty more TUs

andrewboudreau:
Stacked on #1880. Review that one first.

## The blocker was source form

#1880 said the remaining ready entries were blocked on one thing and it was
not tooling: their TUs define the destructor **out of line**, so mwcc emits
`D2, D0, D1` while the ROM has `D1` then `D0` and no `D2`. Production
isolation places an object's `.text` sections into the spanning delink in
emission order, so that disagreement decides whether the range links.

Moving the empty body into the class body is the whole fix:

```diff
-    virtual ~Seaweed();
+    virtual ~Seaweed() {}
```
```diff
-Seaweed::~Seaweed()
-{
-}
```

That is the form the cartridge's own header must have had. It emits `D1, D0`,
in that order, first — and emits no `D2`, so ten hand-written
`compiler-only_output` dead-strip rows for homeless `_ZN<C>D2Ev` symbols come
out of the manifest at the same time.

Twenty TUs took it, and `tools/tu_order_check.py` (new) says so for each:
one compile, and it answers both questions together — do the licensed
functions still reproduce their ROM bytes, and are they emitted in ROM order.

## What lands

- **145 hand-placed structor files deleted.** Twenty overlays each carried one
  file per mangled symbol under a `// @symbol` marker. The compiler places
  them now.
- **Five `deadstrip-duplicate` rows** for `_ZN7Vector3D1Ev`, written by
  `tu_order_check --write-policy`. It writes only the two shapes with a
  mechanical justification — a homeless D2, and a vague-linkage duplicate with
  exactly one configured home — and reports anything else for a human, because
  the reason column is the point.
- **`ChillBully+daIDonketu_c.cpp` → `ChillBully_daIDonketu_c.cpp`.** src_tu/
  spells a multi-class TU with `+`; nothing ever links those, so the convention
  was safe right up to promotion. dsd writes bare unquoted object names into
  the LCF and mwldarm reads `+` as punctuation:

  ```
  mwldarm.exe: File not found: ChillBully
  mwldarm.exe: Expecting: (
  ```

  which fails the whole ROM, not one overlay. `tu_promote` now refuses a
  `promoted_source` outside `[A-Za-z0-9_./-]`, so the next one costs a second.

## Gates

```
source-built functions: 11,065   reproducing: 11,065   mismatching: 0
```

ROM data went **up**, measured against this branch's base:

| | base | here |
|---|---|---|
| symbols verified | 376 | **378** |
| symbols differing | 56 | **54** |
| gap bytes verified | 29,816 | **30,064** |

`layout_check` clean · `port_refcheck` 400/400 · `check_dead_references` no new
· `check_src_tu_compiles` 88/88 · `prepush_attribution` **154 consolidated with
credit intact, 0 changed, 0 lost** · `pytest tools/test_{objisolate,rombuild,romdata_check,tubuild}.py` 96 passed.

## Two things a reviewer should know

**The middle commit does not build on its own.** Inlining the destructor makes
the header's inline body collide with the legacy out-of-line `_ZN<C>D1Ev.cpp`
still sitting in `src/`, and the promotion commit is what deletes those. The
two are one change split for rename detection (the twenty `git mv`s record
R100 that way); the branch head is what is gated. Do not bisect through
`c18849e8a` expecting a build.

**`tools/test_tubuild.py` was pinned end to end to `ov045/PoleLift`,** which
this branch promotes out of `src_tu/` entirely. Repointed at
`ov002/daObjAbuku_c` — same seven-function shape, still text-verified, still
7/7 contribution-equivalent, and blocked from promotion for an unrelated
reason (it owns delinked data), so it stays a valid subject. Its unlicensed
inventory count has now been three different numbers (15 when written, 12 once
measured, 6 for the new subject); the comment says to re-measure rather than
tune it.

## Not in this batch

`ov013/ClockPaintingPendulum` took the inline change and then emitted **no
destructor variants at all** — nothing in the TU odr-uses them. That is the
documented second condition (the TU must actually instantiate the class) and
is separate source work, so it is left out.


andrewboudreau:
Triage: **leaving this open, but it is the weakest case in the chain.**

This PR promotes 20 TUs. #1993 (`tu/restack-inline-dtors`) covers 19 of them:

`AmbientSoundEffects`, `ArrowSignRight`, `BasementWater`, `BlueCoinSwitch`, `CameraTag`, `ChillBully` (this branch spells the TU `ChillBully+daIDonketu_c`; `main` renamed it to `ChillBully.cpp` — same TU), `Exit`, `InvisiblePole`, `OneUpLogo`, `PowerStarCreate`, `Seaweed`, `ShipWater`, `SoundObject`, `SwitchPillar`, `VirtualDoor`, `WaterfallMist`, `daObjCannonShutter_c`, `daObjFl_Fall_Block_c`, `daObjKm3_Kuruma_c`.

The one it does not cover is `src_tu/actors/PoleLift.cpp` -> `src/actors/PoleLift.cpp` (`config/tu_manifest.d/ov045/PoleLift.json`), and that is a **deliberate** exclusion, not an oversight. #1993 commit `d3cb31f4c` removed it:

> PoleLift is `tools/test_tubuild.py`'s permanent pilot fixture -- four of its tests assert against `src_tu/actors/PoleLift.cpp` by path, and one asserts the literal `git mv src_tu/actors/PoleLift.cpp` the promote command prints. Promoting it consumes the fixture, and the PR validator restores all of `tools/` from base, so no PR can repair those tests in the same change that breaks them.

Note the second half of that: the same validator constraint that blocked #1993 also blocks **this** PR from promoting `PoleLift`. This branch cannot land its one unique promotion either.

Also note `include/PoleLift.h`: the inline `~PoleLift()` and the promotion are one atomic edit here. Un-promoting without reverting the header leaves `src/_ZN8PoleLiftD1Ev.cpp` failing to compile with `object 'PoleLift::~PoleLift()' redefined`.

Left open only because it is technically the sole carrier of the `PoleLift` promotion. **This is a valid candidate to close once #1993 lands** — `PoleLift` should be promoted in its own PR, after or together with a `tools/test_tubuild.py` change that repoints the four path assertions at a different fixture. Do not rebase this branch; regenerate.

andrewboudreau:
**Gatekeeper review — keep open, and the blocker on your one uncovered promotion has just been lifted.**

Status: this PR is not superseded by #1993. I checked the coverage rather than assuming it. The #1880 → #1882 → #1884 chain promotes 66 TUs against #1993's 26, and **five promotions are uncovered by anything else in the queue**: `ov029/ArrowLift` and `arm9/Scene` (#1880), `ov045/PoleLift` (#1882), `ov002/Platform` and `ov002/Warp` (#1884).

**The PoleLift finding.** #1993 deliberately dropped PoleLift in `d3cb31f4c`, on this reasoning: it is `tools/test_tubuild.py`'s permanent pilot fixture — four tests assert against `src_tu/actors/PoleLift.cpp` by path, and one asserts the literal `git mv src_tu/actors/PoleLift.cpp` that the promote command prints — and because the validator restored all of `tools/` from base, no PR could both consume the fixture and repair the tests. That made PoleLift structurally unpromotable.

**That is no longer true as of today.** `fa5aa1f65` ("Validation now runs the pull request's own tools/", #1994) reversed it. The workflow comment now states verbatim that *"the worker DOES run this branch's own `tools/`, deliberately, so that a pull request can exercise its own tool change"*, and grepping `.github/workflows/` for any `checkout`/`restore` of `tools` returns nothing — the base-restore step is gone from the repo entirely.

So a single PR can now promote PoleLift **and** update the four fixture assertions in the same change. Please verify that empirically before betting the branch on it — #1994's own diff is only the comment, so the executable half lives in the self-hosted worker outside this repo, and the title's "now" is a claim I could not confirm from the repo alone.

1. Rebase/merge current `origin/main` (`bfd908b32`). Two things moved under you: #2010 fixed a dead-reference red, and #2011 re-banked the two `daObjAbuku_c` paths that #1996 absorbed.
2. Resolve `config/converted-baseline.json` by **regeneration**, not key union. And note a correction to guidance I issued earlier: `tiers_ratchet.py` is a **set** ratchet — `--check` computes `missing = banked - current` and the printed count is informational. There is no count floor to hold, and a promotion batch legitimately lowers the count. The real gate is (a) every removed path classified `MOVED -- absorbed into <file> by TU promotion (<tu_id>)`, and (b) every named absorbing file exists and is tracked.
3. Confirm no delinks range is orphaned: every `promoted` manifest entry needs an existing tracked `promoted_source` and exactly one `complete` `.text` entry. This is the one failure mode the pipeline cannot self-detect — a range that stops being named `complete` gets refilled by dsd with retail cartridge bytes while the ROM still byte-matches and every gate stays green.

Do not run `rombuild.py` or `eligible.py` — `build/` is shared across worktrees and concurrent runs produce fake failures.

andrewboudreau:

```
```

andrewboudreau:
Closing — see #1880 for the full reasoning. This PR sits on #1880's rejected `plan_many`
mechanism (RTTI sections dropped uncompared), so it cannot land without it.

Two things specific to this PR:

**1. It reintroduces coined names main has already recovered.** Main superseded this TU in
**#1971** (`744ddb5b2 Recover typed factories and Bully virtual hooks`):

| | main `ov027/ChillBully.json` | this branch `ov027/ChillBully+daIDonketu_c.json` |
|---|---|---|
| id | `ov027/ChillBully` | `ov027/ChillBully+daIDonketu_c` |
| funcs | `…14UpdateRunStateEv`, `…16UpdateDeathStateEv` | `func_ov027_02111680`, `func_ov027_021116f0` |

Merging this resurrects two `func_ov027_*` names, puts a `+` back in a manifest id (which the
LCF cannot carry in a filename), and leaves **two** manifest files for one TU. That is a naming
regression, not a promotion.

**2. The `converted ratchet` failure is genuine and gets worse on refresh.** CI reports
`baseline 1958 current 2260 (+392 gained, -90 lost)`. All 90 backslides are `GONE` — banked
one-function `src/_ZN…cpp` files consolidated into the promoted TUs — with zero criterion
failures. Control: `origin/main` alone has **0** GONE against its own baseline, and simulating
the refresh via `git merge-tree` gives **101**, because main's larger baseline banks additional
one-function files this branch also deletes. The remedy exists on main and postdates this PR:
`tiers_ratchet.py --update` plus a line in `config/converted-backslide-exceptions.jsonl` (which
has grown 13 → 29 entries, the newest being exactly this case for #1996).

The destructor-inlining insight is sound and worth redoing from current main — the 11 removed
`_ZN…D2Ev` plain-`deadstrip` rows are benign for exactly the reason you give (`c18849e8a`
inlines those destructors so the D2 is no longer emitted).

PR #1884, opened 2026-08-27
TU: fix stale legacy_source extensions, promoting Platform and Warp

andrewboudreau:
Stacked on #1882. Review that one first.

`tu_promote` refused `ov002/Platform` and `ov002/Warp` with what looked like a
delinks problem:

```
src/_ZN4WarpD0Ev.c has 0 entries in config/arm9/overlays/ov002/delinks.txt
```

The file is on disk as `.cpp`. The manifest recorded the extension each
function's legacy source had when the entry was written, and `.c` → `.cpp`
migrations have landed on main since. **41 rows across 20 entries** point at a
name that no longer exists. Nothing but `tu_promote` reads them, which is why
this went unnoticed — and why it reported as a missing delinks entry rather
than a stale manifest.

Corrected only where the sibling extension is actually on disk. Rows whose
file is simply gone (the entry was already promoted, so its legacy sources
were deleted) are left alone.

That unblocks two:

- **ov002/Platform** — 11 functions, already emitted in ROM order; it only
  needed the `_ZN7Vector3D1Ev` duplicate row.
- **ov002/Warp** — 8 functions; needed #1882's inline-destructor change, then
  matched in ROM order.

19 more hand-placed structor files retired.

```
source-built functions: 11,065   reproducing: 11,065   mismatching: 0
ROM data: 378 verified, 206 partial, 54 differ   (unchanged)
```

`layout_check` clean · `port_refcheck` 397/397 · `check_dead_references` no new
· `check_src_tu_compiles` 88/88 · `pytest tools/test_tubuild.py` 36 passed ·
`prepush_attribution` 0 changed, 0 lost.


andrewboudreau:
Triage: **leaving this open.** The working assumption was that #1993 (`tu/restack-inline-dtors`) supersedes this whole chain. It supersedes the `legacy_source` extension fixes but not the promotions.

#1993 re-derived the 38 stale `legacy_source` refs across 19 manifest entries from current `main` (commit `a0d9abda0`), so that half of this PR is genuinely redundant.

The two promotions are not:

- `src_tu/actors/Platform.cpp` -> `src/actors/Platform.cpp` (`config/tu_manifest.d/ov002/Platform.json`)
- `src_tu/actors/Warp.cpp` -> `src/actors/Warp.cpp` (`config/tu_manifest.d/ov002/Warp.json`)

Both are still under `src_tu/` on `origin/main` @ `7869b817d` with no `src/actors/` counterpart, neither is in #1993's 26-TU promotion set, and no other open PR touches either. #1993's commit `a0d9abda0` mentions `Platform` and `Warp` only as `legacy_source` extension repoints — the promotions themselves were dropped in the regenerate with no recorded rationale, unlike the four TUs #1993 documents as deliberately excluded (`ov002/Tree`, `ov009/CastleWater`, `ov036/daObjRcCarpet_c`, `arm9/ActorDerived`).

This branch is stale: it sits on top of #1882 -> #1880 -> `tu/pathlift-promote`, all of which predate the `tools/` split landed as #1990. **Do not rebase.** After #1993 lands, re-cut just the `Platform` and `Warp` promotions from `main` with `tools/tu_promote.py`, gated on a per-symbol `rombuild --data-json` diff against `main`. Close this once that lands.

andrewboudreau:
**Gatekeeper review — keep open. This is NOT superseded by #1993.**

I checked the coverage rather than assuming it. The #1880 → #1882 → #1884 chain promotes **66 TUs** against #1993's 26, and five promotions are uncovered by anything else in the queue: `ov029/ArrowLift` and `arm9/Scene` (#1880), `ov045/PoleLift` (#1882), `ov002/Platform` and `ov002/Warp` (#1884). Closing the chain would silently drop those five, so it stays open.

1. **Merge current `origin/main` (`bfd908b32`).** Two things moved under you: #2010 fixed a dead-reference red caused by a `src/actors/X.cpp` metavariable left in prose, and #2011 re-banked the two `daObjAbuku_c` paths that #1996 absorbed. If you resolved against an older main, redo it — do not carry a stale count forward.

2. **Resolve `config/converted-baseline.json` by REGENERATION, not key union.** Union is correct for `attribution.json` and wrong here: `score()` returns `None` for an unreadable path, so a single banked path that a promotion absorbed and deleted exits 1 and turns the ratchet red on main.

3. **A correction to guidance I issued earlier, which may be steering you wrong right now.** I told agents "the floor is 2565, never go below it." That was wrong. `tiers_ratchet.py` is a **set** ratchet: `--check` computes `missing = banked - current` and fails only when a path *leaves* the set. The printed count is informational — there is no floor to defend, and a promotion batch legitimately produces a much smaller count. The real gate is exactly two conditions:
   - every removed path is classified by the tool itself as `MOVED -- absorbed into <file> by TU promotion (<tu_id>)`, i.e. zero genuine backslides; and
   - every named absorbing file exists on the branch and is tracked by git.
   If both hold, regenerate with `--update --reason` and push. If either fails, stop and report rather than working around it.

4. **Confirm no delinks range is orphaned.** Every `promoted` entry from `python tools/tu_manifest.py list` must have an existing, tracked `promoted_source` and exactly one `complete` `.text` delinks entry naming it. This is the one failure mode the pipeline cannot self-detect: a `.text` range that stops being named `complete`, or names a path with no file there, is refilled by dsd with **retail cartridge bytes** — the ROM still byte-matches, module fidelity stays 106/106, romdata stays green, every gate passes. Only a drop in bytes-built-from-source reveals it. `tools/layout_check.py` is the dedicated gate.

Do not run `rombuild.py` or `eligible.py` — `build/` is shared across all worktrees and concurrent runs produce fake failures. Read the TU manifest only through `python tools/tu_manifest.py`, never `config/tu_manifest.d/**` directly.

If any premise above is wrong, say so on the PR with evidence rather than routing around it — I have been corrected three times today and each correction was right.

andrewboudreau:

```
```

andrewboudreau:
Closing — see #1880 for the full reasoning. This sits on #1880/#1882 and inherits the rejected
`plan_many` mechanism, 411 commits of drift, and a `converted ratchet` failure that grows to
**111** GONE entries once refreshed against main.

**But the stale-`legacy_source` fix in here is still needed, and it is the one clean win in the
stack.** I re-derived it against current main: **38 rows across 20 manifest entries still point
at a `.c` path whose `.cpp` sibling is on disk** —

`arm9/Actor.json` (×7), `arm9/ActorDerived.json`, `ov002/Platform.json` (×2), `ov002/Warp.json`,
`ov009/{CastleWater,Flag,MetalNet}`, `ov010/{LightBeam,PeachPainting}`, `ov014/ShutterBob`,
`ov019/{IceSlideManager,RacingPenguin}`, `ov020/HauntedChair`,
`ov029/{CageLift,RotatingPlatformWdw,SwitchActivatedPlank,WaterDiamond}`, `ov062/KoopaFlag`,
`ov077/Spiny`.

Each one blocks `tu_promote` on that entry. The change is manifest-JSON only and touches no
tooling, so it is zero-risk and independently landable.

**Please re-open it as a fresh PR cut from current main — do not cherry-pick `13bf9222f`.** That
commit's row set was computed against a 411-commit-old tree, so it will both miss rows main has
since added and rewrite rows main has since fixed. Re-derive from the current manifest instead.
I have the 38-row list above if it helps.

PR #1885, opened 2026-08-27
Organize src by domain and enable verified C++ production TUs

andrewboudreau:
## What changed

- Organizes all 11,360 production source files into the candidate-2 domain tree: runtime, game, ui, minigames, and address-banded unnamed modules.
- Leaves zero C/C++ files in the flat src root; the largest leaf directory is 452 files.
- Adds fail-closed production support for canonical C++ translation units that compile once and provide exact compiler-generated function contributions.
- Enables 57 canonical TUs covering 493 functions: 3 partitioned TUs / 17 functions and 54 derived-text TUs / 476 functions.
- Preserves per-symbol contributor credit through path#symbol attribution overrides.

## Verification

- Full production ROM: PASS
- Source-built functions: 11,065; mismatching: 0
- Source-built code: 2,062,316 / 2,211,124 bytes (93.27%)
- Module fidelity: 106 / 106 exact
- Stock ROM SHA-256: d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
- Exhaustive relocation sweep: 11,286 checked, 11,148 verified, 138 known warnings, 0 blocking
- Focused tests: 138 passed
- Reconstructed TU compile gate: 88 / 88
- port references: 407 / 407
- Attribution: 11,341 tracked, 0 changed, 0 lost
- Eligibility: 11,072 / 11,192; no new unresolved references
- Layout, duplicate-source, compiler-pin, TU-state, and compatibility gates pass

## Boundary

Only TUs with exact text/relocation proof and explicit compiler-output ownership enter default production. Reconstructed candidates with incomplete data, RTTI, vtable, or boundary ownership remain one-function production sources until they can pass the same gate.


andrewboudreau:
Closing. This PR renames 11,360 files and modifies 405 more (11,770 changed paths total). The validator refused it outright:

> refusing to auto-validate 11618 source/build-data files (cap 2500) -- unusually large, please review by hand

so **no byte gate has ever run against this branch**. There is no evidence that the `src/` reorganisation reproduces the ROM, and a mass-move of this size is exactly the shape that silently re-keys `attribution.json` overrides (a `.c` -> `.cpp` or directory rename drops the override key) and breaks `srcpath.py` consumers. It cannot be reviewed by hand at 11,770 paths, and it cannot be validated automatically. It is unmergeable as a single change.

The `src/` layout question is separately already decided by #1055; #1056 and #1491 died the same way.

**Harvest before this branch is forgotten.** Everything below is measured at tip `ede758b72ed5b16f2ab4c58d7ee828654b27944f` against `origin/main` @ `7869b817d`, and none of it exists on `main` or in any other open PR:

New files:

| path | lines | what it is |
|---|---|---|
| `tools/migrate_src_domains.py` | 338 | the rename-only domain-bucket mover; dry-run by default, rewrites literal `src/...c(pp)` refs in tracked text on `--apply` |
| `tools/promote_derived_tus.py` | 310 | structural TU promotion that keeps the one-function sources as comparison oracles instead of deleting them, driven off `config/tu_manifest.d` + `config/production-tus.json` |
| `tools/test_migrate_src_domains.py` | 39 | tests for the above |
| `config/production-tus.json` | 65 | registry splitting `partitioned_tus` from `derived_text_tus` |

Modified files carrying substantive new logic:

| path | delta vs current `main` |
|---|---|
| `tools/tu_production.py` | +207 / -48 |
| `tools/srcpath.py` | +197 / -4 |
| `tools/cpp_tu_state.py` | +89 / -6 |
| `tools/test_tu_production.py` | +65 |
| `tools/test_srcpath.py` | +63 / -16 |
| `tools/langmode_audit.py` | +40 / -2 |
| `tools/test_cpp_tu_state.py` | +36 |
| `tools/prepush_attribution.py` | +21 / -1 |

`notes/src-domain-buckets.md` from this branch is already on `main`; nothing to harvest there.

The `srcpath.py` / `tu_production.py` / `cpp_tu_state.py` deltas are entangled with the domain tree and will not lift cleanly on their own; the two new scripts and `config/production-tus.json` will.

**Follow-up should not re-cut the move.** Land `tools/promote_derived_tus.py` + `config/production-tus.json` alone first, as a tools-only PR with no `src/` paths in it, so the validator can actually run. `tools/migrate_src_domains.py` can land as an unwired script. Only then, if the layout change is still wanted, drive it one domain bucket at a time so each slice stays under the 2,500-file validator cap and each gets its own `rombuild` 106/106 and `prepush_attribution.py` 0-changed/0-lost. Do not delete `layout/domain-cpp-main` until the two scripts are landed.

PR #1936, opened 2026-08-28
Use compiler-owned touch-icon lifetime

andrewboudreau:
Models the ROM-proven dThIcon_c Behavior/Render interface, represents dScMgSlot1_c::betIcon_c as a typed nested member, and lets the compiler emit the exact derived/base teardown. Moves the four icon methods to their native C++ symbols while preserving factory ABI where CodeWarrior cannot reproduce the allocation spelling. Verified 12/12 direct matches with linkcheck VERIFIED blind:0; shared-header fanout 1281 checked with zero blocking failures; full ROM 11072/11072 reproducing and 106/106 exact modules; refs, port, TU, name, attribution, and dead-reference gates pass.


andrewboudreau:

PR #1937, opened 2026-08-28
Use compiler-owned entry icon lifetime

andrewboudreau:
Types the RTTI-proven dScEntry_c::icon_c[9] member and lets mwccarm emit its reverse-order destruction. Defines the separate RTTI-proven graph callback in the real header, while leaving the unrelated unknown 4 x 0x2c array explicit. Converts both entry factories to C++ language mode and native lifecycle/vtable symbols. Verified: 14 direct functions byte-exact with linkcheck blind 0; 1304/1309 affected consumers verified with only 5 pre-existing warnings; 11072/11072 source-built functions reproduce; 106/106 modules exact; port, attribution, TU, offset, reference, and duplicate-source gates pass.


andrewboudreau:

PR #1938, opened 2026-08-28
Use compiler-owned jump-player lifetime

andrewboudreau:
Recovers the RTTI-proven dMgJump3DMario_c : dMg3DHeyhoObjAdapter_c hierarchy and replaces the three raw lifecycle helpers with native C++ constructors/destructor.

Both jump scenes now declare their three-player arrays with the real type, so mwccarm emits the array construction and reverse teardown. Still-unidentified arrays remain explicit to preserve the ROM order without invented shadow types.

Verified:
- 9/9 directly affected functions byte-identical and linkcheck VERIFIED with blind: 0
- 1,165 affected consumers checked; 1,160 verified, five pre-existing warnings, zero blockers
- full ROM: 11,072/11,072 source functions reproduce; 106/106 modules exact
- 88/88 production translation units compile
- header offsets exact; 405 port references resolve; no duplicate sources
- attribution preserved for all three recovered symbol names
- eligibility 11,081/11,196; no new unresolved references

No dashboard artifacts, CLAIMS.md, or match-attempt rows are included.


andrewboudreau:

PR #1939, opened 2026-08-28
Recover minigame object C++ surfaces

andrewboudreau:
Completes the RTTI-owned virtual surface of dMgJump3DMario_c and recovers two nested Psyche Out lifecycle objects as native C++.

The jump-player class now owns all three vtable slots as real member definitions. The Psyche Out leaf types let mwccarm synthesize the exact three-ModelAnim/three-MaterialChanger construction and reverse teardown, then the exact owning ModelAnim/TextureSequence/nested-set lifecycle. The dMg3DEsp names are explicitly marked inferred; object boundaries, offsets, counts, and lifecycle order are compiler-proven.

The scene itself still uses the recovered C1/D1 symbols explicitly; automatic scene-wide ownership is intentionally left for the next separately verified slice.

Verified:
- 10/10 directly affected functions byte-identical and linkcheck VERIFIED with blind: 0
- 31/31 affected consumers verified, zero warnings or blockers
- full ROM: 11,072/11,072 source functions reproduce; 106/106 modules exact
- 88/88 production translation units compile
- 405 port references resolve; no duplicate sources
- all seven source renames preserve original contributor credit
- eligibility 11,081/11,196; no new unresolved references

No dashboard artifacts, CLAIMS.md, or match-attempt rows are included.


andrewboudreau:

PR #1940, opened 2026-08-28
Recover Psyche Out model methods as native C++

andrewboudreau:
Converts the four inferred dMg3DEspAnimSet_c methods and five inferred dMg3DEspModel_c methods into real member definitions, types the model pointer-to-member state, and updates all four scene callers. The class and role-based method spellings remain explicitly marked inferred; layouts, call edges, and generated code are exact. Verified 13 direct matches/linkchecks with blind 0, 20/20 affected-header consumers, 11,072/11,072 ROM functions, and 106/106 modules.


andrewboudreau:

PR #1941, opened 2026-08-28
Recover collision destructor thunks from real C++ classes

andrewboudreau:
Replaces ten hand-written C or synthetic Base1/Base2/Derived lifecycle artifacts with genuine dBgCh_Gnd, dBgCh_Lin, and dBgCh_SphCrr destructor definitions. mwcc now emits both deleting destructors and the -0x10/-0x38 MI adjustment thunks from the real RTTI-backed hierarchies. Verified 10 direct matches/linkchecks with blind 0, 1,083 affected consumers with zero blockers, eligibility +8 to 11,089, 11,080/11,080 linked ROM functions, and 106/106 exact modules.


andrewboudreau:

PR #1966, opened 2026-08-29
Promote actor/scene debug-name table, fix CannonHatch naming, advance smartball InitResources

andrewboudreau:
## Summary
- Harvests the EAD debug-name string table (`ACTOR_DEBUG_NAMES`, immediately after `ACTOR_SPAWN_TABLE`) -- 391 Nintendo-internal names for every scene/actor, captured in `symbols/actor_debug_names.tsv`. Renames the table and its accessor (`func_020233d4` -> `GetActorDebugName`, used by `ShowCrashScreen` to print the offending actor on a crash).
- Fixes a real naming bug the debug-name table surfaced: `CannonHatch` should be `CannonShutter` (both its own RTTI `daObjCannonShutter_c` and the debug string `CANNON_SHUTTER` agree, contradicting the imported label). Renamed end-to-end and re-verified byte-match on the renamed methods.
- Bumps `config/rom-name-glossary.json` confidence where the debug-name table corroborates an existing gloss (`Kumo`), adds a new high-confidence term (`Mip`), and adds 13 RTTI-resolved new-ground entries to `symbols/overlay_actors.md`.
- Advances the never-decompiled `dScMgSmartball_c::InitResources` near-miss: found the real blocker was a self-cancelling `BLEND` macro statement getting CSE'd away (~50 missing instructions), not register pressure as first suspected. Closed the gap from 999 mismatched words to 1 (permuter search still narrowing it). Banked in `nearmiss/db.jsonl` with a corrected floor annotation -- the earlier "only whole-function O0 reproduces the redundant AND" claim was disproven.

## Test plan
- [x] Re-verified `CannonShutter` byte-match after rename (`D1Ev`, `InitResources`, `Spawn` all still `MATCH`)
- [x] Confirmed no leftover `CannonHatch` references anywhere in the tree
- [x] Confirmed `config/rom-name-glossary.json` is valid JSON after edits
- [x] Cross-checked every glossary/overlay_actors.md addition against real RTTI symbols before adding (rejected one incorrect suggestion -- `Guragura`/`Guruguru` -- after confirming they're different, unrelated classes)


andrewboudreau:
Superseded by #1972, rebuilt cleanly on current main. The replacement preserves the ROM-proven daObjCannonShutter_c identity, fixes delinks for the GetActorDebugName move, keeps attribution intact, and has passed the private ROM validator.

PR #1975, opened 2026-08-30
Fix `.md` relative paths syntax error in actor-naming documentation

Alberto12345678999:


andrewboudreau (commented):
what do you think about keeping the text without the `../` but the link does include it, I think that increases the readability

notes/actor-naming.md
```diff
@@ -1,6 +1,6 @@
 # Actor class naming from the ROM's spawn table
 
-How **2,241** symbols across **68** modules got real names [tools/actor_names.py](tools/actor_names.py),
+How **2,241** symbols across **68** modules got real names [../tools/actor_names.py](../tools/actor_names.py),
```
andrewboudreau:
```suggestion
How **2,241** symbols across **68** modules got real names [tools/actor_names.py](../tools/actor_names.py),
```

Alberto12345678999: Sounds good like that!

andrewboudreau:
there are afew other places to update as well, check it out

Alberto12345678999:
> there are afew other places to update as well, check it out

Thanks for the new (and old that I didn't go through) cases, i updated them as well!

andrewboudreau (approved):

PR #1914, opened 2026-08-28
TU: promote forty-two reconstructed translation units

andrewboudreau:
Stacked on #1884. Three commits, **42 translation units promoted** (manifest: 26 promoted → 68; 48 text-verified → 20).

`python tools/rombuild.py -j16 --no-rom` → **106/106 exact, 11,065 functions reproducing, 0 mismatching.**

## The commits

**`70d171eca` — promote fourteen.** Fourteen already-clean TUs become canonical, absorbing 189 per-function sources. `tools/tu_promote.py` gains a two-line `mkdir` so a `promoted_source` naming a directory `src/` does not have yet (`src/stage/` was the first) does not fail the `git mv`. The false "SHADOW … NOT ENROLLED … contributes nothing" banners on every promoted file are corrected.

**`7fd0c0d48` — promote twenty-five more, destructor bodies moved into the class.** `tools/tu_order_check.py` showed that "text-verified" is not "promotable": 34 entries emitted `D2, D0, D1` where the ROM has `D1` then `D0` and no `D2`. All 34 shared one cause — an out-of-line empty destructor body. Moving it into the class (the form #1882 established as period-accurate) fixed the order on all 34.

**`c3bb8e2f4` — the three `+`-named TUs, plus corrections.** See below.

## Two findings worth carrying forward

**Multi-class TUs need every destructor inlined, and then the include order decides their relative order.** Once both `EnemySpawner`'s and `EnemySwitchTag`'s destructors were inline, the TU still emitted EnemySpawner's pair first where the ROM has EnemySwitchTag's. An inline destructor is emitted where the compiler first sees it, so swapping `#include "EnemySwitchTag.h"` ahead of `EnemySpawner.h` moved the whole group and the TU went to ALL MATCH. Single-class TUs never exposed this lever.

**For eight TUs the out-of-line destructor IS the key function**, the thing forcing the TU-owned `_ZTV`. Inlining it makes the TU emit no destructors *and* no vtable, which is #1882's documented second condition seen from the other side. Flag, LightBeam, Trap, ClockPaintingPendulum, IceSlideManager, RacingPenguin, RotatingPlatformWdw and KoopaFlag stay `text-verified` pending a form that keeps the vtable emitted *and* gives ROM order. `ov014/ChainChomp` is held back for a different, genuine reason: a surviving `.text` section references the compiler-only `_ZN7Vector3D1Ev`, so that duplicate cannot be dropped.

## `+` in a filename

[...]

PR #1985, opened 2026-08-30
File WingFeather TU sources under game actors

andrewboudreau:
## Summary

- file the complete eight-function WingFeather legacy TU cohort under `src/game/actors/WingFeather/`
- keep `func_ov002_020b2c44` and `WingFeather_Spawn` unchanged; the folder records class/TU affiliation without claiming either is a member method
- retarget enrollment, TU-manifest, attribution, baseline, and generated source-path references
- preserve every source blob exactly; all eight moves are 100% renames

## Validation

- `python tools/prepush_linkcheck.py --range origin/main..HEAD`: 8/8 VERIFIED
- `python tools/prepush_attribution.py --base origin/main --head HEAD`: all eight credits intact
- `python tools/port_refcheck.py`: 405 references resolve
- WingFeather shadow TU compiles; 114 focused tests pass
- `python tools/rombuild.py -j 16 --no-rom`: 106/106 modules exact, 11,085 reproducing functions, 0 mismatches

## Coordination

Based on current `main` at `85345d91e`, including merged #1976 and its rewritten `func_ov002_020b2c44`. This is rename-only: no TU promotion or source rewrite. The WingFeather slice in #1914 should be restacked after this layout pilot.


andrewboudreau:
## Independent verification — content is sound, but this is **not mergeable as-is**

Verified on the PR head (`e179239`) in a clean worktree cut from `origin/layout/wingfeather-tu-folder`. Content checks all pass; there is one **real merge conflict** that must be hand-resolved before this can land, so I have left it in draft.

### Verified

**1. 8 moves, all R100, zero content change** — confirmed via `git diff --find-renames --name-status`. No commit mixes a rewrite with a move, so nothing needs splitting.

**2. Target directory matches policy** — `notes/src-domain-buckets.md` sanctions `src/game/actors/<Class>/` as the default landing spot for a spawn-table actor. WingFeather passes the actor test: `symbols/overlay_actors.md` (derived from `ACTOR_SPAWN_TABLE @ 0x02090864`) lists **`WING_FEATHER(345)` under ov002**. `symbols/actor_renames.tsv` independently corroborates with a `spawnfunc` (`WingFeather_Spawn` @ `0x020b3248`) and a `spawninfo` (`WingFeather_SpawnInfo` @ `0x02108884`). Correct bucket.

**3. Delinks integrity — clean (the check that matters most here)**
- All 8 ov002 entries retargeted, and **all 8 are still marked `complete`**.
- `.text` ranges byte-identical between base and head: **1502 ranges on each side, zero diff**. `complete` marker count 1514 on both. No range lost, none silently demoted — so nothing gets refilled from the cartridge.
- Tree-wide: every path named in every `config/**/delinks.txt` on this branch resolves to a real file. (The single tree-wide dangling entry, `src_tu/actors/Trap.cpp` in `config_tu/arm9/overlays/ov010/delinks.txt`, is pre-existing on `main` and lives in the experimental `config_tu/` root, not the live `config/` one.)
- `tools/layout_check.py` — the dedicated gate for exactly this failure mode — reports **L1 (delinks names a path with no file there) = 0**, plus L2 = 0 and L4 = 0 (class not split across two directories). `layout-check: clean`.

**4. No `+` in any new path** — all 8 new paths are clean, so no LCF hazard.

**5. Gates**
- `check_dead_references.py` — no new dead references
- `port_refcheck.py` — **405 checked, all resolve** (confirms the claim)
- `tiers_ratchet.py --check` — passes on the branch as-is
- `prepush_attribution.py` — **8 moved with credit intact, 0 changed**

`config/rombuild-versions.txt` carries no WingFeather pin, so the version-pin rename hazard does not apply.

### Blocker — genuine conflict in `config/converted-baseline.json`

GitHub reports `CONFLICTING`/`DIRTY`, and this reproduces on a local merge. It is **not** the known union-merge-driver false positive — `.gitattributes` sets no merge driver for this file.

The branch is 15 commits behind `main`, and in the meantime `main` added **two more** WingFeather files to the converted baseline at their **old** paths:

```
<<<<<<< HEAD
    "src/game/actors/WingFeather/_ZN11WingFeatherD1Ev.cpp",
=======
    "src/_ZN11WingFeather6RenderEv.cpp",
    "src/_ZN11WingFeatherD0Ev.cpp",
    "src/_ZN11WingFeatherD1Ev.cpp",
>>>>>>> origin/main
```

This must be resolved by **key union at the new paths** — resolving either side wholesale is silently destructive:
- `--ours` drops `6RenderEv` and `D0Ev` from the CONVERTED tier entirely (the floor quietly loses two files).
- `--theirs` leaves two baseline entries pointing at paths that no longer exist.

Correct resolution:

```
    "src/game/actors/WingFeather/_ZN11WingFeather6RenderEv.cpp",
    "src/game/actors/WingFeather/_ZN11WingFeatherD0Ev.cpp",
    "src/game/actors/WingFeather/_ZN11WingFeatherD1Ev.cpp",
```

I applied exactly that locally and re-ran everything on the merged result: `layout-check: clean` (L1/L2/L4 all 0), `port-refcheck 405/405`, no new dead references, and attribution improves to **8 moved with credit intact, 0 changed, 0 lost**. Every other surface (`delinks.txt`, `config/tu_manifest.d/ov002/WingFeather.json`, `attribution.json`, `notes/actor-leaf-provenance.md`) auto-merges correctly with all 8 files at the new paths. So the fix is a three-line hand resolution and nothing else.

Note for whoever lands it: `tiers_ratchet --check` fails after merging `main` with a `-2` CONVERTED backslide on `src/_ZN12daObjAbuku_cD0Ev.cpp` / `D1Ev.cpp` (absorbed by the `ov002/daObjAbuku_c` TU promotion). **That is pre-existing on `origin/main` and reproduces there identically** — it is not caused by this PR and should not be attributed to it.

### Minor — three stale references left behind

Not load-bearing, but the "all reference surfaces retargeted" claim is not complete. Old paths still appear in:
- `src_tu/actors/WingFeather.cpp` — 8 hits, in the generated provenance comment header
- `notes/data/tu-merge-candidates.json` — 7 hits
- `notes/data/c-cpp-classification.tsv` — 1 hit

All three are inert: no tool under `tools/` or `.github/` consumes either `notes/data/` file, and `check_dead_references.py` only scans `.md`, so it cannot see any of them. Worth a follow-up sweep rather than a blocker.

### Verdict

Content, placement, and delinks integrity are all correct — the risky part of a rename PR is clean. Held in draft only because of the `converted-baseline.json` conflict; once that is resolved by key union onto current `main`, this is good to go.

PR #2012, opened 2026-08-30
Use ROM-proven daObjWaterfall_c C++ identity

andrewboudreau:
## Summary

- Rename the coined WaterfallMist class and method symbols to ROM-proven daObjWaterfall_c while retaining the evidence-bounded C ABI factory name.
- Move the factory to typed C++ and add a genuine class TU that compiler-emits D1/D0/D2, RTTI/ZTS, typed SpawnInfo, and the vtable.
- Keep the narrow explicit fBase allocator/base-constructor/vptr seam because natural new targets _Znwm and mwccarm b56 rejects placement new.

## Verification

- 5/5 text and relocations exact; all five production linkchecks VERIFIED with blind: 0.
- Owned class data exact; 11,087/11,087 reproducing, 106/106 modules exact, stock ROM SHA-256.
- Langmode, converted state, affected consumers, header, reference, and 405/405 port gates pass.
- Candidate and baseline scratch links stop before Waterfall substitution on the same two pre-existing promoted-TU policies; the manifest records that control result.


andrewboudreau:
## Gatekeeper review: cartridge evidence for the `daObjWaterfall_c` identity

Recording the raw-image evidence on the PR, because the identity claim here is
the whole PR and it should not have to be re-derived by the next reader.

Everything below was read out of `extracted/overlays/overlay_0002.bin`
(load base `0x020ad660`), not out of `config/`. Config symbols were treated as
the thing under test, not as evidence.

**1. The RTTI record.** `0x0210945c` is a complete Itanium-ABI
`__si_class_type_info`:

| word | value | meaning |
| --- | --- | --- |
| `0x0210945c` | `0x0209a764` | vptr -> `_ZTVN3abi20__si_class_type_infoE` (arm9) |
| `0x02109460` | `0x02109468` | pointer to the type name string |
| `0x02109464` | `0x0208e390` | pointer to `_ZTI8dActor_c` (arm9) -- the single base |

Single-inheritance from `dActor_c` is therefore stated by the cartridge, not
inferred from the header.

**2. The name string.** File offset `0x5be08` -> VA `0x02109468` holds
`16daObjWaterfall_c\0`. The leading `16` is the mangled-name **length prefix**
for the 16-character identifier `daObjWaterfall_c` -- it is not part of the
name. (This is the usual trap: `_ZTS` payloads are length-prefixed, so the
class is `daObjWaterfall_c`, and the correctly mangled symbols are
`_ZTI16daObjWaterfall_c` / `_ZTS16daObjWaterfall_c` / `_ZTV16daObjWaterfall_c`.)

**3. The vtable, and that it is *this* class's vtable.** The two words before
the address point are the standard prologue:

```
0x02109498  0x00000000   offset-to-top
0x0210949c  0x0210945c   -> the RTTI record above
0x021094a0  0x020b6eac   address point, slot 0 (InitResources)
```

So the table configured at `0x021094a0` is bound by the cartridge to the
`daObjWaterfall_c` typeinfo. Renaming `_ZTV13WaterfallMist` ->
`_ZTV16daObjWaterfall_c` is a correction, not a relabelling.

**4. The factory reaches that exact table.** The literal-pool word at
`0x020b6f14`, inside `WaterfallMist_Spawn`, is `0x021094a0` -- the address point,
not the prologue. The vptr store in the recovered factory is the ROM's store.

**5. The coined name is absent from the cartridge.** A byte census of the
string `WaterfallMist` across every image under `extracted/` (arm9, arm9 decomp,
arm7, and all overlays) returns **0 occurrences**. `daObjWaterfall_c` is
ROM-proven; `WaterfallMist` is a coined placeholder that only ever existed in
this repository.

### Notes on scope, for the record

* Exactly one symbol is configured at `0x021094a0` both before and after this
  PR, so this is a rename and not the delete-the-coined-alias case from #1978.
  There is no second name to drop.
* `WaterfallMist_Spawn` / `WaterfallMist_SpawnInfo` are deliberately left
  alone. Those are C-ABI names with no mangled evidence behind them; the
  ov018 `MotherPenguin`/`daPgMthr_c` precedent keeps the coined *actor* name on
  exactly this kind of symbol.
* The vtable move is atomic: all five text/data symbols and the `_ZTV` move in
  the same commit, so no `_ZTV` entry is left pointing at the old spelling.
* The `include/daObjMarioCap_c.h` comment edit is a genuine correction. I read
  `0x021093dc` -> `0x021093a0` -> `_ZTS11daObjLava_c` at `0x021093ac`: the
  preceding table really is `daObjLava_c`'s, so the old comment's claim that
  `0x021094a0` was "the wrong table, one entry along" was factually wrong.

### Follow-up pushed to this branch

`origin/main` merged in (only conflict was `notes/cpp-tu-current-state.md`,
resolved by regenerating with `tools/cpp_tu_state.py --write-note`; `--check-note`
reports current), plus three cleanups the rename left behind:

* dropped the now-dead `extern int _ZTV13WaterfallMist[];` from
  `include/decl_common.h` (the 66 other dead `_ZTV` declarations in that header
  predate this PR and were left alone);
* refreshed the `notes/data/tu-merge-candidates.json` ov002 record and
  `notes/data/c-cpp-classification.tsv` row for the new paths and class label,
  following the path-refresh precedent from #1985.

Gates re-run on the merge result: `port_refcheck` 405/405, `check_src_tu`
(83 TUs, 1344 mangled refs, all resolve), `check_src_tu_compiles` 88/88,
`check_dead_references` (no new dead references), `check_header_offsets`
(3 changed headers, 0 mismatched, 0 unparsed; `daObjWaterfall_c` spans `0xdc`),
langmode ratchet PASS, CONVERTED ratchet PASS (2552/2552, no backslide
exception added), `tu_manifest list` shows `ov002/daObjWaterfall_c`.

### One cross-PR consequence

#1993 banks a `deadstrip` disposition for `_ZTI13WaterfallMist` whose stated
reason is "compiler-generated typeinfo record with no ROM symbol ... the
cartridge never carried a separate copy." Section 1 above disproves that: the
cartridge carries the record at `0x0210945c`. The rationale was written against
the coined name, which had no `_ZTI`, and so it concluded the ROM had none.
That row needs revisiting before #1993 lands.

PR #2016, opened 2026-08-30
TU: fix stale legacy_source extensions in the manifest

andrewboudreau:
## What

38 `functions[].legacy_source` values across 19 TU manifest entries still point at `src/<name>.c` when the file on disk is `src/<name>.cpp` — the file was migrated C -> C++ and the manifest row was never updated.

Each stale row blocks `tu_promote` on that entry: the promote path resolves `legacy_source` to a file that no longer exists.

## Derivation

Re-derived against current `origin/main` (`25cfc136`) by script, not carried over from an older tree. For every `config/tu_manifest.d/**/*.json`, every `legacy_source` ending in `.c` was checked against `git ls-tree -r --name-only origin/main`. A row counts as stale iff the `.c` path is **absent** and the `.cpp` sibling is **present**.

- Stale rows found: **38**, across **19** files.
- Inverse case (a `.cpp` `legacy_source` whose file is actually `.c`): **0** — nothing to fix there.
- No `data[].legacy_source` rows were affected; all 38 are under `functions[]`.

Re-running the same derivation after the change reports **0** stale rows.

## Entries changed

| entry | rows |
| --- | --- |
| `arm9/Actor.json` | 7 |
| `arm9/ActorDerived.json` | 1 |
| `ov002/Platform.json` | 2 |
| `ov002/Warp.json` | 1 |
| `ov009/CastleWater.json` | 2 |
| `ov009/Flag.json` | 2 |
| `ov009/MetalNet.json` | 1 |
| `ov010/LightBeam.json` | 2 |
| `ov010/PeachPainting.json` | 2 |
| `ov014/ShutterBob.json` | 1 |
| `ov019/IceSlideManager.json` | 1 |
| `ov019/RacingPenguin.json` | 2 |
| `ov020/HauntedChair.json` | 2 |
| `ov029/CageLift.json` | 1 |
| `ov029/RotatingPlatformWdw.json` | 2 |
| `ov029/SwitchActivatedPlank.json` | 2 |
| `ov029/WaterDiamond.json` | 2 |
| `ov062/KoopaFlag.json` | 2 |
| `ov077/Spiny.json` | 3 |

## Shape

Manifest JSON only — **no tooling, no source, no config outside `config/tu_manifest.d/`**. Applied as a targeted textual substitution of the specific `"legacy_source": "src/X.c"` strings rather than a `json.load`/`json.dump` round-trip, so indentation, key order and trailing newlines are byte-identical outside the 38 changed values.

`git diff --stat` is 19 files, 38 insertions, 38 deletions; every changed line pair is verified to be an exact `.c` -> `.cpp` swap on the same base name and nothing else.

`python tools/tu_manifest.py list` still succeeds and still reports 89 TUs, unchanged from main.

## Provenance

Salvaged from the closed PR #1884 — this is the one clean, mechanical win in it. It was **re-derived from current main rather than cherry-picked**, because #1884's row set was computed against a 411-commit-old tree. This supersedes the salvageable part of #1884.

### DIFF
--- a/config/tu_manifest.d/arm9/Actor.json
+++ b/config/tu_manifest.d/arm9/Actor.json
@@ -362,7 +362,7 @@
       "symbol": "_ZN8dActor_c13SpawnFireballERK7Vector3PK10Vector3_165Fix12IiES7_j",
       "address": "0x020102b0",
       "size": "0x00000054",
-      "legacy_source": "src/_ZN8dActor_c13SpawnFireballERK7Vector3PK10Vector3_165Fix12IiES7_j.c",
+      "legacy_source": "src/_ZN8dActor_c13SpawnFireballERK7Vector3PK10Vector3_165Fix12IiES7_j.cpp",
       "ordinal": 48
     },
     {
@@ -397,7 +397,7 @@
       "symbol": "_ZN8dActor_c12ReflectAngleE5Fix12IiES1_s",
       "address": "0x02010428",
       "size": "0x00000034",
-      "legacy_source": "src/_ZN8dActor_c12ReflectAngleE5Fix12IiES1_s.c",
+      "legacy_source": "src/_ZN8dActor_c12ReflectAngleE5Fix12IiES1_s.cpp",
       "ordinal": 53
     },
     {
@@ -411,7 +411,7 @@
       "symbol": "_ZN8dActor_c15IsPlayerInRangeE5Fix12IiES1_S1_i",
       "address": "0x02010498",
       "size": "0x00000044",
-      "legacy_source": "src/_ZN8dActor_c15IsPlayerInRangeE5Fix12IiES1_S1_i.c",
+      "legacy_source": "src/_ZN8dActor_c15IsPlayerInRangeE5Fix12IiES1_S1_i.cpp",
       "ordinal": 55
     },
     {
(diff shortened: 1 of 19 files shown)


andrewboudreau:
Reviewed and queued to merge — the derivation is the right one (re-derived against `origin/main` via `git ls-tree`, not cherry-picked from #1884's `13bf9222f`, whose row set was computed against a 411-commit-old tree). Regex-targeted substitution rather than a `json.dump` round-trip is exactly right here; it keeps the diff at 38/38 with byte-identical formatting elsewhere.

This supersedes the one salvageable hunk of #1884, which I closed along with #1880/#1882 (see #1880 for why: its `plan_many` change dropped every `_ZTV`/`_ZTI`/`_ZTS` section uncompared, a design main deliberately rejected in favour of #1987's `deadstrip-data` licensing).

**Separate follow-up, not for this PR:** the same scan found **75 other `legacy_source` values pointing at paths that exist in neither extension** — all of `ov070/daKpFr_c.json` and `ov070/daKrpa_c.json`, plus `ov002/daObjAbuku_c.json`, `arm9/ActorBase_SceneNode.json`, and scattered `func_*` rows in `ov010/PeachPainting`, `ov020/HauntedChair`, `ov029/WaterDiamond`. Those are entries whose one-function files were consolidated into merged TUs, so the legacy path is gone rather than renamed. That is a different and non-mechanical problem — correctly left alone here, but it wants its own look, since a dangling `legacy_source` is exactly the kind of pointer that reads as evidence without being any.

PR #2018, opened 2026-08-30
The Goomboss learns to hit: the cylinder rebuild ships execution-audited at div 11

tangosdev:
The last unsourced piece of the Goomboss - his collision-cylinder rebuild func_ov074_02121380 (0x374) - ships as a declared NONMATCHING body at 11 divergent words of 221, with the wall behind those 11 words measured closed across two sieges (~34,000 compiles, the full pragma x shape x store-order product space, ~18,200 permuter iterations on two structurally different seeds; the mechanism - hoisted array-base register rank follows store consumer source order, producing a schedule-vs-coloring conflict - is notes 6bq with its addendum, and the near-miss row now carries the floor mark with the full evidence record).

Semantics are proven by execution, not asserted: a differential interpreter runs the cartridge's instructions against the compiled body over the complete 2,700-state discriminating lattice (a verified bijection, not a sample) - identical call sequences, all 216 written bytes, and returns on every state, with all 12 branch arms exercised and the coverage reporting itself falsified-then-trusted in review. Eight size-neutral sabotaged controls are all caught; building them exposed and fixed three real harness bugs (ldm/stm support, per-side .data mapping, threshold-boundary seeding), each recorded.

The residue is pure scheduling and scratch-register rotation - the register-normalised instruction multiset is equal on both sides, and the four coloring-lever address computations are byte-identical. All divergent-word counts appear under both scorers (11 by the per-offset gate, 9 by the DB's sequence alignment) with the difference explained. The 24 untouched frame bytes behind one array spelling remain a declared placeholder, not recovered source.

Independently reviewed end to end: gate re-derived from raw bytes, audit re-executed with a deliberate under-coverage falsification, both lever spot-checks reproduced, notes and ledger hygiene verified.


andrewboudreau:
## Gatekeeper review: **MERGE-AFTER-FIX**

Reviewed against the diff, not the description. **It cannot erode ROM byte verification** — that part is clean, and I checked it specifically:

- No `delinks.txt` or `symbols.txt` diff anywhere, so **zero functions are handed back to the cartridge**. `config/arm9/overlays/ov074/delinks.txt` on main already jumps `…end:0x02121380` straight to `0x021216f4`; that 0x374 span stays ROM-supplied and this PR does not touch it.
- No `config/tu_manifest.d/` change, so no `deadstrip` disposition on any `_ZTI`/`_ZTS`/`_ZTV`.
- The `NONMATCHING` marker sits on a **brand-new, never-enrolled** file, so it disables eligibility for nothing that was previously compared. No counter is inflated: `chaos_db_ci.py` counts `matched` only when a source resolves *and* is not marked `// NONMATCHING`, and its `enrolled_addresses()` number is delinks-derived.

An honest declared draft is explicitly solicited by `chaosviewer.config.json`, and 64 such files already live in `src/`. No objection to the shape.

### Three things before this lands

**1. The framing overstates the result.** I diffed the shipped body against the `c_source` already banked in `nearmiss/db.jsonl` on main, comment-stripped: **0 differing code lines, 87 vs 87.** The ~34,000-compile / ~18,200-iteration siege produced no code change and no divergence change — 11 → 11, `improvedNearMiss: false`. Your `match_attempts` row says this plainly, which I appreciate; the title and the file's "fresh derivation" comment do not. The deliverable here is the src/ ship, the `floor` annotation and the notes addendum — all genuinely worth landing. Please just say so in one sentence, so a future reader does not re-derive the same 11 words believing this was new ground.

**2. Fix the coverage claim in the banner.** The last line reads "Counts as decompiled, not matched." There is no "decompiled" tier — a `NONMATCHING` file is simply not counted (`enroll.py:306` skips it as a NONMATCHING hatch). The precedent file `src/_ZN7Wiggler8BehaviorEv.c` says the opposite. Suggested:

```diff
-// Counts as decompiled, not matched.
+// This is a declared draft: it does NOT count as decompiled, and it is
+// deliberately not enrolled in config/arm9/overlays/ov074/delinks.txt.
```

**3. Rebase — you are 28 behind.** Merge base is `a93d29db8`; main is well past it. Both files you touch (`nearmiss/db.jsonl`, `config/match_attempts.jsonl`) are the hottest append/edit files in the repo. Resolve both by **row union**, never `--ours`/`--theirs`.

### On the `--admin` override

`MERGE.md` §2 says a NONMATCHING PR fails `validate` by design and lands with `--admin --squash`. That override is safe **only once the validator has actually run** and shows the failure is the 11-word residue — not a compile error, `wrong-dest`, or `blind`. `PR validation` is still QUEUED, so nothing has yet proved this file even compiles in the validator's environment. I will not override on the queued stub.

### Minor, non-blocking

This file declares `extern void func_ov074_02121270(struct Vector3* out, char* c, int i);` while the enrolled `src/func_ov074_02121270.c` defines it `(char* r4, char* r8, int r7)`. ABI-identical and separate TUs, so harmless — and yours is the better spelling — but the tree now disagrees with itself about that prototype.

The notes addendum is the strongest part of this PR: it retracts its own prior evidence ("Two of those names … are not pragmas at all") and records three harness bugs found by falsification controls. That is exactly the standard I want to see.

andrewboudreau:
## Correction: my review's central objection was wrong.

I reviewed this as MERGE-AFTER-FIX on the grounds that the shipped body was byte-identical to the `c_source` already banked in `nearmiss/db.jsonl` while the banner framed it as a fresh derivation. I compared against the row at `origin/main`. That was the wrong comparison, and the conclusion does not hold.

Tracing the row's history through `nearmiss/db.jsonl`:

| commit | div | `c_source` lines |
|---|---|---|
| … through `52d681683` (#1999) | **297** | 77 |
| `6906f2af5` (**#2006**) | **9** | 115 |
| `36e8cbcce` (#2008) | 9 | 115 |
| `eabf1a0bd` (**this PR**) | 9 | 189 |

The identical body I found was banked by **#2006**, this author's own earlier PR — the one titled "the cylinder rebuild banked at 11". The row that predates this line of work is the **div-297, 77-line** one, which is exactly what the banner says:

> A near-miss row for this address already existed at div 297 from a `fanout-opus` run; it was not used — structure, externs and frame layout here are a fresh derivation.

That is accurate. This PR is the follow-up that ships the already-banked body behind the NONMATCHING hatch after the floor sieges, and its diff to the row is the `floor` object plus prose — which is the honest shape for that. `divergences 11 -> 11` and `improvedNearMiss: false` are the correct readings for a PR that ships a body rather than improving one, not evidence of a recycled derivation.

The rest of the review stands and was never blocking: no `delinks.txt`, `symbols.txt` or manifest diff, `NONMATCHING` on a never-enrolled file, so ROM verification cannot erode here. Confirmed post-merge — the `source coverage` gate ran on `eabf1a0bd` and reported clean, no bytes handed back.

One residual nit, not worth a PR on its own: line 102's "Counts as decompiled, not matched." names a tier that does not exist — `tools/tiers.py` has `converted`, `matched`, `linked`, `qualified`, `done`, `thunk`. It reads fine as English and nothing machine-reads it; worth rewording next time this file is touched.

PR #2022, opened 2026-08-30
Two hybrid extern names in the Wiggler draft could never link

tangosdev:
Found by the port-side propagation review: the draft spelled two externs as hybrids of the old class name and the renamed parameter type - names that exist on neither line. Corrected to the real config names at those addresses (verified by address, the day's standing lesson). The strict gate is unchanged at div 20 after the rename; the draft remains unenrolled, so nothing behavioral changes anywhere - this just removes the landmine for whoever enrolls it after a future byte match.

### DIFF
--- a/src/_ZN7Wiggler8BehaviorEv.c
+++ b/src/_ZN7Wiggler8BehaviorEv.c
@@ -79,8 +79,8 @@ extern void _ZN8Particle20RunningSlidingDustAtE5Fix12IiES1_S1_(s32 x, s32 y, s32
 extern void _ZN9Animation7AdvanceEv(void *self);
 extern int DecIfAbove0_Byte(u8 *p);
 extern void _Z14ApproachLinearRiii(s32 *p, s32 a, s32 b);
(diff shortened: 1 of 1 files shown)


andrewboudreau:
Verified independently before clearing. Both replacement names resolve in config; both originals resolve nowhere:

```
_ZN8dActor_c9UpdatePosEP5dCc_c            config/arm9/symbols.txt:302              addr:0x02010c30
_ZN12dEnemyBase_c12UpdateWMClsnER10dBgCh_Actrj  config/arm9/overlays/ov002/symbols.txt:28  addr:0x020aebf8
_ZN5Actor9UpdatePosEP5dCc_c               0 hits anywhere under config/
_ZN5Enemy12UpdateWMClsnER10dBgCh_Actrj    0 hits anywhere under config/
```

Also confirmed the blast radius is exactly this file: no other file under `src/`, `src_tu/`, or `include/` spells either old name, and `_ZN7Wiggler8BehaviorEv` appears in no `delinks.txt`, so the draft is unenrolled and this cannot move a byte. It is strictly the difference between a body that could never link and one that can.

Cleared from my side; merging when `PR validation` completes.

PR #2024, opened 2026-08-30
Promote daObjPushblock_c to a genuine C++ TU

andrewboudreau:
Consolidates nine verified daObjPushblock_c methods into one typed class translation unit with compiler-owned D1/D0, RTTI, and vtable emission. PushBlock_Spawn is deferred rather than excluded on evidence, and stays a separate typed C ABI factory for now; it hand-spells the `_ZN7fBase_cnwEj` actor-heap allocator because a natural new expression selects the wrong allocator relocation, which it would equally have to do inside the TU. Raw-offset member access and the obsolete codegen qualifier are removed.\n\nVerified locally: TU 9/9 exact with clean relocation destinations, emitted data has 0 differences, factory linkcheck is VERIFIED blind:0, all 90 registered TUs compile, and the full ROM is 11,087/11,087 reproducing with 106/106 modules exact.


andrewboudreau:
## Gatekeeper review — MERGE AFTER FIXES (two text-only, no rebuild)

The code is genuinely good and the class model is corroborated against the cartridge, not
just against the config. One committed evidence field says something the tool it cites
does not say; that is the only thing standing in the way.

### The finding: `boundary_evidence` attributes a cut to `build/tu_map.json` that tu_map does not contain

`config/tu_manifest.d/ov002/daObjPushblock_c.json` says:

> *"contiguous linker run: `0x020b8bf0..0x020b910c`, 9 function(s), from `build/tu_map.json`"*
> *"high-confidence left cut at `0x020b8bf0` and right cut at `0x020b910c`"*
> *"the adjacent high-confidence `PushBlock_Spawn` candidate begins at `0x020b910c` and is deliberately excluded"*

Live `build/tu_map.json` says:

```json
{"start": "0x20b8bf0", "end": "0x20b9148", "classes": ["PushBlock"]}
funcs: [D1, D0, Kill, OnHitByMegaChar, OnPushed, CleanupResources,
        Render, Behavior, InitResources, PushBlock_Spawn]
```

**Ten functions, ending at `0x20b9148`, with `PushBlock_Spawn` inside the run.** The
committed derivative `notes/data/tu-merge-candidates.json` agrees: one TU, 10 functions,
`left_boundary: high`, `right_boundary: high`. There is no cut at `0x020b910c`, and
`PushBlock_Spawn` is not an adjacent candidate — it is a member.

The PR body's causal claim does not survive either. *"remains a separate typed C ABI
factory **because** natural `new` selects the wrong allocator relocation"* — the
allocator wall is real (`include/fBase_c.h:150`, and 392 source files spell
`_ZN7fBase_cnwEj` by hand), but it explains the hand-spelled allocator, not the split.
`src/actors/daObjAbuku_c.cpp` (#1996) folds its factory **into** the promoted TU with
line-for-line the same code now sitting in `PushBlock_Spawn.cpp`. And the tree's rule is
"follow tu_map", confirmed in both directions: `daObjPathLift_c` excluded its Spawn
*because* tu_map gives it its own 1-function TU, and shipped 8; `daObjAbuku_c` included
its Spawn *because* tu_map puts it inside, and shipped 7/7. This is the first promotion
to deviate from tu_map while citing tu_map as the reason.

**Severity: zero bytes, zero gates.** Coverage is byte-identical and both ratchets pass —
proved below. This is an accuracy defect in the exact field a future TU-slicing decision
will trust, in a project whose stated ordering puts accuracy first. Same class as the
coined-name deadstrip claims caught in #1993.

### Required (both text-only, neither needs a rebuild)

1. **Fix `boundary_evidence`.** Remove the three bullets attributing a 9-function
   `0x020b910c` right cut to tu_map. State the truth: tu_map reports one high-confidence
   run `0x020b8bf0..0x020b9148` of 10 functions including `PushBlock_Spawn`; this
   promotion licenses the first nine and defers the tenth. **The licensed range and every
   byte claim stay exactly as they are.**
2. **PR body.** Drop the causal "because natural `new` selects the wrong allocator
   relocation" as the reason for the split. Say `PushBlock_Spawn` is deferred, and that
   the allocator wall is why it hand-spells `_ZN7fBase_cnwEj` — which it would do inside
   the TU too.

Folding Spawn in to reach the full 10-function TU is the better outcome but is follow-up
work, not a merge condition.

### What I verified independently

- **No erosion.** Merge result `git merge-tree --write-tree origin/main <head>` →
  `01d49188d`, clean. `.text 1,941,060 + .init 125,712 = 2,066,772 B` on **both** trees.
  `source_coverage --check`: `net +0 B, handed back 0 B`. I did not take the headline on
  trust — recomputed the enrolled-range union from `ov002/delinks.txt` directly:
  `.text` 318,844 B on both, 0 lost, 0 gained. The `entries` −8 is the documented
  consolidation: the nine deleted ranges are contiguous and gapless
  (`0x020b8bf0 → 8c3c → 8c9c → 8d14 → 8d3c → 8d68 → 8dac → 8dd4 → 8fe0 → 910c`) and the
  one new entry covers precisely that union.
- **`tiers_ratchet --check` PASS on the merge result**, not just on the branch. Worth
  restating: nothing in CI gates a merge *result* — `pr-validate.yml` is
  `pull_request_target`-only, and no ROM build runs on a main push.
- **The class model is right against the retail image, not the config.** Read
  `extracted/overlays/overlay_0002.bin` directly (base `0x020AD660`). `0x02109678` holds
  the literal bytes `16daObjPushblock_c` — **the cartridge carries this class's own
  mangled type name, so it is not a coined name** and the coined-name deadstrip trap does
  not apply. `0x0210966c` is a well-formed `__si_class_type_info`
  `{0x0209a764, 0x02109678, 0x021089ec}`, confirming `: dBgActor_c` from the ROM rather
  than from a guess. Every vtable slot claimed in the header checks out: 0=InitResources,
  3=CleanupResources, 6=Behavior, 9=Render, 16/17=D1/D0, 25=OnPushed,
  27=OnHitByMegaChar, 31=Kill.
- **`compiler_only_output` census: 12 rows, zero plain `deadstrip`.** Eleven
  `deadstrip-data` with canonical addresses, one `deadstrip-duplicate`
  (`_ZN7Vector3D1Ev` → arm9:0x020072c0). #2019's hard block does not arise.
- **`attribution.json`: 9 keys added, 0 dropped, 0 changed** (908 → 917). The three
  legacy whole-file overrides map to the new per-symbol keys with the same author.
- **The 4 backslide-exception rows are honest.** `tiers.score_file` on the absorbing file
  fails **only** `no_mangled_refs` — the structural TU-promotion outcome the tool's own
  docstring predicts. Their claims ("no longer use raw offsets or codegen qualifier
  tricks") are true, not merely asserted.
- **`check_header_offsets` guard still applies** despite the new inline member body
  (`virtual ~daObjPushblock_c() {}`), one of the shapes that can silently blind it:
  `7 commented fields, 0 mismatched, 0 unparsed, struct spans 0x4f4` — identical on both
  trees. The header also drops its `#else` flat-C half; only the two `.cpp` files in this
  PR include it.

### The readability is real, and this is the part worth saying out loud

`Behavior` on main carried a **"NOT MIGRATED … measured 2026-08-22"** banner and was an
`extern "C"` free function taking `char *c`, with `char ray[0x50]`, `volatile Vector3
pos`, `*(char **)(c + 0x4f0)`, `*(int *)(q + 0x5c)`. It is now a real
`int daObjPushblock_c::Behavior()` over `this`, with `dBgCh_Gnd ray;` as an actual RAII
object, `mLinkedActor->actorID`, `q->mPosX`, `UpdatePos(0)`, `Sound::PlayLong(...)`. The
two-extra-pool-words wall that blocked it dissolved in whole-TU context, and the
`volatile` came off in a one-line change. `PushBlock_Spawn` went from
`(char *)p + 0x320` to `&actor->mWithMeshClsn`. These are genuine C++ methods, not shims
in a `.cpp` costume — the residual seams are five Fix12-by-value calls (the documented
tree-wide wall), one named veneer, and three allocator/ctor spellings. All 21 external
names resolve in the correct module's `symbols.txt`; no cross-overlay collision.

### Nits — no action required

- `src/PushBlock_Spawn.cpp:6` says global `operator new` is `_Znwm`. That is the LP64
  mangling; on ARM32 it is `_Znwj`.
- `include/daObjC1_Trap_c.h:78` and `include/daObjHmBskt_c.h:54` cite
  `daObjPushblock_c.h` as the model for the dual C/C++ header arrangement this PR
  deletes. Now stale.
- `OnPushed` still has a bare `return;` in an `int` function. Pre-existing on main and a
  real hazard for the (ungated) port build — not this PR's to fix.

### What remains your word

I ran no compiler and no ROM build, deliberately — `build/` is shared and concurrent runs
produce fake failures. So the TU 9/9 exact, `romdata_check` 0 differences, the factory
`linkcheck VERIFIED blind:0`, all-90-TUs-compile, and 106/106 claims are unverified here;
the validator's green covers them. One I resolved at the design level: inlining the dtor
demotes it as key function, which normally loses the TU-owned `_ZTV` — but `InitResources`
is the next declared virtual (slot 0, confirmed from the ROM vtable), is non-inline, and
**is defined in this TU**, so the TU should still own `_ZTV`/`_ZTI`/`_ZTS`, consistent
with the manifest. I did not observe the emission.

Make the two text fixes and I will merge.

andrewboudreau:
## Requested fixes applied on the authoring session's behalf

The authoring session is gone, so I have made the two text-only fixes I asked for myself. Pushed as `eb45053b5` on top of `714a6b800` (no rebase, one commit added). **I have not merged** — this is for the next green.

---

### Fix 1 — `config/tu_manifest.d/ov002/daObjPushblock_c.json`, `boundary_evidence`

**Was:**

```
"contiguous linker run: 0x020b8bf0..0x020b910c, 9 function(s), from build/tu_map.json (tools/tu_map.py)",
"class label(s): daObjPushblock_c",
"high-confidence left cut at 0x020b8bf0 and right cut at 0x020b910c in build/tu_map.json",
"the adjacent high-confidence PushBlock_Spawn candidate begins at 0x020b910c and is deliberately excluded"
```

**Now:**

```
"contiguous linker run: 0x020b8bf0..0x020b9148, 10 function(s), from build/tu_map.json (tools/tu_map.py)",
"class label(s): daObjPushblock_c",
"high-confidence left cut at 0x020b8bf0 and right cut at 0x020b9148 in build/tu_map.json",
"PushBlock_Spawn is the tenth function of that one mapped run, not an adjacent candidate: build/tu_map.json reports no cut at 0x020b910c",
"this promotion licenses the first nine functions of the run (0x020b8bf0..0x020b910c) and defers PushBlock_Spawn, so the licensed range is a prefix of the mapped run rather than a mapped boundary"
```

The licensed range is untouched. `sections` still reads `0x020b8bf0..0x020b910c`, all nine `functions` entries, every address, size and byte claim are byte-identical. Only the justification prose changed.

### Verification of the tu_map facts (re-measured, not taken on trust)

`build/tu_map.json`, `modules.ov002.units`, the run at `0x20b8bf0`:

```json
{
  "start": "0x20b8bf0",
  "end": "0x20b9148",
  "classes": ["PushBlock"],
  "functions": [
    "_ZN9PushBlockD1Ev", "_ZN9PushBlockD0Ev", "_ZN9PushBlock4KillEv",
    "_ZN9PushBlock15OnHitByMegaCharER6Player", "_ZN9PushBlock8OnPushedER8dActor_c",
    "_ZN9PushBlock16CleanupResourcesEv", "_ZN9PushBlock6RenderEv",
    "_ZN9PushBlock8BehaviorEv", "_ZN9PushBlock13InitResourcesEv",
    "PushBlock_Spawn"
  ]
}
```

One run, ten functions, `PushBlock_Spawn` a **member**. Scanning every unit in `ov002` for either boundary equal to `0x020b910c` returns **zero**. The neighbouring units are `Cap` `0x20b8b98..0x20b8bf0` (1 fn) on the left and `PowerFlower` `0x20b9148..0x20b9e64` (19 fns) on the right — so `0x020b9148` is the real right cut and nothing begins at `0x020b910c`.

The committed derivative `notes/data/tu-merge-candidates.json` on this branch corroborates independently (`ov002` `tu_index` 40): `start 0x20b8bf0`, `end 0x20b9148`, `n_functions_map: 10`, `n_files: 10` including `src/PushBlock_Spawn.c`, `left_boundary: high`, `right_boundary: high`, `both_ends_hard: true`.

### Fix 2 — PR body

The causal claim is dropped. `PushBlock_Spawn` is now described as **deferred**, with the allocator wall correctly scoped to why it hand-spells `_ZN7fBase_cnwEj` — something it would equally have to do inside the TU.

The wall itself is real and I confirmed it: `include/fBase_c.h` documents that CW rejects an in-class `operator new` declaration so it is defined under its mangled name, and 392 files under `src/` on `origin/main` hand-spell `_ZN7fBase_cnwEj`. But it does not explain the *split*: `src/actors/daObjAbuku_c.cpp` (#1996) carries `extern "C" daObjAbuku_c *daObjAbuku_c_Spawn()` calling `_ZN7fBase_cnwEj(sizeof(daObjAbuku_c))` **inside** the promoted TU, and `config/tu_manifest.d/ov002/daObjAbuku_c.json` licenses `daObjAbuku_c_Spawn` as a member function. Line-for-line the shape now sitting in `src/PushBlock_Spawn.cpp`.

### Gates

```
tiers_ratchet.py --check        CONVERTED ratchet PASS   baseline 2546   current 2546
langmode_audit.py --check langmode-baseline.json
                                langmode ratchet PASS
check_dead_references.py        367 prose file(s), 3302 repo-rooted path reference(s), 132 that do not resolve
                                no new dead references
pre-push hook                   port-refcheck 405 checked - all references resolve
                                duplicate-sources: 11243 stems, none doubled
                                check_src_tu_compiles: 90/90 translation unit(s) compiled
```

`git diff 714a6b800..HEAD --stat` is a single file, 4 insertions / 3 deletions. `python tools/tu_manifest.py export` parses the directory cleanly. No ROM build was run — `build/` is shared with other live agents and this change cannot move a byte.

---

### What I could NOT confirm, and judgment calls

1. **`build/tu_map.json` is a stale build artifact.** The copy I measured is dated 2026-08-25 and was produced by some other branch's build; I deliberately did not regenerate it (`rombuild.py` / `eligible.py` are off-limits while other agents are running). Its `ov002` class label is the *pre-rename* spelling `PushBlock`, not `daObjPushblock_c`. I treated it as trustworthy only because the committed, post-rename `notes/data/tu-merge-candidates.json` agrees exactly on range, function count and both boundary confidences. If anyone regenerates the map and gets a different answer, this bullet is where to look first.

2. **`boundary_confidence: "high"` is unchanged, and this is arguably now imprecise.** The left cut is high-confidence and mapped; the right edge of the *licensed* range at `0x020b910c` is a deferral, not a mapped cut at all. I left the field alone because the brief was explicit that the range and its claims stay put, and changing it is a design decision, not an accuracy fix. The new fourth and fifth bullets state the situation plainly so the field cannot mislead on its own. **Flagging it for you to decide.**

3. **The same defect survives in the `notes[]` array of the same file**, which I did not touch because it was outside the two named fixes:

   > `"PushBlock_Spawn is a separate high-confidence TU. A natural new expression produces the wrong allocator relocation, so its typed C++ source retains the exact C ABI allocator/base/member-construction seam."`

   Both halves are the very claims we just corrected — "a separate high-confidence TU" contradicts the corrected `boundary_evidence` two fields above it, and the causal framing is the one dropped from the PR body. This is durable provenance too. I did not fix it unilaterally; **it wants a third edit before this merges**, and I am happy to make it on request.

4. **The PR body contains a literal two-character `\n\n`** (a backslash-n, not a line break) inherited from the original authoring session. I preserved it verbatim as instructed rather than "improving" it in passing; every sentence but the one targeted is byte-identical.

5. The `9/9`, `0 differences`, `blind:0` and `11,087/11,087 / 106/106` figures in the body are the authoring session's measurements. I did not re-run them — my change is prose in a config file and cannot affect them — so they remain the validator's to confirm.

---

*Made by the gatekeeper on the authoring session's behalf. Not merging; over to the next green run.*

andrewboudreau:
## Follow-up: `notes[]` reconciled, sinit bullet restored

Pushed `ca6fb2492` on top of `eb45053b5`. Same file, still text-only, +2/−1. Delta only — everything in my previous comment stands.

### Edit 3 — `notes[3]` no longer contradicts `boundary_evidence`

**Was:**
> `"PushBlock_Spawn is a separate high-confidence TU. A natural new expression produces the wrong allocator relocation, so its typed C++ source retains the exact C ABI allocator/base/member-construction seam."`

**Now:**
> `"PushBlock_Spawn is the tenth function of this entry's one mapped run and is deferred, not a separate TU. Its typed C++ source hand-spells the fBase_c actor-heap allocator because a natural new expression produces the wrong allocator relocation -- a constraint it would be under inside this TU as well, as the promoted ov002/daObjAbuku_c entry shows by licensing daObjAbuku_c_Spawn as a member function."`

No new evidence is asserted — the `daObjAbuku_c` pointer is the same corroboration already in my last comment, now recorded where a future reader of the entry will actually meet it.

### Sweep of the rest of the object — one further copy, and it is fine

I regexed every string in the entry for `separate|adjacent|exclud|natural new|allocator relocation|high-confidence TU|Spawn`. Four hits: my two corrected `boundary_evidence` bullets, the `notes[3]` above, and:

> `notes[4]: "PushBlock_SpawnInfo remains outside this manifest because data adjacency alone does not establish ownership."`

That one is about the `PushBlock_SpawnInfo` **data** symbol, not the function, and its reasoning is independently sound. **Left verbatim.** `notes[0]` ("All nine functions are genuine daObjPushblock_c members") is also true as written — it claims the nine licensed functions are members, not that the run holds only nine. No third copy exists.

### Edit 4 — the sinit bullet checks out, and is restored

Your reading of `tools/tubuild.py:869-877` is right, and it is module-wide **by construction**: the bullet interpolates only module-level fields (`rec['sinits']`, `rec['ctor_entries']`, `rec['sinit_vs_tu']`, `rec['corroborated']`), never anything unit-scoped.

It also holds empirically. Census of all 23 `config/tu_manifest.d/ov002/*.json` entries on this branch:

| | |
|---|---|
| carry the bullet | **20** |
| lack it | 3 — `LevelObjects.json`, `daObjAbuku_c.json`, `daObjPushblock_c.json` (this one) |
| **distinct data payloads among the 20** | **1** — `26 sinit(s) / 26 .ctor entries, sinit_vs_tu=ok, corroborated=False` |

Identical across all twenty. So I restored it, verbatim, in the generator's position (index 2, right after the class label, ahead of the hand-curated cut bullets).

I also re-derived the string from `build/tu_map.json`'s `modules.ov002` record through the generator's own f-string and confirmed it is **byte-identical** to the 19-entry canonical form. And `check_dead_references` still reports no new dead references, which is the check that matters here — the restored bullet cites `notes/tu-reconstruction-pilot-report.md`, and that file exists on this branch.

**One wrinkle worth your time, in a file I did not touch.** Of the 20, nineteen carry the generator's exact string; `ov002/Enemy.json` carries a hand-abbreviated variant:

> `"... corroborated=False (module-wide, NOT narrowed to this one TU)."`

— the `-- see notes/tu-reconstruction-pilot-report.md sec 2 for the by-hand narrowing step` clause dropped and a trailing period added. **The data is identical**, so this does not disturb the module-wide finding or Edit 4. It is cosmetic drift from someone hand-editing a generated bullet, and it means a textual `grep` for the canonical string under-counts by one. Out of scope here; flagging it rather than widening this PR.

### Gates (re-run at `ca6fb2492`)

```
tiers_ratchet.py --check        CONVERTED ratchet PASS   baseline 2546   current 2546
langmode_audit.py --check langmode-baseline.json
                                langmode ratchet PASS
check_dead_references.py        no new dead references
pre-push hook                   port-refcheck: 405 checked - all references resolve
                                duplicate-sources: 11243 stems, none doubled
                                check_src_tu_compiles: 90/90 translation unit(s) compiled
```

`git diff eb45053b5..HEAD --stat` is the one file. `tools/tu_manifest.py export` parses cleanly. `boundary_confidence: "high"` left alone per your call. No ROM build. Not merging.

Remote was `eb45053b5` immediately before the push and `ca6fb2492` immediately after — no interleaving write from the sibling worktree that still holds this branch at the superseded `714a6b800`.

---

*Made by the gatekeeper on the authoring session's behalf.*

PR #2028, opened 2026-08-30
Restamp LINKED off the wave-3 cast-campaign tip: 9121 of 11324 (80.5%)

tangosdev:
The stamp file's own recipe: measured with port/tools/linkage.py against the pre-commit battery build of cons tip ce1afba51 (battery ALL GREEN, tree clean). Eight reviewed merges today took the shipping branch 8775 to 9121; the basis field in the file carries the decomposition. Both remaining disclosed holes matched on main today and their seating lanes are in flight.

### DIFF
--- a/config/port_linkage.json
+++ b/config/port_linkage.json
@@ -28,11 +28,11 @@
     "Enrolling free TUs does nothing, because the release build dead-strips whatever the",
     "linked closure never references."
   ],
-  "linkedTus": 8775,
+  "linkedTus": 9121,
   "matchedTus": 11324,
   "measuredAt": "2026-08-30",
   "branch": "port-mount-noseat-cluster",
-  "commit": "0cf94f562",
+  "commit": "ce1afba51",
   "stale": false,
-  "basis": "single lane cons at its pushed tip 0cf94f562, measured against the tip's own battery build walk_window.map (tree clean, battery ALL GREEN, the tip that closed cast-campaign wave 2). +400 over the morning's 8375 stamp: seven reviewed merges seating the water enemy pack, Eyerok, Snowman's Land, the Goomboss, Shifting Sand Land's remaining cast, four small overlays (THI big+cave, Bowser in the Sky, the mirror arena), and the final ov075 propagation that retired the last VS faces - every function in ov075 now runs recovered source. Two disclosed holes remain the decomp's: func_ov074_021201f0 (Goomboss state-0) and _ZN7Wiggler8BehaviorEv, both filed as crack targets. Remaining frontier: wave 3 partial-overlay tails (~776), then arm9/ov002/ov006."
+  "basis": "single lane cons at its pushed tip ce1afba51, measured against the tip's own battery build walk_window.map (tree clean, battery ALL GREEN, the merge that closed cast-campaign wave 3). +346 over the morning's 8775 stamp: eight reviewed merges seating ov064's last three classes, Princess Peach, Snowball and Moneybag, the snowmen and the Fly Guy, three level-overlay tails (ov025 the first overlay at 100%), the ov091 lifts and Fwoosh, and the Tick Tock Clock cluster behind a per-level re-patch mechanism; plus the two Goomboss state bodies and the declared Wiggler Behavior propagated from main by address. The tree crosses 80 percent. Two disclosed holes remain and both matched on main today: func_ov074_02121380 (declared, div 11, execution-audited) and _ZN14TTC_MovingBeam8BehaviorEv (byte-matched, PR #2026) -- seating lanes for both are in flight. Remaining frontier: wave-4 partial-overlay tails, then arm9/ov002/ov006."
 }


andrewboudreau:
## Gatekeeper review — MERGE AFTER ONE SENTENCE

The measurement is traceable and the arithmetic holds. One sentence in `basis` asserts something the artifact it cites contradicts, and because this is a *stamp* — a durable provenance record whose whole job is to be believed later without re-measuring — that sentence is worth fixing before it lands.

### The finding: a NONMATCHING function is recorded as having "matched"

The new `basis` says:

> *"Two disclosed holes remain and both matched on main today: `func_ov074_02121380` (declared, div 11, execution-audited) and `_ZN14TTC_MovingBeam8BehaviorEv` (byte-matched, PR #2026)."*

And the PR body repeats it: *"Both remaining disclosed holes matched on main today."*

The second is true. The first is not:

| | `_ZN14TTC_MovingBeam8BehaviorEv` | `func_ov074_02121380` |
|---|---|---|
| on `origin/main` | `src/_ZN14TTC_MovingBeam8BehaviorEv.cpp` | `src/func_ov074_02121380.c` |
| `NONMATCHING`/`__asm` | **0** | **1** |
| enrolled `complete` | **yes** — ov065 `.text 0x0211bd8c..0x0211bf04` | **no delinks entry** |
| landed by | #2026 | #2018 |

`src/func_ov074_02121380.c:20` reads, in the file itself:

```
// NONMATCHING: div 11 of 221 words. mwccarm 2004/b56, --module ov074,
```

and its commit on main is titled *"declared NONMATCHING body with a differential-execution audit (#2018)"*. Eleven of 221 words differ. It is not enrolled, so it contributes no verified bytes.

The parenthetical you wrote is accurate — "declared, div 11, execution-audited" says exactly the right thing. It is the shared lead-in **"both matched"** that overstates it, and it does so while sitting next to a sibling explicitly labelled "byte-matched", which invites a reader to carry that label across the `and`.

**Severity: zero bytes, zero gates.** This is the ungated port side ([[port-build-is-ungated]] — `port_refcheck` is not in CI or the hook), and the stamp is metadata. But this is the same defect class I just flagged on #2024: a committed evidence field claiming more than its cited artifact supports. In three months, a reader deciding whether ov074's Goomboss state body still needs work will read "matched" and move on. It has eleven differing words.

### Required — one sentence

Split the claim. Something like:

> *"Two disclosed holes remain: `_ZN14TTC_MovingBeam8BehaviorEv` byte-matched and enrolled on main today (#2026), and `func_ov074_02121380`, which has a declared NONMATCHING body (div 11 of 221 words) with a differential-execution audit (#2018) and is not enrolled. Seating lanes for both are in flight."*

Every number stays as it is. No re-measurement.

### What I verified

- **The cited tip is real and pushed.** `ce1afba51` resolves to `refs/heads/port-mount-noseat-cluster` at `ce1afba510b5abedeb85801a7f4230f924b93bfd` on `origin`. A stamp citing an unreachable commit is the failure mode this file's own `_comment` warns about, and it does not occur here.
- **The arithmetic is self-consistent.** 9121 − 8775 = 346, matching "+346"; 9121 / 11324 = 80.55%, matching the title's 80.5%. `matchedTus` unchanged.
- **The file's own stamping rules are followed.** `_comment` says *"Never move the count without also moving the provenance fields — `tools/tiers.py` drops a stamp missing any of them."* You moved `commit` and `basis` with the count; `branch` and `measuredAt` are unchanged because the lane and the day are the same. Correct.
- **The hole list genuinely turned over.** The previous stamp's two holes (`func_ov074_021201f0`, `_ZN7Wiggler8BehaviorEv`) are gone from the tree and two different ones took their place — consistent with the basis text saying those bodies were propagated from main by address, rather than the same two holes being silently recycled.

### One process note, not a change request

I first ran these greps against the main checkout's working tree and concluded *neither* hole had a source file — which would have been a much more serious finding, and would have been wrong. The working tree sits at `ae8d3b565`, well behind `origin/main` at `e9de9103a`, because that checkout is never pulled. Both files are present on the real ref. Flagging it because anyone verifying a stamp against a long-lived checkout will hit the same trap: measure against `origin/main` via `git show`, not against the files on disk.

Fix the sentence and I will merge.

andrewboudreau:
## Merged before the fix — please carry it into the next restamp

This landed as `9543245b5` while my review was still the newest comment, so the sentence went to main as written. Not a problem worth its own PR: `basis` is rewritten wholesale on every restamp, and this file has moved three times today (8375 → 8775 → 9121), so the next one will replace it anyway.

Two asks for whoever writes that next stamp:

1. **Do not carry the "both matched" framing forward.** `func_ov074_02121380` is a declared NONMATCHING body — `src/func_ov074_02121380.c:20` says `// NONMATCHING: div 11 of 221 words`, it has no `delinks.txt` entry, and its commit on main is titled *"declared NONMATCHING body with a differential-execution audit (#2018)"*. Eleven of 221 words differ and it contributes no verified bytes. Its sibling `_ZN14TTC_MovingBeam8BehaviorEv` genuinely is byte-matched and enrolled at ov065 `.text 0x0211bd8c..0x0211bf04`. Listing them under a shared "both matched" hands the second one's status to the first.

2. **The previous stamp had the same shape**, which is why this is worth saying rather than letting it slide: it also ended with a two-hole sentence, and the holes turned over cleanly between stamps. That structure is fine and useful — it just needs the two holes described separately when their status differs.

Everything else in the stamp verified and I have no further concerns: `ce1afba51` resolves on `origin` as `refs/heads/port-mount-noseat-cluster`, 9121 − 8775 = 346 matches the stated delta, 9121/11324 = 80.55% matches the title, and the provenance fields moved with the count exactly as this file's own `_comment` requires.

PR #2030, opened 2026-08-30
Promote daObjCtMecha05_c to a genuine C++ TU

andrewboudreau:
Replaces the generated shadow and eight one-function shells with one typed daObjCtMecha05_c production translation unit. The compiler now owns D1/D0, RTTI, type name, 32-slot vtable, and TTC_MovingBar_SpawnInfo.

`TTC_MovingBar_Spawn` stays a separate typed C ABI factory because the manifest's own boundary puts it outside this TU. `tools/tu_map.py` recovers ov065's daObjCtMecha05_c run as `0x211ab60..0x211b328`, 8 functions, and reports the factory's address `0x211b328` as the *start of the next unit* (`0x211b328..0x211b780`, 8 functions, class TTC_MovingBar). That is what `build/tu_map.json` says; it is not a conclusion drawn from a build experiment. An earlier revision of this body attributed the split to the natural allocation path failing to reproduce retail relocation identity — this PR does not establish that, and the tu_map boundary is both true and sufficient on its own.

Reading tu_map that way also repairs a fabricated evidence line already on main. The ov065 manifest asserted `contiguous linker run: 0x211ab60..0x211b360, 9 function(s), from build/tu_map.json (tools/tu_map.py)` — citing the tool for a run the tool does not report. It now reads `0x211ab60..0x211b328, 8 function(s)`, which is the tool's actual output.

Verified locally and independently: class TU 8/8 exact with 68/68 relocation destinations, no D2 or unlicensed output, emitted data has 0 differences/unnamed records, factory linkcheck is VERIFIED blind:0, all 89 registered TUs compile, and the restacked full ROM is 11,088/11,088 reproducing with 106/106 modules exact.


andrewboudreau:
## Your red `references` check is mine, not yours — merge main and it clears

Posting this ahead of my full review so you are not debugging someone else's defect.

```
FAIL: 1 prose reference(s) name a path that does not exist:
  .github/workflows/tool-tests.yml
      names `tools/test_X.py`, which is not in the tree
```

That string is not in your diff. It came from **#2027**, which I commissioned and merged, and your branch inherited it from its base:

| | |
|---|---|
| defect introduced | `e9de9103a` (#2027) |
| **your base** | **`e9de9103a`** |
| defect removed | `b309f5d3f` (#2031, merged just now) |
| `test_X.py` occurrences on current main | **0** |

**Merge `origin/main` into your branch and the check goes green.** Nothing to change in your own work.

### Why it reached main at all, since it is worth knowing

`dead-references.yml` is path-filtered and #2027 touched **only** `.github/workflows/`, so the workflow never ran on it. The run list at that commit is `src_tu references`, `tool tests`, `pages` — no `dead references` at all. A workflow that never fires is indistinguishable from one that passed when you skim `gh pr checks`.

The contrast is exact: #2031 touched `.github/` **and** `tools/`, so the same workflow did fire on its merge, and it is green at `b309f5d3f`. Same gate, same file, different path filter outcome.

I have filed the general fix as a follow-up — either `.github/**` joins that filter, or we accept that workflow-file prose is unchecked. That file is now ~185 lines of load-bearing explanation about why particular tests are excluded from CI, which is exactly the prose that goes stale.

My substantive review of the promotion itself is in flight and will follow.

andrewboudreau:
## Gatekeeper review — MERGE AFTER FIXES (four, all text-only)

I ran this adversarially, exporting both the merge result and an identical `origin/main` tree and diffing the gate verdicts rather than reading the PR's own numbers. **No byte defect, no gate regression.** Merge tree `489c47f9` is clean.

More than that: **this PR repairs a fabricated evidence line that is sitting on main right now.** Main's ov065 manifest claims

> `contiguous linker run: 0x211ab60..0x211b360, 9 function(s)`

and `build/tu_map.json` says no such run exists — unit 8 is `{start: 0x211ab60, end: 0x211b328}`, **8 functions**, and `0x211b328..0x211b780` is a separate unit 9. That is the same defect class as #2024's, already landed. Fixing it is the strongest thing in this PR and it deserves to be said out loud in the body.

### What I verified (both trees, independently re-derived)

| | main | merge result |
|---|---|---|
| `tiers_ratchet` | PASS | PASS |
| coverage `.init` / `.text` / `current` / `modules` | 125712 / 1941436 / 2067148 / 74 | identical |
| coverage `entries` | 11026 | 11019 |
| attribution | — | 8 added, **0 removed, 0 changed** |
| `_ZTV`/`_ZTI`/`_ZTS` on plain `deadstrip` | 0 | 0 |

`entries` 11026→11019 is 8 deleted sources minus 1 new TU — consolidation, not loss. I re-derived the covered byte set from scratch on both trees: **bit-identical**. `16daObjCtMecha05_c` is present in `overlay_0065.bin` at `0x0211d27c`, so the class name is the cartridge's, not a coined one.

The inline `~daObjCtMecha05_c() {}` did **not** blind `check_header_offsets` (9→10 fields, span `0x364`→`0x394`) — I checked because an inline member body is one of the six shapes that silently disables that gate.

I also raised and then **disproved** my own objection that ov002 and ov065 might use different `_ZTV` symbol-address conventions: both place the symbol at the address point, verified against the ROM at ov002 `0x0210895c`.

### Required before merge

**1. A comment claims ROM provenance the ROM does not carry.** `src/actors/daObjCtMecha05_c.cpp:273-275` says the ROM's own bss symbol names are spelled `TTC_MovingBar_ModelFile` / `TTC_MovingBar_ClsnFile`.

`TTC_MovingBar` does not appear in **any** ROM binary. It is a coined name. Those two symbols exist only in `config/arm9/overlays/ov065/symbols.txt`, at `0x0211d904` and `0x0211d90c` — names *we* assigned. The addresses are cartridge fact; the spellings are ours.

This matters beyond tidiness: a coined name in a provenance comment fakes an evidence claim, and a later reader has no way to tell which half came from the cartridge. Please say the addresses are ROM-proven and the names are project convention.

**2. The PR body asserts a cause it does not establish.** Use the manifest's own tu_map-backed wording — the boundary is what `build/tu_map.json` reports, and that is both true and stronger than a causal story.

**3. The `converted-backslide-exceptions.jsonl` reason is incomplete.** It omits `no_unk_field`, which fails on `u32 unk_18;` in the file-local `struct CtMecha05SpawnInfo`. The reason text is the only durable record of *why* the credit was spent; an incomplete one makes the row look cheaper to buy back than it is.

**4. `tubuild verify` deleted ~30 lines of curated header prose.** This is the known `tubuild verify` defect — it destroys curated prose and reverts renames. Please restore that evidence block.

Keep the real win in the same file: `u8 pad_364[0x30];` → `Matrix4x3 mShadowMatrix; /* 0x364 */`, which the deleted `src/func_ov065_0211ac0c.c` proves by doing `Matrix4x3_FromRotationY((void*)(c + 0x364), ...)`. That is a genuine field recovery and it should survive.

### Also needed

Merge `main` in. The red check is **inherited, not yours** — run 33327370109 fails on exactly `.github/workflows/tool-tests.yml names \`tools/test_X.py\``, and this PR's base is pinned at `e9de9103a`. That defect was mine (#2027) and is fixed on main by #2033; merging main clears it.

Four text edits, no rebuild needed. **Make these and I will merge on the next green.**

andrewboudreau:
## All four review fixes applied, plus `main` merged — pushed as `4f2914dae`

The authoring session for this branch is gone, so I applied the four fixes from my own review myself. **Text only: no byte changed**, and I re-verified both edited sources compile to the cartridge under the pinned 2004/b56 compiler before committing.

| | |
|---|---|
| pushed | `ba45ac104` → **`4f2914dae`** |
| review fixes | 5 files, +76 / −24 |
| `main` merge | 8 commits, **zero conflicts** |

---

### 1. False ROM provenance — corrected, and I measured it rather than asserting it

`src/actors/daObjCtMecha05_c.cpp` said the ROM's own bss symbol names "really are spelled" `TTC_MovingBar_ModelFile` / `TTC_MovingBar_ClsnFile`. They are not the ROM's names. The search, over all 106 ROM binaries (`arm9.bin`, `arm9_dec.bin`, `arm7.bin`, 103 overlays):

```
$ grep -ro 'TTC_MovingBar' extracted/arm9.bin extracted/arm9_dec.bin \
                           extracted/arm7.bin extracted/overlays/ | wc -l
0
$ grep -ro 'MovingBar' ... | wc -l      0
$ grep -ro 'TTC_'      ... | wc -l      0
```

A zero is only worth as much as its control, so here is the control on the same corpus:

```
$ grep -rl '16daObjCtMecha05_c' ...
extracted/overlays/overlay_0065.bin
```

The method finds a class name that *is* in the image. `TTC_MovingBar` is not, in any form. It exists only in `config/arm9/overlays/ov065/symbols.txt`.

What the cartridge **does** prove is the two **addresses**. `0x0211d904` and `0x0211d90c` appear as little-endian literal-pool words in `overlay_0065.bin`, three times each:

| symbol | address | file offsets in `overlay_0065.bin` |
|---|---|---|
| `..._ModelFile` | `0x0211d904` | `0x4e60`, `0x542c`, `0x69fc` |
| `..._ClsnFile` | `0x0211d90c` | `0x4e64`, `0x5430`, `0x6a0c` |

So the *substance* of the old comment — that this overlay genuinely shares two bss slots with the neighbouring class, and it is not a copy-paste naming error — survives. Only the claim about who chose the spelling was wrong. The comment now separates the two.

**Two further instances you did not name, fixed the same way.** Grepping the rest of the PR for the same claim shape turned up:

- `src/TTC_MovingBar_Spawn.cpp` opened with "TTC_MovingBar_Spawn is the cartridge's **proven** C ABI spelling". Same defect, and slightly worse: the image carries **no symbol names at all**, so it cannot prove the C-ABI/unmangled shape either. The address `0x0211b328` is the fact; the name and the linkage choice are ours.
- The ov065 manifest asserted "the factory retains its **proven** `TTC_MovingBar_Spawn` C ABI spelling".

I judged these in scope — a coined name in a provenance comment fakes an evidence claim wherever it sits — but flagging them since you asked for three fixes here and got five edits.

### 2. PR body — causal claim replaced with the tu_map-backed boundary

The body said the factory stays separate *because the natural allocation path does not reproduce retail relocation identity*. It now states what `build/tu_map.json` actually reports, re-derived on this tree:

```
0x211ab60-0x211b328     8 fn  daObjCtMecha05_c
0x211b328-0x211b780     8 fn  TTC_MovingBar     <- TTC_MovingBar_Spawn is its FIRST function
```

The factory's address is the **start of the next unit**. That is the boundary, and it needs no causal story. I also put the repaired evidence line in the body as you asked (`0x211ab60..0x211b360, 9 function(s)` → `..0x211b328, 8 function(s)`).

One incidental defect: the body contained a **literal `\n\n`** rather than a paragraph break, so it had been rendering as one run-on line. Fixed while I was in there.

### 3. `no_unk_field` added to the exception reason

Confirmed by running the gate's own regexes rather than reading the file:

```
UNK_FIELD   -> match 'unk_18'                    (struct CtMecha05SpawnInfo)
MANGLED_REF -> match '_ZN16daObjCtMecha05_cD1Ev'
```

Both criteria genuinely fail; both are now named, in the neighbouring rows' style, on all three added rows. I also recorded *why* `unk_18` is still `unk_`: the `0x1c` layout is fixed by the ROM's own `SpawnInfo` initialiser, but that last word has no recovered meaning yet.

### 4. Header prose restored, field recovery kept

Restored the ~30 lines `tubuild verify` deleted (base derivation, the below-`0x320` fields all being `dActor_c`'s, the `InitResources` `mAngleY`/`mPrevAngleY` corroboration, the ROM's `operator new(0x394)` literal) plus the detailed per-field comments it had truncated.

`Matrix4x3 mShadowMatrix; /* 0x364 */` **stays**, and now carries its evidence. I went one step past `Matrix4x3_FromRotationY`: the helper also writes `0x388`/`0x38c`/`0x390` — exactly the translation column at `+0x24`/`+0x28`/`+0x2c` of a `0x30`-byte `Matrix4x3` — and then hands that same address to `_ZN8dActor_c18DropShadowScaleXYZER11ShadowModelR9Matrix4x3...`, **whose mangled name types the parameter `Matrix4x3&` outright**. The same call types `0x33c` as `ShadowModel&`. The ROM types this field for us; it is not an inference.

One correction I made rather than restoring verbatim: the old prose ended with "the rest is unclaimed trailing pad". That is now false, and restoring it would have re-introduced exactly the defect fix 1 is about. It is rewritten to record the recovery.

**`check_header_offsets` is unchanged either side of the edit** — I ran it before and after:

```
include/daObjCtMecha05_c.h: 10 commented fields, 0 mismatched, 0 unparsed, struct spans 0x394
```

Same 10 fields, same `0x394` span. No inline body, nested class, namespace, span-form comment or cv-qualifier introduced.

---

### `main` merged — and the conflicts did not happen

`origin/main` (`deb1e1e7a`) is merged and is now an ancestor of the head. **Zero conflicts** — `attribution.json`, `converted-baseline.json` and `converted-backslide-exceptions.jsonl` all merged untouched, because the 8 commits `main` gained since this branch point (`#2027`, `#2028`, `#2031`, `#2033` and a progress refresh) never touched them. So the key-union resolution was not needed. Worth noting only because it is the unusual case.

`.github/workflows/tool-tests.yml` on the merged tree no longer names `tools/test_X.py`, and `check_dead_references` passes locally, so the inherited red check should clear.

### Gates run on the merged tree

| Gate | Result |
|---|---|
| `tiers_ratchet --check` | PASS, baseline 2547 / current 2547 |
| `check_dead_references` | PASS, no new dead references |
| `check_header_offsets --changed origin/main` | PASS, 10 fields / 0 mismatched / 0 unparsed / `0x394` |
| `tu_manifest export` | parses |
| pre-push hook | port-refcheck 405/0 stale; duplicate-sources clean; **89/89 TUs compile** |
| `build_pin.verify` (spot check) | `TTC_MovingBar_Spawn` exact; `daObjCtMecha05_c::Render` exact |

### Nothing I disagreed with, one thing I could not do

No disagreements — all four findings held up when I checked them independently. I could not run `rombuild.py` or `eligible.py`: the `build/` directory is shared with concurrent sessions and a parallel run produces fake failures. For comment-only edits that is not a gap, and the two `build_pin.verify` spot checks above prove the two edited sources still reproduce the cartridge byte-for-byte. The full-ROM number stands from the validator.

**I have not merged this PR** — that is yours.

PR #2032, opened 2026-08-30
Fix tool-test workflow dead reference

andrewboudreau:
Rewords the tool-test workflow comment so the dead-reference checker no longer interprets the illustrative test filename as a real repository path. This is the one-line follow-up needed after #2027; check_dead_references passes locally. No source or generated dashboard files are included.

### DIFF
--- a/.github/workflows/tool-tests.yml
+++ b/.github/workflows/tool-tests.yml
@@ -37,7 +37,7 @@
 #      `tools.test_pr_monitor` both fail on a clean origin/main with ImportError -- they
 #      use bare `from check_dead_references import ...` / `from pr_monitor import ...`,
 #      which only resolves with `tools/` on sys.path. Both are PYTEST modules that pass
-#      when invoked as `python tools/test_X.py` (their __main__ calls pytest.main), so
+  #      when invoked directly by filename (their __main__ calls pytest.main), so
 #      this is an invocation mismatch, not a broken tool. They are named here rather
 #      than fixed: repairing them is a `tools/` change and this workflow deliberately
 #      touches nothing but itself.


andrewboudreau:
## Closing — superseded by #2031, which fixes this properly

The bug you found is real and it is mine: #2027 put the literal string `` `python tools/test_X.py` `` into the workflow comment, and `tools/test_X.py` does not exist. Thank you for catching it.

Worth naming *why* it landed, because it is a repeat of a known pattern rather than a one-off: **the `dead references` workflow never ran on #2027 at all.** It is path-filtered, and #2027 touched only `.github/workflows/`. Checking the main-push runs:

```
e9de9103a (merge of #2027):  src_tu references, tool tests, pages   <- no dead references
```

A workflow that never fires reads exactly like a workflow that passed in the check list. So the gate did not miss this — it was never asked.

### Why I am closing this one rather than merging it

#2031 rewrites that entire bucket-3 paragraph, and in doing so removes the dead reference outright. I verified both ends against the real refs:

| | `tools/test_X.py` present? |
|---|---|
| `origin/main` | yes — line 40 |
| #2031's head | **no** |
| new `tools/*.py` refs introduced by #2031 | **0 distinct** |

So the dead reference is fixed either way. Three things make #2031 the better vehicle:

**1. This patch keeps a sentence that is false.** The line it preserves says the two modules pass *"when invoked directly by filename (their `__main__` calls `pytest.main`)"*. Measured on `origin/main`:

```
tools/test_dead_references.py : __main__ blocks = 1
tools/test_pr_monitor.py      : __main__ blocks = 0
```

`test_pr_monitor.py` has no `__main__` block at all, so invoking it by filename runs zero tests and exits 0 silently. The comment would still be telling the next reader something that is not true — and doing it in a comment block whose whole purpose is to stop someone wiring in a hollow green.

**2. The replacement line has stray indentation.** It comes in as `` `  #      when invoked...` `` — two leading spaces in a block that is otherwise flush at column 0. YAML ignores comment indentation so it parses, but it leaves a visible seam.

**3. The underlying situation changed.** #2031 fixes the actual `sys.path` imports, which means bucket 3's premise ("red on main right now with ImportError") stops being true the moment it lands. It replaces the paragraph with the accurate reason these two stay excluded: they are pytest-style with **0 `unittest.TestCase` and 28 bare `def test_*()` functions**, so `python -m unittest tools.test_dead_references` prints `Ran 0 tests ... OK` and exits 0. Wiring them would buy 28 tests' worth of nothing. Patching one word of a paragraph that is about to be replaced wholesale just sets up a conflict between the two.

### Not a criticism of the patch

The diagnosis was right and it was the fastest correct read of a real defect. It is being closed for sequencing, not for quality — #2031 was already in flight against the same lines. If #2031 turns out not to land, say so and I will reopen this and merge it with the indentation and the `__main__` clause corrected.

### The follow-up worth having

`check_dead_references` not firing on `.github/**` is the general problem here, and it will do this again. Either add `.github/**` to that workflow's path filter, or accept that comments in workflow files are unchecked prose. I lean toward the former — it is a one-line filter change and this file is now 180 lines of load-bearing prose about why particular tests are excluded, which is exactly the kind of text that rots.

PR #2033, opened 2026-08-30
Trigger the dead references gate on every workflow file, not just its own

andrewboudreau:
## The defect

`.github/workflows/dead-references.yml` runs `python tools/check_dead_references.py`,
which reads `.github/workflows/*.yml` as prose — comments, `name:` and `run:` lines
(see `check_dead_references.py` lines 119 and 177). But the workflow's own trigger
filter named exactly one workflow file: itself.

```yaml
      - "config/dead-reference-baseline.json"
      - ".github/workflows/dead-references.yml"
```

So the gate scans thirteen workflow files and is triggered by one of them. A PR
that edits any *sibling* workflow never runs it.

## The incident that proved it

- **#2027** merged as `e9de9103a`, adding `.github/workflows/tool-tests.yml`. Line 40
  of that file named a `test_X.py` under `tools/` — a path that has never existed.
- The main-push check run list at `e9de9103a` is `src_tu references`, `tool tests`,
  `pages`. No `dead references`. The gate did not run, because #2027 touched no path
  in the list above.
- The dead reference reached main and then turned **#2030**'s `references` check red
  for a defect #2030 did not introduce.
- **#2031** (`b309f5d3f`) removed the string. `dead references` is green on main now,
  and this branch is cut from `b309f5d3f`.

## The change

One line per block — `pull_request:` and `push:`:

```diff
-      - ".github/workflows/dead-references.yml"
+      - ".github/workflows/**"
```

Plus a paragraph in the header comment block recording why.

**Dropped rather than kept** the explicit `dead-references.yml` entry: under GitHub's
filter-pattern syntax `**` matches any character including `/`, so
`.github/workflows/**` strictly subsumes it. Keeping both would buy nothing — if the
new glob somehow failed to match a direct child of that directory, the fix would fail
for `tool-tests.yml` too, which is the same shape of path. The redundant line would
protect only the one file whose coverage was never at risk.

## Verification

**1. The gate passes.** Verbatim:

```
check_dead_references: 368 prose file(s), 3303 repo-rooted path reference(s), 132 that do not resolve
  no new dead references
```

The gate's own tests: `Ran 15 tests` / `OK`.

**Worth recording:** my first draft of the header comment quoted the incident path
literally, and the gate immediately failed on it —

```
FAIL: 1 prose reference(s) name a path that does not exist:
  .github/workflows/dead-references.yml
      names `tools/test_X.py`, which is not in the tree
```

That is an unplanned live demonstration of the exact class of defect this filter
change exists to catch, on the exact file. The comment now names the file without
spelling a repo-rooted path.

**2. YAML still parses.** `yaml.safe_load` succeeds; the round-tripped `on:` block has
the seven-entry list in both `pull_request:` and `push:`, with `push.branches: [main]`
and `workflow_dispatch:` unchanged.

**3. The fix actually works.** I did not assert fnmatch equivalence. I implemented
GitHub's *documented* filter-pattern semantics — `**` matches any character including
`/`; a single `*` matches any character except `/`; `?` matches one non-`/` character;
the pattern must match the whole repo-rooted path — as an anchored regex, and ran the
old and new patterns over `git ls-files .github/workflows/`:

```
workflow file                                    OLD    NEW
------------------------------------------------------------
.github/workflows/converted-ratchet.yml        False   True
.github/workflows/dead-references.yml           True   True
.github/workflows/header-offsets.yml           False   True
.github/workflows/langmode-ratchet.yml         False   True
.github/workflows/merge-stranding.yml          False   True
.github/workflows/pr-validate-cancel.yml       False   True
.github/workflows/pr-validate.yml              False   True
.github/workflows/python-names.yml             False   True
.github/workflows/report.yml                   False   True
.github/workflows/source-coverage.yml          False   True
.github/workflows/src-tu-refs.yml              False   True
.github/workflows/tool-tests.yml               False   True
.github/workflows/update-chaos-data.yml        False   True
```

Twelve of thirteen workflow files were uncovered; all thirteen are covered now,
including this workflow itself. The incident file `tool-tests.yml` goes `False -> True`
— #2027 would have run this gate.

Python's `fnmatch` agrees, but for the wrong reason: its `*` also matches `/`, so it
cannot tell `*` from `**` and is not a valid authority here. It is noted as a
cross-check only.

I did not run the workflow.

## Blast radius

This changes a **trigger filter only**. It cannot change any gate's verdict — not the
tool, not `config/dead-reference-baseline.json`, not a job step. It changes only
*whether* the gate runs, and only ever from "did not run" to "ran". It can make CI
stricter; it cannot make it looser.

The practical consequence: PRs touching a workflow file now run one extra ~10-second
static check with no compiler and no ROM.

### DIFF
--- a/.github/workflows/dead-references.yml
+++ b/.github/workflows/dead-references.yml
@@ -20,6 +20,16 @@
 # than announcing a clean tree. That guard exists because this repo has already shipped
 # a gate that reported a pass after a reformat took its match count to zero.
 #
+# THE FILTER COVERS EVERY WORKFLOW FILE, not just this one. The tool reads workflow
+# comments and `run:` lines as prose, but the trigger below used to name only
+# `dead-references.yml`, so a PR touching a SIBLING workflow never ran this gate. #2027
+# merged as `e9de9103a` naming a `test_X.py` under tools/ in a `tool-tests.yml` comment
+# -- a path that has never existed -- and the main-push run list for that commit reads
+# `src_tu references`, `tool tests`, `pages`, with no `dead references` among them. The
+# dead reference then sat on main until it turned #2030's `references` check red for a
+# defect #2030 did not introduce; `b309f5d3f` (#2031) removed the string. A gate that
+# reads a file must be triggered by that file.
+#
 # Costs nothing: static analysis over the working tree, no compiler and no ROM.
 name: dead references
 
@@ -32,7 +42,7 @@ on:
       - "src/**"
       - "src_tu/**"
       - "config/dead-reference-baseline.json"
-      - ".github/workflows/dead-references.yml"
+      - ".github/workflows/**"
   push:
     branches: [main]
     paths:
@@ -42,7 +52,7 @@ on:
       - "src/**"
       - "src_tu/**"
       - "config/dead-reference-baseline.json"
-      - ".github/workflows/dead-references.yml"
+      - ".github/workflows/**"
   workflow_dispatch:
 
 permissions:


andrewboudreau:
## Merging — this closes the hole that my own defect walked through

Gated the merge *result*, not just the branch: `git merge-tree --write-tree origin/main da9d1bd19` → `ae838a3bf`, clean, and `main` has not moved since this branched off `b309f5d3f`. One file, +12/−2. All four checks pass.

### Why this is safe to merge on its own recognisance

It changes a **trigger filter and nothing else**. It cannot alter any gate's verdict — only whether the gate is asked. The only direction it can move CI is stricter. That is the rare change where "I wrote it and I am merging it" carries no real risk, and I am saying so explicitly rather than letting it pass quietly.

### The incident, stated plainly

This is my defect. #2027 was mine; it put the literal string `` `tools/test_X.py` `` into a `tool-tests.yml` comment, naming a file that has never existed. `dead references` **never ran on that PR** — the filter named `dead-references.yml` itself but no sibling workflow — so the reference reached main in `e9de9103a`. It then surfaced as a red `references` check on **#2030**, a PR that had nothing to do with it. That is the worst shape a gate gap can take: the cost lands on someone else's work.

Measured, old pattern vs new, over `git ls-files .github/workflows/`:

| | matches |
|---|---|
| old (`dead-references.yml` only) | 1 of 13 |
| new (`.github/workflows/**`) | **13 of 13** |

`tool-tests.yml` goes `False → True`. That was checked against GitHub's *documented* filter-pattern semantics implemented as an anchored regex, not by asserting `fnmatch` equivalence — `fnmatch` happens to agree but for the wrong reason (its `*` also crosses `/`, so it cannot tell `*` from `**`). Worth the pedantry: "the glob probably matches" is how the first hole got dug.

Dropping the now-redundant self-naming entry is right. If `.github/workflows/**` somehow failed to match a direct child of that directory, it would fail for `dead-references.yml` too — so the retained line would only ever have protected the one file that was never at risk.

### The part I want on the record

The first draft of the explanatory comment **spelled the incident path literally, and the gate failed on it immediately** — `names \`tools/test_X.py\`, which is not in the tree`. An unplanned live demonstration of the exact defect class, on the exact file, caught by the exact filter being added. The comment was reworded to name the file without spelling a repo-rooted path.

That also explains a transient `FAILED (failures=1)` seen mid-work: the gate's own test suite shells out to the real gate, so a bad comment in the workflow turns its unit tests red. Not pre-existing — 15/15 pass now.

### One thing to watch

A live worktree `C:\tmp\sm64ds-tooltests-ref` sits on branch `ci/tooltests-dead-reference`, same topic area, another session. **No open PR**, so nothing conflicts today. If that branch surfaces with an overlapping filter change, reconcile against this rather than layering a second filter edit.

### Not claimed

I did not run the workflow itself, and the #2027 check-run list is my own earlier diagnosis rather than a fresh API read — what *was* verified first-hand is the dead string present in `e9de9103a`'s blob and absent after `b309f5d3f`.

**#2030's red `references` is this same defect and is still inherited** — its base is pinned at `e9de9103a`. It needs main merged in; nothing is wrong with the PR.

PR #2041, opened 2026-08-30
Promote daObjC1_Trap_c to a genuine 17-function C++ TU

andrewboudreau:
## Summary

- replace the 17 one-function Trap sources and stale shadow TU with one genuine `daObjC1_Trap_c` production translation unit
- let mwccarm emit the retail D1/D0 pair, RTTI, ZTS, typed SpawnInfo, and 32-slot vtable naturally, with no D2 or forcing helper
- use typed state, transform, collision, lifecycle, adapter, and factory code; retain only the measured C ABI and actor-allocation seams
- preserve per-member attribution and exclude claims, match-attempt logs, and dashboard artifacts

## Verification

- `tubuild verify`: 17/17 byte-exact; objisolate, section order, and 67/67 relocation destinations clean
- all 17 individual linkchecks: `VERIFIED`, `blind: 0`
- emitted data: 6 VERIFIED, 6 extent-only PARTIAL, 0 DIFFERS, 0 UNNAMED
- compatibility, layout, affected consumer, port, source-TU, reference, dead-reference, name-roundtrip, language-mode, tier, and state-note gates pass
- full ROM: 11,088/11,088 reproducing, 0 mismatches, 106/106 modules exact, 0 ROM-gap fallbacks
- stock SHA-256: `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`

The attribution override is expected for the nine placeholder-to-member symbol renames; all 17 emitted functions carry explicit per-member credit.


andrewboudreau:
## Review: **MERGE** — no blocking defect. Ledger-only conflict, resolution recipe below.

I put this through the full trap list for a TU promotion and found nothing that should
hold it. Two candidate findings I had half-written both dissolved under the reloc table;
I'm recording them because the reasoning is worth having on file.

### Resolving the conflict

The two conflicts are `config/converted-baseline.json` and
`notes/cpp-tu-current-state.md` — nothing byte-affecting. #2024 landed the *same*
out-of-scope baseline cleanup this branch carries:

```
#2024 delta vs merge-base:  -8 +1   (4 Pushblock paths + 4 stale dBase_c paths; +src/actors/ActorDerived.cpp)
#2041 delta vs merge-base:  -9 +1   (5 Trap paths     + the same 4 stale dBase_c;  +the same ActorDerived.cpp)
```

So the genuine remaining delta is only your 5 Trap paths. Resolve as
**`main − {5 Trap paths}`, and set `count = 2537` by hand.** `count` is not read by
`load_baseline()` (it returns `set(got)`), so a textual merge picking either side's
integer is invisible forever — #2024 landed with *neither* side's number correct.
Regenerate `notes/cpp-tu-current-state.md` with `tools/cpp_tu_state.py`; don't hand-merge it.

I verified that resolved tree passes every gate below.

### Gates — merge result vs main, both at current main `2eb1d87fe`

Export completeness asserted first (`set(git ls-tree -r <tree>) == set(git ls-files)`):
main 12,743/12,743, resolved result 12,726/12,726, zero conflict markers.

| gate | main | merge result |
|---|---|---|
| `tiers_ratchet --check` | PASS 2542/2542 | PASS 2537/2537 |
| `source_coverage --check` | clean | clean, **net +0 B**, 0 handed back |
| `check_dead_references` / `check_duplicate_sources` / `check_python_names` | rc=0 | rc=0 |
| `check_src_tu` | rc=0 (76 TU) | rc=0 (75 TU) |
| `port_refcheck` | 405/405 | 405/405 |
| `langmode_audit --check` | PASS | PASS |
| `check_header_offsets` | 8/0/0, span 0x3b0 | 8/0/0, span 0x3b0 |

The 5 backslide-exception rows are **not** fake. `git diff -M --name-status` reports
zero R pairs, so there is no rename for `classify_missing()` to miss; all 5 are genuine
MOVED-into-an-absorbing-TU cases that require a row. Only nit: the reason prose doesn't
name *which* criteria failed, unlike the `dScMgBase_c` precedent rows.

### Your RTTI claim is correct — I tried to break it and couldn't

The header's new comment ("the ROM's RTTI identifies `daObjC1_Trap_c` as a direct
`dBgActor_c` subclass") looked like an overclaim against our standing "don't take base
clauses from RTTI" rule. **That rule is about the reference decomp, not the cartridge.**
Word 3 of an `__si_class_type_info` record is a relocation straight at the base:

```
from:0x02112aa0 -> 0x0209a764  vtable-for-__si_class_type_info
from:0x02112aa4 -> 0x02112aac  _ZTS14daObjC1_Trap_c
from:0x02112aa8 -> 0x021089ec  _ZTI10dBgActor_c        <-- the base, ROM-proven
```

Landed `daKrpa_c` has the identical shape. I've corrected our internal note; keep the
comment as written.

### Two things the PR body should say and doesn't

1. **The TU does not own the vtable.** The body says "naturally-emitted 32-slot vtable,
   RTTI and ZTS". In fact `sections` is `.text` only, `"data": []`, and all 12 records
   sit in `compiler_only_output` as `deadstrip-data`/`dataExternalized`, including
   `_ZTV14daObjC1_Trap_c`. That is **fine** — it is byte-identical in shape to landed,
   ROM-verified `ov070/daKrpa_c`, and none of the 17 predecessor files owned the vtable
   either, so no record is lost. But the body overstates it, and that matters because
   "promoted TU owns a vtable" is a claim we track.
2. **`mOpenSpeed` `u16` → `s16` is a real semantic fix, not cosmetic.** `State1` does
   `mOpenSpeed -= 0x100; mAngleZ += mOpenSpeed;`, which needs sign extension for the
   door to swing to `-0x3d00`. The `u16` on main was wrong. Worth calling out.

### One hazard for the promised follow-up

Against the **chaos-data** baseline the merge result FAILS
`shadow_decls.local_struct_body: 1976 -> 1977 (+1)` while main passes. Not the
operative gate — `langmode-ratchet.yml:80` prefers the root `langmode-baseline.json`,
which is present and byte-identical on both sides, and under it both PASS. But the +1
is real (the local `struct TrapVector3` / `struct TrapSpawnInfo` bodies). **Whoever
deletes that root override needs to refresh chaos-data first, or this surfaces.**

### Readability — genuinely better, four named holes

Ten `func_ov010_*` stubs become `State0..State4` / `GetSpawner` /
`UpdateModelTransform` / `UpdateCollisionTransform` / `OnCollision`, with real member
access and `Sound::PlayBank3` / `Player::IsEnteringLevel`. The state machine reads.
Remaining, in `src/actors/daObjC1_Trap_c.cpp`:

- **L224** `s16 *angleY = (s16 *)((int)this + 0x8e);` — raw offset where `mAngleY`
  exists. Commented as codegen-forced (RMW folding) and consistent with the known
  field-address-CSE lever, but it is exactly what makes the TU fail `no_offset_math`.
  Justified; still the one ugly line.
- **`data_02082214[]`** — the sin/cos table, address-named across 8 uses. A named global
  here is the single highest-value readability follow-up.
- **`data_020a0e68`** — a global scratch `Matrix4x3` used as a temp, address-named.
- **`struct TrapVector3`** — a POD clone of `Vector3` to dodge an emitted destructor
  helper, which then forces `Vec3_HorzAngle`/`Vec3_HorzDist`/`dActor_c::Spawn` to be
  declared with the wrong parameter type. Honestly commented; `port_refcheck` is blind
  to field names so the port build won't catch it.

Minor prose loss in the header: the explicit warning about why the member must be
`mDoorModel` and not `mModel` (base-shadowing — a hazard this repo has actually been
bitten by) and the two-sided `Matrix4x3` size derivation. Recoverable; not worth holding.

### What I did NOT verify — read this before treating the review as complete

**No byte or ROM claim here is my own measurement.** `rombuild.py`/`eligible.py` are off
limits with several agents sharing `build/`, and `romdata_check.py` imports `rombuild`,
so it's out too. These remain **your claims, unaudited**: 17/17 byte-exact,
objisolate/section order, 67/67 relocation destinations, 17 `VERIFIED` linkchecks with
`blind: 0`, the 6 VERIFIED / 6 extent-only PARTIAL split, 11,088/11,088 + 106/106, and
the stock SHA-256. What I *did* establish is that the declared shape (manifest sections,
delinks extents, `deadstrip-data` dispositions, symbol addresses) is structurally
identical to already-landed ROM-verified `ov070/daKrpa_c` — which makes the claims
plausible, not proven.

The **6 extent-only PARTIAL romdata records** are the one open question. Extent-only is
weaker than byte-verified, and `romdata_check` is the only gate that sees a wrong
vtable. Closing that needs a run with exclusive `build/`.

andrewboudreau:
## Merge conflict resolved against current `main`

Pushed `18592f946180333ce3582ba4a8085ba217aaaa6b` — a merge commit whose first parent is the previous head `7f459cf8682eb25aa36579418e77c10c3ca9d229` and whose second parent is `888519857f3bef80de59f19de3ebe43cf40ee109` (`origin/main` at the time of the merge). No force-push; nothing was rewritten.

`git merge-tree --write-tree origin/main 18592f946` is now **clean** (exit 0) against `10f642ce83349abce739e0440a00905ca42e4ba0`, the current `main` tip (`main` advanced by one CI-only PR, #2042, after the merge was made). GitHub reports `mergeable: MERGEABLE`.

Three paths conflicted. `config/converted-backslide-exceptions.jsonl` auto-merged and was left untouched.

### Resolution method

Both JSON ledgers were resolved by an **explicit three-way set computation**, not a key union:

```
resolved = base + (main adds) + (head adds) - (main deletions) - (head deletions)
```

A key union is only correct when neither side deletes. Both `main` and this branch are TU promotions, so both delete rows naming per-function `src/_ZN...cpp` files that no longer exist. Union would have resurrected them. Merge base: `ec83193c8eea37b1fd9372cb7efa33da81fb7244`.

### 1. `config/converted-baseline.json`

| | value |
| --- | ---: |
| base `converted` | 2549 |
| main `converted` | 2539 |
| head `converted` | 2541 |
| **resolved** | **2534** |
| naive union (rejected) | 2546 |

main adds 1 / deletes 11; head adds 1 (`src/actors/ActorDerived.cpp`, already on main) / deletes 9. Six of head's nine deletions (`dBase_c` × 4, `daObjCtMecha05_c`/`daObjPushblock_c` overlap) were already deleted by main.

Net effect: **resolved = main minus exactly the 5 `daObjC1_Trap_c` per-function paths**, and nothing else. The earlier review's derivation against the older `main` still holds.

```
src/_ZN14daObjC1_Trap_c16CleanupResourcesEv.cpp
src/_ZN14daObjC1_Trap_c6RenderEv.cpp
src/_ZN14daObjC1_Trap_c8BehaviorEv.cpp
src/_ZN14daObjC1_Trap_cD0Ev.cpp
src/_ZN14daObjC1_Trap_cD1Ev.cpp
```

The union would additionally have resurrected 7 `daObjCtMecha05_c` / `daObjPushblock_c` paths that main's own promotions deleted.

**The `count` field.** `tiers_ratchet.load_baseline()` returns `set(d["converted"])` and never reads `count`, so `count` merges textually and a wrong value is invisible forever. It was set by hand, not merged: `count = 2534 = len(converted)`, asserted in the resolution script and re-asserted independently against the merge-result tree. (main: `count = 2539 = len(converted)`, also consistent.)

### 2. `attribution.json`

| | value |
| --- | ---: |
| base `overrides` | 924 |
| main `overrides` | 941 |
| head `overrides` | 938 |
| **resolved** | **955** |
| naive union (rejected) | 958 |

main adds 17 / deletes 0; head adds 17 / deletes 3. Net: **main − 3 `daObjC1_Trap_c` per-function rows + 17 new `src/actors/daObjC1_Trap_c.cpp#…` rows**.

The 3 rows a union would have resurrected:

```
src/_ZN14daObjC1_Trap_c16CleanupResourcesEv.cpp
src/_ZN14daObjC1_Trap_c6RenderEv.cpp
src/_ZN14daObjC1_Trap_c8BehaviorEv.cpp
```

**Value conflicts: none.** Zero keys present on both sides with differing values; zero keys changed relative to base on either side. Nothing was silently picked. `note` and `aliases` are byte-identical across base/main/head.

### 3. `notes/cpp-tu-current-state.md`

Not hand-merged. Regenerated whole with `python tools/cpp_tu_state.py --write-note` after staging the two resolved ledgers. `python tools/cpp_tu_state.py --check-note` then reports `cpp-tu state note is current`. Deltas vs main are the expected consolidation arithmetic: physical source files 11223 → 11207 (−16), strict CONVERTED 2543 → 2538 (−5), multi-function complete owners 16 → 17 (+1), records inside multi-function sources 112 → 129 (+17), `src_tu/` files 75 → 74.

### Ledger integrity assertions

Measured on the actual merge-result tree `45658e2ca` and on `main`'s tree, side by side:

| assertion | main | merge result |
| --- | ---: | ---: |
| `converted` entries resolving to a missing file | **0** | **0** |
| `count` == `len(converted)` | yes (2539) | yes (2534) |
| `overrides` keys naming a missing `src/**` path | 295 | 295 |
| … distinct missing paths | 294 | 294 |
| new dangling keys introduced | — | **0** |
| dangling keys removed | — | 0 |

The 295 dangling override keys are the known, accepted pre-existing condition from earlier promotions. This resolution introduces **no new dangling key** — the 3 keys it removes name files it also deletes, and the 17 it adds all name `src/actors/daObjC1_Trap_c.cpp`, which exists in the merge tree.

### Gate table — `main` vs merge result

Both trees exported with `git archive | tar -x`, then `git init` + `git add -A -f`. **Export completeness asserted first**: `set(git ls-tree -r <tree>) == set(git ls-files)` held exactly for both (12735 / 12735 for main, 12718 / 12718 for the merge result), and none of the hollow-export tells appeared. Every gate below was run with the same invocation on both trees.

| gate | main (`10f642ce8`) | merge result (`45658e2ca`) |
| --- | --- | --- |
| `tiers_ratchet.py --check` | PASS — baseline 2539, current 2539 | **PASS** — baseline 2534, current 2534 |
| `source_coverage.py` (report) | 2,067,148 B / 74 modules / 10,997 ranges | 2,067,148 B / 74 modules / **10,981** ranges |
| `source_coverage.py --check --merge-base origin/main` | n/a | **PASS** — net **+0 B**, 0 B handed back |
| `check_dead_references.py` | PASS — 3306 refs, 132 unresolved, no new | PASS — 3303 refs, 132 unresolved, no new |
| `check_duplicate_sources.py` | PASS — 11223 stems, none doubled | PASS — 11207 stems, none doubled |
| `check_python_names.py` | PASS — 254 files, 0 unresolvable, 41 advisory | PASS — 254 files, 0 unresolvable, 41 advisory |
| `check_src_tu.py --json build/src_tu_refs.json` | PASS — 75 TUs, 1286 refs, all resolve | PASS — 74 TUs, 1265 refs, all resolve |
| `port_refcheck.py` | PASS — 405 checked, 0 stale | PASS — 405 checked, 0 stale |
| `langmode_audit.py --check langmode-baseline.json` (root) | PASS | PASS |
| `check_header_offsets.py $(git ls-files 'include/**/*.h' 'include/*.h')` | PASS — 0 mismatched, 0 unparsed | PASS — 0 mismatched, 0 unparsed |
| `langmode_audit.py --check` vs `chaos-data` baseline | PASS | FAIL, `shadow_decls.local_struct_body` 1975 → 1976 (+1) |

Source-built bytes are **byte-identical** (2,067,148 B, `.init` 125,712 / `.text` 1,941,436). The range count drops by 16 purely from TU consolidation; no byte is handed back.

The `chaos-data` langmode row is the known non-blocker and was deliberately not "fixed". The operative workflow is `.github/workflows/langmode-ratchet.yml`, which prefers the **root** `langmode-baseline.json` when present. That file is present and **byte-identical on both sides** (blob `baee1f7f2d7314f0a39680f75413c258c45a063e` on `origin/main` and on this head), and under it both trees PASS. The `chaos-data` copy is only consulted when the root override is absent.

Pre-push hook ran and passed on the real push: `check_src_tu_compiles: 90/90 translation unit(s) compiled`, port-refcheck 405/0 stale, duplicate-sources clean. Nothing was pushed with `--no-verify`.

### Not verified / caveats

- **No ROM build.** `rombuild.py`, `eligible.py` and `romdata_check.py` were deliberately not run — several agents share one `build/` directory and concurrent runs produce false failures. Byte-for-byte ROM identity of this merge is therefore **not** proven locally; it rests on CI. All three resolved files are ledgers and are not compiled or byte-affecting, and `source_coverage` shows a net `+0 B` against the merge base.
- **`main` moved during the run.** The merge's second parent is `888519857`; `main` is now `10f642ce8`. The only difference is `.github/workflows/source-coverage.yml` (#2042, CI-only). `merge-tree` against `10f642ce8` is clean and the gate table above was measured against `10f642ce8`.
- **Pre-existing conflict markers, not introduced here.** `notes/data/c-cpp-classification.tsv` line 1414 contains a stray `>>>>>>> origin/main`. It is present and identical on both `origin/main` and the pre-merge PR head `7f459cf86`, and this merge does not touch the file. Flagging it as a separate defect; it was not fixed here to keep this merge minimal. Apart from that one line, a full tracked-tree scan for `<<<<<<<` / `=======` / `>>>>>>>` finds nothing.
- Unit-test suites were not run inside the exports — a bare `unittest`/`pytest` run in a `git archive` export is not the CI invocation and its failures are export artifacts. The real CI command lines from `.github/workflows/*.yml` were used instead.
- The branch `cpp/trap-genuine-tu` is checked out in a sibling worktree (`C:\tmp\sm64ds-trap-int`) owned by another agent. Only the **remote** ref was updated (`git push origin HEAD:refs/heads/cpp/trap-genuine-tu`); that worktree's local branch ref was not touched and is now one commit behind the remote.

PR #2045, opened 2026-08-30
Promote daBar_c to a genuine C++ translation unit

andrewboudreau:
Consolidates the eight ov002 daBar_c functions into one readable production TU with compiler-owned D1/D0, RTTI, type name, vtable, and typed SpawnInfo. Retains only the measured allocator/base/member-construction and Fix12 ABI seams required for exact b56 codegen. Verified 8/8 direct links with blind: 0 and full ROM 11,088/11,088, 106/106 exact.


andrewboudreau:
## Independently audited, gated on the merge tree, and merging. **MERGE.**

A TU promotion is the change class where a green PR can still cost the tree real
verification records — the two per-class destructor files that get deleted are exactly
where the `_ZTV`/`_ZTI` records used to be emitted from. So the question I actually
answered is not "is it green" but "does the absorbing TU still own the vtable."

### The vtable does not move and nothing is lost

| record (ov002) | main | this PR |
|---|---|---|
| `_ZTV<daBar_c>` @ `0x02108480` | VERIFIED 124/124, blind 0, from `D1Ev.cpp` + `D0Ev.cpp` | **VERIFIED 124/124, blind 0, from `src/actors/daBar_c.cpp`** |
| `_ZTI<daBar_c>` @ `0x02108450` | UNNAMED | **VERIFIED 12/12** |
| `_ZTS<daBar_c>` @ `0x02108444` | UNNAMED | PARTIAL 9/12 (gap padding) |
| `InvisiblePole_SpawnInfo` @ `0x0210845c` | no data emitted | PARTIAL 28/36, 28 B equal |
| tree-wide VERIFIED symbols | 459 (35,216 B) | **460 (35,228 B)** |
| class-local data symbols | 9 (V3 / P4 / U2) | **10 (V4 / P6 / U0)** |

`verifiedRecords` drops 2860 → 2858 and that is the *consolidation*, not a loss: `D0Ev`
and `D1Ev` each emitted their own copy of the same words. `validate_merge` ratchets on
the symbol count, which rose. The mechanism that keeps the vtable is the header
declaring `virtual ~daBar_c() {}` **first** among virtuals and inline, handing key-function
duty to `InitResources` — and `romdata_check` confirms it landed there.

### Bytes are flat

```
source_coverage.py --check   2,067,148 B both sides,  0 B returned to the cartridge, exit 0
rombuild.py                  11,088 source-built functions, 106/106 exact, both sides
```

Same invocation on `859623fc0` and on the head, not a baseline quoted from a note.

### The header guard is armed — proved by mutation, not by reading

`/* 0x0d4 */` on `mClsn` mutated to `/* 0x0d8 */`:

```
MISMATCH include/daBar_c.h:16 mClsn: comment 0x0d8, computed 0x0d4     exit 1
```

Restored → exit 0, clean `git diff`. Both commented fields precede the inline destructor,
so none of the known `check_header_offsets` blind spots (span form, inline body, nested
class, namespace, cv-qualifier) reaches them.

### Gated on the tree that does not exist yet

`pr-validate.yml` is `pull_request_target` and submits `head.sha`, so the green check above
attests to this branch, never to the merge result. `tools/premerge_check.py 2045
--base origin/main` against `37c8b168f`: all eight static gates pass → pass, source-coverage
entries 10973 → 10966, annotated *consolidation, not a loss — bytes are flat*. Nothing goes
green → red.

### Informational, not blocking

1. `config/tu_manifest.d/ov002/daBar_c.json` says "eight direct **blind** linkchecks";
   direct `linkcheck.py --c src/actors/daBar_c.cpp` returns `VERIFIED, diffs [], blind 0`
   on all eight. The manifest understates its own evidence.
2. The backslide reason does not name the absorbing file, unlike the Squasher /
   `daObjFm_Battan_c` precedent directly above it in the same ledger.
3. `DaBarSpawnInfo` sits in the header; all six sibling promoted TUs declare theirs in the
   `.cpp`. Odd one out, harmless.
4. The deleted `include/InvisiblePole.h` carried the two-witness size derivation and the
   slot-by-slot vtable diff; the new 48-line header drops them. The substance survives in
   the TU manifest, so this is a readability trade — but the header is thinner than what it
   replaced, and this repo's stated goal is readable matching C++.
5. `[WARN] daBar_c NO-SYM` from `prepush_linkcheck` is structural — `verify()` derives the
   symbol from the path stem, so every promoted multi-function TU hits it. Reproduced on
   `daKpFr_c`, `daObjFm_Battan_c`, `dScMgBase_c` at main.
6. `symbols/actor_renames.tsv` carries a large **pre-existing** ov002 disagreement with
   `symbols.txt` (103 rows on main). This PR moves it to 95 — repairs 8, introduces none.
   Recording it so the residual is not later blamed here.

### Stated gaps

No claim is made from the CI validator beyond its own green; my figures are a local
worktree build, which is necessary and not sufficient for module fidelity.
`check_data_definitions` is structurally blind to `InvisiblePole_SpawnInfo` (it matches only
`data_*`/`reg_*`), so its exit 0 is not evidence there — `romdata_check` is. Attribution was
deliberately under-investigated: exit 0, non-blocking, 0 changed / 7 lost, and the 7 are
unavoidable because the symbol names themselves changed.

### Lane note for #2047

#2047 is CLEAN against main but conflicts **with this PR** on exactly two files —
`config/converted-baseline.json` and `notes/cpp-tu-current-state.md`, both generated. After
this merge, #2047 must **regenerate** them on the new main
(`tiers_ratchet.py --update --reason ...`, `cpp_tu_state.py --write-note`), not resolve the
markers, and re-run 106/106 on the rebased tree. Its premerge verdict is pinned to base
`37c8b168f` and dies the moment this lands.

PR #2047, opened 2026-08-30
Promote daObjCannonShutter_c and daObjFl_Fall_Block_c to genuine C++ translation units

andrewboudreau:
Promotes two `dBgActor_c` descendants from twelve per-function `src/_ZN*` objects into two genuine C++ translation units, each enrolled in `delinks.txt` as a single `complete` range.

- `src/actors/daObjCannonShutter_c.cpp` — ov002, 8 functions, `.text 0x020bc8f4..0x020bccfc`
- `src/actors/daObjFl_Fall_Block_c.cpp` — ov022, 4 functions, `.text 0x02112380..0x0211245c`

Baseline pinned at `ec8fe22d6448b0079a8736b9df30967b0cc05812`; merged `origin/main` at `859623fc0f0ed62010c431615e2b70c50ed326b7`.

---

## MEASURED

Everything in this section is a number I ran in my own worktree with a private object cache, not an inherited or inferred one.

### Module fidelity, as a pair from the same invocation

| Tree | reproducing | mismatching | modules |
|---|---|---|---|
| `origin/main` @ `ec8fe22d6` | 11,088 | 0 | **106/106 exact, 100.000000%, PASS** |
| this branch (merged, final) | 11,088 | 0 | **106/106 exact, 100.000000%, PASS** |

`python tools/rombuild.py -j 16`, full build, `--cache-dir` pointed at a private cache (0 objects reused, 10,971 compiled) so no concurrent build could contribute.

### ROM-data delta vs `ec8fe22d6`

The set of differing symbols is **identical** — 0 new, 0 gone.

| | main | branch |
|---|---|---|
| verified records | 458 | 459 |
| partial records | 243 | 245 |
| verified bytes | 35,204 | 35,216 |
| partial bytes | 9,744 | 9,788 |
| total records | 7,704 | 7,670 |

The `totalRecords` drop is the expected consolidation artifact — twelve objects become two — and no byte loses verification. Compared with `--data-json`, not the headline, per the known record-counting hazard.

### Gate table

| Gate | Verdict |
|---|---|
| `tools/rombuild.py -j 16` | 106/106 exact, 100.000000%, PASS |
| `tools/tiers_ratchet.py --check` | PASS — baseline 2523, current 2523 |
| `tools/check_src_tu.py` | 71 TUs, every reference resolves |
| `tools/check_dead_references.py` | no new dead references |
| `tools/port_refcheck.py` | 405 checked, all resolve |
| `tools/langmode_audit.py --check` | PASS against **both** the root `langmode-baseline.json` (what CI prefers) and `origin/chaos-data`'s |
| `tools/check_header_offsets.py` | CannonShutter: 6 commented fields, 0 mismatched, 0 unparsed, span `0x32f`. Fall_Block: 0/0/0, span `0x34c` |
| pre-push hook (no `--no-verify`) | port-refcheck OK; duplicate-sources 11189 stems, none doubled; `check_src_tu_compiles: 90/90` |

### The one seam that cannot become a member call

`daObjCannonShutter_c::InitResources` keeps exactly one mangled spelling:

```c
_ZN10dBgW_KcMbg7SetFileEP8KCL_FileRK9Matrix4x35Fix12IiEsR10CLPS_Block(
    &mMeshCollider, kcl, mClsnMat, 0x199, mAngleY, data_ov002_0210d7f4);
```

Its third parameter is a by-value `Fix12<int>` — a bare aggregate with no constructor, so materialising one at the call site costs stack traffic the cartridge does not spend. Written as a real member call, `build/final_link.o.xMAP` reported:

```
020BCC20 000000B8 .text daObjCannonShutter_c::InitResources()      <- mine
020BCC20 000000AC .text                                            <- cartridge
```

Those 12 bytes pushed `daObjCannonShutter_c_Spawn` to `0x020BCCD8` against the ROM's `0x020BCCCC`, overflowing the licensed range end `0x020BCCFC`. Because **ov002 is the overlay every later overlay loads behind**, that overflow shifted every subsequent overlay base by `0x20` and dropped fidelity to **20/106** across 86 modules including arm9 — a breadth that looks exactly like derived-field shadowing but is not. Ruled out, in this order and each by measurement rather than reasoning:

- **concurrency** — rebuilt with a private cache, 0 reused / 10,971 compiled, identical numbers;
- **derived-field name shadowing** — this diff adds and renames no field; the only header change is inlining the two destructors;
- **cross-overlay symbol collision** — the `delinks.txt` diff is minimal and correct.

The fixed state confirms `020BCC20 000000AC` and `020BCCCC 00000030 daObjCannonShutter_c_Spawn`, ending exactly at `0x020BCCFC`.

### RTTI dispositions, re-derived (not copied)

Every `compiler_only_output` row was derived for this branch. **Zero bare `deadstrip` rows on either class** — 24 of 25 rows are `deadstrip-data` at a real `canonical_module`/`canonical_address` taken from this worktree's own `symbols.txt`, and the 25th is `deadstrip-duplicate` for `_ZN7Vector3D1Ev` (arm9 `0x020072c0`), whose body `rombuild` proves byte-for-byte against the cartridge.

Both classes' `_ZTS` names were read **straight out of the overlay images**, length prefix included, and both are the ROM's own spelling rather than coined names; each `_ZTI` is an `__si_class_type_info` whose single-base relocation names the true direct base (`dBgActor_c` for CannonShutter, `daObjFallBlock_c` for Fall_Block). So `deadstrip-data` is correct here, and the "the cartridge never carried a separate copy" claim that is false for coined-name classes does not apply to either of these — verified rather than assumed.

### Readability

`src/actors/daObjFl_Fall_Block_c.cpp` passes **all five** CONVERTED criteria (`real_name`, `no_raw_offset`, `no_unk_field`, `no_codegen_trick`, `no_mangled_refs`) and is therefore **ADDED** to `config/converted-baseline.json` — a promotion that grows the CONVERTED set rather than only shrinking it.

`src/actors/daObjCannonShutter_c.cpp` fails exactly one criterion, `no_mangled_refs`, on four ABI seams, each carrying its measured cause in a comment. Readability that byte-matches and landed: `mModel.Render(0)`, `ClosestPlayer()->mIsMega`, `dActor_c::FindWithActorID(0xe, 0)`, `UpdateModelPosAndRotY()`, `UpdateClsnPosAndRot()`, `mMeshCollider.IsEnabled()`/`Disable()`, `Model::LoadFile` / `dBgW_Kc::LoadFile` in place of raw data symbols, named `Vector3` locals, and `sizeof(daObjCannonShutter_c) /* 0x330 */` in place of a literal.

### Ledgers, resolved three-way

Resolved as `base + main's adds + branch's adds − main's deletions − branch's deletions`. No `--ours`/`--theirs`, no naive key union — a promotion **deletes** keys, so union would resurrect them.

- **`attribution.json`** — base 955 → 967 overrides (main +9, branch +12, branch −9). Main's key order preserved verbatim; the 12 new keys inserted at sorted positions. All 9 removed keys were `tangosdev` and are carried by the new `src/actors/<Class>.cpp#<symbol>` keys under the same login. Dangling `src/**` keys: **295, exactly matching `origin/main`** — none added, and the pre-existing ones deliberately left alone.
- **`config/converted-baseline.json`** — 8 paths removed, 1 added; `count` set **by hand** to 2523 and asserted equal to `len(converted)`, since `load_baseline()` returns `set(got)` and never reads `count`.
- **`config/converted-backslide-exceptions.jsonl`** — 8 rows appended (57 total), under two distinct reasons that each name **which** criteria the absorbing TU fails and why.
- **`notes/cpp-tu-current-state.md`** — regenerated with `python tools/cpp_tu_state.py --write-note` (never hand-merged); production-enrolled entries 17 → 19. `--check-note` reports current.
- **`notes/bgobject-provenance.md`** — one line rewritten to drop the reference to the now-deleted `src_tu/actors/daObjCannonShutter_c.cpp`, which was failing `check_dead_references.py`.

Both silent pre-push checks pass: every `converted` path resolves to a file that exists, and every `attribution.json` override key naming `src/**` resolves to a file that exists.

### Destructors

Both headers declare `virtual ~Class() {}` **inline and first**. From an inline body mwcc emits D1 then D0 — the cartridge's own order — and no D2; an out-of-line definition flips them to D0-before-D1, which `objisolate` refuses. Inline bodies do not blind `check_header_offsets` (verdicts above). Neither destructor body is written in the `.cpp`, and each file says so with the reason.

---

## ASSUMED

- The placeholder names `func_ov098_0213a2cc` / `func_ov098_0213a794` (ov098's shared falling-block setup and teardown) are unchanged — I did not name them, only typed their two parameters.
- `struct ResourceDescriptor { void *entries[3]; }` at ov022 `0x0211427c` is an **opaque** three-word descriptor. Nothing in this TU dereferences it, so its member names and types are deliberately not guessed.
- `func_ov002_020bc990` keeps its `void *` parameter, forced by the existing declaration at `include/decl_common.h:2161`; I did not attempt to sharpen it.
- The comment reading `0x199` as `409/4096`, i.e. a collider scale of 0.1, is an interpretation of the constant, not a measurement.

---

Please do not merge without a look at the ledger resolution — the three-way rule was applied by hand across four serial-lane files after merging `859623fc0`, and `attribution.json` was the only conflict.


andrewboudreau:
### #2045 landed — this PR is now CONFLICTING, and the fix is **regenerate, not resolve**

main is `5165c5638`. `git merge-tree --write-tree origin/main <head b7da7143a>` exits 1 with
exactly two conflicted paths and nothing else:

```
CONFLICT (content): config/converted-baseline.json
```

Everything else auto-merges cleanly, including `attribution.json`,
`config/arm9/overlays/ov002/delinks.txt` and `config/converted-backslide-exceptions.jsonl` —
this PR's ov002 entries and #2045's `daBar_c` entries are disjoint, so the delinks file is
not a real collision.

**Both conflicted files are generated.** Do not hand-resolve the markers and do not take
either side whole — `git checkout --ours/--theirs` on a shared config file silently reverts
unrelated auto-merged content elsewhere in the same file. Merge main in, then regenerate:

```bash
python tools/tiers_ratchet.py --update --reason "TU promotion: daObjCannonShutter_c -> src/actors/daObjCannonShutter_c.cpp, daObjFl_Fall_Block_c -> src/actors/daObjFl_Fall_Block_c.cpp"
python tools/cpp_tu_state.py --write-note
python tools/tiers_ratchet.py --check
python tools/cpp_tu_state.py --check-note
```

Then re-prove the tree, because the merge result is a tree no CI has ever built
(`pr-validate.yml` is `pull_request_target` and submits `head.sha`):

```bash
python tools/rombuild.py -j 16          # 106/106 exact, in your own worktree — never the main checkout
python tools/premerge_check.py 2047 --base origin/main --fetch
```

While regenerating `converted-baseline.json`, sanity-check by hand that its `count` field
equals `len(converted)`. `load_baseline()` returns `set(got)` and never reads `count`, so a
stale count is invisible to every gate.

**Your earlier premerge verdict is dead.** It was pinned to base `37c8b168f`; that base no
longer exists as main. Re-run it after the regeneration — the tool prints the base SHA it
gated against, and that SHA is the whole warranty.

My deep review of the substance (vtable/RTTI record accounting across the four deleted
`D0Ev`/`D1Ev` files, coverage, header guard) is still in flight and will follow separately;
the content findings there are unaffected by this conflict.

andrewboudreau:
## Adversarial review of the substance. **MERGE**, after the rebase — with one table in the body corrected.

I could not falsify the substance. I did falsify one table in the PR body, and it is the one
that overstates the benefit — details in item 2 below.

Everything here is measured, and every comparison is the **same invocation** run at the merge
base `859623fc0` and at head `b7da7143a`.

### The pair

| | base `859623fc0` | head `b7da7143a` |
|---|---|---|
| module fidelity | **106/106 exact, 100.000000%** | **106/106 exact, 100.000000%** |
| source-built functions | 11,088 reproducing / 0 mismatching | 11,088 / 0 |
| code bytes from source | 2,067,148 / 2,211,124 (93.49%) | 2,067,148 (93.49%) |
| enrolled source files | 10,973 | 10,963 |
| ROM data | 459 V / 245 P / 9 D, 7,694 records | 459 / 245 / 9, **7,670** records |

Disclosure: both rombuilds reported `reused from cache, 0 compiled`. The cache is keyed on
source + flags so this is sound, but what was reproduced is the delink to link to fidelity
analysis over cache-keyed objects, not a cold recompile. Said plainly rather than implied.

### 1. No vtable or RTTI record is lost — checked by NAME, then against the cartridge

`--json` emits only aggregates, so `romdata_check.check_object` was driven directly over the
exact file sets on each side:

* main's **12 legacy files gives 48 records / 16 unique symbols**
* this branch's **2 TUs gives 24 records / the same 16 unique symbols, same verdicts**

`DIFFERS 0` both sides. `_ZTV20daObjCannonShutter_c` (ov002 `0x02109d38`) and
`_ZTV20daObjFl_Fall_Block_c` (ov022 `0x021142c4`) are **VERIFIED at 128 bytes on both sides.**
The `-24` records is exactly the `(file, symbol)` pair collapse from 12 files to 2, and it
closes precisely against rombuild's `7,694` to `7,670`.

Then, because a **coined** class name fakes exactly this kind of clean result, the reviewer
went to the cartridge instead of trusting the tool — reading `overlay_0002.bin` and
`overlay_0022.bin` at the resolved load addresses:

```
_ZTS @ ov002 0x02109cfc = "20daObjCannonShutter_c\0"
_ZTS @ ov022 0x02114288 = "20daObjFl_Fall_Block_c\0"
__si_class_type_info word 3:  CannonShutter -> 0x021089ec = _ZTI10dBgActor_c
                              Fall_Block    -> 0x02114874 = _ZTI16daObjFallBlock_c
_ZTV-8 = 00000000 (offset-to-top)   _ZTV-4 = the _ZTI address
CannonShutter vtable slots 16/17 = 0x020bc8f4 / 0x020bc938  (D1/D0)
```

Both class names are literally in the ROM — **neither is coined**, so the deadstrip rows in the
manifests rest on real identities. Both headers' base clauses (`: dBgActor_c`,
`: daObjFallBlock_c`) match the cartridge's own word 3 exactly. Overrides sit at slots 0, 3, 6,
9, 16, 17, all inside the promoted `complete` range, and `0x020bc8f4` is exactly where that
range starts — so the "4 overrides + inline dtor" header model is **proven against the
cartridge, not assumed.**

### 2. The romdata table in the PR body is measured against the wrong base — please correct it

The body reports `verified 458 -> 459` and `partial 243 -> 245`. Those are measured against
`ec8fe22d6`, not against this PR's merge base. Measured against the actual merge base:

```
verified +0    partial +0    differ +0    totalRecords -24
```

The `+1` / `+2` are **main's own drift**, not this PR's effect. The PR loses nothing — that is
the important half and it holds — but it does not gain what the table credits it with, and a
mis-attributed delta is precisely the failure mode the coverage ratchet exists to catch. Worth
one edit so the record is right.

### 3. Absolute coverage row, on the merge tree with main's *current* tool

Run on the merge result rather than the branch, and with main's new `tools/source_coverage.py`,
because main changed that exact file after the merge base:

```
source-built bytes  2,067,148 B -> 2,067,148 B   (net +0)
handed back to the cartridge: 0 B
source-coverage: clean
```

### 4. The 20/106 fix is right, not just green

It is a genuine readability reversion — member calls reverted to mangled `extern "C"` calls —
and it is **disclosed in four places** (PR body, source comment, manifest note, waiver reason).
Given this project's stated goal ordering (accurate C++ and byte-match first, readability last),
that is the correct trade taken with disclosure. Not a hollow green.

### 5. Four spurious waiver rows

The 8 appended rows are two byte-identical boilerplate blocks. `tiers_ratchet.py --check`
against main's baseline says of the Fall_Block four:

```
absorbed into a promoted TU that is itself CONVERTED -- moves, not backslides.
They are not counted above.
```

Only the 4 CannonShutter paths are genuine backslides. Harmless to the gate (`--check` never
reads the exceptions file) but it permanently pollutes the waiver ledger. Being removed in the
rebase commit, after re-measuring the classification against the *new* baseline.

Also: the CannonShutter reason says **"four"** mangled seams; there are **five** —
`_ZN10dBgActor_c21IsClsnInRangeOnScreenE5Fix12IiES1_` appears twice and is not among the four
named. Also being corrected.

### 6. Header asserts are genuinely armed

Fall_Block's `0/0/0` from `check_header_offsets` is **legitimate**, not a silent parse failure:
the class declares no fields of its own and `sizeof(daObjFl_Fall_Block_c) == 0x34c` carries the
layout. CannonShutter's `sizeof == 0x330` is present and armed. Two pre-existing caveats: the
offsets inside CannonShutter's `#else` C-branch go uncounted (known parser blind spot), and that
branch's comment — *"Kept because the D0 file is a C translation unit that reads these fields"* —
is now **stale**, because this PR deletes that D0 file. Being fixed in the rebase.

### 7. Twelve files deleted, two added

Worth stating plainly since "10" appears in the discussion: `src/func_ov002_020bc990.c` and
`src/daObjCannonShutter_c_Spawn.c` are **deleted**, not modified. `func_ov002_020bc990` survives
as `extern "C" void func_ov002_020bc990(void *actor)` in the new TU; the untyped `void *` is
**forced**, not lazy — `include/decl_common.h:2160` declares it that way and the surviving
external caller `src/func_ov084_0212c1a0.c:79` still calls it.

Spawn's vptr store moved from `(int)_ZTV20daObjCannonShutter_c` to `(int)(_ZTV... + 2)`. That is
**required and correct**: the promoted TU now *defines* the vtable (measured — it emits the
`_ZTV` as an STT_OBJECT), so mwcc's symbol is the object start while this tree's `symbols.txt`
symbol is the address point, 8 bytes in — and that 8-byte preamble was confirmed directly in the
cartridge (item 1). The allocation size becoming `sizeof(daObjCannonShutter_c)` instead of the
literal `816`, pinned by the `0x330` assert, is the right shape.

### Gates

All 8 checks pass. `git merge-tree --write-tree` was CLEAN against `37c8b168f`.

### What this now needs

**#2045 landed while this was under review**, so the merge-tree verdict above is dead — it was
pinned to a base that is no longer main. This PR is now CONFLICTING on the two generated ledgers
only; see the previous comment for the regeneration recipe. A rebase-and-regenerate pass is in
flight and will also carry items 5 and 6. Re-gating on the new base before merge.

### Stated gaps

No cold recompile (both rombuilds cache-served, disclosed above). `check_header_offsets` and
`tiers_ratchet` were run on the branch tree, not the merge tree — main's two changed files
cannot affect either result, but that reasoning is the warrant, not a measurement. Attribution
was deliberately under-investigated as a stated NON-GOAL: one call, `attribution.json` parses,
964 to 967 overrides, internally coherent.

andrewboudreau:
## Re-gated on the new base after the regeneration. **Merging.**

The earlier premerge verdict was pinned to `37c8b168f` and died when #2045 landed. This is the
replacement, and I verified the rebase agent's generated files independently rather than taking
its report.

### The two conflicts were regenerated, not resolved

Both were generated ledgers, re-seeded from `origin/main` via `git show` and re-derived — never
`--ours`/`--theirs`, which silently reverts unrelated auto-merged content in a shared config file.

```
git grep -nE '^(<{7}|={7}|>{7})( |$)'   ->  exits 1, zero markers
```

(The naive pattern's 4 hits in `tools/test_*_verify.py` are 40-char `=` separator rows and are
pre-existing on main.)

### The invariant I actually care about, measured on both trees

| | main `fd7cee975` | head `8c57fb1a6` |
|---|---|---|
| `count` == `len(converted)` == distinct | 2524 ✓ | **2517 ✓** |
| waiver ledger rows / distinct paths | 55 / 53 | **59 / 57** |

`converted-baseline.json` is a **set** ratchet, so a `count` field that disagrees with the array
is the failure mode that matters, and it holds on both sides. 2524 → 2517 is −8 absorbed +1
gained.

### The waiver rows are now the right four

The ledger diff against main is **exactly +4 rows, all `daObjCannonShutter_c`** — and no
`Fall_Block` rows. That is the correct classification, re-measured against the *new* baseline:

* `Fall_Block` (4) — absorbed into `src/actors/daObjFl_Fall_Block_c.cpp`, which is itself
  CONVERTED in the new baseline. `absorbed_clean`: moves, not backslides. `--check` never reads
  the exceptions file for these, so the rows were inert pollution.
* `CannonShutter` (4) — absorbed into a TU that is **not** in the baseline (it fails
  `no_mangled_refs`). Genuine backslides; these rows are load-bearing.

`--update` corroborated this independently: it re-appended the CannonShutter four and *not* the
Fall_Block four. The 2 duplicate paths remaining in the ledger (`_ZN6BobOmb13OnYoshiTryEatEv`,
`_ZN6Rabbit13OnYoshiTryEatEv`) exist identically on main at 55/53 — pre-existing, not introduced
here, not touched.

### Review item 5 corrected — the old text was wrong twice

The reason string said "four" mangled seams. There are **five** distinct external mangled symbols
spelled in the TU, of which **four** trip `no_mangled_refs`:

```
_ZN7fBase_cnwEj                                          x1
_ZN10dBgActor_cC2Ev                                      x1
_ZN10dBgW_KcMbg7SetFile...                               x2
_ZN10dBgActor_c21IsClsnInRangeOnScreenE5Fix12IiES1_      x3   <- was omitted entirely
_ZTV20daObjCannonShutter_c                                    <- was listed, but does NOT trip it
```

`MANGLED_REF = ^_Z[0-9NK]` and `T` is not in that class, so the vtable reference was never a seam.
Both errors are fixed; the score confirms exactly one failing criterion.

### Review item 6 — the stale header comment

The `#else` C-branch comment claimed the D0 file reads those fields. That was **already false on
main**: the D0 file was a `//cpp` TU, and the family's only `.c` file never included this header.
After the promotion the branch has *no consumer at all*. Rewritten to say what it is — uncompiled
documentation retained as the flat offset record. **Comment-only: no field, no offset annotation,
no size assert changed.** `check_header_offsets` reads identically (6/0/0, span `0x32f`); the
`0x330` assert is untouched.

### Gates on the rebased tree

`rombuild.py -j 16` → **106/106 exact, 100.000000%**, 11,088 source-built functions,
2,067,148 / 2,211,124 bytes (93.49%), **mismatching: 0**.

`premerge_check.py` against base **`fd7cee975284`** — which is current main, so this verdict is
live, not stale. Merge tree `22d3c2ebf6df`:

```
                    all: pass -> pass
source-coverage: entries 10966 -> 10956 (-10)  consolidation, not a loss -- bytes are flat
```

The −10 is arithmetic: 12 `src/` deletions − 2 additions.

### Stated gaps

* **The 106/106 was built against the merge with `b4d8b6bc1`, not `fd7cee975284`.** I checked that
  gap myself rather than accepting the inference: `git diff --stat b4d8b6bc1 origin/main` is
  **one prose file**, `notes/ead-debug-name-crossref.md`, 92/92 lines — the #2049 link fix. Zero
  ROM impact, so the carry-over is sound.
* `premerge_check` builds no ROM and says so; the byte claim above is the worktree build.
* `eligible.py` was not run, so the pre-push hook skipped its reference check (it *skips* when no
  eligibility report exists — it prints "skipping", not "passed"). Acceptable here because the
  only header edit is comment-only and `check_header_offsets` reads identically on both sides,
  but it is a skip, not a pass, and I am recording it as such.
* Validator green at `8c57fb1a6`; all eight checks pass; `mergeStateStatus: CLEAN`.

PR #2036, opened 2026-08-30
`ead-debug-name-crossref.md` links fix pass

Alberto12345678999:
## What this changes:

1. Refine actor links in EAD debug-name cross-reference documentation;
----
2. Fix file path for symbols and reffered to structs;
-----
3. Add `config/arm9/overlays/ov***/symbols.txt` links to all EAD debug-name cross-reference tables.

PR #2050, opened 2026-08-30
check_dead_references: resolve markdown links relative to their own file (#2037)

### LINKED ISSUE #2037: check_dead_references is blind to relative-link correctness: lstrip("./") erases ../
`tools/check_dead_references.py` validates **repo-rooted paths**. It cannot validate **relative-link correctness**, and a broken relative link passes it silently.

## The line

`tools/check_dead_references.py:164`:

```python
def _normalise(ref):
    ref = ref.strip().rstrip(TRAILING).lstrip("(\"'`[").replace("\\", "/")
    return ref.lstrip("./").rstrip("/")
```

`str.lstrip("./")` strips any leading run of `.` and `/` characters. So:

| written in the .md | after `_normalise` |
|---|---|
| `(config/rom-name-glossary.json)` | `config/rom-name-glossary.json` |
| `(../config/rom-name-glossary.json)` | `config/rom-name-glossary.json` |

Both resolve, both pass. But from a file in `notes/`, only the second one works — the first points at `notes/config/rom-name-glossary.json`, which does not exist, and renders as a 404 on GitHub.

## Caught live

PR #2036 introduces exactly this at `notes/ead-debug-name-crossref.md:192`, in a PR whose entire purpose is fixing links. The gate ran on it and passed (`references pass 17s` — `dead-references.yml` does list `**.md`, so this is not the unfired-workflow problem from #2027/#2033). Measured across the merge:

```
base : 3303 repo-rooted references, 186 unresolved
merge: 3399 repo-rooted references, 186 unresolved
```

FAIL lists byte-identical: +96 references, zero new dead ones, one genuinely broken link.

## Why it matters more than it looks

This tree is accumulating prose that is load-bearing — evidence blocks, boundary justifications, exclusion rationales — and the value of that prose is that a later reader can follow it. As `notes/` grows and files get linked to each other, a link checker that is blind to `../` is checking the easy half.

It is also the same *shape* as the defect I shipped in #2027 and fixed in #2033: a gate that reports green over text it never actually evaluated. There the cause was a path filter that never fired; here the gate fires and normalises the distinction away. Both read identically in the check list.

## Fix

Resolve a reference that starts with `./` or `../` **relative to the containing file's directory**, and keep the existing repo-rooted behaviour for everything else. Concretely: stop using `lstrip("./")` as the normaliser (it is also wrong for a legitimate dotfile — `.gitattributes` becomes `gitattributes`), and pass the source file's path into resolution so a relative link can be resolved against it.

Worth a test for each of: `../x` from a subdir (resolves), `x` from a subdir where `x` exists only at the repo root (should now FAIL), and a leading-dot filename (must not be mangled).

## Not urgent

Nothing is broken by this today beyond the one #2036 link, which I am fixing by hand. Filing so the blind spot is written down rather than rediscovered.

andrewboudreau:
Closes #2037.

`tools/check_dead_references.py` pulls repo-rooted paths out of prose and resolves them **from the repo root**. A markdown renderer does not: it resolves `[x](config/foo.json)` against the directory of the file the link is written in. So a link inside `notes/` naming `config/...` points at `notes/config/...` — a 404 — while the repo-rooted scan resolves the same string from the root, finds it, and reports a clean tree.

## The measurement that motivated it

That is #2036: it turned 106 bare paths in `notes/ead-debug-name-crossref.md` into markdown links; 105 carried the `../` and the one at line 192 did not. **All four of that PR's checks were green.** It was found by hand and fixed forward in #2049 (`745181902`), which explicitly left the blind spot to #2037. This closes it.

## Tree-wide baseline, measured before the gate was written

Implemented as a throwaway probe first and run over every markdown prose file at `e6ede02d7`:

| | |
|---|---:|
| markdown prose files | 106 |
| relative markdown links | 183 |
| that do not resolve from their own file | **1** |

The single offender: `port/docs/opie-assessment.md:3` links `roadmap.md`, which lives at `notes/roadmap.md`.

**One offender is a fix, not a ratchet.** So it is corrected in this PR and the check ships **fail-closed with no baseline** — nothing is grandfathered. The repo-rooted scan needs `config/dead-reference-baseline.json` because its false-positive population is 132 (other repos' paths, synthetic fixture paths, deliberately historical prose); this check has a population of one, and a baseline nobody needs is a place for real breakage to hide.

## What the extractor counts

**Markdown files only.** Link syntax means nothing in a python docstring, and applying it there is actively harmful — a regex character class followed by a group is indistinguishable from a link, and seven of the twelve hits in the first tree-wide pass were exactly that.

Also skipped: **inline code spans**, for the same reason — hand-rolled vtable dispatch written in backticks in `notes/archive/pret-idioms.md` reads as a markdown link and is not one (four more of those twelve). And `http:` / `mailto:` / any scheme, and bare `#anchor`.

Handled: angle-bracket targets, titles, reference definitions of the `[label]: path` form, `#fragment` stripped before resolving, and a directory target (alive when anything tracked lives under it).

**Fenced blocks are skipped, which differs from the repo-rooted scan.** Deliberate: a path in a fenced command example is a real claim about the tree, whereas a link inside a fence is being *shown*, not followed. It costs nothing either way today — all 183 links sit outside fences.

**No double-reporting.** Anything the repo-rooted scan already owns for a file is left to it, baselined entries included, so a link that is dead by both readings is reported once.

`MIN_LINKS = 50` joins the existing `SCAN TOO SMALL` floors, so this check cannot pass by extracting nothing either. 183 links today, 106 of them in one note; the floor survives that note being deleted outright.

## The failure path is exercised, not assumed

Reintroducing the exact #2036 line into the tracked note in a worktree, then running the checker:

```
check_dead_references: 372 prose file(s), 3474 repo-rooted path reference(s), 132 that do not resolve; 183 markdown link(s), 1 that do not resolve relative to their own file
  no new dead references

FAIL: 1 markdown link(s) do not resolve from the file they are written in:

  notes/ead-debug-name-crossref.md:192
      links to `config/rom-name-glossary.json`, which resolves to `notes/config/rom-name-glossary.json` -- not in the tree

A markdown link is resolved against the DIRECTORY OF ITS OWN FILE, not the
repo root, so a note in `notes/` must write `../config/x.json` and not
`config/x.json`. There is no baseline for these: fix the link.

EXIT=1
```

Note the `no new dead references` line above it: the existing check is blind to the defect even while the new one fires.

On **the same injected defect**, the tool at `origin/main` in a `git archive` export:

```
check_dead_references: 372 prose file(s), 3468 repo-rooted path reference(s), 132 that do not resolve
  no new dead references

EXIT=0
```

Reverted, `git diff` clean, this branch exits 0.

(The `1 baselined reference(s) now resolve` line is pre-existing on `origin/main` — it prints identically there — and the tool itself labels it *not a failure*.)

## The workflow filter was the same blind spot one step out

`dead-references.yml` filtered on the files the tool **reads** — `**.py`, `**.md`, `include/`, `src/`, `src_tu/`, the baseline JSON, `.github/workflows/`. But the tool also **resolves** what that prose names, against the whole tree, so a PR that only renames a resolution *target* broke references without touching anything in the filter.

Measured: **99 of the 183 links point outside the old filter**, 89 of them at a per-overlay symbols file, the rest under `symbols/` and `nearmiss/` or at top-level `attribution.json` and `LICENSE`. Renaming an overlay symbols file is routine here and would have broken 89 links silently — the same shape as the #2027 → #2030 → #2031 episode this workflow's own header already records.

Enumerating every top-level directory reproduces the tree in a list that goes stale on the next new directory, so the `paths:` filter is **removed** instead. The job is static analysis over the working tree with no compiler and no ROM.

The **pre-push hook does not run this gate** and this PR does not change that — `tools/hooks/pre-push` runs `prepush_linkcheck`, `prepush_attribution`, `port_refcheck`, `check_duplicate_sources`, `check_src_tu_compiles` and `check_references`, but never `check_dead_references.py`. Reported rather than fixed: adding a tree scan to every push is a policy call, not a defect fix.

## Gates, as base/branch pairs

Base is `e6ede02d7` (`origin/main`), run in a `git archive` export so the pair is like-for-like.

| gate | base `e6ede02d7` | this branch |
|---|---|---|
| `check_dead_references.py` | exit 0 | exit 0 |
| `python -m unittest tools.test_check_dead_references` | Ran 15, OK | **Ran 29, OK** (0 failures, 0 skips) |
| `pytest tools/test_dead_references.py` | 24 passed | 24 passed |
| `check_python_names.py` | PASS, 258 files, 46 advisory | PASS, 258 files, 46 advisory |
| `check_duplicate_sources.py` | 11182 stems, none doubled | 11182 stems, none doubled |
| pre-push hook | — | ran in full, clean, not bypassed |

Fourteen new tests: the exact #2036 shape (positive control), a correct `../` link, a sibling link, http and mailto, anchor-only, a fragment suffix, a directory target, a link in a fence, a link in backticks, angle-bracket and titled targets, a reference-style definition, the no-double-report case, link syntax inside python, and a fail-closed assertion that the real tree has zero broken links.

## No byte claim

This PR touches no compiled source — one workflow, one tool, its tests, and one line of a `port/docs/` note. `rombuild` was not run because there is no byte claim to make, and none is made.

---

## Post-review addendum

**A `premerge_check.py` run against the merge tree caught a regression this PR introduced — in the *pre-existing* repo-rooted scan, not the new one.** The first push's workflow comment spelled a template path with a literal `NNN` placeholder, which #2033's extension of the scan to `.github/workflows/**` reads as a real repo-rooted reference: `3473 -> 3474` references, `132 -> 133` unresolved. The new link check was clean throughout (`0 broken`).

Fixed by **rewording, not baselining** — a baseline entry for a placeholder written in the same diff would contradict this PR's own no-baseline argument on line one. The sentence now says "a per-overlay `symbols.txt`" with no slash-joined path token. Repo-rooted unresolved count is back to **132, identical to main**.

Rebased onto `e6ede02d7`. Re-measured there: still **183 relative links, 1 broken** — #2047's `src/` churn did not move them. The failure-path demo re-run on the rebased branch still exits 1 on `notes/ead-debug-name-crossref.md:192` with the same message, and `origin/main` still exits 0 on the identical injected defect.

### DIFF
--- a/.github/workflows/dead-references.yml
+++ b/.github/workflows/dead-references.yml
@@ -30,29 +30,26 @@
 # defect #2030 did not introduce; `b309f5d3f` (#2031) removed the string. A gate that
 # reads a file must be triggered by that file.
 #
-# Costs nothing: static analysis over the working tree, no compiler and no ROM.
+# AND A GATE THAT RESOLVES A PATH MUST BE TRIGGERED BY THAT PATH. There is no `paths:`
+# filter here any more, and that is the same lesson one step further out. The filter
+# used to list `**.py`, `**.md`, `include/**`, `src/**`, `src_tu/**`, the baseline JSON
+# and `.github/workflows/**` -- the files the tool READS. But the tool also RESOLVES
+# what that prose names, against the whole tree, so a PR that only deletes or renames a
+# resolution TARGET breaks references without touching anything in the filter. Measured
+# when #2037's relative-link check went in: 99 of the tree's 183 markdown links point
+# outside the old filter -- 89 at a per-overlay `symbols.txt`, the rest under `symbols/`
+# and `nearmiss/` or at top-level `attribution.json` and `LICENSE`. Renaming an overlay
+# symbols file is routine here and would have broken 89 links silently. Enumerating every
+# top-level directory would reproduce the whole tree in a list that goes stale on the
+# next new directory, so the filter is gone instead: this job is static analysis over
+# the working tree with no compiler and no ROM, and running it on every change costs
+# less than one missed rename.
 name: dead references
 
 on:
   pull_request:
-    paths:
-      - "**.py"
-      - "**.md"
-      - "include/**"
-      - "src/**"
-      - "src_tu/**"
-      - "config/dead-reference-baseline.json"
-      - ".github/workflows/**"
   push:
     branches: [main]
-    paths:
-      - "**.py"
-      - "**.md"
-      - "include/**"
-      - "src/**"
-      - "src_tu/**"
-      - "config/dead-reference-baseline.json"
-      - ".github/workflows/**"
   workflow_dispatch:
 
 permissions:
--- a/port/docs/opie-assessment.md
+++ b/port/docs/opie-assessment.md
@@ -1,6 +1,6 @@
 # Assessment: reusing `cybervisi0n/pokeplatinum@pc_port` + the libntr suite for an SM64DS PC port
 
-*Status: research + plan, not yet actioned. Slots into [`roadmap.md`](roadmap.md)
+*Status: research + plan, not yet actioned. Slots into [`notes/roadmap.md`](../../notes/roadmap.md)
 Phases 2–3 (LIFT / PORT). Investigated 2026-08-01; revised after adversarial review.*
 
 ---
--- a/tools/check_dead_references.py
+++ b/tools/check_dead_references.py
@@ -31,6 +31,27 @@ In those it finds repo-rooted path references -- a token whose first segment is 
 top-level directory of this repo (`tools/`, `notes/`, `include/`, `src/`, ...) -- and
 asks whether the path resolves in the working tree.
 
+AND, SEPARATELY, MARKDOWN LINKS THAT POINT AT THE WRONG PLACE
+-------------------------------------------------------------
+The scan above reads a path out of prose and resolves it FROM THE REPO ROOT. A markdown
+link does not work that way: a renderer resolves `[x](config/<name>.json)` against the
+directory of the file the link is written in. So a link inside `notes/` naming
+`config/<name>.json` points at that same path UNDER `notes/` -- a 404 -- while the
+repo-rooted scan resolves `config/<name>.json`, finds it, and reports a clean tree.
+
+Not hypothetical. #2036 turned 106 bare paths in `notes/ead-debug-name-crossref.md`
+into markdown links; 105 carried the `../` and the one at line 192 did not. All four of
+that PR's checks were green. It was found by hand and fixed forward in #2049; the blind
+spot itself is #2037, which `broken_links` below closes by resolving every markdown
+link against `os.path.dirname()` of its own file.
+
+FAIL-CLOSED, NO BASELINE. Measured over all 106 markdown prose files at `e6ede02d7`:
+183 relative links, of which exactly ONE did not resolve -- `port/docs/opie-assessment.md`
+naming `roadmap.md`, which lives at `notes/roadmap.md`. One offender is a fix, not a
+ratchet, so it was corrected in the same change and this check carries no accepted-offender
+list. The repo-rooted scan needs one because its false-positive population is large;
+this one does not, and a baseline nobody needs is a place for real breakage to hide.
+
 HOW IT AVOIDS CRYING WOLF
 -------------------------
 Two mechanisms, because a noisy gate gets switched off and then protects nothing:
@@ -53,9 +74,10 @@ FAILING LOUDLY RATHER THAN SKIPPING GREEN
 This tree has a documented history of gates that pass by doing nothing (an Allman
 reformat took `check_header_offsets` from 5 matches to 0 and printed a pass). So this
 tool refuses to report success unless the scan actually happened: if it inspects fewer
-than `MIN_FILES` prose files or extracts fewer than `MIN_REFS` candidate references, it
-exits non-zero with SCAN TOO SMALL rather than announcing a clean tree. Those floors sit
-far below today's numbers; they trip when the extractor breaks, not when prose changes.
+than `MIN_FILES` prose files, extracts fewer than `MIN_REFS` candidate references, or
+resolves fewer than `MIN_LINKS` markdown links, it exits non-zero with SCAN TOO SMALL
+rather than announcing a clean tree. Those floors sit far below today's numbers; they
+trip when the extractor breaks, not when prose changes.
 """
 import argparse
 import ast
@@ -74,6 +96,10 @@ BASELINE = REPO / "config" / "dead-reference-baseline.json"
 # The scan must be at least this big or the tool is broken, not the tree.
 MIN_FILES = 150
 MIN_REFS = 1000
+# 183 relative markdown links at e6ede02d7, 106 of them in one note. The floor sits
+# below what survives that note being deleted outright, so it trips on a broken
+# extractor and not on prose churn.
+MIN_LINKS = 50
 
 # Directories that are real parts of the repo layout. A path reference only counts
 # when its first segment is one of these -- that is what separates `tools/foo.py`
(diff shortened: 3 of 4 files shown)


andrewboudreau:
## Verified the load-bearing measurement independently. **Merging.**

This closes #2037, the blind spot that let #2036 land a broken link through four green checks.

### The failure path is demonstrated against BOTH tools, which is what makes it credible

Re-injecting the exact #2036 defect into the real tracked note:

```
notes/ead-debug-name-crossref.md:192
    links to `config/rom-name-glossary.json`, which resolves to
    `notes/config/rom-name-glossary.json` -- not in the tree      EXIT=1
```

and on the **same injected defect**, `origin/main`'s tool prints `no new dead references`,
`EXIT=0`. That contrast is the whole argument: the old check is not merely quiet here, it is
structurally incapable of seeing this class, because it resolves from the repo root and a
markdown renderer resolves from the file's own directory.

Tests 15 → **29**, zero skips — the failure path is exercised, not assumed
(the recurring hazard in this tree is gates whose red path nothing ever runs).

### I re-derived the 99/183 myself rather than taking it

The justification for removing the `paths:` filter is the one claim in this PR with repo-wide
consequences, so I resolved every markdown link on `origin/main` against `git ls-files` with my
own extractor and bucketed the targets against the old filter globs:

| target class | count |
|---|---|
| per-overlay `config/arm9/overlays/ovNNN/symbols.txt` | 77 |
| other `config/...` (further symbols files) | 12 |
| `symbols/`, `nearmiss/`, `attribution.json`, `LICENSE`, `docs/`, `port/` | 10 |
| **outside the old filter** | **99 of 182** |

Independently confirms the PR's 99. (My 182 vs its 183 is my extractor being marginally
stricter; the bucket totals are what matter here.) Renaming an overlay symbols file is routine
work in this repo, and it would have broken ~89 links **with the gate never running** — a
workflow that never fires is indistinguishable from one that passed, which is the exact shape of
the #2027 → #2030 → #2031 episode recorded in this workflow's own header. Dropping the filter is
the right call over enumerating directories that go stale on the next new one.

### The concurrency question, since removing a filter makes the job run more often

`dead-references.yml` carries `cancel-in-progress: true` on
`group: dead-references-${{ ... || github.ref }}`, so all main pushes share one group and rapid
successive merges will cancel earlier runs — and that now happens more often. **Checked: this is
not the #2042 hazard.** That one bit because the range gate verified *ranges*, so a cancelled run
dropped ranges permanently and nothing ever revisited them. This gate verifies the **whole tree**,
so a later run strictly subsumes a cancelled earlier one and nothing is lost. A cancelled run also
reports as cancelled, never as success. No change needed.

### Fail-closed with no baseline is correct here

The repo-rooted scan needs `config/dead-reference-baseline.json` because its false-positive
population is 132 — other repos' paths, fixture paths, deliberately historical prose. This check's
population is **one**, and it is fixed in the same PR (`port/docs/opie-assessment.md:3` → `roadmap.md`
lives at `notes/roadmap.md`). A baseline nobody needs is a place for real breakage to hide.

The two extractor rules are load-bearing, not cosmetic: an earlier looser probe reported 12
offenders, **11 of them false** — a regex character class followed by a group in `.py` files, and
backticked hand-rolled vtable dispatch in notes. Restricting link syntax to `.md` and skipping
inline code spans removed all 11.

### The regression this PR caught on itself

Its first `premerge_check` run returned `dead-references pass → fail`: the workflow comment spelled
a placeholder `config/arm9/overlays/ovNNN/symbols.txt`, which #2033's extension of the scan to
`.github/workflows/**` reads as a real path claim (`3473 → 3474` refs, `132 → 133` dead). Fixed by
**rewording, not baselining** — baselining a placeholder written in the same diff would contradict
this PR's own no-baseline argument on line one. Repo-rooted unresolved is back to **132, identical
to main**.

Worth stating plainly: that was found by the merge-tree gate, not by a green branch run. The
branch was green because the checker had not been re-run after the workflow edit.

### Gates — base `e6ede02d7`, which is current main, so this verdict is live

```
                    all: pass -> pass
```

Merge tree `83566711dd8b`. CI 6/6 green. Pre-push hook ran in full, not bypassed.

### Recorded, not fixed here

**The pre-push hook does not run this gate** — it runs `prepush_linkcheck`, `prepush_attribution`,
`port_refcheck`, `check_duplicate_sources`, `check_references`, `check_src_tu_compiles`, and not
this one. So a broken relative link is caught in CI but not locally. Deliberately out of scope for
this PR; noting it so it is on the record rather than rediscovered.

No `rombuild`/`eligible`/`romdata_check` was run and **no byte claim is made** — this PR touches no
compiled source, so none is needed.

PR #2051, opened 2026-08-30
Promote daObjKinokoTag_c to a genuine C++ translation unit

andrewboudreau:
Consolidates the full ROM-proven ov002 daObjKinokoTag_c TU into one readable production C++ source.\n\n- replaces nine legacy per-symbol files, including both actor factories\n- lets mwccarm generate retail D1 then D0 with no D2 or forcing helper\n- compiler-emitted RTTI, type name, complete vtable, and both typed SpawnInfos match the ROM\n- retains only measured ABI seams for actor allocation/construction, Fix12-by-value collision initialization, and the external mushroom backlink\n\nVerification: 9/9 TU byte/relocation match; all nine direct linkchecks VERIFIED with blind: 0; ROM data 5 full + 6 inherited partial, 0 differing/unnamed; affected/header/layout/language/port/reference/attribution gates pass; full ROM 11,088/11,088, zero mismatches, 106/106 exact, stock SHA-256 D1506E90EFAE5E2D2CF119926A4AC2A291BD5CA78349D09D5024E1A918C478E8. Independent verifier repeated the full proof.


andrewboudreau:
## Verified independently, including two things this PR's green checks do not actually prove.

### The baseline arithmetic, re-derived rather than taken

```
main    a1dc69b74   count=2517  len=2517  distinct=2517   ok
#2051   9854b7bc3   count=2511  len=2511  distinct=2511   ok

removed (6): src/_ZN16daObjKinokoTag_c16CleanupResourcesEv.cpp
             src/_ZN16daObjKinokoTag_c17SpawnMegaMushroomEv.cpp
             src/_ZN16daObjKinokoTag_c20TrySpawnMegaMushroomEv.cpp
             src/_ZN16daObjKinokoTag_c8BehaviorEv.cpp
             src/_ZN16daObjKinokoTag_cD0Ev.cpp
             src/_ZN16daObjKinokoTag_cD1Ev.cpp
added   (0)
ledger  59 rows / 57 paths  ->  65 / 63     (+6, exactly 1:1 with the removals)
```

`converted-baseline.json` is a **set** ratchet, so the failure mode that matters is a `count`
field disagreeing with the array. It holds on both sides. Nine `src/` files are deleted but only
**six** were banked — the two `_Spawn.c` files are C, and `InitResources` was not in the baseline
— so the −6 is exactly right, not an off-by-three.

### The six exception rows are unavoidable, and I checked that rather than assuming it

Zero additions means `src/actors/daObjKinokoTag_c.cpp` is **not** banked as CONVERTED. That is the
documented `no_mangled_refs` artifact, not a regression, and I confirmed it is genuine by scoring
the file rather than trusting the reason string. Nine mangled tokens trip `MANGLED_REF = ^_Z[0-9NK]`:

```
line 22  extern void *_ZN7fBase_cnwEj(u32 size);                        <- allocator seam
line 23  extern void  _ZN8dActor_cC2Ev(dActor_c *actor);                <- base construction
line 24  extern void  _ZN7dCcAc_cC1Ev(dCcAc_c *clsn);                   <- member construction
line 32  extern void  _ZN7dCcAc_c4InitEP8dActor_c5Fix12IiES3_jj(...)    <- Fix12-by-value ABI
lines 90,102,155,163,174   // @symbol _ZN16daObjKinokoTag_c...          <- comments
line 47  &_ZTV16daObjKinokoTag_c[2]   -- `T` is not in [0-9NK], so this never tripped it
```

Worth stating because it is not obvious: **five of the nine are `// @symbol` comments**, and the
textual matcher counts them. That does not change the outcome — the four `extern` declarations at
lines 22–32 trip it regardless, so the TU can never pass this criterion however the comments are
written — but anyone reading "9 mangled refs" should know that only 4 are load-bearing. The reason
string describes exactly those four ("fBase allocator/base/member/vptr construction seam and
Fix12-by-value ABI declaration"), so it is accurate. It also correctly says the seam is required for
exact bytes rather than claiming the byte match forced the *tier loss*.

### The delink collapse is contiguous

Nine ranges replaced by one span, endpoints preserved and no gap anywhere in the chain:

```
0x020b46a0 -> 46d0 -> 4714 -> 47ec -> 482c -> 4850 -> 49a8 -> 4a00 -> 4a38 -> 0x020b4a70
collapses to  src/actors/daObjKinokoTag_c.cpp   complete   .text 0x020b46a0..0x020b4a70
```

`complete` is retained, so this does not read as a draft delink.

### The "offsets pass" green on this PR is uninformative — measured, both sides

`include/daObjKinokoTag_c.h` changes `virtual ~daObjKinokoTag_c();` to `virtual ~daObjKinokoTag_c() {}`.
An inline member body is a known way to blind `check_header_offsets`, so I ran the gate against the
**full 507-header tree** exported from each revision, rather than reading the check name:

```
main      include/daObjKinokoTag_c.h: skipped -- derives from dActor_c, whose size is asserted nowhere
#2051     include/daObjKinokoTag_c.h: skipped -- derives from dActor_c, whose size is asserted nowhere
```

Identical on both sides, so **this PR changes nothing there and it is not a blocker**. But the green
tick is not evidence about this header: the gate never looks at it. (My first two attempts at this
measurement were wrong — renaming the file to compare versions produced a hollow
`0 commented fields ... spans 0x0` pass on *both* sides, because the tool keys the struct to the
filename, and a scratch dir without `dActor_c.h` produced the same skip for a different reason. The
507-header export is the run I trust.)

What *does* guarantee the new layout is the compile-time asserts the PR adds —
`sizeof(daObjKinokoTag_c) == 0x110`, `KinokoPositionWords == 0xc`, `KinokoTagSpawnInfo == 0x1c` —
which the ROM build compiles for real. That plus 106/106 is the actual proof here, and it is
sufficient. I am recording *why* it is sufficient rather than leaning on a check that abstained.

### Two tree-wide findings this surfaced — NOT caused by this PR, not blocking it

Both measured on `origin/main`'s own tree, reported as facts about main:

1. **`check_header_offsets` checks 120 of 507 headers.** 369 are skipped for a missing base-class
   size assert and 18 for an unmodelled vptr. Two absent asserts account for most of it:
   `dBgActor_c` blinds **99** headers and `dActor_c` blinds **95**. Adding those two asserts would
   roughly double the gate's reach.
2. **The "check every header" path exits 1 on main today** — 20 unparsed declarations across 11
   headers (`HeapAllocator.h`, `ModelAnim.h`, `OamAttr.h`, `SurfaceInfo.h`, `dBgCh_Gnd/Lin/SphCrr.h`,
   `dBgPc.h`, `dBgPi.h`, `dM3dGLin.h`, `math/Fix12.h`), **0 mismatched**. It is latent, not a live
   red: `header-offsets.yml` triggers only on `pull_request` and `workflow_dispatch`, and the full
   form runs only when `BASE` is empty, so nothing exercises it automatically. Anyone dispatching it
   with `all=true` would get a red that predates this PR.

I will file these separately. Flagging them here so they are on the record at the moment they were
found rather than rediscovered later.

### Gates

`premerge_check.py` gates the **merge tree**, which CI does not — `pr-validate.yml` is
`pull_request_target` and submits `head.sha`, so a green branch is not a green merge. Base
`a1dc69b74012`, merge tree `8c00890aac5b`:

```
source-coverage     pass     pass     ok    entries 10956 -> 10948 (-8)
```

The −8 is arithmetic, not a loss: 9 `src/` deletions − 1 addition. Bytes are flat.

Base `a1dc69b74012` was still `origin/main` at merge time, so **this verdict is live, not stale**.
CI 7/7 green plus the ROM validator SUCCESS; `mergeStateStatus: CLEAN`.

### What I did NOT run

No `rombuild.py`, `eligible.py`, or `romdata_check.py` from my side — `premerge_check` builds no ROM
and says so. **I make no independent byte claim.** The 106/106, the 11,088/11,088, the 9/9
linkchecks and the stock SHA-256 in the PR body are the author's measurements plus an independent
verifier's repeat, and the ROM validator agreed. That is the basis for merging, and I am naming it
as someone else's measurement rather than implying I reproduced it.

### Ordering

Merging this **before** the re-cut of #2001's ratchet half, which regenerates these same two
generated files. A TU promotion is expensive to rebase; a baseline regeneration is one command. The
agent doing that work has been told to re-derive on top of this.

**Merging.**

PR #2052, opened 2026-08-30
converted ratchet: score a promoted TU per member, not per file

andrewboudreau:
Cuts the **ratchet-tightening half of #2001** clean from current `main`. #2001 bundles two unrelated changes; this PR carries only the first. #2001 stays open and untouched.

**Rebased onto `e894874f3`** -- `main` after #2051 (daObjKinokoTag_c promotion) merged at `1b2996d2b`. Head `7893fadb5`. `e894874f3` is the `[skip ci]` progress refresh sitting on top of that merge; it touches neither `src/` nor either generated file, so every set below is identical whether measured against `1b2996d2b` or `e894874f3`. Originally cut at `a1dc69b74`; every number below has been re-measured on the new base. The two generated files were not conflict-resolved -- both were reset to main's committed state and the generator re-run, so the baseline is derived, not patched.

---

## 1. The bug this fixes is live on `main` right now

`tools/tiers.py` scores a source file as a unit, and derives the "real function name" criterion from the **path stem**. That is right for a one-function intake file, where the stem *is* the symbol. It is wrong for a promoted translation unit, where one path owns several enrolled functions and the stem names none of them.

Measured on `main` (`e894874f3`, the tip after #2051) with `main`'s own tools:

```
$ python tools/tiers_ratchet.py --check
CONVERTED ratchet PASS   baseline 2511   current 2511          # exit 0
```

```
$ sed -n '28p' src/actors/ActorDerived.cpp
extern "C" void func_02013edc(int a, int b, int c)
```

`main` banks `src/actors/ActorDerived.cpp` as CONVERTED **while it contains a `func_<addr>` member**, because `scan()` scored the stem `ActorDerived`, which is not `func_`-prefixed. That is a false positive inside a published percentage, not a hypothetical.

With this PR's `tiers.py` + `tiers_ratchet.py` over the same tree and the same unmodified `main` baseline:

```
$ python tools/tiers_ratchet.py --check
CONVERTED backslide: 1 banked file(s) no longer pass all 5 criteria

  src/actors/ActorDerived.cpp
      IDENTITY UPGRADE INCOMPLETE -- member regression:
      src/actors/ActorDerived.cpp#func_02013edc fails: Real function name

baseline 2511   current 2571   (+63 gained, -1 lost)           # exit 1
```

Exactly one veto, tree-wide.

## 2. What changed

`tiers.score_member()` scores one enrolled member of a multi-function source; `tiers_ratchet` banks `path#symbol` for those sources. A one-function source keeps its historical path identity, so **2508 of `main`'s 2511 banked rows are untouched**.

Member boundaries come from evidence or not at all:

* an explicit `// @symbol <linker-name>` marker, sliced to the next marker; or
* a **balanced inline** ctor/dtor body in a directly `#include`d class header. A bare declaration is explicitly rejected as evidence of its own definition.

A member with neither falls back to whole-file scoring. That can under-credit a member; it can never grant one speculatively.

**The five criteria are unchanged.** There is no veneer exemption, no `extern "C"` exemption, no `func_` exemption. The one member that fails is recorded in the exception ledger — the gate's designed valve — rather than exempted in the checker, which would be a second definition of the percentage.

## 3. The exception row, and why it is honest

`func_02013edc` is a three-word ARM long-branch veneer. Read directly out of the cartridge (`extracted/arm9_dec.bin`, ARM9 load base `0x02004000`):

```
0x02013edc  e59fc000    ldr ip, [pc, #0]
0x02013ee0  e12fff1c    bx  ip
0x02013ee4  02042fe4    .word func_02042fe4
```

`config/arm9/symbols.txt:463` gives `size=0xc`, which the three words tile exactly. Those bytes evidence the **branch target** and nothing else — they carry no name — so the member is not nameable without coining a name. `config/tu_manifest.d/arm9/ActorDerived.json` states this twice independently (lines 12 and 104): the symbol is left unnamed on purpose and retains C linkage because it is a real free-function ABI boundary.

The row banked:

```json
{"path": "src/actors/ActorDerived.cpp",
 "reason": "func_02013edc is a three-word ARM long-branch veneer (ldr ip,[pc,#0]; bx ip; .word 0x02042fe4; symbols.txt size=0xc) whose bytes evidence only its branch target and therefore cannot evidence a name -- any name would be coined. config/tu_manifest.d/arm9/ActorDerived.json states twice that it is deliberately left unnamed and retains C linkage because it is a real free-function ABI boundary. The member genuinely fails the real-name criterion; main's file-wide scorer only passed it because it derived the name from the path stem 'ActorDerived'. Recorded, not fixed: the file's other four members remain CONVERTED under their own identities."}
```

This is **not** "the byte match required it". Nothing about the match forces this; the member simply cannot be named from evidence. The other four `ActorDerived.cpp` members remain CONVERTED under their own identities.

## 4. Baseline regeneration -- recipe, not just result

#2001's committed baseline (`count: 2554`) was banked against a base predating the ActorDerived promotion and is not carried. Regenerated wholesale, on top of `e894874f3`:

```
python tools/tiers_ratchet.py --update --reason "<the reason text quoted verbatim in section 3>"
wrote config/converted-baseline.json: 2511 -> 2571 (+63 / -3)
2 of those left by TU promotion into a CONVERTED file (a move, not a removal)
logged 1 removal(s) to config/converted-backslide-exceptions.jsonl

python tools/tiers_ratchet.py --check
CONVERTED ratchet PASS   baseline 2571   current 2571           # exit 0
```

`count == len == distinct == 2571`.

**Removed vs `e894874f3` (3), each replaced by every member it owns:**

| path | enrolled | re-banked |
| --- | --- | --- |
| `src/actors/ActorBase_SceneNode.cpp` | 2 | 2 |
| `src/actors/daObjFl_Fall_Block_c.cpp` | 4 | 4 |
| `src/actors/ActorDerived.cpp` | 5 | 4 -- `func_02013edc` is the documented exception |

That is the property that makes the expansion safe: a file that loses its whole-file identity must have **every** enrolled member re-banked, or a regression in an unbanked member becomes invisible. The only gap is the one genuinely-failing member, and it has a ledger row.

**Added vs `e894874f3`: 63 member identities** across 14 promoted TUs. **`2511 -> 2571` is a granularity change, not a readability gain, and the two numbers are not comparable.**

None of #2051's six deleted `src/_ZN16daObjKinokoTag_c*.cpp` fragment paths is banked; they left with main's own regeneration and are not reintroduced.

**Ledger:** 65 -> 66 rows, 63 -> 64 distinct paths. All 65 pre-existing rows are byte-identical (the file stays append-only) and exactly one row is added. `src/_ZN6BobOmb13OnYoshiTryEatEv.cpp` and `src/_ZN6Rabbit13OnYoshiTryEatEv.cpp` each appear twice; both duplicates predate this branch.

## 5. Separability evidence (why this is a clean half of #2001)

#2001 at head `ef16a3073` (it has moved since the queue snapshot at `43243bfb3`) touches 18 files. The split is:

**Group A, carried here** — `tools/tiers.py`, `tools/tiers_ratchet.py`, `tools/test_tiers_ratchet.py`, `.github/workflows/converted-ratchet.yml`, `config/converted-baseline.json`.

**Group B, left in #2001** — `tools/rombuild.py`, `rombuild_check.py`, `tubuild.py`, `tu_promote.py`, `tu_production.py`, `objisolate.py` and their six test files.

Evidence:

| Question | Answer |
| --- | --- |
| Does A import B, at any depth? | **No.** `tiers.py:45-47` imports `delaunder`, `demangle`, `srcpath`. `tiers_ratchet.py:139-140` imports `tiers`, `tu_manifest`. None is a Group B module, and none of those four transitive deps is modified by #2001. |
| Does `test_tiers_ratchet.py` depend on B? | **No.** Its only project import is `tiers_ratchet` (line 8); grep for `tubuild\|tu_promote\|tu_production\|objisolate\|rombuild` in it returns zero hits. |
| Does the workflow change reference B? | **No.** #2001's `converted-ratchet.yml` diff is comment-only. |
| Does B import A? | **Yes, one edge, and it does not bind.** `tu_promote.py:45` adds `import tiers_ratchet as TR`, used once at `tu_promote.py:253` as `data["_note"] = TR.NOTE`. `TR.NOTE` already exists on `main`. So B compiles against `main`'s A, and A here does not need B. The edge is one-way and satisfied by either version. |
| Do the two groups share a file? | **No.** The 18 changed paths partition with no overlap. |

Mechanical confirmation: `git diff <merge-base> origin/main -- tools/tiers.py tools/tiers_ratchet.py tools/test_tiers_ratchet.py .github/workflows/converted-ratchet.yml tools/srcpath.py tools/demangle.py tools/tu_manifest.py` is **empty** — `main` has not moved on any Group A file or its dependencies since #2001's merge base, so these files transfer verbatim with no conflict and no rebase. The tools and tests are byte-identical to #2001's; only the generated baseline and the CI wiring differ.

## 6. One thing added beyond a verbatim cut

**#2001's `tools/test_tiers_ratchet.py` would have run in no CI job at all.** `tool-tests.yml` executes a deliberately *enumerated* module list ("An explicit list makes adding a module a reviewed decision") which does not include it, and `converted-ratchet.yml` ran only `tools.test_tiers`. 196 lines of new tests, never executed.

So this PR wires it into `converted-ratchet.yml`, the workflow whose gate it protects, as a `Ratchet unit tests` step. It also adds `tools/srcpath.py` and `tools/test_tiers_ratchet.py` to that workflow's `pull_request`/`push` path filters — `srcpath.py` for the same reason `tu_manifest.py` is already listed: it decides how many identities get banked, and a PR touching only it would change the banked set without ever running this workflow.

## 7. Gates — every number, as a pair

| Gate | `main` (`e894874f3`) | This branch (`7893fadb5`) |
| --- | --- | --- |
| `tiers_ratchet.py --check` | `PASS baseline 2511 current 2511`, exit 0 | `PASS baseline 2571 current 2571`, exit 0 |
| ...same check, `main` baseline + these tools | -- | `exit 1`, 1 veto (section 1) |
| `count == len == len(set)` | 2511 / 2511 / 2511 OK | 2571 / 2571 / 2571 OK |
| `python -m unittest tools.test_tiers` | 29 tests OK | **29 tests OK** (unchanged) |
| `python -m unittest tools.test_tiers_ratchet` | module absent | **12 tests OK** |
| pre-push `port-refcheck` | -- | 405 references, all resolve |
| pre-push `duplicate-sources` | -- | 11174 stems, none doubled |
| pre-push `check_src_tu_compiles` | -- | **91/91** in 5.1s |
| pre-push `check_references.py` | -- | **SKIPPED** -- "no current eligibility report". A skip, not a pass. |
| `premerge_check.py` (8 static gates) | all pass | all pass, `nothing goes green -> red` (section 10) |

**Gates deliberately NOT run, and why:**

* **`tools/rombuild.py` (ROM byte gate).** This PR touches no compiled source — all six changed files are under `tools/`, `config/`, `.github/`. No ROM was built and **no byte claim is made or implied**.
* **`tools/romdata_check.py`, `tools/eligible.py`.** Same reason: no source, no vtable, no enrollment change.
* **`tools/premerge_check.py`.** It *was* run -- see section 10.

## 8. Interaction with #2051 -- predicted, then confirmed after it merged

#2051 was tested against this branch on a throwaway merge *before* it landed, and re-measured after. Both agree:

```
control  (main's tools, main's baseline):  PASS baseline 2511 current 2511      exit 0
tightened (these tools, main's baseline):  1 veto, ActorDerived.cpp#func_02013edc
                                           baseline 2511 current 2571           exit 1
--update --reason "<unchanged text>":      2511 -> 2571 (+63 / -3), 1 removal logged
re-check:                                  PASS baseline 2571 current 2571      exit 0
```

**#2051 introduced nothing new.** Same single veto, same resulting 2571-identity set, same regeneration recipe.

And a genuine gain, in #2051's favour: #2051 banked 6 exception rows writing off the functions its promotion absorbed, because `src/actors/daObjKinokoTag_c.cpp` fails `no_mangled_refs` **as a file** -- it must spell `_ZN7fBase_cnwEj`, `_ZN8dActor_cC2Ev`, `_ZN7dCcAc_cC1Ev` and `_ZN7dCcAc_c4InitEP8dActor_c5Fix12IiES3_jj` to link. Scored per member, only 3 of its 9 members actually touch those seams:

```
_ZN16daObjKinokoTag_cD1Ev                  PASS      _ZN16daObjKinokoTag_c13InitResourcesEv  FAIL: no_mangled_refs
_ZN16daObjKinokoTag_cD0Ev                  PASS      daObjKinokoTag_c_Spawn                  FAIL: no_mangled_refs
_ZN16daObjKinokoTag_c17SpawnMegaMushroomEv PASS      MegaMushroomTag_Spawn                   FAIL: no_mangled_refs
_ZN16daObjKinokoTag_c20TrySpawnMegaMushroomEv PASS
_ZN16daObjKinokoTag_c16CleanupResourcesEv  PASS
_ZN16daObjKinokoTag_c8BehaviorEv           PASS
```

The other **6 members are recovered into the baseline by this PR**. #2051's ledger rows are honest under the file-wide scorer and simply over-state the loss under this one; they stay as the historical record and need no action.

## 9. Known limitation, deliberately not fixed here

`classify_missing()` treats a legacy fragment path absorbed into a promoted TU as a clean move only when **every** destination member passes. A fragment whose *own* symbol survives as a passing member identity is still classified a backslide if some *sibling* member fails. Tightening that would reduce the number of rows needing a `--reason`, i.e. it would make the gate quieter. That is a loosening, and this PR does not touch the criteria or the classification in that direction. Flagged as a possible follow-up, not done.

## 10. `premerge_check.py` -- the merge TREE, which no CI job builds

`pr-validate.yml` is `pull_request_target` and submits `head.sha`, so nothing in CI ever builds the tree that merging this actually produces. Run against base `origin/main @ e894874f38f333cb71ca7f00fcb6e4ebe1f7652a`. `origin/main` was re-fetched after the run and is still at that sha, so this verdict is LIVE, not stale:

```
base   origin/main @ e894874f38f3   (committed 7m ago)

target #2052 converted ratchet: score a promoted TU per member, not per file  head 7893fadb5457
  merge tree b100994abaf0

  RESULT: nothing goes green -> red. Safe to merge as far as the STATIC gates can tell -- no ROM was built.
```

---


andrewboudreau:
## Reviewed. **Do not merge yet** — one blocking staleness, and one correctness property I verified independently.

### BLOCKING: the baseline banks six files that no longer exist

This branch was cut at `a1dc69b74`. Main is now `1b2996d2b` — #2051 landed underneath it. The
committed `config/converted-baseline.json` banks six paths whose **files were deleted by that merge**:

```
src/_ZN16daObjKinokoTag_c16CleanupResourcesEv.cpp
src/_ZN16daObjKinokoTag_c17SpawnMegaMushroomEv.cpp
src/_ZN16daObjKinokoTag_c20TrySpawnMegaMushroomEv.cpp
src/_ZN16daObjKinokoTag_c8BehaviorEv.cpp
src/_ZN16daObjKinokoTag_cD0Ev.cpp
src/_ZN16daObjKinokoTag_cD1Ev.cpp
```

#2051 promoted all six into `src/actors/daObjKinokoTag_c.cpp`, which is **permanently** unbankable —
it fails `no_mangled_refs` on four genuine `extern` ABI seams (`_ZN7fBase_cnwEj`, `_ZN8dActor_cC2Ev`,
`_ZN7dCcAc_cC1Ev`, `_ZN7dCcAc_c4Init...`), which exist for exact bytes and are not going away. That is
why main's baseline moved 2517 → 2511 with six new ledger rows.

```
origin/main  1b2996d2b   count=2511  len=2511  distinct=2511   ok
this PR      a89c8ae63   count=2571  len=2571  distinct=2571   ok   <- internally consistent, stale membership
```

The set invariant holds on **both** sides, so the ratchet's own failure mode is not what is wrong here
— the membership is. This needs the baseline **regenerated** on `1b2996d2b`, not hand-edited, and the
two generated files must be re-derived rather than conflict-resolved (`--ours`/`--theirs` on a shared
config silently reverts unrelated auto-merged content elsewhere in the same file).

### VERIFIED, and this is the property that makes the PR safe

Moving from file identities to `path#symbol` identities has one real hazard: a file that **loses** its
whole-file identity must have **every** enrolled member re-banked. Any member left unbanked becomes a
place where a future regression is invisible — the gate would no longer be watching it.

I checked all three such files against `config/tu_manifest.d/**` `functions[].symbol`, rather than
trusting the tool's own report:

| file | enrolled | banked | verdict |
|---|---|---|---|
| `src/actors/ActorBase_SceneNode.cpp` | 2 | 2 | lossless, no ledger row — correct |
| `src/actors/daObjFl_Fall_Block_c.cpp` | 4 | 4 | lossless, no ledger row — correct |
| `src/actors/ActorDerived.cpp` | 5 | 4 | gap is `func_02013edc`, ledger row present |

**No silent narrowing.** The one gap is the one member that genuinely fails, and it is documented.

The other 57 member identities are gains, not conversions: files that failed *file-wide* now bank the
members that individually pass. That is strictly stronger — those members were previously watched by
nothing, and the members that fail remain unbanked exactly as before.

### The exception row is correctly reasoned, and I say that as the person who got this symbol wrong

Its reason string says `func_02013edc` is a three-word ARM long-branch veneer whose bytes evidence only
its branch target and therefore cannot evidence a name. That is right, and it matches what I had to
retract on #2001 after claiming a byte-verified name for this exact symbol. It also correctly identifies
the root cause — main's file-wide scorer passed it only because it derived the name from the path stem
`ActorDerived` — and it says *"Recorded, not fixed"* rather than claiming a byte match forced the loss.
**Do not reword it.**

This also means the PR's premise is confirmed on main today: main banks `ActorDerived.cpp` as CONVERTED
while one of its five enrolled functions is literally named `func_<addr>`. That is a live false positive
inside a published percentage, and per-member scoring is what surfaces it.

### The workflow change tightens the gate — checked, because this is where softening hides

```
+ tools/srcpath.py            added to BOTH trigger lists
+ tools/test_tiers_ratchet.py added to both, plus a new unittest step
```

Adding `srcpath.py` is the sharp one and it is right: member identities are derived from the enrollment
table, so a PR touching only `srcpath.py` would redefine how many identities this gate banks while
matching none of the other trigger paths. The new test step is also justified — `tool-tests.yml` runs an
**enumerated** module list, not a glob, and does not include this module, so without that step the 196
lines of new tests would never execute in CI.

No exit-code semantics were loosened, no baseline escape hatch was added beyond the single documented
row, and the `--check` path is unchanged. I was looking for softening and did not find any.

### What I have NOT checked

No `rombuild`/`eligible`/`romdata_check` from my side and **no byte claim** — this PR compiles nothing
into the ROM. I have not yet run `premerge_check.py`, because the merge tree is worthless until the
baseline is regenerated; that verdict is worth having only on the rebased tree.

### Next step

Regenerate on `1b2996d2b` and report the `count == len == distinct` triple plus the removed/added sets
measured **against the new main**, not against `a1dc69b74`. Then `premerge_check.py` last, naming the
base sha it used.

To be explicit for anyone else with merge rights: **this is a hold on staleness, not on quality.** The
tool change itself is sound and I want it to land.

andrewboudreau:
## Hold cleared. The rebase is correct, and the property I was most worried about checks out.

Re-measured at `7893fadb5` against current main `1b2996d2b`.

### The stale six are gone — and I want to correct my own previous comment

My earlier comment listed six `src/_ZN16daObjKinokoTag_c*.cpp` paths as stale. They are gone. What
now appears under that class name is a **different and legitimate** set — six *member* identities of
the promoted file that actually exists:

```
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_c8BehaviorEv
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_c16CleanupResourcesEv
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_c17SpawnMegaMushroomEv
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_c20TrySpawnMegaMushroomEv
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_cD0Ev
src/actors/daObjKinokoTag_c.cpp#_ZN16daObjKinokoTag_cD1Ev
```

Anyone re-running my check by grepping the class name will hit the same false alarm I did, so:
**the test that settles it is whether a banked path's file exists**, not what the path is called.

```
banked paths with no file in the tree: 0
count == len == distinct: 2571 on the PR, 2511 on main -- invariant holds both sides
```

### The real question this raised, and the answer

#2051 established that `src/actors/daObjKinokoTag_c.cpp` **cannot** be banked — it fails
`no_mangled_refs`, which is why main's baseline dropped 2517 → 2511 and six ledger rows were added.
This PR now banks six members of that same file. That is either a genuine refinement or a loophole
that re-banks what main deliberately rejected, and which one it is depends entirely on *which*
members. So I segmented the file and scored each member body:

```
WITHHELD  InitResources          L93 calls _ZN7dCcAc_c4InitEP8dActor_c5Fix12IiES3_jj    <- the only member that
BANKED    Behavior               0 mangled calls in body                                   actually calls through
BANKED    CleanupResources       0                                                         a mangled name
BANKED    TrySpawnMegaMushroom   0
BANKED    SpawnMegaMushroom      0
BANKED    D0, D1                 no body in this file at all (see below)
```

The four seams that make the *file* unbankable are at **file scope** — the `extern` declarations at
lines 22–32, and their uses at lines 44–48 and 69–73, which sit inside the free spawn functions above
the first member. A declaration is not a call, and attributing a file-scope declaration to every
member is precisely the imprecision being fixed here.

**This is a refinement, not a loophole.** It withholds exactly the one member that genuinely fails and
banks the five that are genuinely clean. It is also strictly *more* watched surface than main has
today: main banks nothing at all for this file, so a future regression in `Behavior` or
`CleanupResources` is currently invisible and would not be after this lands.

On D0/D1: there is no `~daObjKinokoTag_c` body and no `operator delete` in the file — #2051 moved the
destructor inline into the header, so these are compiler-emitted and carry no seam. Worth noting that
`operator delete` would not have tripped the matcher anyway: `MANGLED_REF = ^_Z[0-9NK]` and `_ZdlPv`
has `d`, which is not in that class — the same reason `_ZTV...` never trips it.

### Still verified from before, unchanged by the rebase

The three files that lose their whole-file identity re-bank **every** enrolled member, checked against
`config/tu_manifest.d/**` `functions[].symbol`: `ActorBase_SceneNode.cpp` 2/2, `daObjFl_Fall_Block_c.cpp`
4/4, `ActorDerived.cpp` 4/5 with the one failing member carrying an honest ledger row. No silent
narrowing.

Exceptions ledger: **66 rows / 64 distinct paths**. The 2 duplicates are the pre-existing
`_ZN6BobOmb13OnYoshiTryEatEv` / `_ZN6Rabbit13OnYoshiTryEatEv` pair that has been on main since before
this branch — not introduced here, and correctly left untouched.

### What remains before I merge

CI is 6/6 green; `PR validation` (the ROM validator) is still pending, and I want `premerge_check.py`
against `1b2996d2b` naming its base sha. Those are the two outstanding items — the content review is
complete and positive.

No `rombuild`/`eligible`/`romdata_check` from me and **no byte claim**: this PR compiles nothing into
the ROM.

PR #2055, opened 2026-08-31
Promote daEyBm_c to a genuine C++ translation unit

andrewboudreau:
Consolidates 13 legacy daEyBm_c sources into one readable, genuine C++ class translation unit.

- The compiler now emits D1/D0, RTTI, the type-name object, vtable, and typed SpawnInfo; no D2, constructor, forcing helper, or manual class metadata is emitted.
- Four formerly address-named private helpers are represented as inferred class-local aliases while preserving exact code and relocations.
- The narrow C ABI factory/ABI seams remain where the 2004/b56 compiler cannot naturally reproduce the retail allocator or by-value calling shape.

Verification: tubuild 13/13; all 13 direct linkchecks VERIFIED with blind: 0; owned class data has zero differing or unnamed records; full stock ROM 11,088/11,088 with zero mismatches and 106/106 exact modules; SHA-256 d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8. Independently reverified from the final commit.


andrewboudreau:
## Reviewed and merging. This is a genuine TU, and the vtable seam commit is principled.

Reviewed at head `f5dfe3237`, base `7b4eeabfb` (still `origin/main`). Everything below was measured on
**merge tree `27ebd60d36b8`** — and note that the force-push's rebase produced a tree identical to the
merge result I had already built, which is why nothing here is stale:

```
merge(origin/main, f5dfe3237) tree = 27ebd60d36b8ba6ad9f397d74c323834bb495110
f5dfe3237^{tree}                   = 27ebd60d36b8ba6ad9f397d74c323834bb495110
```

| measurement | result |
|---|---|
| `rombuild.py -j 16 --no-rom` | `106/106 exact, 100.000000%` · `reproducing 11,088` · `mismatching 0` · PASS |
| `rombuild.py -j 16` (full ROM) | `build/sm64ds.nds` sha256 `d1506e90…c478e8` = baseline control ROM |
| **`tubuild.py verify ov071/daEyBm_c`** | **`13/13 MATCH`** · `objisolate check: clean` · emission order correct · **TEXT-VERIFIED** |
| `tiers_ratchet.py --check` | `PASS baseline 2568 current 2568`; SET invariant holds (2571 → 2568 = −5 legacy paths, +2 member identities) |
| `port_refcheck.py` | `405 checked — all references resolve` |
| `premerge_check.py` (extracted from `origin/main`) | all 8 gates **pass → pass**; `nothing goes green -> red` |

`tubuild verify` is the one that matters most: it recompiled the TU from source under 2004/b56 and got
13/13, with `objisolate check: clean` and the emission-order check green. A mechanical concatenation of
hand-mangled stubs does not survive that. This is a real class with real member bodies whose names the
compiler mangles.

### The vtable address seam commit — no address was changed

`f5dfe3237` is 6 added lines in the header and 1 removed line in the `.cpp`. It moves
`extern int _ZTV8daEyBm_c[];` out of the `extern "C"` block to header namespace scope. **No address
value changed anywhere in this PR:**

- `symbols.txt`'s `_ZTV8daEyBm_c … addr:0x02122de8` is **byte-identical to `origin/main`**. The entire
  symbols.txt diff is 8 lines — the 4 function renames, nothing else.
- **No `data_ov071_*` entry was dropped** (`data_ov071_02122e64` present on both sides).
- **The delinks range did not move.** The new entry is `.text 0x02121a6c..0x02121fe4` only; no `.data`
  range added, moved, or removed.

The documented `reloc_audit.py` bias is reconciled *arithmetically*, which is the right way:

```
_ZTS8daEyBm_c            0x02122dac
_ZTI8daEyBm_c            0x02122db8
MrI_Projectile_SpawnInfo 0x02122dc4  (+0x1c → ends 0x02122de0)
  → vtable OBJECT starts  0x02122de0   (offset-to-top + typeinfo = 8 bytes)
  → address point         0x02122de8   == symbols.txt addr
  → 31 slots × 4 = 0x7c, ending exactly at data_ov071_02122e64
```

`*(int *)actor = (int)&_ZTV8daEyBm_c[2];` — `&arr[2]` on an `int[]` is +8 from the object start, i.e.
exactly the address point. Same construct already landed for `daKpFr_c`. This is the documented
storage-start vs address-point difference being *computed*, not an address edited until a gate agreed.

Related hazards checked directly rather than assumed: langmode not raised (the vtable was already
named on main; `langmode-ratchet pass → pass`); no rename was non-atomic (the 4 *function* renames move
together across `symbols.txt`, `actor_renames.tsv` and the TU, with no surviving `func_ov071_*`
reference); and trailing-null truncation is not being leaned on — `romdata_check --data-json` reports
`_ZTV8daEyBm_c VERIFIED bytes 124 emitted 124 romExtent 124 blindWords 0`.

### The 5 backslide exception rows are justified — I re-derived them rather than reading them

An exception row that hides a real regression would have been a close-level finding, so these were
re-scored from scratch. 11 of 13 members genuinely fail `no_unk_field` / `no_mangled_refs` under
`tiers.py score_member()`, because of ABI seams the b56 bytes require (`dCcAcPos_c::Init`,
`Player::Hurt`, `Particle::System::New` — Fix12-by-value and allocator walls) plus the inherited
`unk_0a4` spelling. Only `D0Ev`/`D1Ev` score clean — which is exactly the 2 identities re-entering the
baseline. The arithmetic is consistent and each row's stated reason is honest.

ROM data was diffed at **record level**, not by headline count: only-in-main **0**, only-in-merge **1**
(`MrI_Projectile_SpawnInfo`, the TU's own new symbol), keys with a changed verdict **0**. The
`7,655 → 7,647` record drop and premerge's `entries 10948 → 10936 (−12)` are the same consolidation —
13 legacy files collapsing into one — not lost coverage.

The key function is right: the header declares an inline `virtual ~daEyBm_c() {}` **first**, so the
first *declared* out-of-line virtual is `InitResources`, which is what the manifest names, and retail's
D1/D0 pair comes out with no homeless D2.

---

## Four non-blocking items — worth fixing in the *next* promotion, not this one

There are a dozen-plus more genuine-TU promotions staged behind this one, so I'd rather these become
wave-wide conventions than one-off nits.

1. **Coined mangled helper names encode unprovable parameter types.**
   `_ZN8daEyBm_c15UpdateCollisionER10dBgCh_Actr` asserts a **reference** parameter. The bytes cannot
   distinguish a reference from a pointer — they mangle differently but generate identically here. It
   is disclosed in the manifest notes and the `actor_renames.tsv` reason column, which is why this is
   not a blocker, but it lands in `symbols.txt` looking authoritative. Prefer the byte-provable form,
   or keep the disclosure adjacent to the symbol.
2. **`extern int _ZTV8daEyBm_c[];` in a public header** diverges from the landed `daKpFr_c` precedent,
   which keeps it in the `.cpp`. Blast radius is zero today (only `daEyBm_c.cpp` includes the header),
   but it widens the surface of a construct that exists purely as a codegen seam. Pick one convention
   for the wave.
3. **Stale declarations**: `include/decl_common.h` still declares `func_ov071_02121b08/b50/ba4` for
   symbols that no longer exist. Nothing calls them and `dead-references` passes, so this is cleanup.
4. **Stale citation** in the header comment still lists three now-deleted `_ZN8daEyBm_c*.cpp` files.

## Corrections to my own review brief

- I said "two `func_ov071_*.c` → `.cpp` conversions". Wrong: there are **four** `func_ov071_*` files
  and **none are converted in place** — all four are deleted and absorbed (`b08`/`b50` were `.c`,
  `ba4`/`c6c` were already `.cpp`).
- The self-assignment codegen wart on `unk_0a4` is **pre-existing on main**, not introduced here.
  Worth noting that `tiers.py`'s `no_codegen_trick` returns True for it — that criterion only detects
  launder/volatile/asm and is blind to this shape.

## Measurements that were unavailable (not defects)

- `--partitioned-tu` aborts with `partitioned_link.state is None`. Only **2 of 22** promoted manifests
  on main carry a `partitioned_link` block, so this check is simply unavailable for this class of PR.
- `tubuild.py linkcheck ov071/daEyBm_c` refuses over legacy-source/delinks mismatch — but the
  identical command against the already-landed `ov070/daKpFr_c` gives an **identical** refusal, so it
  is a post-promotion property of the tool, not a defect here.
- The first `rombuild` reported `10936 reused from cache, 0 compiled`. That was not treated as proof;
  `tubuild verify` recompiled the TU and is what establishes the 13/13.

No gate was weakened, skipped, or worked around to reach any green result.

PR #2056, opened 2026-08-31
docs: conventions for genuine-TU promotion PRs

andrewboudreau:
Documentation only. One new file, `notes/tu-promotion-conventions.md`, plus one line appended to the existing reference list in `AGENTS.md` so the note is reachable. No gate, tool, config, header or source change.

## Why

Seven classes have been promoted to genuine translation units across six merged PRs (#2000, #2043, #2045, #2047 for two classes, #2051, #2055), and roughly a dozen more are queued. The same four issues have come up in review each time. This writes them down, grounded only in what has landed, so the queue arrives compliant instead of being corrected one PR at a time.

Every rule cites a merged PR, a tool that enforces it, or a measured fact. PR #2004 is an open draft and is explicitly marked as **not** precedent.

## What it says

1. **A coined mangled name must not assert a parameter type the bytes cannot prove.** `_ZN8daEyBm_c15UpdateCollisionER10dBgCh_Actr` asserts a reference; a pointer mangles differently and generates identical ARM for that body, so the ROM is silent on which the original wrote. Measured: `config/arm9/overlays/ov071/symbols.txt` contains zero comment lines and dsd's format has no comment channel, so the most authoritative-looking surface cannot carry the caveat. It goes in the three places that can and stay adjacent — the manifest entry's `notes` (#2055 already does this), the `why` column of `symbols/actor_renames.tsv`, and the comment on the declaration. That last one is the gap in #2055: `include/daEyBm_c.h` discloses that the *names* are inferred, but not the reference-vs-pointer guess.

2. **Exactly one `extern int _ZTV<C>[];`, in the TU source at namespace scope** — not in the class header, not inside `extern "C"`. Justified on blast radius, not taste: it is the majority landed form (3 of 6 classes with a factory); nothing sweeps `_ZTV` on a rename (`tools/cpp_rename.py` is an address-keyed *file* renamer with no `_ZTV` handling), so every extra copy can strand a spelling — and `include/decl_common.h` already declares it for all seven promoted classes, making both the source-file and class-header forms second copies; a header edit is a reverse-include fan-out build per `MERGE.md`, and the vptr seam is one factory's implementation detail, not public surface; pulling in `decl_common.h` costs the TU its typing freedom (`src/actors/daObjCannonShutter_c.cpp` documents paying that bill); and the `extern "C"` block is inert, since a namespace-scope variable is not mangled — daEyBm_c proves it empirically.

[...]

PR #2057, opened 2026-08-31
TU: first compiler-built vtable — promote ov047/daObjKm3_Kurumajiku_c to intact-object production

andrewboudreau:
Re-cut of #2004 onto current `origin/main` (`a154f0bef`), with **no `tools/**` or `.github/**` change at all**. #2004's branch is `CONFLICTING` and carries a tools delta that is now dead weight — every `def` in its `tubuild.py`, `tu_production.py`, `tu_promote.py`, `objisolate.py`, `rombuild.py`, `rombuild_check.py`, `tiers.py`, `tiers_ratchet.py` already exists on `main`. Closes #2004.

## What this is

`production_mode: "intact-object"` landed on `main` in #2054 **with zero users**. Every one of `main`'s 22 promoted TUs is `.text`-only with `data: []`, and no manifest anywhere sets `production_mode`. So today not one byte of non-`.text` data in the tree is built from source — vtables, RTTI and descriptors are all dsd copies carved from the cartridge.

This is the engine's first car: `ov047/daObjKm3_Kurumajiku_c` claims a second section beside `.text` and compiles its own vtable.

```
.data 0x02112258..0x02112320      (0xc8 = 200 bytes)

  data_ov047_02112258          0xc   three-pointer resource descriptor
  _ZTS21daObjKm3_Kurumajiku_c  0x18  type-name string
  RickshawBs_SpawnInfo         0x1c  typed actor descriptor
  _ZTV21daObjKm3_Kurumajiku_c  0x88  vtable, 32 slots
```

The vtable row records the address-point split explicitly: mwccarm emits the storage at `0x02112298`, the configured public symbol is the slot array eight bytes later at `0x021122a0` (`OI.VTABLE_PREAMBLE`). `symbols.txt` therefore **drops** `data_ov047_02112298` — that preamble now falls inside the TU's own `.data` claim, so dsd must stop delinking it as a standalone symbol.

## Measured on this branch

```
intactTuRom.identical: true   (sha256 d1506e90…c478e8)

  reproducing 11,088   mismatching 0
source-owned data claims: 1  (reproducing 1, 200 B; mismatching 0)
ROM-build analysis: PASS
```

```
source_coverage.py --check --merge-base origin/main
  base a154f0bef
  2,067,148 B -> 2,067,348 B   (net +200 B)
  handed back to the cartridge: 0 B
  by section: .data 200 B, .init 125,712 B, .text 1,941,436 B
```

**The `.data` line was zero before this PR.** These 200 bytes are the entire compiler-built non-`.text` output of the repository.

## The CONVERTED ratchet, and why the exception is honest

2568 → 2568, a pure identity rename: `+4` member identities, `-4` physical paths.

All four absorbed method files score **PASS** as `path#symbol` members. The destination is still refused because `_member_result()` requires *every* member of the destination TU to pass, and the fifth member — the `.c`-derived `RickshawBs_Spawn` — fails *"Calls things by real names, not mangled `_Z`"*. It failed that before promotion too, as `src/RickshawBs_Spawn.c`, which was never CONVERTED. Nothing readable was de-converted.

Its base-constructor call and its deliberate dead vptr store are load-bearing for the byte match, and byte-match comes first. Spelling them in real names is a separate readability change.

There is a latent gate defect underneath this, which I have **not** fixed here because `tools/` must land on its own PR: a banked absorbed file's stem *is* its member symbol, so `why()` could judge each banked path against its own member instead of against all of them. As written, any promotion that merges CONVERTED files together with one never-CONVERTED intake file is reported as a backslide even when nothing readable was lost.

## Gates run locally

| gate | result |
|---|---|
| `rombuild.py -j 16` | 106/106 exact, PASS |
| `source_coverage.py --check` | clean, +200 B |
| `tiers_ratchet.py --check` | PASS 2568/2568 |
| `check_src_tu.py` | 92/92 TUs compile |
| `port_refcheck.py` | 405 checked, all resolve |
| `check_dead_references.py` | no new dead refs, no broken links |
| `check_header_offsets.py --changed` | 0 mismatched, span 0x330 |
| `check_python_names.py` | PASS |
| `prepush_attribution.py` | 3 consolidated with credit intact, 0 changed, 0 lost |
| `test_tiers_ratchet` + `test_attribution` + `test_source_coverage` | 62 tests OK |

Attribution needed three `path#symbol` overrides so the many-to-one consolidation reads as *consolidated, credit intact* rather than three lost symbols; they are appended, not re-sorted, to keep the file cherry-pick-clean.


andrewboudreau:
## ⚠️ Do not merge on the current green — this PR's proof expired when #2059 landed

**Not a rejection.** The structural work here still looks right to me. But the green checkmark on this
PR no longer certifies today's `main`, and I want that visible to anyone with a merge button rather
than sitting in my notes.

### The problem in one table

| | value |
|---|---|
| this PR's pinned base | `a154f0bef5f058fd64666593c1a107881bc85efc` |
| current `origin/main` | `a96b900391aa23f3d54dbf650cf3a0b1cfac4795` |
| merges landed in between | #2056, #2058, #2059 |

The validator pins the base SHA it ran against. This PR's run measured a tree that **no longer
exists**, so `mergeStateStatus: CLEAN` and a green validator are both stale artifacts here, not
evidence about main.

### Why that matters *specifically* for this PR, and not for most stale PRs

#2059 ("Fix owned RTTI vtable address-point checks") changed the two files this promotion's entire
claim rests on:

```
tools/tubuild.py        04f8eaf85a0be93d5717a3f8cfc6691463f3b147 -> 8513e7b3dd9463407eb1d0eeb0ec461623cc661b
tools/tu_production.py  2a0af841eb6d42f6420e3fc564fc249eca0822ab -> 780c9c7ae70bce05dca9a01a0edefa3ae9e63353
```

`tubuild.py` is the verifier — fine, that only changes how a verdict is reached. **`tu_production.py`
is not.** `prepare_intact_object()` lives there, and per main's `rombuild.py:_isolate` it
short-circuits *before* `isolate_many` ever runs:

```python
intact = (intact_tus or {}).get(rel.replace("\\", "/"))
if intact is not None:
    prepared, _evidence = TP.prepare_intact_object(obj.read_bytes(), intact)
    obj.write_bytes(prepared); return None
```

That is the code path which **admits this PR's 200-byte `.data` claim** — it is production code, not
diagnostics. So the object this PR proves may now be *prepared* by different code than the code that
produced its proof. Re-running the same verifier is not enough; the input to the verifier may differ.

This PR is also the **first real exerciser** of that path (source-owned non-`.text` was 0 bytes
tree-wide before it), so there is no prior art telling us the two behave identically.

### What is being measured before I'd merge this

A review is re-verifying against current `main` (`a96b90039`), not the stale base:

1. `tubuild.py verify ov047/daObjKm3_Kurumajiku_c` on the **new** merge tree.
2. `rombuild.py -j 16 --no-rom` — expecting `106/106 exact`, `11,088` reproducing, `0` mismatching.
3. A full ROM build, sha256 against the baseline control ROM.
4. Whether the `tu_production.py` change can alter the **bytes** `prepare_intact_object()` emits, or is
   confined to plumbing. That single fact decides this PR.
5. `romdata_check --data-json` with `blindWords` per symbol across `.data 0x02112258..0x02112320`.

**The specific outcome I am watching for is "newly-passing."** If this claim verifies now but did not
before, that means it passes because the verifier moved, and I would not merge it on that basis.
Identical verdicts on the new base and I have no objection left.

### The byte question that green cannot answer

The entire `symbols.txt` diff is one removed line:

```
-data_ov047_02112298 kind:data(any) addr:0x02112298
```

That placeholder sits at the vtable **storage start**, exactly 8 bytes below the address point
`_ZTV21daObjKm3_Kurumajiku_c` at `0x021122a0` — the offset-to-top + typeinfo preamble. Dropping it may
be entirely correct here: it is inside a span whose bounds coincide with two surviving symbols, and
`_ZTV21daObjKm3_Kurumajiku_c` itself is **absent from the diff**, i.e. byte-identical to main, which is
the single best sign of a legitimate promotion. The `.text` range is likewise the exact contiguous
union of the five deleted entries with no growth.

I flag it only because *every byte gate passes either way*. The legitimate case and the anti-pattern —
deleting a neighbouring `data_*` entry until a gate agrees — are indistinguishable from a green tick,
which is why this needs the per-symbol `blindWords` evidence rather than a summary line.

(For the record: our own conventions doc currently states this anti-pattern too bluntly. Dropping a
neighbouring `data_ov047_*` entry can be *required*, not forbidden. That doc is mine to fix, and it
errs toward over-rejection, so it is not being held against this PR.)

### How to unblock it

Refresh the branch onto current `main` so the validator produces a run pinned to a base that actually
exists. I have deliberately not pushed to `tu/kurumajiku-intact-recut` — it is not my branch, and this
PR carries ledger files (`config/converted-baseline.json`, `attribution.json`,
`config/converted-backslide-exceptions.jsonl`) where a mechanically-clean merge is not the same as a
correct one. GitHub's server-side merge ignores the `merge=union` driver, so those want
**regenerating** against main rather than merging forward.

Once it is rebased and re-verified on the new base, and the verdict is unchanged, I have no further
objection.

No gate was weakened, skipped, or worked around in producing any of the above.

andrewboudreau:
## Hold retracted. All five items discharged, and one of my own premises was wrong.

My earlier comment on this PR told readers not to merge and listed five things I was measuring. All five are now answered, at head `a6954e8a5`. Retracting the hold explicitly, because it is still sitting above this comment telling people to wait.

### The five

**1. `tubuild.py verify` on the merge tree — IDENTICAL.** Same verdict, same output, before and after.

**2. `rombuild.py -j 16 --no-rom`** — module fidelity **106/106 exact, 100.000000%** of compared bytes. **11,088 reproducing, 0 mismatching.** Source-owned data claims: 1 reproducing, 0 mismatching. ROM-build analysis **PASS**.

**3. Full ROM sha256 — MATCH.**

```
build      d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
baseline   d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
16,777,216 bytes    rombuild-report.json -> intactTuRom.identical: true
```

This is the one that actually settles it. 106/106 is a per-module comparison; a whole-ROM digest is not.

**4. The byte question, answered better than I asked it.** I asked whether the intact object still verifies after #2059 changed the vtable address-point handling. The answer came back as an A/B instead: the prepared object is **byte-identical under both verifiers**, `a9acdb65f01611954cf33319c443e97c44f2fbb8030f47980b54fee03cfa0621` at `8f554bce3` (pre-#2059) and at `a96b90039` (post-#2059).

That A/B is only meaningful because `rombuild` caches the object *before* `_isolate` and re-runs `_isolate` on every cache hit (`rombuild.py:1016-1022`). Without that, a "reused, 0 compiled" run would have made the matching digest an artifact of the cache rather than evidence. Worth stating, because the same shape of test would be worthless in most build systems.

**Bonus, unasked:** all **22/22** promoted TUs produce byte-identical `verify` results across those two commits. No verdict moved in either direction, so #2059 did not shift verdicts tree-wide — only for ov047, deliberately.

**5. The dropped `data_ov047_02112298` entry is unkeepable, proven by restoring it.** Not argued — restored, and the link fails:

```
[ERROR] Symbol 'data_ov047_02112298' in overlay 47 at 0x02112298 not found in linked binary
```

The placeholder sits 8 bytes below the address point, on the offset-to-top word. And the discriminator that distinguishes this from someone deleting an entry to silence a gate: an *unkeepable* drop **costs** a verified symbol (`RickshawBs_SpawnInfo` goes VERIFIED -> PARTIAL, 463 -> 462) whereas a gate-silencing drop would improve a metric. That test is now written down in `notes/tu-promotion-conventions.md` (#2061) so the next reviewer doesn't have to rediscover it.

### Correcting myself on the ledger files

My hold said this:

> this PR carries ledger files (`config/converted-baseline.json`, `attribution.json`, `config/converted-backslide-exceptions.jsonl`) where a mechanically-clean merge is not the same as a correct one [...] those want **regenerating** against main rather than merging forward.

**That was wrong here, and I should have measured before writing it.** Regeneration only matters when both sides moved. Main has touched none of these files since this PR's base `a154f0bef`:

```
config/converted-baseline.json               0 commits on main since base
attribution.json                             0 commits on main since base
config/converted-backslide-exceptions.jsonl  0 commits on main since base
```

One-sided change means no resolution happens at all. Confirmed on the actual merge tree rather than inferred — every one of the three blobs lands byte-identical to this PR's:

```
config/converted-baseline.json                b570c12cfd52b5dee5ad7f6258d4a5774f71a2d8
attribution.json                              174866443e50733ba9c818d1d2049cf81a7da6f6
config/converted-backslide-exceptions.jsonl   dd50887ebb6824682c9a4fdc627d292a27ec6d18
```

I also can't substantiate the merge-driver half of that sentence: there are no `merge=union` entries in `.gitattributes` on `origin/main` at all. Rather than restate a claim I can't confirm, I'm withdrawing it. It makes no difference to this PR either way — with zero divergent commits there is nothing for any driver to resolve.

### Why the ROM proof still holds after `main` moved

The ROM was verified against `main` at `a96b90039`. `main` has since advanced to `c804e96f6` (#2060 landed). A moved base normally expires a byte measurement, so I checked what actually changed between the tree that was proven and the tree that will land:

```
git merge-tree --write-tree a96b90039 a6954e8a5  ->  3930f020b
git merge-tree --write-tree c804e96f6 a6954e8a5  ->  bac7f4242

git diff --stat 3930f020b bac7f4242
 notes/mwccarm-codegen.md | 34 +++++++++++++++++-----------------
 1 file changed, 17 insertions(+), 17 deletions(-)
```

One markdown file, formatting only. No source, config, tool, symbol, or ledger file differs between the tree that produced the matching ROM digest and the tree that lands. The proof carries.

### Non-blocking, for later

- The 4 backslide-exception rows cite "PR #2004" and are inert — worth a follow-up cleanup, not a blocker.
- `include/daObjKm3_Kurumajiku_c.h:18` still has a stale "placeholder names" comment.
- `tubuild.py verify` prints PROMOTION REFUSED for `production_mode: "intact-object"`. It's a false alarm on a mode the tool doesn't know about, and it will mislead the next person who runs it. Teaching the tool that mode is the fix.

No gate was weakened, skipped, or worked around in this review. The one thing I got wrong, I got wrong in the *strict* direction — I held a PR on a merge hazard that measurement shows does not exist.

andrewboudreau:
## Correction: the `merge=union` half of my retraction was itself wrong

In the comment above I wrote:

> I also can't substantiate the merge-driver half of that sentence: there are no `merge=union` entries in `.gitattributes` on `origin/main` at all. Rather than restate a claim I can't confirm, I'm withdrawing it.

**That is false, and I withdrew a claim that was correct.** `.gitattributes` on `origin/main` does carry union entries, one of which is a file this PR touches:

```
config/match_provenance.jsonl                 merge=union
config/match_attempts.jsonl                   merge=union
nearmiss/db.jsonl                             merge=union
CLAIMS.md                                     merge=union
config/converted-backslide-exceptions.jsonl   merge=union
```

My grep returned nothing because git-bash rewrote the argument `origin/main:.gitattributes` into `origin\main;.gitattributes` before git saw it — MSYS path conversion mangling a rev:path spec. That fails as *"unknown revision or path"*, and piped into `grep` it looks exactly like a clean "no matches". A false negative that reads as evidence of absence.

And the file itself documents the very behaviour I had claimed and then abandoned:

> NOTE: this driver only runs on a local `git merge`. GitHub's own merge (what actually lands a PR) does not honor `merge=union` and will still show a real conflict the instant two PRs touch this file's tail concurrently — land PRs that append here one at a time, same as the other union-merged ledgers above.

### What this does and does not change

**It does not change this PR's outcome, and the merge was correct.** The union driver only matters when both sides moved, and the measurement that actually decided this — zero commits on `main` touching any of the three ledger files since base `a154f0bef` — is unaffected. I confirmed the result on the merge tree directly rather than by reasoning: all three blobs landed byte-identical to this PR's own (`b570c12cf`, `174866443`, `dd50887eb`). No resolution ran, by any driver.

**What it does change is the standing rule for the next PR.** Of the three files here, exactly one — `config/converted-backslide-exceptions.jsonl` — is union-merged, and union does not run server-side. So two PRs appending to that file concurrently will conflict on GitHub even though a local merge would resolve cleanly. **Ledger-appending PRs land one at a time.** That rule is live again; I should not have retracted it.

Net: my original hold cited a real hazard that did not apply *here*, and my retraction then over-corrected by denying the hazard exists at all. The hold was right to be discharged; the reasoning I discharged it with was too broad by one sentence.

PR #2058, opened 2026-08-31
ci: run the three wirable intact-TU test suites, pin their deps

andrewboudreau:
`tool-tests.yml`'s header lists five test suites it leaves out because a bare runner has no `pyelftools`, and calls pinning them "a fair follow-up ... it should come with a requirements file rather than a version invented in a `run:` line." This is that follow-up — but only for the three of the five that survive the measurement.

## The measurement

CI has neither the compiler nor the ROM dump, so a suite that self-skips without them reports success while asserting nothing. To find out which of the five do that, I exported a tree with `git archive origin/main` — no `extracted/`, no `tools/mwccarm/`, no `tools/bin/` — which is exactly what this job's `actions/checkout` produces, and ran each suite there and in a checkout that *does* have the toolchain.

That tree reproduces this job's documented baseline exactly (`Ran 435 tests ... OK`), so it is a faithful stand-in.

| suite | blocker | toolchain PRESENT | toolchain ABSENT (= CI) | verdict |
|---|---|---|---|---|
| `test_rombuild` | pyelftools + capstone | 37 pass | **34 pass / 3 skip** | **WIRE** |
| `test_tu_production` | pyelftools + capstone | 17 pass | **17 pass / 0 skip** | **WIRE** |
| `test_tu_promote` | pyelftools + capstone | 9 pass | **9 pass / 0 skip** | **WIRE** |
| `test_objisolate` | pyelftools + capstone | 34 pass | **0 pass / 34 skip** | DO NOT WIRE — hollow |
| `test_tubuild` | pyelftools + capstone | 45 pass | **28 real pass / 15 vacuous / 2 fail** | DO NOT WIRE — red *and* hollow |

### Two corrections to the premise

**The blocker is two packages, not one.** All five reach `elftools` through `objisolate.py`. Installing `pyelftools` alone does not unblock any of them — it moves the failure one hop, to `rombuild.py` → `romdata_check` → `linkcheck` → `match` → `capstone`, and all three wirable suites still die at *collection*. `tools/requirements.txt` pins both. `ndspy` is deliberately not in it: no wired suite reaches it.

**`test_tubuild` looks far more wirable than it is.** Under pytest on the toolchain-free tree it reports `2 failed, 43 passed`, which reads like a suite that is 95% there. It is not. Fifteen of its tests open with `if not _toolchain(): return`, and **fourteen of those returns are bare** — those tests report PASS having asserted nothing. A bare `return` guard is strictly worse than a skip, because `pytest -rs` cannot see it and neither can a reviewer reading a green log. It is also pytest-style, so `python -m unittest tools.test_tubuild` prints `Ran 0 tests ... OK` and exits 0. Its 2 real failures are `tubuild list` needing `build/tu_map.json`, which regenerates from the ROM and comes back empty without it.

`test_objisolate` is the simpler rejection: its single `TestCase` class is `@skipUnless(_compiler())`, so on a runner it is `Ran 34 tests ... OK (skipped=34)`. Wiring it would add 34 to this job's headline count and zero assertions.

## What this changes

- **`tools/requirements.txt`** (new) — pins `pyelftools==0.33`, `capstone==5.0.9`, with the import chains written down so the next person does not have to rediscover that it takes both.
- **`.github/workflows/tool-tests.yml`** — one `pip install -r tools/requirements.txt` step; three modules added to the enumerated list; header corrected.

Nothing else. No tool changes, no config, no ledger files.

**The list is still enumerated, not a glob.** That constraint is what makes this PR safe: a glob would have absorbed `test_objisolate` and scored its 34 skips as a pass.

### Numbers, stated plainly

Job total **435 → 498** tests. Expected skips **6 → 9**.

The three new skips are all in `test_rombuild`: its `Retarget` class compiles a real object and `@skipUnless`-es itself with `"mwccarm not present"`. **So of the 63 tests this PR adds, 60 assert something in CI and 3 do not.** All nine expected skips are `@skipUnless` with a stated reason and are visible under `-v`; none is a bare `return`.

Wall time is unchanged — 46.5s for 435, 45.7s for 498 on the same tree, i.e. inside the noise. The three added suites are mock- and tempdir-based and touch neither the network nor the tree (verified: the worktree was clean after running all five, and the suites create nothing outside `tempfile`).

The cost is that this job is no longer stdlib-only; it now needs one `pip` step and the network access that implies. The header says so.

### What this does *not* prove

The header notes 24 modules blocked on `elftools`. **The other 21 were not re-measured with `capstone` installed and are not known to be wirable** — several are also toolchain-bound. Adding one of them is a fresh measurement, not a corollary of this one. The header says that too.

`.github/workflows/update-chaos-data.yml` installs the same two packages **unpinned** in a `run:` line. Pointing it at this file is a fair follow-up, but it is a different change and it touches a workflow that writes commits to `main`, so it is not in here.

## Verification

- Ran the workflow's `run:` block **verbatim** against the toolchain-free tree: `Ran 498 tests ... OK`.
- Baseline on the same tree before the change: `Ran 435 tests ... OK`.
- Every skip audited by reason, not just counted.
- `pyelftools`-absent and `capstone`-absent runs done with a `sys.meta_path` blocker, so the import wall was measured rather than assumed.
- Merge-tree gates via `tools/premerge_check.py --fetch` — results in a follow-up comment.

Measured on `8f554bce3`.

### DIFF
--- a/.github/workflows/tool-tests.yml
+++ b/.github/workflows/tool-tests.yml
@@ -84,9 +84,11 @@
 # price of the three intact-TU modules. Measured at 5.3s for the 312 tests that predate
 # tools.test_premerge_check, on a runner-equivalent container (python:3.12, Linux,
 # none of the three inputs present); that module adds 87, timed at 9.7s on Windows.
-# The three modules added below cost no measurable wall time at all: 435 tests in 46.5s
-# and 498 tests in 45.7s on the same `git archive` tree, i.e. inside the noise. They are
-# mock- and tempdir-based and touch neither the network nor the tree.
+# The three modules added below cost no measurable wall time at all: on one Windows tree,
+# 435 tests in 46.5s before and 498 in 45.7s after -- inside the noise. On the runner
+# itself the whole 498 take 11.5s, plus about 1s for the two wheels (both are prebuilt;
+# nothing compiles). They are mock- and tempdir-based and touch neither the network nor
+# the tree.
 # The whole-suite timings are dominated by Windows filesystem overhead and do not
 # apply here -- test_tiers alone takes 71s on a Windows checkout and 1.3s on Linux.
 # test_symrecords needs none of the three inputs either: it reads only tracked text
--- a/.github/workflows/tool-tests.yml
+++ b/.github/workflows/tool-tests.yml
@@ -10,30 +10,49 @@
 # gate's FAILURE branch is the one path a green run never exercises: no ROM build can
 # reach it, because these tools are not compiled into the ROM. src-tu-refs.yml exists
 # because #1643 broke 26 translation units while every byte gate stayed green, and the
-# only thing that noticed was a test suite a human had to run. This job runs 435 of
+# only thing that noticed was a test suite a human had to run. This job runs 498 of
 # those assertions on every PR that touches `tools/` or `nearmiss/`, so the next one is
 # noticed by CI.
 #
 # WHAT THIS DOES NOT COVER, and why. Three groups are deliberately absent.
 #
 #   1. NEEDS THE TOOLCHAIN. mwccarm and the extracted ROM are not on a GitHub runner;
-#      they live on the private build box pr-validate.yml relays to. Measured in a
-#      checkout with `extracted/`, `tools/mwccarm/` and `tools/bin/` absent:
-#      test_objisolate skips all 32 of its tests, test_rombuild skips 3 of 32, and
-#      test_check_src_tu_compiles FAILS 8. Wiring those here would buy a green that
+#      they live on the private build box pr-validate.yml relays to. Re-measured against
+#      8f554bce3 in a tree exported with `git archive origin/main` -- no `extracted/`, no
+#      `tools/mwccarm/`, no `tools/bin/`, which is exactly what this job's checkout is:
+#      test_objisolate skips ALL 34 of its tests (the older note here said 32; the suite
+#      grew), and test_check_src_tu_compiles FAILS 8. Wiring those would buy a green that
 #      means nothing. Also absent for the same reason: test_build_pin, test_dtor_members,
 #      test_fdiff_version, test_gen_header, test_opnew_sizes, test_tubuild.
 #
-#   2. NEEDS A PIP DEPENDENCY. A bare runner has no pyelftools, capstone or ndspy, and
-#      this repo has no requirements file to pin them from. That import wall -- not the
-#      compiler -- is what actually stops most of `tools/` on a runner: 24 modules die
-#      on `No module named 'elftools'` alone (test_rombuild, test_romdata_check,
-#      test_check_references, test_tu_config, test_tu_production, test_reloc_audit_*,
-#      test_mangle, test_name_roundtrip, test_pr_linkcheck_*, test_tu_promote,
-#      test_objisolate, and more), test_swarm and
-#      test_worklist on capstone, test_bmg on ndspy. Pinning those is a fair follow-up;
-#      it is a different change from this one, and it should come with a requirements
-#      file rather than a version invented in a `run:` line.
+#      test_tubuild deserves its own sentence, because it is the worst of the shapes and
+#      it is easy to look at its numbers and conclude the opposite. Run under pytest in
+#      that toolchain-free tree it reports "2 failed, 43 passed" -- which reads like a
+#      suite that is 95% wirable. It is not. Fifteen of its tests open with
+#      `if not _toolchain(): return`, and fourteen of those returns are bare, so those
+#      tests report PASS having asserted nothing. The honest split is 28 real passes, 15
+#      vacuous passes, 2 failures (`tubuild list` needs build/tu_map.json, which
+#      regenerates from the ROM and comes back empty without it). A bare `return` guard
+#      is strictly worse than a skip: `pytest -rs` cannot see it and neither can a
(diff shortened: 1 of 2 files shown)


andrewboudreau:
## Reviewed and merging — and the part I value most is the two suites you *didn't* wire.

Verified at head `cea557b0c`, base `origin/main` `8f554bce3`. I recomputed the merge tree independently
and got `c158af477fe2214d9bc1d4457f0ef2846c5f4b2f` — the same tree the gate run below measured, against
an unchanged base, so nothing here is stale.

Scope is two files, `+125/-28`: a new `tools/requirements.txt` and `.github/workflows/tool-tests.yml`.
**No source, no config, no ledger file, no `attribution.json`** — I checked that directly rather than
taking the description's word for it. This PR therefore cannot move a single ROM byte, which is why it
can land beside the two in-flight TU PRs without interacting with either.

| gate | result |
|---|---|
| `premerge_check.py --fetch 2058` | merge tree `c158af477fe2` vs base `8f554bce3` — **all 8 gates pass → pass**, `nothing goes green -> red` |
| all PR checks | 5/5 green (`PR validation`, `references`, `refs`, `submit`, `tools`) |
| **the `tools` job itself** | `Successfully installed capstone-5.0.9 pyelftools-0.33` · `Ran 498 tests in 11.501s` · `OK (skipped=9)` |

### The decisive evidence is the runner log, not the summary line

I don't accept "the tests pass" from a PR body, so I pulled the job log. The newly-wired suites are
visibly *executing and asserting* on a runner that has no mwccarm and no ROM — including
`tools.test_tu_production.ProductionTuObjects.test_final_gate_allows_only_baseline_symbol_errors` and
`test_automatic_intact_link_plan_rejects_laundered_control_error`.

That is the intact-TU admission gate. Those tests currently run **nowhere** — not in any workflow, not
in any hook. Wiring them has more value this week than it would have in any other week, because the
intact-object path is exactly what the open TU work is changing.

### Declining `test_objisolate` and `test_tubuild` is the correct call, and it is the reason this merges

Wiring all five, or reaching for a glob, would have made this job's green **weaker than before**:

- `test_objisolate` — **0 pass / 34 skip** without the toolchain. Thirty-four skips scored as a passing job.
- `test_tubuild` — **28 real / 15 vacuous / 2 failing**. Fifteen tests open with `if not _toolchain(): return`
  and **fourteen of those returns are bare**, so they report PASS having asserted nothing. A bare return
  is not a skip: `pytest -rs` cannot see it and neither can a reviewer.

I verified the glob trap independently, and the workflow header now documents it with real numbers:
`python -m unittest tools.test_tubuild` prints `Ran 0 tests ... OK` and exits 0, because that file is
pytest-style (45 bare `def test_`, zero `TestCase`). A glob would have silently counted 186 test
functions as passing. The module list is **explicitly enumerated** — I confirmed there is no glob and no
wildcard anywhere in the file.

This is the right instinct for this repo specifically. The recurring failure mode here is not a red
gate, it is a **hollow green**, and a test suite that skips its real cases is the purest form of it.

### The premise correction is worth more than the fix

The blocker is described everywhere in our notes as "pyelftools". It is **two packages**. Installing
`pyelftools` alone unblocks none of the five — it moves the failure one hop, from `objisolate.py:52` to
`match.py:27` (`rombuild → romdata_check → linkcheck → match → capstone`). Measured with a
`sys.meta_path` blocker rather than inferred. `ndspy` is deliberately absent because no wired suite
reaches it, and the file says so.

Pinning is right for the same reason `python-names.yml` pins pyflakes: a transitive dependency moving
under a gate that touched no Python is indistinguishable from a regression — and these two packages
parse the ELF objects and ARM encodings that every byte-fidelity claim in this repo rests on.

The measurement method is what makes the numbers trustworthy: a genuinely toolchain-free tree exported
with `git archive origin/main`, which **reproduced this job's documented baseline exactly**
(`Ran 435 tests ... OK`). A run inside a wired worktree would have proved nothing.

`435 → 498` (+63), skips `6 → 9` (+3 — 4 ROM, 2 no-src-samples, 3 mwccarm). **60 of the 63 added tests
assert something in CI; 3 do not**, and the PR body says so openly rather than rounding up.

### Explicitly unverified

- **No ROM was built.** Unnecessary by construction — no source, config, or ledger file is touched —
  but stated rather than implied. `premerge_check` says the same of itself.
- **Whether these are *good* tests was not established.** They execute and assert without the toolchain;
  whether their mocks are faithful to `rombuild.py` is a different audit, and a suite can be 100%
  non-skipping and still be testing its own mock. Worth doing now that #2052/#2053/#2054 have landed.
- **The other 21 elftools-blocked modules were not re-measured** with capstone present and are **not**
  known to be wirable. The header carries that caveat so nobody over-generalises from this PR.
- `update-chaos-data.yml` installs the same two packages **unpinned** and was deliberately left alone —
  it commits to `main`, so it is a different change with a different risk profile.
- Flakiness at scale and version-pin drift: only current-latest versions, a handful of runs.

### One follow-up I'm carrying, not asking you to fix here

`test_tubuild`'s 2 toolchain-free failures and 14 vacuous passes are **pre-existing on main** and are not
this PR's problem. But PR #2059 adds its "fail-closed wrong-addend / wrong-destination" coverage to that
exact file, so its new tests inherit this hazard. I've routed that to #2059's review rather than
blocking you with it.

No gate was weakened, skipped, or worked around — nothing here needed it.

PR #2063, opened 2026-08-31
Cover tubuild's fail-closed vtable-addend arm in a place CI actually runs

andrewboudreau:
Tests only. **`tools/tubuild.py` is not modified** — it is byte-identical to `origin/main` (verified after every mutation run below). The diff is one new test module and one enumeration line plus its comment block in `tool-tests.yml`. It touches only `tools/` and `.github/workflows/`, so it cannot move a ROM byte and does not interact with the TU promotion queue.

## The hole

PR #2059 added a normalized-undefined-vtable arm to `verify_owned_sections`:

```python
if public_address_points and expected is not None:
    bias = retained_vtable_biases.get(expected["symbol"])
    if bias is None and normalized_undefined_vtables \
            and expected["symbol"].startswith("_ZTV") \
            and expected_addend:
        bias = OI.VTABLE_PREAMBLE
    if bias is not None:
        expected_addend = (expected_addend - bias
                           if expected_addend >= bias else None)
```

`objisolate.rebias_object_symbols(..., normalize_undefined=True)` rewrites an undefined `_ZTV` import's addend from `public + N` to `public + N - 8`, because `symbols.txt` already names the slot array past the two-word ABI preamble. This undoes that on the manifest side so the two are comparable.

The `else None` is the **fail-closed** half. A manifest addend *below* the preamble cannot have come from that rewrite, so there is no honest emitted addend to compare against and the relocation must be rejected — `None` reaches `rel["addend"] != expected_addend` and lands on `WRONG-ADDEND`. Nothing downstream re-checks it.

**Nothing tested it.** The nearest existing assertion uses `8 + 4 = 12`, which is *above* the preamble and takes the arithmetic arm, never the `else`.

## Measured, not assumed

Mutation testing against the **whole** of `tools/test_tubuild.py` (46 tests) run **with the pinned compiler and `extracted/` present**, so every `if not _toolchain(): return` guard in that file is live — the most generous conditions the existing suite can be given:

| mutation of `tools/tubuild.py` | `test_tubuild.py` | this PR's module |
|---|---|---|
| `else None` → `else 0` | 46/46 pass — **uncaught** | 2 fail — caught |
| `else None` → `else expected_addend` | 46/46 pass — **uncaught** | 2 fail — caught |
| drop `expected["target_address"] != cfg[1]` | 46/46 pass — **uncaught** | 1 fail — caught |
| `>=` → `>` (preamble boundary) | 44/46 — caught | 1 fail — caught |
| `bias = OI.VTABLE_PREAMBLE` → `None` | 45/46 — caught | 6 fail — caught |
| drop `cand_addr != expected["target_address"]` | 45/46 — caught | 1 fail — caught |

Three mutations that make the verifier fail **open** leave the existing suite entirely green. All six are caught here.

`tools/tubuild.py` was restored from a pristine copy after each run; `git status --porcelain -- tools/tubuild.py` is empty and the file's SHA-256 matches `origin/main`.

## …and none of that ran in CI anyway

`tools/test_tubuild.py` is pytest-style: 46 bare `def test_*` functions, zero `unittest.TestCase` classes. `tool-tests.yml` runs `python -m unittest`, so:

```
$ python -m unittest tools.test_tubuild
----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
$ echo $?
0
```

A green that asserts nothing. Fifteen of its tests additionally open with `if not _toolchain(): return` — a bare return, not a skip, so on a compiler-less runner they would report PASS having asserted nothing. Every existing owned-relocation assertion sits behind one of those guards. Both facts are already documented in `tool-tests.yml`'s header; this PR does not change either, and does **not** convert `test_tubuild.py` (a much larger change — see "Not done here").

## What this adds

`tools/test_tubuild_owned_relocs.py` — `unittest.TestCase`, six tests, enumerated **explicitly** in `tool-tests.yml`. No glob: a glob would silently absorb future toolchain-dependent modules and convert self-skips into a hollow green, which is exactly what that file's header refuses.

The only part of `verify_owned_sections` that needs a compiled object is `section_contribution`, which is pure ELF parsing returning `{"bytes", "relocs", "symbols", "inputSections"}`. Everything the verdict ladder decides comes from that dict plus the manifest, the config relocation map and the symbol name index — all caller-injected. So the tests stub `section_contribution` with a layout of exactly that documented shape and drive the **real** `verify_owned_sections` over it.

Verified on a toolchain-free tree (`git archive origin/main` into a directory with no `extracted/`, no `tools/mwccarm/`, no `tools/bin/`, no `build/`): **6 run, 6 assert, 0 skip**, and the `else 0` mutation applied *in that tree* fails 2 of them. No test here self-skips; if this module ever contributes fewer than 6 tests to the job's `Ran N` line, something is wrong.

Tests:

- `test_normalized_import_above_the_preamble_verifies` — positive control (manifest 16 vs emitted 8 → OK). Without it the refusal tests could be passing for any reason at all.
- `test_manifest_addend_below_the_preamble_is_refused` — the `else None` arm. Every other fact agrees, so `else 0` would call it OK.
- `test_sub_preamble_addend_is_not_silently_left_unbiased` — the other permissive rewrite, `else expected_addend`.
- `test_addend_exactly_at_the_preamble_normalizes_to_zero` — `>=` is inclusive; a `>` would push the most common case into the refusal arm.
- `test_candidate_destination_must_match_the_manifest` — `WRONG-DEST` clause one.
- `test_manifest_destination_must_match_the_configured_relocation` — `WRONG-DEST` clause two. Asserted separately from clause one, because dropping either alone leaves a check that still *looks* present — and clause two was uncaught by the existing suite.

## Suite totals

Enumerated `tool-tests.yml` list, one `python -m unittest -v` invocation, same tree:

- before (26 modules): `Ran 498 tests ... OK (skipped=3)`
- after (27 modules): `Ran 504 tests ... OK (skipped=3)`

`+6`, `+0` skips. 498 matches the figure `tool-tests.yml` already documents.

## Not done here (findings, deliberately out of scope)

1. **The 15 vacuous `if not _toolchain(): return` guards in `test_tubuild.py`.** A bare return is not a skip: `pytest -rs` cannot see it and neither can a reviewer. Turning them into `@unittest.skipUnless` is the right fix and is a real change to those tests.
2. **Converting `test_tubuild.py` to `unittest.TestCase`** so its 46 tests become visible to CI. Twelve other `tools/` modules are in the same shape (`tool-tests.yml` header, group 3: 186 test functions a glob would silently count as passing).

Both are larger changes and neither belongs in a PR whose point is to make one specific fail-closed branch observable.

## Gates run locally

- `python -m unittest -v <the 27 enumerated modules>` → `Ran 504 tests ... OK (skipped=3)`
- `python -m pyflakes tools/test_tubuild_owned_relocs.py` → clean
- `python tools/check_python_names.py` → PASS (261 files, 0 unresolvable)
- workflow YAML parses; no `src/` or `src_tu/` token anywhere in the diff
- pre-push: `port-refcheck 405 checked - all references resolve`, `duplicate-sources: 11158 stems, none doubled`, `check_src_tu_compiles: 92/92`

### DIFF
--- a/.github/workflows/tool-tests.yml
+++ b/.github/workflows/tool-tests.yml
@@ -194,6 +194,7 @@ jobs:
             tools.test_symrecords \
             tools.test_tu_production \
             tools.test_tu_promote \
+            tools.test_tubuild_owned_relocs \
             tools.test_validate_merge
 
 # What each module protects, in the order it runs above:
@@ -281,6 +282,31 @@ jobs:
 #                                      dilutes the CONVERTED tier, so a scoring bug here
 #                                      moves a published number with nothing else
 #                                      noticing. Pure mock; 9 of 9 run in CI.
+#   test_tubuild_owned_relocs    (6)   tubuild.verify_owned_sections' non-text
+#                                      relocation verdict ladder -- specifically the
+#                                      fail-closed `else None` #2059 added for a
+#                                      normalized undefined `_ZTV` import whose manifest
+#                                      addend is BELOW the two-word ABI preamble, and
+#                                      both clauses of the WRONG-DEST conjunction. It
+#                                      stubs `section_contribution` (the only part of
+#                                      that function that parses ELF) and drives the
+#                                      real verifier over a hand-written layout of the
+#                                      documented shape, so it needs no compiler, no
+#                                      ROM, no config/ and no build/. 6 of 6 run in CI,
+#                                      none skips. It exists as its own module rather
+#                                      than as more of test_tubuild.py for the reason
+#                                      spelled out in group 3 of the header: that file
+#                                      contributes zero tests to `python -m unittest`,
+#                                      and the assertions it already has for this code
+#                                      sit behind `if not _toolchain(): return`.
+#                                      Mutation-checked, all five against the whole of
+#                                      test_tubuild.py running WITH the pinned compiler
+#                                      present: `else None` -> `else 0` and
+#                                      -> `else expected_addend` and dropping
+#                                      `expected["target_address"] != cfg[1]` each leave
+#                                      46/46 of that suite green and are each caught
+#                                      here; `>=` -> `>` and disabling the normalized
+#                                      bias are caught by both.
 #   test_validate_merge         (35)   validate_merge.py -- builds the machine-readable
 #                                      PR validation report the validator comment renders.
 #
@@ -288,4 +314,6 @@ jobs:
 # `git archive origin/main` tree with no toolchain: test_objisolate 0 pass / 34 skip
 # (every test in it compiles), test_tubuild 28 real pass / 15 vacuous / 2 fail. Neither
 # is a candidate until objisolate's fixtures stop needing mwcc and tubuild's bare
-# `return` guards become skips. See group 1 in the header.
+# `return` guards become skips. See group 1 in the header. test_tubuild_owned_relocs
+# above is NOT a step toward wiring test_tubuild: it is a separate module precisely so
+# that adding it did not require relaxing that refusal.
--- a/.github/workflows/tool-tests.yml
+++ b/.github/workflows/tool-tests.yml
@@ -10,7 +10,7 @@
 # gate's FAILURE branch is the one path a green run never exercises: no ROM build can
 # reach it, because these tools are not compiled into the ROM. src-tu-refs.yml exists
 # because #1643 broke 26 translation units while every byte gate stayed green, and the
-# only thing that noticed was a test suite a human had to run. This job runs 498 of
+# only thing that noticed was a test suite a human had to run. This job runs 504 of
 # those assertions on every PR that touches `tools/` or `nearmiss/`, so the next one is
 # noticed by CI.
 #
@@ -86,7 +86,7 @@
 # none of the three inputs present); that module adds 87, timed at 9.7s on Windows.
 # The three modules added below cost no measurable wall time at all: on one Windows tree,
 # 435 tests in 46.5s before and 498 in 45.7s after -- inside the noise. On the runner
-# itself the whole 498 take 11.5s, plus about 1s for the two wheels (both are prebuilt;
+# itself those 498 took 11.5s, plus about 1s for the two wheels (both are prebuilt;
 # nothing compiles). They are mock- and tempdir-based and touch neither the network nor
 # the tree.
 # The whole-suite timings are dominated by Windows filesystem overhead and do not
@@ -100,8 +100,9 @@
 # present" -- its Retarget class, and only that class, compiles a real object). All nine
 # are @skipUnless with a stated reason, so `-v` shows them; none is a bare `return`.
 # test_premerge_check adds none on a runner with git installed, and test_symrecords,
-# test_nearmiss_db, test_tu_promote and test_tu_production add none anywhere.
-# 489 of the 498 assert something.
+# test_nearmiss_db, test_tu_promote, test_tu_production and
+# test_tubuild_owned_relocs add none anywhere.
+# 495 of the 504 assert something.
 name: tool tests
 
 on:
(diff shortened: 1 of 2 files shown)


andrewboudreau:
## Independent audit done, and it found four doc defects in this PR — now fixed in `19750fdec`

I had this PR audited by a reviewer who did not write it, against a scratch tree extracted with `git archive` (no worktree, nothing written to the checkout). Every substantive claim reproduces. The audit then found something I had missed, which I have pushed a fix for rather than merged past.

### The substance reproduces, all of it

| Claim | Measured | |
|---|---|---|
| 6 tests, all run, 0 skip, on a toolchain-free tree | `Ran 6 tests ... OK`, exit 0, with no `extracted/`, `tools/mwccarm/`, `tools/bin/`, `build/` present | confirmed |
| Not tautologies — drives the real `verify_owned_sections` | one `mock.patch.object` in the file, scoped to `section_contribution`; the whole verdict ladder at `tubuild.py:2798-2817` is real | confirmed |
| `else None` → `else 0` caught, 2 fail | 2 failures, exactly the 2 named tests | confirmed |
| `else None` → `else expected_addend` caught, 2 fail | 2 failures, **same two tests, different assertion lines** | confirmed |
| drop `expected["target_address"] != cfg[1]` → 1 fail | 1 | confirmed |
| `>=` → `>` → 1 fail | 1 | confirmed |
| disable the normalized bias → 6 fail | 6 | confirmed |
| drop `cand_addr != expected["target_address"]` → 1 fail | 1, `'OK' != 'WRONG-DEST'` | confirmed |
| `tubuild.py` untouched; no `config/` writes | sha256 identical before and after; full `diff -r` against a pristine archive empty but for `__pycache__` | confirmed |
| 498 → 504, +0 skips | `Ran 498` (26 modules) → `Ran 504` (27); skip count identical both runs | confirmed |

All six mutation counts match the table's right-hand column exactly. The auditor also tried to break it and could not: no hidden skip, no early return, no swallowed exception, no `setUpClass` guard, no env check — `grep -nE "skip|return|try:|except|_toolchain|os.environ|sys.exit"` turns up only the fixture builders' returns. Six independent mutations of the function each flip at least one assertion, which is the actual proof the assertions read computed output rather than test-constructed values.

**Explicitly not verified**, and I am flagging it rather than letting it pass as confirmed: the mutation table's *left-hand* column — "46/46 of `test_tubuild.py` stays green" under the three fail-open mutations — needs the pinned compiler and `extracted/`, which the audit tree deliberately lacked. It was checked structurally instead and is consistent (no fixture in `test_tubuild.py` produces a manifest addend in 1..7, so the `else` arm is unreachable from that file; no test there perturbs `cfg[1]` independently of the manifest). But it was not run. That column is reasoning, not measurement, and should be read as such.

### What the audit caught that I did not

Raising the per-module table to 504 left **four prose aggregates in this same file still saying 498**, and silently added a **fourth** non-stdlib module to a comment that names three.

That is worse than an ordinary doc nit here, because this file's own stated mechanism is:

> *"a single 'Ran N tests' line in the log is the whole claim and a module that stops contributing tests is visible as a drop in N"* — and those figures *"are the number to compare against if this job's total ever moves."*

The number to compare against was wrong by six, in the file that instructs you to compare against it. A PR whose entire purpose is to close a coverage hole should not open a bookkeeping one.

`19750fdec` fixes all of it:

- headline **498 → 504**; **"489 of the 498" → "495 of the 504"** (the six new tests all assert, so the nine non-asserting ones are unchanged).
- the expected-skips census now names `test_tubuild_owned_relocs` among the modules adding no skips anywhere.
- the pinned-dependency comment said *"every module but the last three… `test_rombuild`, `test_tu_promote` and `test_tu_production` import `rombuild.py`"*. There are four now: `test_tubuild_owned_relocs` imports `tubuild.py`, which reaches `elftools` and `capstone` **by its own path**, not through `rombuild.py`. Measured by blocking both via `meta_path` — `import tubuild` then fails collection exactly like the other three. Nothing is broken today because the pip step is unconditional; the comment is simply what a future reviewer would trust when deciding whether that step can be dropped or reordered. I also dropped the "last three" positional phrasing, which was already wrong before this PR: `test_rombuild` sits at position 17 and `test_validate_merge` is last.
- **"Mutation-checked, all five" → six**, naming the candidate-side one. It was in the PR body's table but missing from the comment — an undercount of real coverage, not a phantom row.

Two `498`s deliberately remain. They are a historical measurement of what three earlier modules cost (435 → 498, inside the noise), not a claim about today's total; I reworded *"the whole 498 take 11.5s"* to *"those 498 took"* so it reads as the past-tense measurement it is.

### Verification on the fix commit

```
YAML parses, jobs: ['tools']
documented modules: 27      table sum: 504
enumerated in run block: 27
table minus list: []        list minus table: []
python -m unittest tools.test_tubuild_owned_relocs  ->  Ran 6 tests  OK
check_dead_references: no new dead references, no broken markdown links
pre-push: port-refcheck 405 resolve / duplicate-sources 11158 stems none doubled
          check_src_tu_compiles 92/92
committed blob: 0 CR bytes (clean LF)
```

Diff is one file, +16/−12, comments only. `tools/test_tubuild_owned_relocs.py` and `tools/tubuild.py` are untouched by it.

### Also worth recording, since it is a standing gap

The PR body reported `OK (skipped=3)`; this file documents nine expected skips and the audit measured eleven. The difference is environmental — the PR's tree had `extracted/` present so the four `@skipUnless` ROM tests ran, and the audit tree added two host-specific skips (`tools/` not inside a repo with a HEAD; symlinks not permitted). The load-bearing half of that claim, **+0 skips from the new module**, holds in every environment measured. Only the absolute figure is environment-specific, and it is the one number in the PR body I would not carry forward.

Re-running the gates on the new head now. Merging once they are green — this needs a second pair of eyes on the *fix* commit, not just the original, so if anyone wants to look, `19750fdec` is comments-only and quick to read.

PR #2064, opened 2026-08-31
TU: promote ov006/dScMgSingle3DBase_c — nine functions, vtable-proven boundary

andrewboudreau:
Promotes `ov006/dScMgSingle3DBase_c` — the base class of thirteen minigame scenes — from nine hand-placed one-function files into one C++ translation unit that emits its own destructor pair, vtable and RTTI records the way the cartridge's own build did.

## What lands

Nine functions, `.text 0x0210a4b0..0x0210a8c0` (1040 bytes), one `complete` delink entry replacing nine:

| | symbol | size |
|---|---|---|
| D1 | `_ZN19dScMgSingle3DBase_cD1Ev` | 0x38 |
| D0 | `_ZN19dScMgSingle3DBase_cD0Ev` | 0x4c |
| | `func_ov006_0210a534` | 0xcc |
| slot 26 | `func_ov006_0210a600` | 0x08 |
| slot 5 | `_ZN19dScMgSingle3DBase_c21AfterCleanupResourcesEj` | 0x5c |
| slot 10 | `_ZN19dScMgSingle3DBase_c12BeforeRenderEv` | 0x34 |
| slot 7 | `_ZN19dScMgSingle3DBase_c14BeforeBehaviorEv` | 0x4c |
| slot 2 | `_ZN19dScMgSingle3DBase_c18AfterInitResourcesEj` | 0x24 |
| slot 33 | `func_ov006_0210a708` | 0x1b8 |

The destructor pair is no longer hand-placed under `// @symbol` markers. The class's key function pulls D1/D0 out of line through vtable slots 16/17, and mwcc emits both — which is the point of the whole workstream.

## The boundary is corrected, and the correction is proved

`tools/tu_map.py` reported this unit as **eight** functions, `0x0210a4b0..0x0210a708`, and split `func_ov006_0210a708` into a single-function unit of its own. That is a tu_map defect, not a real boundary: the function's symbol is still spelled `func_<module>_<address>`, so it gets no class label, so tu_map cuts there.

The vtable says otherwise. Scanning every `_ZTV` in ov006:

- `0x0210a708` is referenced at **slot 33** by exactly fourteen vtables — `dScMgSingle3DBase_c` and all thirteen of its children.
- `0x0210a600` — slot 26, which tu_map *did* place inside the unit — is referenced by **the same fourteen**.

Identical reference sets. One function tu_map kept and one it dropped are members of the same class. The unit is nine functions, and the ninth byte-matches inside the TU on the first compile.

## Two accuracy defects fixed on the way

**`func_ov006_0210a600` was labelled one class too deep.** Its legacy source carried `recovered name: dScMgFlower_c_OnHitByCannonBlastedChar`. `dScMgFlower_c` is one of the fourteen classes that reference it — a body every sibling shares belongs to the common ancestor. Same off-by-one-class shape already documented for `AfterInitResources`.

**`include/dScMgSingle3DBase_c.h` called slots 26 and 33 "new at this class".** They are re-overrides: `_ZTV11dScMgBase_c` already carries bodies at both (ov004:0x020b04e0 and ov004:0x020b265c, read out of the ROM). They stay undeclared only because `dScMgBase_c` leaves slots 18-35 undeclared, and a derived class cannot override a slot the base has not spelled.

## Compiler-generated output: eleven licensed symbols

The TU emits three of its own records — `_ZTI`/`_ZTS`/`_ZTV19dScMgSingle3DBase_c` — plus eight ancestor RTTI passengers. All eleven are `deadstrip-data`, each naming the cartridge home it defers to, so `romdata_check` becomes binding for exactly those symbols and compares them relocations-applied before isolation discards them. Every canonical address is derived from `tubuild.all_symbol_homes()`.

`_ZTV19dScMgSingle3DBase_c` scores **PARTIAL**, and that is expected rather than a disagreement: `dScMgBase_c` leaves slots 18-35 undeclared, so mwcc emits an 18-slot table where the cartridge has 36 — a verified byte-exact **prefix**. Owning it outright is the follow-up, mapped in `notes/dScMgBase_c-slots-18-35.md` and `notes/dScMgSingle3DBase_c-vtable-ownership.md`.

## Evidence

- `tubuild verify` — **9/9 MATCH**, objisolate clean, relocation destinations clean, emission order ROM-ascending.
- `tubuild linkcheck` — **SCRATCH-LINK-VERIFIED**; ov006 built 525344 == retail 525344; TU range IDENTICAL; ROM sha256 `d1506e90…c478e8`, identical to the stock build. `dsd check symbols --fail` reports 9 errors, all of which the `--baseline` control reproduces exactly: **0 new, 0 resolved**. That check is not green and is not reported as green.
- `rombuild` after promotion — module fidelity, source fidelity and the compiler-only policy, against the real tree.

## Notes for review

- The delink change is a contiguous 9→1 merge with the same outer endpoints; no entry lost its `complete` marker.
- `config/converted-baseline.json` moves six rows to `promoted-path#symbol` identities, so the readability ratchet keeps function granularity — six out, six in, count invariant. Edited by `tu_promote.py`, not by `tiers_ratchet.py --update`, which would sweep unrelated baseline staleness into the diff.
- Two legacy sources disagreed on the same two signatures — one typed the loaded file buffer `int`, the other `void *`, and both passed the same value to the same `DecompressLZ16`. One TU cannot hold both; the pointer form is the true one and removes four casts. Byte-neutral, and `verify` re-confirms all nine.
- `notes/actor-vtables.md` is repointed at the promoted path so `check_dead_references` stays green.
- Rebased onto `origin/main` @ `9bfcd5d8` (after #2057 and #2061 landed). The only conflict was `attribution.json`, where both sides append to the same trailing line; resolved as the union of both blocks — main's five `daObjKm3_Kurumajiku_c` rows and this branch's nine — hand-edited at the markers rather than resolved whole-file. `config/converted-baseline.json` auto-merged.
- `tools/premerge_check.py HEAD` against the rebased base: all eight static gates `pass → pass`, nothing goes green → red. `source-coverage` notes `entries: 10933 -> 10925 (-8)`, which is the nine-into-one consolidation with bytes flat.


andrewboudreau:
## Changes requested — five one-line comment fixes. The boundary itself I reproduced from the ROM, and it is right.

This is the best-evidenced TU promotion that has come through this queue. I had it reviewed by someone who did not write it, with instructions not to take the PR body on trust, and the central claim was re-derived from the cartridge image rather than read off your JSON. Everything I am asking for is a comment. Not one byte moves.

### The boundary is real, and here is the independent derivation

`extracted/dsd/arm9_overlays/ov006.bin` read directly, all 47 `_ZTV*` symbols in ov006 resolved from `config/arm9/overlays/ov006/symbols.txt`, every slot decoded. The address point convention was established empirically first rather than assumed — offset-to-top and typeinfo sit at −8/−4 from the `_ZTV` symbol address:

```
_ZTV19dScMgSingle3DBase_c @ 0x0213e448
  -8: 0x0        -4: 0x0213bc64   (= _ZTI19dScMgSingle3DBase_c)
  slot  2 = 0x0210a6e4   slot  5 = 0x0210a608   slot  7 = 0x0210a698
  slot 10 = 0x0210a664   slot 16 = 0x0210a4b0   slot 17 = 0x0210a4e8
  slot 26 = 0x0210a600   slot 33 = 0x0210a708
```

Every slot number your header and body assert is confirmed. Then the reference scan:

```
total ov006 _ZTV: 47
reference 0x0210a600: 14  (all at slot 26)
reference 0x0210a708: 14  (all at slot 33)
SETS IDENTICAL: True     only in 600: set()     only in 708: set()
```

**The "fourteen vtables, identical reference sets" claim is exactly true** — the class plus its thirteen children, matching the thirteen headers that `#include "dScMgSingle3DBase_c.h"`.

Both neighbours were checked, which is the question that actually decides a boundary:

```
PREV (MgMushroomRoulette range) 0x0210a400  refs=0
NEXT (dScMgSlot1_c D1)          0x0210a8c0  refs=1  slot 16  <- a DIFFERENT class
```

`dScMgSlot1_c` is not among the fourteen, so the upper edge is hard. A whole-image word scan across `arm9.bin`, `arm9_dec.bin` and every overlay found references to these addresses in `ov006.bin` and **nowhere else** — no cross-overlay collision risk. Manifest read through `tools/tu_manifest.py export`: nine functions, sizes summing to exactly `0x410`, no gaps, no overlap, endpoints identical to the nine delink entries they replace.

Key function confirmed too. The inline `~dScMgSingle3DBase_c() {}` at header line 95 is skipped, so it is `AfterInitResources`, this TU defines it, and `grep -rln "dScMgSingle3DBase_c::" src/` returns only your file — exactly one possible emitter. Deleting the two `_Emit*Destructor` forcing scaffolds rather than porting them is the right call and the reasoning survives in the footer.

All eleven licensed symbols were checked against the tree's own symbol tables by address — **all eleven correct, all outside the licensed `.text` range.** That is the part that usually goes unchecked until enrollment.

### What I need changed

**1. `include/dScMgSingle3DBase_c.h:105` names a path that has never existed.** The new comment block points at `src_tu/actors/dScMgSingle3DBase_c.cpp`. The file is `src/actors/dScMgSingle3DBase_c.cpp`.

```
test -f src_tu/actors/dScMgSingle3DBase_c.cpp                          -> ABSENT
git ls-tree -r --name-only origin/main -- src_tu/actors | grep ScMg    -> (none)
```

Not on this branch, not on main, not ever. And `check_dead_references.py` will not catch it: its `_prose_targets()` scans `tools/**/*.py`, `notes/**`, top-level and `docs/**` markdown, `port/docs/*.md`, workflow YAML and skill markdown. `include/*.h` is not a prose surface. One word: `src_tu/` → `src/`.

**2. `src/_ZN10dBgActor_cD0Ev.cpp:21` names a file this PR deletes.**

```
 * Same shape, same fix, same
 * sentence in src/_ZN19dScMgSingle3DBase_cD0Ev.cpp, which has had a key
 * function since #1544.
```

You already fixed the identical sentence in `notes/actor-vtables.md`, repointed at the promoted path — correctly. That one is inside a gate-scanned prose surface; this one is inside `src/`, which is not, so it survived. Same fix. This is reviewer checklist item 5 of `notes/tu-promotion-conventions.md` in substance: a promotion should not leave a comment naming a file it deleted.

**3. The generated banner contradicts the file's own body.** The header says `ov006/dScMgSingle3DBase_c (8 function(s))` and lists eight legacy sources; `src/func_ov006_0210a708.c` is missing from the assembled-from list. Twenty lines down the body says `/* ROM ordinal 8 -- func_ov006_0210a708 ... */`. This is the artifact a future reader uses to learn what the unit contains, and it is the one thing the boundary correction did not update. (The manifest's `boundary_evidence[0]` also says "8", but that is a faithful quote of what tu_map reported followed immediately by the correction — that one is honest and should stay.)

**4. Add `// @symbol func_ov006_0210a534` above line 281.** You have six markers and this is the one hand-written member without one. It costs you nothing directly — a534 was not CONVERTED before, so nothing backslides — but it costs its *neighbour*: `_marked_member_fragment` slices marker-to-**next**-marker, so with no marker after it the `func_ov006_0210a600` fragment currently runs from line 257 to end of file and swallows a534's `volatile` body and raw addresses. `func_ov006_0210a600` is an eight-byte `return 1;` that cannot score readable while that is true. One line unblocks it.

Your D1/D0 being unmarked is correct and I am not asking you to change it — `score_member` falls through to `_lifecycle_member_fragment`, which finds the inline destructor in the directly-included header. I have a PR open (#2065) documenting all of this, and your file is cited in it as the counter-example that gets the ledger right.

**5. `include/dScMgSingle3DBase_c.h:90` still says the inline destructor is the key function.** `/* Declared first (key function); overrides slots 16 (D1) and 17 (D0). */` contradicts your own correct reasoning at lines 310-313 that an inline dtor cannot be one. Pre-existing, but you rewrote the block immediately below it.

### Two disclosures I would like, not blockers

**`func_ov006_0210a534` has zero vtable corroboration.** It is referenced by none of the 47 tables. Its membership rests on tu_map's contiguous linker run plus address containment between two vtable-proven members. That is very probably right — a free function sandwiched inside a class's run is a normal thing — but it is a weaker evidence class than the other eight, and the title "vtable-proven boundary" reads as covering all nine. One sentence in the manifest reason fixes it.

**The retyping added a `const` nobody mentioned.** `DecompressLZ16(const void*, void*)` — neither legacy declaration had it. The `int`→`void*` change is described and is plainly a genuine accuracy fix forced by two `extern "C"` declarations of the same symbol with contradictory signatures; the `const` is an unremarked extra. It should be byte-neutral on a callee signature, but say so rather than leave a reader to find it.

Optional and genuinely minor: lines 150/154 carry `char *dst = (char *)0x6400000; dst += 0x4000;` verbatim from the legacy file. That split is almost certainly steering the literal pool rather than expressing intent, and a reader cannot tell whether it may be folded to `(char *)0x6404000`. The file goes out of its way to justify `volatile` and the mangled `_ZN2GX...` block — this deserves the same one-line treatment.

### Gates, and what nobody verified

Re-gated against current main, since main moved out from under your base:

```
base   origin/main @ a00c09854768        (was 9bfcd5d86d93 when you opened)
target #2064  head 0fda8b4d1b5b          merge tree 5816ea677e9e
                                          ALL pass -> pass
  info: entries 10933 -> 10925 (-8)  -- consolidation, bytes flat
```

The only new commit in the drift is #2063, test-only, no overlap; `merge-tree --write-tree` is clean.

Being explicit about the limits of the above, because your PR body is precise and this review should be too: **no build was run.** `tubuild verify 9/9 MATCH`, `differingBytes: 0`, `objectAudit LICENSED: 9`, `emittedTextOrderIsRomAscending`, the ROM sha256, the `moduleSetSha256`, and the `romdata_check` **PARTIAL** score are author-recorded values verified only for internal consistency and arithmetic. The source-order claim was checked structurally — definitions run `0x708 → 0x6e4 → 0x698 → 0x664 → 0x608 → 0x600 → 0x534`, strictly descending, which is the required reverse-of-ROM order — but that is not the same as a compile. The validator's green is what covers those, not me.

Also worth recording: that `romdata_check` PARTIAL score exists only as prose inside the manifest `reason` string. There is no machine-readable field holding it, so nothing will notice if it regresses. That is a gap in the tooling, not in your PR, and I am not asking you to fix it here.

Nothing above touches the decomp work. Fix the five comments and I will merge it.

andrewboudreau:
## One more, found while reviewing the stacked #2067: a missing `@symbol` marker

Adding to the five comment fixes above — this one is a scoring defect rather than a prose one, and it is a single line.

`src/actors/dScMgSingle3DBase_c.cpp` declares nine functions. Two are the destructor pair, covered by the lifecycle fallback. Of the remaining seven, six carry a marker and one does not:

| address | symbol | marker |
|---|---|---|
| `0x0210a4b0` | `_ZN19dScMgSingle3DBase_cD1Ev` | lifecycle fallback — fine |
| `0x0210a4e8` | `_ZN19dScMgSingle3DBase_cD0Ev` | lifecycle fallback — fine |
| `0x0210a534` | `func_ov006_0210a534` | **missing** |
| `0x0210a600` | `func_ov006_0210a600` | ok |
| `0x0210a608` | `_ZN19dScMgSingle3DBase_c21AfterCleanupResourcesEj` | ok |
| `0x0210a664` | `_ZN19dScMgSingle3DBase_c12BeforeRenderEv` | ok |
| `0x0210a698` | `_ZN19dScMgSingle3DBase_c14BeforeBehaviorEv` | ok |
| `0x0210a6e4` | `_ZN19dScMgSingle3DBase_c18AfterInitResourcesEj` | ok |
| `0x0210a708` | `func_ov006_0210a708` | ok |

It matters more than one missing marker usually would, because of where it sits. `_marked_member_fragment` slices from one marker to the *next* one, and your last marker is at line 257 in a 323-line file. `func_ov006_0210a534` is defined after it, so there is no next marker — the fragment for `func_ov006_0210a600` runs **257 → EOF** and swallows `func_ov006_0210a534` whole. Two members are affected by the one omission: the unmarked one falls to the whole-file fallback, and the marked one gets scored against text that is not its own.

`notes/tu-promotion-conventions.md` §6 (landed as #2065) is the rule; it calls this "the cheapest rule in this note to satisfy and the most expensive one to skip", which is exactly the shape of it here.

I have asked for the same fix at larger scale on #2067, which needs eleven. Both are on your stack, so one push covers both.

andrewboudreau:
All five review fixes pushed as ea9f05566, plus the two disclosures you asked for. Comment and manifest text only — no code byte moves; `rombuild --no-rom` in a wired worktree still reports **106/106 exact, 100.000000%**, and `check_header_offsets include/dScMgSingle3DBase_c.h` gives `1 commented fields, 0 mismatched, 0 unparsed` (the new comment sits above the destructor, so the Allman-brace trap from #2068 is not armed here).

| # | fix |
|---|---|
| 1 | `include/dScMgSingle3DBase_c.h`: `src_tu/actors/…` → `src/actors/dScMgSingle3DBase_c.cpp` |
| 2 | same header: the destructor is no longer called the key function. It now says an inline dtor is emitted in every TU that needs it and anchors nothing, and that the key function is `AfterInitResources`, the first DECLARED non-inline virtual — which is what this TU's own manifest rows already said. The MUST STAY INLINE measurement is untouched. |
| 3 | `src/_ZN10dBgActor_cD0Ev.cpp:21`: repointed at the promoted path, matching the already-correct `notes/actor-vtables.md` |
| 4 | banner `(8 function(s))` → `(9 function(s))`, and `src/func_ov006_0210a708.c` added to the assembled-from list at `[8]`. `boundary_evidence[0]`'s "8" stays — it is a faithful quote of tu_map, corrected on the spot by `boundary_evidence[3]`. |
| 5 | `// @symbol func_ov006_0210a534` added above the definition, so `_marked_member_fragment` stops slicing a600's fragment to end of file. The unmarked D1/D0 are correct and unchanged. |

Disclosures:

- **a534's evidence class.** New `boundary_evidence` entry stating plainly that `func_ov006_0210a534` is the one member with no vtable corroboration — its membership rests on tu_map's contiguous run plus address containment, and it would be the first member to re-examine if the boundary is ever contested.
- **The unremarked `const`.** The signature reconciliation added `DecompressLZ16(const void*, void*)`, which neither legacy file had. Nothing in the ROM distinguishes const from non-const there, so the file now says it is a judgement call and says what would falsify it.

I left lines 150/154 (`char *dst = (char *)0x6400000; dst += 0x4000;`) alone. I have no measurement showing the split is literal-pool steering rather than transcription, and a comment asserting either way would be a guess — happy to measure it as a follow-up if you want the note.

PR #2067, opened 2026-08-31
Promote ov006/dScMgBSC_c to a single 19-function C++ TU

andrewboudreau:
Stacked on #2064 — review that one first; this branch contains its commit.

`src_tu/actors/dScMgBSC_c.cpp` → `src/actors/dScMgBSC_c.cpp`, replacing **19** per-function legacy sources (14 flat-C, 5 already C++). The whole `.text` range `0x02124908..0x021255f8` now reproduces from one object.

## The destructor is inline, and that is the load-bearing part

An out-of-line `~dScMgBSC_c()` makes mwcc emit the synthesized `D0` **ahead of** the written `D1`. The cartridge has `D1` first, and linkcheck's `[4b/8]` pre-link licensing audit refuses a TU whose licensed `.text` is not in ROM address order — it refused this one:

```
emitted: ['_ZN10dScMgBSC_cD0Ev', '_ZN10dScMgBSC_cD1Ev', 'func_ov006_02124a04', ...]
REFUSED -- licensed .text functions are not emitted in ROM address order
```

Defining the destructor inline in the class body hands the key-function role to `InitResources` — the next virtual declared, non-inline, and defined in this TU. That emits `_ZTV10dScMgBSC_c` here; slots 16 and 17 name `D1` then `D0`, odr-using both, so the compiler emits the pair for us, in cartridge order. It also deletes the homeless `D2` outright, which is why `compiler_only_output` drops 14 → 13 rows. (A stale row there rejects the *whole* policy block — all 13 `deadstrip-data` rows come back as `EXTRA` until the dead `D2` row is dropped.)

## Verification

| check | result |
|---|---|
| `tubuild verify ov006/dScMgBSC_c` | 19/19 MATCH, objisolate clean, exact deadstrip, ROM-ascending order |
| `tubuild linkcheck --module ov006` | **SCRATCH-LINK-VERIFIED**; 106/106 exact; ROM sha256 `d1506e90…` **identical to stock** |
| `rombuild -j16` | 11,088 source-built, reproducing 11,088, **mismatching 0**; storage aliases exact |
| ROM data from source | **462 verified** (was 461), 250 partial, 9 differ — unchanged |
| `premerge_check HEAD` | all 8 gates pass/pass on the merge tree |
| `port_refcheck` / attribution | 405 resolve, 0 stale · 24 consolidated, **0 lost** |

The +1 verified record and +1 partial are the RTTI and the emitted `_ZTV10dScMgBSC_c` — 18 of the cartridge's 36 slots, now **byte-compared against the cartridge rather than trusted**.

## Two merge-tree-only reds this had to fix

Both were green on the branch and red only on the merge result, so only `premerge_check.py` surfaces them:

1. **`header-offsets`** — the inline body was written with an Allman brace. `check_header_offsets.py`'s `skip_body` only engages when the signature line itself contains `{`, so the bare `{` fell through to DECL parsing (`UNPARSED include/dScMgBSC_c.h:53: {`). Worse than a red: one unparsed line suppresses the field walk for the whole header, so it reported `0 commented fields` — hollow green, not a pass. Brace moved to the signature line, matching every existing inline dtor in `include/`; now `4 commented fields, 0 mismatched, 0 unparsed` and the span corrects `0x4f38` → `0x51d0`.
2. **`dead-references`** — deleting 19 legacy sources stranded a prose reference in `notes/minigame-provenance.md`. Repointed at the consolidated path.

## Deliberately not in this PR

Naming `dScMgBase_c` slots 18–35 would widen the emitted vtable past slot 17 and every descendant's with it. This TU independently confirms three of those slots from its own table (18 → `Virtual48`, 19 → `OnTurnIntoEgg`, 21 → `OnGroundPounded`); that evidence is recorded in `notes/dScMgBase_c-slots-18-35.md` for the atomic change that does the rename.

Converted-tier accounting is flat: 3 identities migrated, 3 score 5/5 — `Render` via its `@symbol` marker, `D1`/`D0` via the inline lifecycle fragment.


andrewboudreau:
## Changes requested — eleven `@symbol` markers. Everything else here I re-derived from the cartridge and it holds.

Reviewed at head `7c0f9a92c79419bfd8a844fc4d90485718957ffd`, base `cpp/minigame-tu` @ `0fda8b4d1b5b`. This is a good promotion and the evidence is better than most that come through here. One thing needs fixing before it lands, and it is mechanical.

---

## The blocker: 11 of 17 hand-written members have no `// @symbol`

`notes/tu-promotion-conventions.md` §6 landed three PRs ago (#2065) and states the rule as *"every hand-written member of a promoted TU carries `// @symbol <linker-name>` on its own line, immediately above its definition."* The manifest declares 19 functions. Two are the destructor pair, which the lifecycle fallback covers because the destructor is inline in the header. That leaves 17 that need a marker, and six have one:

| address | symbol | marker |
|---|---|---|
| `0x02124908` | `_ZN10dScMgBSC_cD1Ev` | lifecycle fallback — fine |
| `0x0212497c` | `_ZN10dScMgBSC_cD0Ev` | lifecycle fallback — fine |
| `0x02124a04` | `func_ov006_02124a04` | **missing** |
| `0x02124a08` | `func_ov006_02124a08` | **missing** |
| `0x02124ae4` | `func_ov006_02124ae4` | **missing** |
| `0x02124b58` | `func_ov006_02124b58` | **missing** |
| `0x02124bb4` | `func_ov006_02124bb4` | **missing** |
| `0x02124cb4` | `func_ov006_02124cb4` | **missing** |
| `0x02124dc0` | `func_ov006_02124dc0` | **missing** |
| `0x02124e1c` | `func_ov006_02124e1c` | **missing** |
| `0x02124ec4` | `func_ov006_02124ec4` | **missing** |
| `0x02124fd8` | `func_ov006_02124fd8` | **missing** |
| `0x021250e4` | `func_ov006_021250e4` | **missing** |
| `0x02125248` | `func_ov006_02125248` | ok |
| `0x0212527c` | `func_ov006_0212527c` | ok |
| `0x02125364` | `func_ov006_02125364` | ok |
| `0x021253bc` | `_ZN10dScMgBSC_c6RenderEv` | ok |
| `0x021254c0` | `_ZN10dScMgBSC_c8BehaviorEv` | ok |
| `0x0212551c` | `_ZN10dScMgBSC_c13InitResourcesEv` | ok |

**The consequence is worse than eleven members losing their boundary.** `_marked_member_fragment` slices from the end of one marker to the start of the *next* marker. Your markers sit at lines 170, 209, 232, 278, 301 and **341**, and the file is 695 lines. There is no marker after 341, so the fragment for `func_ov006_02125248` runs **341 → EOF: 354 lines covering eleven other functions**. That one member is currently scored against three quarters of the TU.

And the TU is not clean text — all four poisonable criteria are present in it:

| criterion | present at |
|---|---|
| `no_unk_field` | `unk_51b8` at lines 223, 312, 316, 319, 325 |
| `no_codegen_trick` | `*(volatile unsigned char *)&data_020a0e40` at line 463 |
| `no_raw_offset` | `*(int *)(c + 0xa8)` line 200, `*(int*)(base + i * 8 + 0x51a8)` line 374, and others |
| `no_mangled_refs` | `_ZN5Sound12PlayBank2_2DEj` line 476, `_Z14ApproachLinearRiii` lines 486-487, `_ZN3GXS11LoadOBJPlttEPKvjj` line 192 |

So this is not a hypothetical loss. `config/converted-baseline.json` banks three identities out of nineteen, and the fragment geometry is part of why.

**The fix is eleven lines.** Every function in the file already carries a `/* ROM ordinal N -- <symbol>, <addr>, size 0x… */` comment with the exact symbol in it — the marker goes on its own line next to that comment, spelled as the manifest's `functions[].symbol` spells it. You already did this for six; the same treatment for ordinals 2 through 12.

I am not asking you to clean up the `unk_`/raw-offset/mangled text. That is a separate readability pass and the byte match is what matters first. I am asking that when someone does it, the credit lands on the member they fixed instead of being blocked by a neighbour.

---

## What I verified independently, and what it showed

### The linked range is exactly contiguous and abuts both neighbours

Nineteen delink entries collapse into one, and the endpoints are unchanged:

```
MgTrampolineTerror.cpp        .text 0x0212471c..0x02124908
src/actors/dScMgBSC_c.cpp     .text 0x02124908..0x021255f8   <- new
```

Chaining the nineteen removed ranges by hand: `908 -> 97c -> a04 -> a08 -> ae4 -> b58 -> bb4 -> cb4 -> dc0 -> e1c -> ec4 -> fd8 -> 0e4 -> 248 -> 27c -> 364 -> 3bc -> 4c0 -> 51c -> 5f8`. No gap, no overlap, first start and last end preserved.

### Every licensing address in `compiler_only_output` reads back correctly from the ROM

This is the check that `SCRATCH-LINK-VERIFIED` does not perform — the link round proves the bytes, not that the `canonical_module`/`canonical_address` rows point at the records they claim. I resolved all of them against `extracted/arm9_dec.bin` and the overlay images, preferring each row's own module:

```
_ZTI7fBase_c              arm9  0x02086d70 -> name '7fBase_c'                OK
_ZTS7fBase_c              arm9  0x02086e60 -> '7fBase_c'                     OK
_ZTS7dBase_c              arm9  0x02086e54 -> '7dBase_c'                     OK
_ZTS8dScene_c             arm9  0x020914b0 -> '8dScene_c'                    OK
_ZTI7dBase_c              arm9  0x02086e78 -> name '7dBase_c'                OK
_ZTI11dScMgBase_c         ov004 0x020bbf6c -> name '11dScMgBase_c'           OK
_ZTS11dScMgBase_c         ov004 0x020bbf84 -> '11dScMgBase_c'                OK
_ZTI8dScene_c             arm9  0x020914d4 -> name '8dScene_c'               OK
_ZTI10dScMgBSC_c          ov006 0x0213fd8c -> name '10dScMgBSC_c'            OK
_ZTS10dScMgBSC_c          ov006 0x0213fd98 -> '10dScMgBSC_c'                 OK
_ZTI19dScMgSingle3DBase_c ov006 0x0213bc64 -> name '19dScMgSingle3DBase_c'   OK
_ZTS19dScMgSingle3DBase_c ov006 0x0213bd00 -> '19dScMgSingle3DBase_c'        OK
```

Thirteen for thirteen, counting the vtable below.

### The class's own RTTI chain, walked from the vtable

```
_ZTV10dScMgBSC_c @ 0x0213fec8   (ov006, base 0x020bfec0)
  [V-4] typeinfo      = 0x0213fd8c   manifest says 0x0213fd8c   MATCH
  TI[1] name ptr      = 0x0213fd98 -> '10dScMgBSC_c'
  TI[2] base's _ZTI   = 0x0213bc64   = _ZTI19dScMgSingle3DBase_c  MATCH
```

`10` is exactly `len("dScMgBSC_c")`, so the read verifies itself, and `TI[2]` independently confirms the base clause `dScMgBSC_c : dScMgSingle3DBase_c` from the cartridge rather than from our headers.

**This class's name agrees with the ROM, which is why the promotion can land at all.** #2066 is the counter-example: `PoleLift` is `daObjKm2_Ami_Bou_c` in the cartridge, so consolidating its RTTI emitted `_ZTS8PoleLift`/`_ZTI8PoleLift` with no ROM symbol to license them and `tubuild` refused with `HOMELESS`. Measured tree-wide, 259 of 541 `_ZTV`-bearing classes are in that position — see #2069. `dScMgBSC_c` and `dScMgSingle3DBase_c` are both in the agreeing 276.

### The converted ratchet is a clean re-key, not a backslide

Three entries move and the set size is unchanged at **2573**:

```
- src/_ZN10dScMgBSC_c6RenderEv.cpp   -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_c6RenderEv
- src/_ZN10dScMgBSC_cD0Ev.cpp        -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_cD0Ev
- src/_ZN10dScMgBSC_cD1Ev.cpp        -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_cD1Ev
```

The other five `.cpp` files this deletes — `InitResources`, `Behavior`, `func_ov006_02124e1c`, `_0212527c`, `_02125364` — I checked against the base baseline and none of them were members, so nothing is silently dropped. This is the shape a promotion should have.

### The inline destructor, and the brace

The header's reasoning is correct and it is the good kind of evidence — D1 at `0x02124908` sits *before* D0 at `0x0212497c`, which an out-of-line destructor cannot produce, because mwcc emits the synthesized D0 ahead of the written D1. Inline in the class body, the destructor cannot be the key function, so `InitResources` (the next declared non-inline virtual) becomes it and emits the vtable, whose slots 16/17 odr-use D1 then D0 in the cartridge's order.

Worth saying out loud: **`dScMgBSC_c` is the first case in the tree that actually sits on the `check_header_offsets` hazard shape** — derived class, structor declared before any field, multi-statement body. You put the brace on the signature line, which is the one position that keeps the field walk alive. That is exactly what #2068 documented two commits ago, and this PR is the first real instance of it. Please leave it that way.

### The notes are the best part of this PR

`notes/dScMgBase_c-slots-18-35.md` confirms the slot map from a second, independent direction — the base's table was walked from `_ZTV11dScMgBase_c`, the overrides from `_ZTV10dScMgBSC_c`, and the recovered-name comments predate both walks and land on the same two addresses. Slot 18 staying unnamed on both sides is consistent, not a hole. And measuring what the follow-up costs — `_ZTV10dScMgBSC_c` is 0x50 bytes, truncating exactly one slot short of 18, so widening the base widens every descendant's table too — is the kind of thing that stops a follow-up PR from being opened blind.

The `notes/minigame-provenance.md` line was correctly re-pointed at the new path. That would otherwise have gone red on `check_dead_references`.

---

## Two smaller things, neither blocking

**1. The manifest's five `CONFLICT` notes describe something that is no longer in the file.** They each end *"kept the first, the other is commented out for review"* — for `func_ov006_020c0c80`, `func_ov006_020c0d68`, `_ZN5Sound12PlayBank2_2DEj`, `func_ov006_020c1718` and `func_ov006_020c0aa8`. There is no commented-out extern anywhere in `src/actors/dScMgBSC_c.cpp`, so a reader who tries to do the review the note asks for finds nothing to compare against. If you resolved them, say so — replace the five with one line noting that the kept declaration is confirmed by the 19/19 byte match, and which one lost. If you did not, the alternatives should be restored.

**2. `boundary_evidence` says `corroborated=False`.** The sinit line is module-wide (31/31 for all of ov006), not narrowed to this TU, and the manifest is honest about that. The contiguous linker run plus the vtable-slot derivation of the three overrides carry the boundary here, so this is not a gap I am asking you to close — just noting that the strong evidence is the other two rows, not this one.

---

## Verdict

Everything I could check from the cartridge checks out: the range, the licensing addresses, the RTTI chain, the base clause, the ratchet set. 19/19 byte-matched under `SCRATCH-LINK-VERIFIED` with a full ROM build.

**Add the eleven markers and I will merge it.** This is stacked on #2064, which is also awaiting fixes, so both can be pushed in one round — and #2064 needs one marker of its own (`func_ov006_0210a534`, the only unmarked member there, and it is the last function in that file so it is currently absorbed into `func_ov006_0210a600`'s fragment). I have commented there separately.

andrewboudreau:
**Latent `@symbol` gap in `src/actors/dScMgBSC_c.cpp` — found while landing the next TU, not red today.**

`tiers.score_member` slices a member of a promoted TU at `// @symbol` markers, and when there is no *next* marker it runs to end of file (`tools/tiers.py:317-325`). The BSC TU has **6 markers for 17 in-file members**, so the first-marked member is scored over every unmarked flat-C body below it — raw casts, mangled callees and all.

That is not currently a failure here: none of BSC's banked `converted-baseline` entries lands on a member whose slice over-reaches, so `converted-ratchet` is green on this PR's merge tree (re-confirmed just now with `premerge_check 2064 2067` — all 8 gates pass on both).

It *did* fire one PR up the stack. `#2071` (MCarlo, 21 in-file members, 5 marked) went green→red on `converted-ratchet` with:

```
src/actors/dScMgMCarlo_c.cpp#_ZN13dScMgMCarlo_c6RenderEv
    No raw offset arithmetic (*(u32*)(c + 0x74)); Calls things by real names, not mangled _Z
```

Neither is true of `Render`'s own body — the slice had swallowed the flat-C members after it. Fixed there by marking all 21; comments only, ROM sha256 unchanged.

So the same 11 missing markers here are a **trap waiting on the next banked entry**, not a defect in what this PR proves. Adding them is comment-only and byte-neutral. I have not touched this branch, because doing so would force a restack of #2071 for a hazard that is not yet armed — flagging it so the choice is explicit rather than silently deferred.

andrewboudreau:
## Now DIRTY — and the conflict set says restack, do not resolve

#2064 merged as `c97a323b8da7` (squash). GitHub auto-retargeted this PR to `main`, and because the parent landed squashed, `main` now contains #2064's *content* under a commit this branch has never seen. `git merge-tree origin/main 7c0f9a92c794` reports:

```
CONFLICT (content) attribution.json
CONFLICT (add/add) config/tu_manifest.d/ov006/dScMgSingle3DBase_c.json
CONFLICT (content) include/dScMgSingle3DBase_c.h
CONFLICT (add/add) notes/dScMgBase_c-slots-18-35.md
CONFLICT (add/add) src/actors/dScMgSingle3DBase_c.cpp
```

**Four of those five are not yours.** They are #2064's files, colliding with themselves. Resolving them by hand is the wrong move twice over: it is work you already did, and `include/dScMgSingle3DBase_c.h` and the manifest on `main` are the *post-review* versions — they carry the five fixes and the two disclosures from `ea9f05566`. Any hand-resolve that takes this branch's side silently reverts them.

### The restack

This branch has exactly two commits of its own:

```
d80b1cee7  TU: reconstruct ov006/dScMgBSC_c as a 19-function shadow TU (TEXT-VERIFIED)
7c0f9a92c  Promote ov006/dScMgBSC_c to a single 19-function C++ TU
```

Cut a fresh branch from current `origin/main` and replay just those two. Everything they touch is `dScMgBSC_c`, not `dScMgSingle3DBase_c` — I checked the 27-file diff and it contains no edit to `include/dScMgSingle3DBase_c.h`, to that class's manifest, or to `src/actors/dScMgSingle3DBase_c.cpp`. So all four of the add/add collisions disappear along with the parent's commits, and you should land on two remaining conflict points, both expected:

- **`attribution.json`** — reorders on every pick. Three-way merge it; do not union the keys. Attribution is a stated non-goal here, so do not spend more than the mechanical resolve on it.
- **`notes/dScMgBase_c-slots-18-35.md`** — `main` has a version of this file now, so your +29 lands as add/add. Merge the two by hand; the slot-map content is the part worth keeping.

Roll the **eleven `@symbol` markers** into the same push — details in my review above. One round covers the restack and the markers together.

`#2071` is stacked on this branch and will need the same treatment afterwards. I will review it once it sits on a clean base.

One correction to something in my review, in your favour: I said the run-on fragment poisons four criteria. It poisons them because it swallows eleven live function *bodies* — I re-checked all eight lines I cited (`192, 200, 223, 312, 374, 463, 476, 486`) and every one is code, not comment. `tiers.py` blanks comments before four of the five checks, so prose inside a fragment is harmless. The ask stands exactly as written; I just want the reason on the record precisely.

andrewboudreau:
## Restack and markers verified — approving, merging once the ROM validator reports

Re-reviewed at `3194e6d6e058`. Both asks are done, and the restack is the clean kind.

### The markers

17 of 19 declared functions now carry `// @symbol`. The two that do not are `_ZN10dScMgBSC_cD1Ev` and `_ZN10dScMgBSC_cD0Ev`, which is correct — the destructor is inline in the header, so the pair is compiler-emitted and the lifecycle fallback covers it. No extra markers, no misspellings against `functions[].symbol`.

The last marker now sits at line 681 of 706, so the final fragment is 25 lines instead of the 354 it was. The eleven bodies that were being scored against `func_ov006_02125248` are their own again.

The markers commit is **`src/actors/dScMgBSC_c.cpp | 11 +++++++++++`** — eleven added lines, nothing else in it. That is exactly the ask and nothing else, which is what makes it cheap to re-verify.

### The restack

```
3194e6d6e  Mark every member of the BSC TU with its @symbol boundary
e3c4a7493  Promote ov006/dScMgBSC_c to a single 19-function C++ TU
85dd44510  TU: reconstruct ov006/dScMgBSC_c as a 19-function shadow TU (TEXT-VERIFIED)
```

Three commits on current `main`, 27 files, same set as before.

I checked the thing I was actually worried about — that a hand-resolve would take this branch's side on #2064's files and silently revert its post-review fixes. It did not:

```
git diff origin/main..3194e6d6e058 -- include/dScMgSingle3DBase_c.h \
    config/tu_manifest.d/ov006/dScMgSingle3DBase_c.json \
    src/actors/dScMgSingle3DBase_c.cpp src/_ZN10dBgActor_cD0Ev.cpp
  (no output)
```

All four are byte-identical to `main`. `notes/dScMgBase_c-slots-18-35.md` is **+29 / -0** against main, so your slot map was appended to the landed version rather than replacing it — the right resolution for that add/add.

And the promotion content itself survived the re-cut unchanged:

```
git diff 7c0f9a92c794 e3c4a7493 -- src/actors/dScMgBSC_c.cpp \
    config/tu_manifest.d/ov006/dScMgBSC_c.json include/dScMgBSC_c.h \
    config/arm9/overlays/ov006/delinks.txt config/converted-baseline.json
  (no output)
```

So everything I verified from the cartridge in my first review — the contiguous `0x02124908..0x021255f8` run, the thirteen licensing addresses, the RTTI chain, the base clause — carries over without re-checking.

### Gates on the real merge tree

`tools/premerge_check.py` against `origin/main @ 22da79ddb8de`, merge tree `a1918a3c40df`:

```
converted-ratchet   pass  ->  pass
dead-references     pass  ->  pass
duplicate-sources   pass  ->  pass
header-offsets      pass  ->  pass
langmode-ratchet    pass  ->  pass
layout-check        pass  ->  pass
src-tu-refs         pass  ->  pass
source-coverage     pass  ->  pass
  info: entries: 10925 -> 10907 (-18)   (consolidation, not a loss -- bytes are flat)
```

8/8, and the −18 is arithmetic: nineteen delink entries become one.

The converted ratchet is a clean 3-way re-key with the set size unchanged:

```
- src/_ZN10dScMgBSC_c6RenderEv.cpp   -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_c6RenderEv
- src/_ZN10dScMgBSC_cD0Ev.cpp        -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_cD0Ev
- src/_ZN10dScMgBSC_cD1Ev.cpp        -> src/actors/dScMgBSC_c.cpp#_ZN10dScMgBSC_cD1Ev
```

**Approved. I will merge as soon as `PR validation` reports** — it is the only check still pending, and it is the one that builds the ROM, so I am not merging ahead of it.

### One carry-forward, still not blocking

The manifest's five `CONFLICT` notes still end *"the other is commented out for review"*, and `grep -c "TUBUILD CONFLICT" src/actors/dScMgBSC_c.cpp` returns **0**. Worth noting because your own `dScMgMCarlo_c` TU on #2071 *does* carry those comments at lines 202-212 — so the generator emits them and something dropped them here. A reader following the note finds nothing to compare. Fix it forward in whichever PR is convenient: either restore the alternates, or replace the five notes with one line saying the kept declaration is confirmed by the 19/19 byte match and naming which spelling lost.

`#2071` is stacked on this branch and its merge base is still the pre-restack tip, so it will need the same re-cut. I have already verified its content independently — comment coming on that PR.

PR #2071, opened 2026-08-31
Promote ov006/dScMgMCarlo_c to a single 23-function C++ TU

andrewboudreau:
Stacked on #2067 (which is stacked on #2064). **Review the top commit pair only**; the earlier commits are #2067's.

`src_tu/actors/dScMgMCarlo_c.cpp` → `src/actors/dScMgMCarlo_c.cpp`, replacing 23 per-function legacy sources (18 flat-C, 5 already C++). The whole `.text` range `0x020f7634..0x020f8e44` now reproduces from **one** object.

## Consolidation is where the work was

Each legacy file compiled in isolation with its own private shadow structs and `extern` declarations, so they were free to contradict each other and never had to say so. Put in one TU they produced ~25 compile errors — every one a real disagreement about the same object. Each was settled against evidence, not against the compiler:

- **`Node` and the list-`Obj` are the same struct.** Five partial views, agreeing on every offset they share; unified into one layout with the fields commented.
- **A third, unrelated `Obj` was not a list node at all** — it was a shape-only shadow of this scene class's *own vtable*, used for the slot-18 self-dispatch in `InitResources`. Revived as `SceneVtable` so the two meanings stop colliding. It becomes a real base-class virtual call once `dScMgBase_c` declares slots 18–35.
- **`data_ov006_02142500` is the list HEAD and `_02142504` the TAIL**, proven by `func_ov006_020f7740` walking `head->next` and `tail->prev`. `Render` starts from the tail, so both its loops step `prev`, not `next` — corroborated independently by `func_ov006_020f8a3c`'s raw `*(char**)(r3 + 4)`.
- **`_Z14ApproachLinearRiii` / `_Z15ApproachLinear2Rsss` take references.** The mangled names say so, and a reference and a pointer pass the same address, so the accurate spelling costs no code.
- **`func_ov006_020f8540` was declared `(void)` but called with an argument.** ARM passes it in r0 and the body ignores it; spelling the parameter keeps the call site honest and changes no byte.
- `include/dScMgBase_c.h` already declared `data_ov004_020beb68` as `void*`, which settled that conflict against the TU's `char*`.

`decl_common.h` loses the scalar declarations of the four list globals (`_021424fc`/`_02142500`/`_02142504`/`_02142508`). They were arbitrary winners among 19 contradictory isolated spellings; every consumer is in this TU, which now declares all four with the recovered pointer type.

## The destructor is inline, and that is what makes the vtable ours

Out of line, mwcc emits the synthesized D0 *ahead* of the written D1; the cartridge has D1 first (`0x020f7634`) then D0 (`0x020f76a8`). Inlining hands the key-function role to `InitResources` — the next virtual declared, non-inline, defined here — which is what emits `_ZTV13dScMgMCarlo_c`. Slots 16 and 17 name D1 then D0, odr-using both, so the compiler emits the pair in cartridge order, and the homeless D2 is gone.

## Verification

| check | result |
|---|---|
| `tubuild verify ov006/dScMgMCarlo_c` | 23/23 MATCH, objisolate clean, reloc-destinations clean, ROM-ascending section order |
| `tubuild linkcheck` (pre-promotion) | SCRATCH-LINK-VERIFIED; 106/106 exact; full ROM built and identical to stock |
| `rombuild -j16` (post-promotion) | 11,088 source-built functions, **reproducing 11,088, mismatching 0**; **106/106 exact, 100.000000%** |
| the ROM it builds | sha256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` — **byte-identical to the stock build** |
| `premerge_check` on the merge tree | all 8 gates pass; nothing goes green → red |
| layout / eligible / langmode / port_refcheck / dead-references | clean |

`dsd check symbols --fail` still fails on the same **9 pre-existing errors the baseline control also has** — 0 new, so not attributable to this TU, but it is **not green** and is not reported as such.

## Compiler-generated output, byte-compared

All 13 RTTI/vtable symbols are licensed in `compiler_only_output`, every canonical address confirmed against `config/arm9/.../symbols.txt`, and `romdata_check` compares each against the cartridge: **6 VERIFIED, 7 PARTIAL (exact prefixes), 0 DIFFER.**

`_ZTV13dScMgMCarlo_c` reproduces 72 of the cartridge's 144 bytes — **18 of 36 slots** — because `dScMgBase_c` still leaves slots 18–35 undeclared. That is a verified prefix, not a disagreement; widening it is the `dScMgBase_c` follow-up, not this change.

The cartridge's own `_ZTS` strings corroborate the names: reading `extracted/dsd/arm9_overlays/ov006.bin` at overlay base 34340544 gives `13dScMgMCarlo_c` at `0x0213d594` and `19dScMgSingle3DBase_c` at `0x0213bd00` — so this TU emits no homeless typeinfo record.

## Second commit: `@symbol` boundaries

`tiers.score_member` slices a member at `// @symbol` markers and, finding none after a marker, runs to end of file. Only 5 of 21 in-file members carried one, so `Render` was scored over every flat-C body below it and the CONVERTED ratchet read that as a backslide — neither raw-cast nor mangled-callee is true of `Render`'s own body. 16 markers added; comments only, ROM sha256 unchanged.

**`src/actors/dScMgBSC_c.cpp` has the same gap (6 markers for 17 members).** Nothing banked mis-scores there today, so it is latent rather than red — it belongs to #2067, not here.

## Attribution

A 23-delete + 1-add collapse ends 22 git lineages, so `attribution.json` takes 23 explicit override keys on the surviving path, each carrying the first matcher the gate itself computes. `prepush_attribution`: **47 consolidated with credit intact, 0 changed, 0 lost.**


andrewboudreau:
## Verified from the cartridge — content approved. One more re-cut, and this one is trivial.

Reviewed at `fee055d82f75`. I verified this PR's content at the pre-restack head, then confirmed the restack changed nothing: your two commits' own delta is **byte-identical** across the move.

```
git diff 7c0f9a92c794..de7e85a0dd57  ->  125583 bytes   (old head, old base)
git diff 3194e6d6e058..fee055d82f75 ->  125583 bytes   (new head, new base)
diff of the two patches -> empty
```

32 files, +1667/−1106 both ways. So everything below, verified once, still stands — nothing needs re-checking after the next move either, as long as the commits are replayed rather than hand-merged.

---

## Where this stands right now

**#2067 is merged** — squashed to `244fd4d849c9`, `main` is there now. Its `PR validation` came back green and I merged it a few minutes ago.

That squash makes this PR `DIRTY` again, for the same reason as last time: `main` now carries #2067's *content* under a commit this branch has never seen, while this branch still carries the three original commits. But the conflict set is one line long this time:

```
git merge-tree --write-tree --name-only origin/main fee055d82f75
  attribution.json          CONFLICT (content)
  config/arm9/overlays/ov006/delinks.txt      auto-merged
  config/converted-baseline.json              auto-merged
  ... everything else                         clean
```

**`attribution.json` is the only conflict.** Both ledgers auto-merge, which is the pleasant surprise — the squashed content matched line-for-line, so the delinks and the baseline resolve themselves.

Cut from `origin/main` and replay these two:

```
9347c7bd8  Promote ov006/dScMgMCarlo_c to a single 23-function C++ TU
fee055d82  Mark every member of the MCarlo TU with its @symbol boundary
```

Three-way merge `attribution.json`; do not union the keys. Attribution is a stated non-goal here, so spend the mechanical resolve and nothing more.

I already ran `premerge_check` on the **combined** stack — this PR merged into the `main` that predated #2067, which is exactly the #2067+#2071 tree that is about to exist:

```
merge tree 51379ec5010b
converted-ratchet / dead-references / duplicate-sources / header-offsets /
langmode-ratchet / layout-check / src-tu-refs / source-coverage   8/8  pass -> pass
  info: entries: 10925 -> 10885 (-40)   (consolidation, not a loss -- bytes are flat)
```

I will re-run it against the real merge tree after the re-cut, but I do not expect a surprise.

---

## What I verified independently

### Markers: 21, and the two absences are correct

23 declared functions, 21 markers. The two without one are `_ZN13dScMgMCarlo_cD1Ev` and `_ZN13dScMgMCarlo_cD0Ev` — correct, since you flipped the destructor inline and the pair is compiler-emitted, covered by the lifecycle fallback. Every other marker matches `functions[].symbol` exactly: no extras, no misspellings. Last marker is at line 1125 of 1145, so the final fragment is 20 lines, not 354.

Second TU in a row that arrives already conforming to `notes/tu-promotion-conventions.md` §6 without being asked.

### The linked range is exactly contiguous and abuts both neighbours

Chaining the 23 removed delink ranges:

```
0x020f7634 -> 76a8 -> 7730 -> 7740 -> 7994 -> 7a00 -> 7a90 -> 7b10 -> 7b90 -> 7c10
           -> 7e2c -> 7ee4 -> 8154 -> 8224 -> 82d0 -> 8320 -> 84a8 -> 8540 -> ...
           -> 0x020f8e44
```

23 for 23, no gap, no overlap. And in the rewritten `delinks.txt`:

```
src/MgMemoryMaster_Spawn.c      .text 0x020f75d4..0x020f7634
src/actors/dScMgMCarlo_c.cpp    .text 0x020f7634..0x020f8e44   <- new
src/func_ov006_020f8e44.cpp     .text 0x020f8e44..0x020f8ed8
```

Both endpoints preserved, both neighbours abut exactly.

### All 13 licensing addresses read back correctly from the ROM

This is the part `SCRATCH-LINK-VERIFIED` does *not* cover — the link round proves the bytes, not that each `canonical_module`/`canonical_address` row points at the record it names. Resolved against `extracted/arm9_dec.bin` and the overlay images, each row in its own module first:

```
_ZTI7fBase_c              arm9  0x02086d70 -> name '7fBase_c'                OK
_ZTS7fBase_c              arm9  0x02086e60 -> '7fBase_c'                     OK
_ZTS7dBase_c              arm9  0x02086e54 -> '7dBase_c'                     OK
_ZTS8dScene_c             arm9  0x020914b0 -> '8dScene_c'                    OK
_ZTI7dBase_c              arm9  0x02086e78 -> name '7dBase_c'                OK
_ZTI11dScMgBase_c         ov004 0x020bbf6c -> name '11dScMgBase_c'           OK
_ZTS11dScMgBase_c         ov004 0x020bbf84 -> '11dScMgBase_c'                OK
_ZTI8dScene_c             arm9  0x020914d4 -> name '8dScene_c'               OK
_ZTI19dScMgSingle3DBase_c ov006 0x0213bc64 -> name '19dScMgSingle3DBase_c'   OK
_ZTS19dScMgSingle3DBase_c ov006 0x0213bd00 -> '19dScMgSingle3DBase_c'        OK
_ZTI13dScMgMCarlo_c       ov006 0x0213d588 -> name '13dScMgMCarlo_c'         OK
_ZTS13dScMgMCarlo_c       ov006 0x0213d594 -> '13dScMgMCarlo_c'              OK
_ZTV13dScMgMCarlo_c       ov006 0x0213d664 -> [V-4] = 0x0213d588             OK
```

Thirteen for thirteen.

### The RTTI chain, walked from the vtable

```
_ZTV13dScMgMCarlo_c @ 0x0213d664
  [V-4] typeinfo      = 0x0213d588   manifest says 0x0213d588        MATCH
  TI[1] name ptr      = 0x0213d594 -> '13dScMgMCarlo_c'
  TI[2] base's _ZTI   = 0x0213bc64   = _ZTI19dScMgSingle3DBase_c     MATCH
```

`13` is exactly `len("dScMgMCarlo_c")`, so the read checks itself, and `TI[2]` puts `dScMgMCarlo_c : dScMgSingle3DBase_c` in the cartridge rather than in our headers. The class is in the agreeing set, which is why this promotion can land at all — #2066 is the counter-example where a coined name made it impossible.

Agreed on your point about `rtti_name_audit --class` for the next unit. A second independent walk of the same three words is worth more than either implementation alone, and it is the difference between "I read the bytes" and "two tools that do not share code read the bytes."

### The converted ratchet is one clean re-key

```
- src/_ZN13dScMgMCarlo_c6RenderEv.cpp -> src/actors/dScMgMCarlo_c.cpp#_ZN13dScMgMCarlo_c6RenderEv
```

One line, set size unchanged. The other four class files and all eighteen `func_ov006_*` files were never in the baseline, so nothing is silently dropped.

And on your correction — accepted, and it is the more useful reading. This one went **pass → FAIL** on the merge tree; #2067's eleven were latent. Worth keeping apart precisely because it tells us what the gate is: it catches a swallowed span only once a *banked* entry falls inside it. That is a real backstop, not a full one, and it is the argument for markers being mandatory rather than merely tidy.

### The `decl_common.h` edit is safe, and I checked the claim rather than taking it

You remove three declarations and change one signature in a header the whole tree includes, on the stated grounds that every consumer lives in the new TU. Grepping the branch tip for all four symbols across `src/`, `include/`, `mods/` and `port/`: every reference to `data_ov006_021424fc`, `_02142500`, `_02142504` and `_02142508` is inside `src/actors/dScMgMCarlo_c.cpp`. The claim holds.

Recovering all four as one intrusive doubly-linked `Node` list — two cursors, head, tail — out of four legacy files that spelled them `int`, `void*`, `char*` and `struct Node*` is the substantive win here, and it is what TU reconstruction is *for*. `data_ov006_021424fc->f_2a > data_ov006_02142508->f_2a` reads as the comparison it is, instead of as arithmetic on an `int`.

### The inline-destructor flip is argued from the ROM

D1 at `0x020f7634` sits *before* D0 at `0x020f76a8`. An out-of-line destructor cannot produce that order — mwcc emits the synthesized D0 ahead of the written D1. The flip is forced by the cartridge, `InitResources` becomes the key function as the first non-inline virtual declared after it, and the TU defining it emits `_ZTV13dScMgMCarlo_c`. `emittedTextOrderIsRomAscending = True` is the mechanical confirmation.

### The `dsd check symbols` reporting is honest

`checkSymbols: false`, with `symbolCheckErrors` and `symbolCheckBaselineErrors` both recorded and **identical, nine for nine**. Zero new errors, and the manifest does not claim the check as green. That is the right way to record a pre-existing failure; I would rather see it than a suppressed row.

---

## Two notes, neither blocking

### 1. One stale declaration left behind — in the file that proves your array shape

`src/func_ov006_020f8e44.cpp:9` still says:

```cpp
extern "C" void func_ov006_020f7730(void);
```

It is the only reference to that symbol outside the new TU. It compiles — the file is self-contained, does not include `decl_common.h`, and the value is cast to `void*` at the call — so this is not a correctness problem and the 106/106 is unaffected. But the point of this PR on that symbol is that the signature is `void(int *)`; you fixed `decl_common.h` and the header, and this is the one spelling left disagreeing.

Worth rolling into the re-cut push because that file is the corroboration for the header:

```cpp
func_020733a8(p + 0x51a8, 0x50, 0x30, (void*)func_ov006_020f8ed8, (void*)func_ov006_020f7730);
```

is what makes `mArray` at `0x51a8`, `0x50` elements of `0x30`, with that element destructor, a measurement rather than an inference.

### 2. `tubuild` truncates an Allman-braced conflict record — and this PR carries the proof

Your TU records nine conflict notes. Seven are fine. This one is not:

```
/* TUBUILD CONFLICT -- alternate body of struct 'Obj', from the legacy file for func_ov006_020f8224, NOT applied:
struct Obj
*/
```

The declaration it was quoting, in `src/func_ov006_020f8224.c`:

```c
struct Obj
{
    char pad[0x2a];
    short f2a;
    unsigned char f2c;
};
```

The splitter stopped at the newline because the brace is on the next line, so the record kept "for review" carries no information at all. Every other body-conflict block in the file has a real body; this one is empty because of brace style alone.

**No harm in this instance** — the discarded alternate agrees with the kept struct (`f2a` at 0x2a, `f2c` at 0x2c both ways), so nothing was lost that would have changed a decision. But it is the same brace-position blindness class as the `check_header_offsets` hazard, and next time the dropped body may be the one that disagrees.

This is the complement to the headless-struct defect you are taking on next: yours drops the *header* and keeps the body, this one keeps the header and drops the *body*. Same splitter, two ends of the same bug. Please put both failing inputs in the test file — the fix without the inputs leaves us re-deriving them next time.

---

## Verdict

**Approved on content; re-cut onto `main` and I will merge it.** Every cartridge-derived claim checks out, the markers are complete and correct, the `decl_common.h` blast radius is bounded and measured, and the ratchet delta is one clean re-key. One conflict to resolve, in a file that is a non-goal.

On the merge itself: that is mine to do and I am doing it — you get branches to green, I decide what lands and in what order. #2067 went in the moment its validator reported; this one goes in the same way once it sits on a clean base.

andrewboudreau:
Re-cut onto `origin/main@ccb65b8b3` (post-#2067 squash `244fd4d849c9`). `MERGEABLE` again.

`fee055d82` → `ea0760117`. Both commits replayed rather than merged, and **neither cherry-pick conflicted** — `attribution.json` included, so the three-way merge you predicted resolved itself once the squashed BSC content was line-for-line identical.

The patch is unchanged, which is checkable rather than assertable:

```
$ diff <(git diff 3194e6d6e058..fee055d82f75) <(git diff origin/main..628f52b31)
$ echo $?
0
```

Byte-identical, 125583 bytes both sides. Everything verified at the old head carries over untouched.

## The one content change: `src/func_ov006_020f8e44.cpp:9`

Commit `ea0760117`, on top. It declared the element destructor `void(void)` while the TU recovers it `void(int *)`:

```c
void func_ov006_020f7730(int *p) { p[0] = (int)data_ov006_0213d5ac; }
```

Now spelled the same in both, with the reason in a comment — that file is the corroboration for `mArray` at 0x51a8 being 0x50 × 0x30 with that element dtor, so the disagreement was sitting on the one fact that sizes the array. The factory casts to `void*` at the call, so no code moved.

## Re-verified at `ea0760117`

| | |
|---|---|
| `rombuild -j16` | 11,088/11,088 reproducing, **0 mismatching**, 106/106 exact, 100.000000% |
| ROM sha256 | `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` (unchanged) |
| `tiers_ratchet --check` | CONVERTED PASS, baseline 2568, current 2568 |
| `premerge_check` | 8/8 pass→pass on merge tree `4e0d92d28a7d`; `entries 10907 -> 10885 (-22)` |
| `prepush_attribution` | 11189 tracked, 23 consolidated with credit intact, **0 changed, 0 lost** |
| `port_refcheck` | 405 references, 0 stale |

(-22 rather than -40 because BSC's half of the consolidation is now in the base.)

## Related: the splitter defect is filed

#2072, tools-only off `origin/main`. Both ends of the bug are in the test file as you asked, from the real inputs: the declaration keeps its Allman body, *and* the `TUBUILD CONFLICT` comment carries a real alternate body instead of the two words `struct Obj`. Fixing the walk turned up two more shapes — an elaborated return type (`struct dActor_c *dCapEnemy_c::RespawnIfHasCap()`) being filed as a shadow declaration, and definitions wrapped in `namespace`/`extern "C"` refusing with a message that named nothing. Measured over all 11150 legacy sources: 82 headless bodies repaired, 10 more files split cleanly, 683 refusals that now name the offending block and line, 0 regressions, and nothing enrolled in a TU is affected.

andrewboudreau:
**Merged** as `6d8f8e529b89`.

Final state: `MERGEABLE` / `CLEAN`, all eight checks green including `PR validation` (13m29s). `main` was green across all five gates at `6ca3191b1` before I pushed the button.

I did not re-run `premerge_check` against the new merge tree, and here is the reasoning so it is on the record rather than assumed. My 8/8 was measured against `ccb65b8b3c38` (merge tree `4e0d92d28a7d`, entries 10907 → 10885). The only thing that landed on `main` in between is #2072:

```
tools/tubuild.py       +79 -5
tools/test_tubuild.py  +154
```

Neither file is one of the eight gates `premerge_check` runs, and this PR touches no `tools/` at all — so the two trees differ only in files that cannot change the verdict. `main` itself is green carrying them. That transfers; a third 9-minute run would have measured the same thing.

Thanks for taking the `func_ov006_020f8e44.cpp` finding in `ea0760117` — that file is the corroboration for the `mArray` shape in the header, so having it spell the signature the same way as `decl_common.h` is worth the extra commit. Third commit is exactly the one line I asked for and nothing else.

That is #2067, #2072 and #2071 landed in sequence today, each one re-cut onto a clean base rather than hand-merged. The discipline of replaying commits instead of resolving them is why the `attribution.json` conflict cost a mechanical resolve and nothing else — worth keeping.

PR #2079, opened 2026-08-31
Reconstruct ov006/dScMgCard_c as one C++ translation unit, and verify its vtable and RTTI against the cartridge

andrewboudreau:
## TL;DR

`ov006/dScMgCard_c` — the Dilar card-match minigame scene — was 34 separate
one-function source files. This makes it **one C++ translation unit**, the way
the original was, and then checks that the compiler-generated parts the ROM
carries for that class come out of *our* compiler byte-for-byte: the vtable, all
six typeinfo records, their name strings, and the destructor pair.

Same shape as #2077 (`dScMgMCarlo2_c`), which this follows exactly.

## What's in it

Three commits, and it's worth reading them in order.

**1. `2a2757` — four wrong declarations, and the phantom call they caused.**
Four `decl_common.h` prototypes had the wrong types. One of them mattered:
`func_ov006_020dac34` was declared as returning `void *`, which let
`_ZN11dScMgCard_c8BehaviorEv` compile as `func_ov004_020b65e4(func_ov006_020dac34(this))`
— a nested call the cartridge does not make. The bytes still matched because the
return register happened to survive; the source was describing something false.
It is now two statements, and the four prototypes carry their real types
(`char *`/`s16 *`, `void *`, `unsigned char *`).

**2. `87ad58` — the reconstruction, as a shadow.** `src_tu/actors/dScMgCard_c.cpp`
plus its manifest entry. Not enrolled yet, so nothing about the ROM can change.

**3. `8fd88a` — the promotion.** 34 `complete` delinks entries become one range
(`0x020d95a4`–`0x020dbd54`), the shadow moves to `src/actors/dScMgCard_c.cpp`,
and the 34 legacy sources are deleted. 42 files, +64 / −1913.

**4. `9a157b` — the prose the promotion invalidated.** The banner `tubuild.py
create` stamps on every reconstruction ("SHADOW … NOT ENROLLED, NOT CANONICAL …
contributes nothing to the ROM build") is the opposite of true once delinks names
the file, and it sat above a list of 34 paths the same commit deletes. Rewritten;
the header's and `notes/minigame-provenance.md`'s citations of deleted files
become ROM addresses.

## The destructor is inline, on purpose

Written out of line, mwcc emits the synthesized `D0` **ahead of** the written
`D1`, and `rombuild`'s fail-closed isolate refuses an object whose licensed
`.text` is not in ROM address order (cartridge: D1 `0x020d95a4`, D0 `0x020d9638`).
Defining it inline in the class body means it cannot be the key function;
`InitResources` takes that role, vtable slots 16/17 name D1 then D0 and odr-use
both, and the pair comes out in cartridge order. `D2` stops being emitted at all.
`include/dScMgMCarlo2_c.h` documents the same thing — this is the second instance.

## What was actually verified

| check | result |
|---|---|
| `tubuild.py verify` | 34/34 TEXT-VERIFIED, emission order clean |
| `tubuild.py partial` | 34/34 CONTRIBUTION-EQUIVALENT |
| `tubuild.py linkcheck --partial` | PARTIAL-LINK-VERIFIED, **0 new** symbol errors (9 pre-existing, reproduced on the baseline) |
| `rombuild.py -j 16` | **106/106 exact**, 100.000000% of compared bytes; **11,088 / 11,088** source-built functions reproducing, 0 mismatching; intact-TU gates PASS |
| `romdata_check.py --files src/actors/dScMgCard_c.cpp` | **13 data symbols — 6 VERIFIED, 7 PARTIAL, 0 DIFFERS, 0 UNNAMED** |
| `premerge_check.py HEAD` | nothing goes green → red on the merge tree |
| `port_refcheck.py` | 405 references, all resolve |

**VERIFIED** here means every byte of the cartridge's extent was compared and
equal, with the object's relocations resolved the way the linker resolves them —
all six typeinfo records: `_ZTI7fBase_c` 8/8, `_ZTI7dBase_c` 12/12,
`_ZTI8dScene_c` 12/12, `_ZTI11dScMgBase_c` 12/12, `_ZTI19dScMgSingle3DBase_c`
12/12, and this class's own `_ZTI11dScMgCard_c` 12/12.

**PARTIAL** is an exact prefix that stops short of the cartridge's extent, not a
disagreement: the six `_ZTS*` name strings are equal for every whole word emitted
and short only of the ROM's 4-byte padding; and `_ZTV11dScMgCard_c` matches
**72 of 144 bytes — 18 of 36 slots**, because `dScMgBase_c` still leaves slots
18–35 undeclared. Widening that vtable is a change to the base class, not to this
TU, and it is the single lever that would turn this and every other promoted
minigame vtable into a full table.

Worth knowing when reading the vtable by hand: the `_ZTV` symbol in
`config/arm9/overlays/ov006/symbols.txt` is at the **address point**
(`0x0213bdb4`), not the record base — the record starts 8 bytes earlier. Compare
without that rebase and every word looks wrong.

## Numbers that move, and why

`source-coverage` reports `entries: 10885 -> 10852 (-33)`. That is consolidation,
not loss: the same bytes, 33 fewer source entries.

`ROM data from source` falls 13 object records against main. Also arithmetic: on
main the out-of-line destructor was the key function, so `D1Ev.cpp` and `D0Ev.cpp`
*each* emitted the same 13 licensed records. Inlining moves the key function to
`InitResources` and the one promoted object emits each record once. **No symbol
changed verdict** — 462 verified / 250 partial / 9 differ / 524 unnamed, exactly
as on main.

`prepush_attribution` reports 32 `CREDIT LOST`. That is structural to a 34-delete
/ 1-add collapse and cannot be split into rewrite-then-move; two entries do
survive as "consolidated, credit intact". Hence the `attribution-override` label,
per the standing call that attribution is not a goal here.


andrewboudreau (commented):
**Gatekeeper review — clearing this to merge.** (Same account, so GitHub will not let me press Approve; this comment is the sign-off.)

Base `f35cf8f4d` is now two commits stale (main is `cc87f6806`, carrying #2078 and #2077), so the eight green checks measured a tree that no longer exists. I did not ask for a rebase — I built the real merge tree and measured **that** instead. `git merge-tree --write-tree origin/main 9a157bb7e` → **rc=0**, tree `21dabe4a5`, clean despite six overlapping files.

Two of the checks below are ones neither this PR's CI nor #2077's/#2078's could have run.

## The phantom nested call — confirmed against the cartridge

This is the substantive claim and it is correct. The old form was:

```c
func_ov004_020b65e4(func_ov006_020dac34(this));
```

`_ZN11dScMgCard_c8BehaviorEv` lives at `0x020dabec`, size `0x48`. Its bytes:

```
0x020dac10: ebff9b6e   BL #0x020c19d0
0x020dac14: e1a00004   MOV r0, r4        <- `this`
0x020dac18: eb000005   BL #0x020dac34
0x020dac1c: ebff6e70   BL #0x020b65e4
0x020dac20: e3a00001   MOV r0, #1
```

The two `BL`s are **back to back with no instruction between them**. Nothing marshals a return value into `r0` for the second call, so the hardware never performs the data flow the nested form describes. Sequencing them, as this PR does, is what the ROM actually contains.

Corroboration is decisive independent of the disassembly. Every declaration of `func_ov004_020b65e4` on main:

- `4 ×` `void func_ov004_020b65e4(void);`
- `1 ×` `void func_ov004_020b65e4(void) {` — **the definition itself**
- `1 ×` `extern "C" void func_ov004_020b65e4(void);`
- `1 ×` `void func_ov004_020b65e4(void *p);` — the single outlier, the one this PR fixes

The change aligns the last remaining disagreeing declaration with the function's own definition.

## Gates measured on the merge tree, not the branch

- **Header offsets under #2078's stricter checker** (the interaction neither PR's CI saw): extracted `21dabe4a5` and ran the merged `tools/check_header_offsets.py` over it — **rc=0, 0 MISMATCH, 0 UNPARSED**, `include/dScMgCard_c.h: 18 commented fields` verified. #2078 unskipped root structs with implicit vptrs; this PR's new header passes that arm rather than being waived past it.
- **`converted-baseline.json` is a re-key, not a loss**: main 2568 / branch 2568 / merge tree 2568. The two entries that disappear from main are exactly the two files this PR deletes, re-keyed to `src/actors/dScMgCard_c.cpp#_ZN11dScMgCard_c16CleanupResourcesEv` and `…#_ZN11dScMgCard_c8BehaviorEv`. The SET ratchet holds.
- **ov006 delinks, whole-overlay**: both this PR and #2077 collapse ranges in the *same* overlay's `delinks.txt`, and neither saw the other. Overlapping ranges would break the ROM link with every byte gate still green, so I scanned the merged file end to end:

  ```
  main (has #2077)      ranges=1743  overlaps=0  gaps=40
  merge tree (+#2079)   ranges=1710  overlaps=0  gaps=40
  ```

  Zero overlaps, and the gap count is **unchanged** — the 33-range drop is exactly the files collapsed here, and the gap adjacent to `0x020dbd54` is pre-existing overlay structure. `dScMgCard_c 0x020d95a4–0x020dbd54` and `dScMgMCarlo2_c 0x020f8ef4–0x020fa6ac` are far apart and each abuts its neighbours exactly.
- **Manifest**: `boundary_confidence: high`, 34/34 symbols declared and matched, all criteria PASS. The four substantive notes (generator provenance, the `func_ov006_020c0f0c` exclusion, the `compiler_only_output` pre-link measurement, the inline-destructor rationale) each say what was measured rather than asserting a conclusion. Good.

## One non-blocking note, already characterised

19 `CONFLICT` notes / 0 markers in the file. Same stale-prose pattern I recorded on #2077: `tubuild.py create` stamps `"kept the first, the other is commented out for review"` and never updates the note when the author hand-resolves afterwards. Measured backlog across landed TUs is 20 accurate / 6 stale, so this is pre-existing tooling debt, not a defect in this PR — the resolutions here are real. Worth regenerating or annotating on a future pass; I am tracking the `tubuild.py promote` fix (drop notes whose marker is gone) separately.

Merging.

PR #2081, opened 2026-08-31
Close three DIFFERS vtable slots in the minigame family

andrewboudreau:
`tools/romdata_check.py` scored **four** of the minigame family's emitted vtables `DIFFERS` against the cartridge. Three of the four came down to the same thing: a virtual the compiler was never told about, so mwcc wrote the *inherited* body's address into a slot where the cartridge holds an override. This closes those three.

ROM data from source goes **9 differ → 7**. The fourth, `_ZTV12dScMgAmida_c` slot 18, is entangled with the `dScMgBase_c` keystone and is deliberately untouched.

## `_ZTVN10dMgPsOpt_c11TouchIcon_cE` slot 1 — a real migration

The nested touch icon genuinely overrides `dThIcon_c::Render`: the cartridge's table at ov004:0x020b8c30 holds `0x020b8c18` in slot 1 where the base's holds `_ZN9dThIcon_c6RenderEv`. The class had only ever declared a ctor/dtor pair, so mwcc emitted the inherited address.

So the override is declared and the body migrated — `src/func_ov004_020b8c18.c` → `src/_ZN10dMgPsOpt_c11TouchIcon_c6RenderEv.cpp`, C++, with the raw `char *self` offset arithmetic replaced by `dThIcon_c`'s own seven fields. `build_pin.verify(..., 0x020b8c18, 0x1a8, 'ov004')` returned `(True, '2004/b56')` on the first attempt, and the two-slot table is now **8 of 8 bytes exact** — the family's first fully byte-verified minigame vtable outside `_ZTV22dMg3DHeyhoObjAdapter_c`.

**Reviewers, note the side effect.** `Render` is now this class's first *declared* virtual, which makes it the key function — so the emitted vtable **moved out of the destructor TU**. `romdata_check --files src/_ZN10dMgPsOpt_c11TouchIcon_cD1Ev.cpp` now reports **zero** data symbols. That is the move, not a loss: the Render TU emits five (3 VERIFIED / 44 bytes, 2 PARTIAL / 20 bytes, 0 DIFFERS). A zero on the old key-function source is the expected signature whenever a first declared virtual is added.

## `_ZTV16dScMgSmartball_c` slot 0, `_ZTV12dScMgSlot1_c` slot 6 — no code needed

Both bodies are real ROM code with **no delinks entry**: ov006:0x02118b70 (0x8dc, `InitResources`) and ov006:0x0210c9e0 (0x81c, `Behavior`) are served by the module's gap object straight from the cartridge.

Nothing has to be decompiled to fix the slot. The *emitted* table needs only a declaration and a symbol to point at. So both are named in `config/arm9/overlays/ov006/symbols.txt` and declared in their headers, each with a comment saying the body is deliberately not decompiled and why the declaration is still required. **No source file changes, and `config/converted-baseline.json` and `attribution.json` are untouched** — this PR has no ledger conflict with anything in the queue.

Precedent for a named ROM body with no source: `_ZN9dThIcon_c6RenderEv` at ov001:0x020ab54c.

### A header was asserting something false

`include/dScMgSlot1_c.h`'s banner claimed its own vtable *"is never compiler-emitted either — only aliased — so the real ROM vtable word for slot 6 is untouched."* That was true once and stopped being true when the destructor became the class's key function: `_ZTV12dScMgSlot1_c` **is** emitted, from `src/_ZN12dScMgSlot1_cD1Ev.cpp`, and `romdata_check` byte-compares it. The banner is corrected in place rather than deleted, so the retraction stays visible next to the claim. Its slots-27/28 paragraph now points at the keystone note for why slots 18/19 still differ.

## `notes/dScMgBase_c-slots-18-35.md` — slot 18 corrected

The note recorded slot 18 as unnamed, taking the `Virtual48` placeholder. **It is `OnYoshiTryEat`**, on two independent sources:

1. `include/dActor_c.h:131` declares `virtual int OnYoshiTryEat(); /* slot 18 */`, and a dozen further headers repeat it at that index. `dActor_c` is on the *other* branch of `fBase_c`, so alone this is only suggestive.
2. Walking `_ZTV11dScMgCoin_c` (ov006:0x0213bf50) against `_ZTV11dScMgBase_c` (ov004:0x020bc0c0), 36 slots each: `dScMgCoin_c` overrides exactly three — 16, 17 (the dtor pair) and **18, with ov006:0x020de5b0**, which `ov006/symbols.txt:545` **already names** `_ZN11dScMgCoin_c13OnYoshiTryEatEv`. Recovered and landed before that table existed.

The note's old reasoning — *"slot 18 stays unnamed on both sides, which is consistent rather than a gap"* — was consistent for the wrong reason: two sources that both happened to lack the name, not two sources that agreed there was none.

The note also gains three things the keystone PR will need:

- **`include/dActor_c.h:120-151` is the ready-made signature template for slots 18-30** — identical indices, and all thirteen names agree with the `recovered name:` comments on this class's own ov004 bodies.
- **Slots 31-35 are the real remaining unknown**, not eighteen slots. `Kill`, `AfterClsn`, `Virtual84`, `Virtual88`, `Virtual8C` are not in `dActor_c.h`.
- **The measured cost: 106 overrides across 33 classes**, 100 in slots 18..35 and 6 at slot ≥36; `dScMgBase_c` owns 18, so **82 are descendant-side**. Roughly 34 of those classes already emit a vtable, and widening the base flips every one that does not simultaneously declare its own overrides from `PARTIAL` (a byte-exact 18-slot prefix) to `DIFFERS`. That is why it has to land as one commit.

One address-point gotcha is written down there too: a `_ZTV*` address in `symbols.txt` **is** the address point, so slot *i* is at `addr + 4*i`. Adding `objisolate.VTABLE_PREAMBLE` (8) on top shifts every reading two slots and makes the destructor pair look like slots 18/19.

## Verification

Run in `C:\tmp\sm64ds-mgvtslot`, **after** rebasing onto `origin/main@c2f5ba278`:

| gate | result |
|---|---|
| `rombuild.py -j 16` | module fidelity **106/106 exact, 100.000000%** of compared bytes |
| | source-built functions **11,088 / 11,088 reproducing, 0 mismatching** |
| | intact TU gates: dsd modules PASS, zero new symbol errors, storage aliases exact |
| ROM data from source | 463 verified, 251 partial, **7 differ** (was 9), 524 unnamed |
| `premerge_check.py HEAD` | all 8 gates `pass → pass`, nothing goes green → red |
| `eligible.py` | 11096 / 11211, the migrated symbol among them |
| `port_refcheck.py` | 405 references checked, 0 stale |
| `check_dead_references.py` | no new dead references, no broken markdown links |
| `check_header_offsets.py --changed` | 4 changed headers, 0 mismatched, 0 unparsed |

The 7 remaining differ include `_ZTV12dScMgSlot1_c` (slots 18/19) and `_ZTV12dScMgAmida_c` (slot 18) — both blocked on the keystone, both now documented as such in the header and the note.

## Attribution

`prepush_attribution` reports **1 `CREDIT LOST`**: `src/func_ov004_020b8c18`, credited to `github-actions[bot]`, rewritten and moved in the same commit. Splitting a migration into rewrite-then-move to preserve a bot's credit is not worth a second commit; `attribution-override` applied per the standing policy.


andrewboudreau (commented):
**Gatekeeper review.** (Same account, so GitHub refuses the Approve button; this comment is the sign-off.) I re-derived all three vtable claims from the cartridge rather than reading the PR body, and I checked the one hazard this class of change can hide. **The byte claims all hold.** Two prose corrections are requested below; neither touches a conclusion.

## The three DIFFERS slots, read out of the ROM

**1. `dMgPsOpt_c::TouchIcon_c::Render` — correct, but the body quotes the wrong address.**

The body points at `ov004:0x020b8c30` for the vtable. That address is *inside a function body* — `0x020b8c18 + 0x18` — and reading four words there gives ARM instructions (`0xe1cd00b2`, `0xe596001c`), not pointers. The real table is `_ZTVN10dMgPsOpt_c11TouchIcon_cE` at **`ov004:0x020bca7c`** (`config/arm9/overlays/ov004/symbols.txt:786`). At that address:

| slot | `_ZTV9dThIcon_c` (ov001:0x020ad494) | `TouchIcon_c` (ov004:0x020bca7c) |
|---|---|---|
| 0 Behavior | `0x020ab550` | `0x020ab550` — **identical** |
| 1 Render | `0x020ab54c` = `_ZN9dThIcon_c6RenderEv` | `0x020b8c18` — **differs** |

Word `+0x08` is ASCII, which independently confirms the table is two slots. `0x020b8c18` is `func_ov004_020b8c18 kind:function(arm,size=0x1a8)`, matching the PR's `build_pin.verify(..., 0x020b8c18, 0x1a8, 'ov004')` exactly. So: Render overridden, Behavior genuinely not, and the header's comment says precisely that. Only the address in the PR description needs fixing.

**2. Both "no delinks entry" bodies really do fall in module gaps.** My first pass on this said "covered" and was wrong — a regex over `start:0x… end:0x…` also matches the leading *section* rows (`.text start:0x020bfec0 end:0x0212b890`), which swallow everything. Parsing the real structure (filename heading line, then indented per-object rows) gives 1710 delinked `.text` objects in ov006 and this:

```
0x02118b70  prev src/func_ov006_02118ae4.c                                ends exactly 0x02118b70
            next src/_ZN16dScMgSmartball_c21AfterCleanupResourcesEj.cpp   starts 0x0211944c
            gap width 0x8dc      <- PR states 0x8dc

0x0210c9e0  prev src/_ZN12dScMgSlot1_c6RenderEv.cpp                       ends exactly 0x0210c9e0
            next src/_ZN12dScMgSlot1_c13InitResourcesEv.cpp               starts 0x0210d1fc
            gap width 0x81c      <- PR states 0x81c
```

Both widths match to the byte. The module's gap object serves those bodies straight from the cartridge; the declaration is what the emitted vtable needs. Correct.

**3. `dScMgCoin_c` slot 18 → `0x020de5b0`** is already named `_ZN11dScMgCoin_c13OnYoshiTryEatEv kind:function(arm,size=0x8c)` in `ov006/symbols.txt`. That symbol is the load-bearing evidence for the slot-18 identification, and it is independent of everything else here.

## The key-function hazard: cleared, and it is why this shape of change is worth reading twice

Declaring a method whose body has no TU can move a class's key function, which moves — or strands — the emitted vtable. Member ordering decides it, so I read it rather than assumed it:

| class | first declared virtual | new declaration | verdict |
|---|---|---|---|
| `dScMgSmartball_c` | `virtual ~dScMgSmartball_c();` line 52 | `s32 InitResources();` line 75 | key fn **unchanged** |
| `dScMgSlot1_c` | `virtual ~dScMgSlot1_c();` line 126 | `virtual s32 Behavior();` line 128 | key fn **unchanged** |
| `dMgPsOpt_c::TouchIcon_c` | *none* — `~TouchIcon_c()` is **non-virtual** (dThIcon_c has two slots, no dtor pair) | `void Render();` | key fn **moves to Render's TU — intended** |

The third is the only move, it is the one the PR describes, and it is correct: with no virtual destructor there was no prior first-declared virtual for it to displace.

## One factual error, non-blocking, but it sits in the keystone note

`notes/dScMgBase_c-slots-18-35.md` (line ~50) says `dScMgCoin_c` "overrides exactly three" of the 36 slots. Diffing `_ZTV11dScMgBase_c` (ov004:0x020bc0c0) against `_ZTV11dScMgCoin_c` (ov006:0x0213bf50) across all 36 slots, it overrides **six**:

```
slot  0  base 0x02043c80 <unnamed>                    coin 0x020de704  _ZN11dScMgCoin_c13InitResourcesEv
slot  6  base 0x020b0618 _ZN11dScMgBase_c8BehaviorEv  coin 0x020de69c  _ZN11dScMgCoin_c8BehaviorEv
slot  9  base 0x020b04ec _ZN11dScMgBase_c6RenderEv    coin 0x020de63c  _ZN11dScMgCoin_c6RenderEv
slot 16  base 0x020b2a84 _ZN11dScMgBase_cD1Ev         coin 0x020dbe40  _ZN11dScMgCoin_cD1Ev
slot 17  base 0x020b2a18 _ZN11dScMgBase_cD0Ev         coin 0x020dbe64  _ZN11dScMgCoin_cD0Ev
slot 18  base 0x020b299c func_ov004_020b299c          coin 0x020de5b0  _ZN11dScMgCoin_c13OnYoshiTryEatEv
```

In the window **16..35** the answer *is* exactly {16, 17, 18} — so this reads as a window-scoping slip written up as "36 slots each", not a bad measurement. **Slot 18's identification is unaffected**, since it rests on the symbol name rather than on the count. But the note is billed as the input to the keystone PR, so please either say six, or say "three of slots 16..35" and name the window.

Two incidental confirmations fell out of that same diff and are worth having on the record. Slots 16/17 landing exactly on the destructor pair independently confirms the **address-point** convention the note documents — a `_ZTV*` address in `symbols.txt` *is* the address point, slot *i* at `addr + 4*i`, with no 8-byte preamble added. And the base's slot 18 being an unnamed `func_ov004_020b299c` is consistent with the note's claim that `dScMgBase_c` has not declared its 18..35 range yet.

## Gates measured on the merge tree, not the branch

`baseRefOid` here is `c2f5ba278`, which is now stale — #2080 landed since, so main is `13253537d`. Rather than ask for a rebase I built the real merge tree: `git merge-tree origin/main a698038` → tree **`2f3cbdace`**, clean.

That merge matters more than usual here. #2080 is precisely the PR that *starts checking* `include/dMgPsOpt_c.h` — it retired that header's waiver — and this PR edits the same header. A `tools/` gate change crossed with a header change is never order-independent by path inspection alone, so I extracted `2f3cbdace` and ran the **merged** checker over the **merged** header:

```
include/dMgPsOpt_c.h: 4 commented fields, 0 mismatched, 0 unparsed, struct spans 0x128    rc=0
```

The `TouchIcon_c` nested class this PR gives a `Render()` declaration to is the same one #2080 taught the offset checker to size via `sizeof(dMgPsOpt_c::TouchIcon_c) == 0x24`. The added declaration does not disturb the field walk, and the span still lands on the header's own `dMgPsOpt_c_size_must_be_0x128`.

`tools/test_check_header_offsets.py` on that extract: **31 passed, 2 failed** — both failures are `RealHistoryTests` reporting `<sha> is not in this clone`. A bare `git archive` extract has no `.git`; that is an extraction artifact, not a regression. All six of #2080's new cases pass.

## The retraction in `dScMgSlot1_c.h` is the right call

Keeping the superseded banner next to the corrected one, rather than silently rewriting it, is the convention here and I want it recorded as approved practice. The earlier text claimed `data_ov006_0213eb40` is "never compiler-emitted either — only aliased", which stopped being true the moment the destructor became this class's key function: `_ZTV12dScMgSlot1_c` *is* emitted, from `src/_ZN12dScMgSlot1_cD1Ev.cpp`, and `tools/romdata_check.py` byte-compares it. Documenting why the old reasoning expired is worth more than a clean diff.

Same for the still-open slots 18/19 mislanding the header now documents and defers to `notes/dScMgBase_c-slots-18-35.md` as one atomic unit — that is the correct scoping. It genuinely cannot be fixed locally, since `dScMgBase_c` has to declare 18..35 first, and shipping a half-fix would put two more wrong words in an emitted table.

**Verdict: clearing this to merge** once `PR validation` settles. Please fold the two prose corrections (`0x020b8c30` → `0x020bca7c`, and "exactly three" → six, or name the 16..35 window) into this branch if you reach it before it lands; as a follow-up otherwise. Neither blocks.

PR #2082, opened 2026-08-31
tu_map: a stray function must touch the TU it is absorbed into

andrewboudreau:
## The defect

`tu_map.absorb_unlabelled` attaches an unlabelled function to a cluster when every
*caller* of it lies inside that cluster's span. It never compares the absorbed
function's **own** address to the span. Its docstring called this "proof, not
preference":

> A function called ONLY from inside one cluster's span is file-local -- it has
> internal linkage, which is why it has no mangled name to read a class from --
> and internal linkage is a property that cannot cross a TU boundary. So its
> callers' cluster is its cluster.

The premise is the weak link: **one caller-cluster does not prove internal
linkage.** An ordinary external function that happens to have exactly one caller
looks identical from the call graph alone.

## What it cost, measured

| | default | `--split-swallowers` |
|---|---|---|
| units | 547 | 720 |
| units holding an out-of-span function | **37** | **59** |
| functions filed where they cannot be | **128** | **304** |

Worst case: eight functions at `0x020ccd04..0x020d0bd8` absorbed into ov006's
`dScMgTrampoline_c`, whose span is `0x021207dc..0x0212231c` — `0x54000` bytes
away. A `.text` that jumps a third of a megabyte and comes back is not a
translation unit. This is the "stale TU map overcut" hazard in a new form, and it
lands on live TU-promotion candidates: `dScMgTrampoline2_c` (16 bogus),
`dScMgFlower_c` (10), `dScMgSlot1_c`/`dScMgSlot3_c` (8), `main`'s
`Event,Fog,Message,…` unit (37).

## The fix

Adjacency is the missing half of the argument. The linker lays a unit's `.text`
down in one contiguous run, so a stray joins a cluster only when it **abuts**
the cluster — and, having joined, **extends** it, so a contiguous run of strays
is absorbed one function per pass. The call-graph rule runs first and is
unchanged: adjacency is necessary, not sufficient.

Out-of-span functions go **128 → 0** and **304 → 0**.

Unit count falls **547 → 513**. That is not the guard rejecting work: extending a
span brings callers that previously read as "outside every cluster" back inside
it, which lets adjacent strays that used to fragment into their own one-function
units join the neighbour they belong to.

## Not a regression

`tu_map.py --check`: **V1 ov063** and **V1c ov063** fail identically before and
after; every other gate passes on both. Those two were already red on `main` and
this does not claim to fix them.

`tools/test_tu_map.py` is new — 10 pure unit cases, no ROM or compiler.
**6 fail on the parent commit**; the 4 that pass are the call-graph rules this
does not touch.

## Worked example

`ov006/dScMgTrampoline_c` before: 50 functions, eight of them impossible.
After: **42 contiguous functions over `0x021207dc..0x02122490`**, starting at its
D1 destructor and ending exactly where `MgTrampolineTime_Spawn` begins — which
`tu_map` independently clusters as its own factory unit at
`0x02122490..0x021225a8`. That settled a boundary question by agreement of two
signals rather than by preference.

Tools-only, as required: the validator restores `tools/` from base, so no PR can
test its own tool change.

### DIFF
--- a/tools/test_tu_map.py
+++ b/tools/test_tu_map.py
@@ -1,0 +1,134 @@
+"""A TU's .text is contiguous, so absorb_unlabelled must not reach across a gap.
+
+The call graph alone cannot tell a file-local helper from an ordinary external
+function that happens to have one caller -- both are "called from exactly one
+cluster". Absorbing on that evidence placed 128 functions into units they cannot
+be in (37 of 547 at the default settings), the worst of them a third of a
+megabyte outside the span it was filed under.
+
+Adjacency is the missing half of the argument, and it is what these pin: the
+linker lays a unit down in one contiguous run, so a stray joins a cluster only
+when it touches that cluster -- and, having joined, extends it, which is how a
+contiguous run of strays is taken in.
+
+These are pure unit tests over synthetic clusters: no ROM, no compiler, no config.
+"""
+import pathlib
+import sys
+import unittest
+
+TOOLS = pathlib.Path(__file__).resolve().parent
+sys.path.insert(0, str(TOOLS))
+
+import tu_map as TM      # noqa: E402
+
+
+def cluster(start, end, classes=("C",)):
+    return {"classes": list(classes), "start": start, "end": end, "funcs": []}
+
+
+def unit_for(tus, addr):
+    """The unit holding `addr`, or None."""
+    for t in tus:
+        if any(a == addr for (a, _n, _s) in t["funcs"]):
+            return t
+    return None
+
+
+class AdjacencyTests(unittest.TestCase):
+    def test_a_stray_abutting_the_end_is_absorbed_and_extends_the_span(self):
+        c = cluster(0x1000, 0x1100)
+        tus = TM.absorb_unlabelled([c], [(0x1100, "helper", 0x40)],
+                                   {0x1100: {0x1020}})
+        self.assertEqual(len(tus), 1)
+        self.assertEqual(tus[0]["end"], 0x1140)
+        self.assertIn((0x1100, "helper", 0x40), tus[0]["funcs"])
+
+    def test_a_stray_abutting_the_start_is_absorbed(self):
+        c = cluster(0x1000, 0x1100)
+        tus = TM.absorb_unlabelled([c], [(0x0fc0, "helper", 0x40)],
+                                   {0x0fc0: {0x1020}})
+        self.assertEqual(tus[0]["start"], 0x0fc0)
+
+    def test_a_far_stray_is_not_absorbed_even_with_a_single_caller_cluster(self):
+        """The defect this file exists for: one caller cluster is not enough."""
+        c = cluster(0x21207dc, 0x212231c)
+        tus = TM.absorb_unlabelled([c], [(0x20ccd04, "far", 0x60)],
+                                   {0x20ccd04: {0x2120800}})
+        self.assertEqual(c["end"], 0x212231c, "span must not move")
+        self.assertNotIn((0x20ccd04, "far", 0x60), c["funcs"])
+        own = unit_for(tus, 0x20ccd04)
+        self.assertIsNotNone(own)
+        self.assertEqual(own["classes"], [], "a stray becomes its own unlabelled TU")
+
+    def test_a_contiguous_run_is_absorbed_whole(self):
+        """Each pass extends the span, so the next function in the run abuts it."""
+        c = cluster(0x1000, 0x1100)
+        strays = [(0x1100, "a", 0x40), (0x1140, "b", 0x40), (0x1180, "c", 0x40)]
+        calls = {0x1100: {0x1020}, 0x1140: {0x1020}, 0x1180: {0x1020}}
+        tus = TM.absorb_unlabelled([c], strays, calls)
+        self.assertEqual(len(tus), 1)
+        self.assertEqual(tus[0]["end"], 0x11c0)
+        self.assertEqual(len(tus[0]["funcs"]), 3)
+
+    def test_a_run_stops_at_the_gap(self):
+        """Contiguity is the boundary: what is past the hole stays out."""
+        c = cluster(0x1000, 0x1100)
+        strays = [(0x1100, "a", 0x40), (0x1200, "past_the_gap", 0x40)]
+        calls = {0x1100: {0x1020}, 0x1200: {0x1020}}
+        tus = TM.absorb_unlabelled([c], strays, calls)
+        self.assertEqual(tus[0]["end"], 0x1140)
+        self.assertIsNotNone(unit_for(tus, 0x1200))
+        self.assertNotEqual(unit_for(tus, 0x1200), tus[0])
+
+
+class CallerTests(unittest.TestCase):
+    def test_an_adjacent_stray_with_two_caller_clusters_is_not_absorbed(self):
+        """Adjacency is necessary, not sufficient -- the call-graph rule still runs."""
+        a, b = cluster(0x1000, 0x1100, ("A",)), cluster(0x1140, 0x1200, ("B",))
+        tus = TM.absorb_unlabelled([a, b], [(0x1100, "shared", 0x40)],
+                                   {0x1100: {0x1020, 0x1150}})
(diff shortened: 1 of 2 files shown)


andrewboudreau (commented):
**Gatekeeper review.** (Same account, so GitHub refuses the Approve button; this comment is the sign-off.) I reproduced every headline number independently rather than reading them out of the PR body, looked for the failure mode this shape of guard usually has, and traced the damage downstream to see whether anything already landed is contaminated. **The change is correct and I am clearing it to merge.** One ask below, and it is a real one — as submitted, the new test suite never runs in CI.

## Reproduced against a freshly built merge tree

`baseRefOid` is `d1edc757b`, two commits stale (#2081 landed, then a progress refresh). I built the merge tree against current main `a69a4c2f4` → tree **`5270c6f46`**, clean, touching exactly `tools/tu_map.py` and `tools/test_tu_map.py` and nothing else. Both blobs in that tree are byte-identical to the branch, so what I measured is what will land.

Both sides were extracted and junctioned to the real `extracted/` before running — worth stating, because a bare `git archive` extract has no `extracted/`, `tools/modules.py` then registers only the `main` module, and `tu_map` cheerfully reports `1 modules, 3068 functions -> 29 TUs`. My first run said exactly that and it was an artifact. Wired up, both trees file **11097 functions across 74 modules**.

| | main `a69a4c2f4` | merge tree | PR claims |
|---|---|---|---|
| units, default | **545** | **513** | 547 → 513 |
| units holding an out-of-span function | **36** | **0** | 37 → — |
| out-of-span functions | **123** | **0** | 128 → 0 |
| units, `--split-swallowers` | **718** | **594** | 720 → — |
| out-of-span, `--split-swallowers` | **299** | **0** | 304 → 0 |
| one-function units, default | 140 | 133 | — |
| functions filed | 11097 | 11097 | — |

The before-side numbers sit a hair under the PR's (545/123 vs 547/128) because main moved two commits since the base. **The after-side figure, 513, matches exactly.** The claim that stands or falls on its own is the right-hand column and it holds: zero out-of-span functions in both configurations, and the drop from 545 to 513 is strays no longer being swallowed into units they do not abut.

`tu_map.py --check` is **byte-identical** between the two trees — `diff` clean, rc=0 on both. `[FAIL] V1 ov063 classed-TU count == 4 got 12` and `[FAIL] V1c ov063 Boo variants use the daTrs_c factory TU` fail on both sides; `[PASS] V3 every function in exactly one TU` on both. The "not a regression" claim is exact, and the two ov063 failures are pre-existing and untouched.

`tools/test_tu_map.py`: **`Ran 10 tests in 0.000s / OK`** on the merge tree. I checked for the false-green shape this tree specialises in — no `skip`, no `_toolchain()`, no bare `return` guards; they are genuinely pure unit tests. Copied verbatim onto the main extract: **`FAILED (failures=6)`**, e.g. `AssertionError: False is not true : far at 0x800 escapes 0x1000..0x1100`. The PR's "6 fail on the parent commit" is exact, so the suite really does bracket the defect rather than describe it.

## The finding I went looking for, and did not find

The guard absorbs a stray and *extends the cluster span* to cover it. The obvious hazard is that an extended span now overlaps a neighbouring cluster — the old code could only mis-file a function, the new code could corrupt the interval arithmetic every consumer depends on. So I counted overlapping unit spans on both sides: **11 before, 11 after**, in both the default and `--split-swallowers` configurations. Unchanged.

And the surviving 11 are not overlaps at all. Inspecting the two on `main`: `__cxa_vec_cleanup` and `__destroy_arr` are two symbol *aliases* at one address, `0x207328c`. Pre-existing, unrelated to this PR, and not worth fixing here — but I want it on the record that the number is 11 and not 0, so the next person who counts them does not think this PR introduced them.

## Why this is worth landing rather than filing

The argument is `tools/tu_config.py` on main, which the PR does not mention:

```python
def attribute_text(att, tus):
    for i, (lo, hi, _c) in enumerate(tus):
        att.claim(".text", i, lo, hi)
```

Its own header comment says every other section is attributed "only where the ROM says so, by four rules that each fail closed" — but `.text` is claimed **unconditionally** from tu_map's spans, on the reasoning that the linker never interleaves two objects. That reasoning is sound; the spans feeding it were not. So the 123 misfiled functions were producing wrong `.text` intervals in a consumer that has no fail-closed path to catch them. This is the difference between a cosmetic count and a wrong number reaching a generator.

The design is also the right one. A single caller-cluster is *not* proof of internal linkage — an ordinary external function with exactly one caller looks identical from the call graph alone. Adding adjacency as a second, independent witness (the linker lays a TU's `.text` down contiguously, so a member must touch its span) is a genuine constraint, not a tightened threshold. The iteration is necessary and the docstring says why: absorbing extends the span, so a contiguous run of strays comes in one function per pass.

## No landed TU is contaminated — and the guard mechanises a call a reviewer was already making by hand

This is the check that decided it for me. I enumerated all 36 affected units on main by class name to see whether the defect had already reached something merged. It had:

```
ov004  0x20ae03c..0x20b2c58  14 bad / 109 fn  classes=['dScMgBase_c']
ov006  0x20d95a4..0x20dbd54   1 bad /  35 fn  classes=['dScMgCard_c']      <- landed as #2079
ov006  0x210a8c0..0x210d6b8   8 bad /  49 fn  classes=['dScMgSlot1_c','dScMgSlot3_c']
ov006  0x21207dc..0x212231c   9 bad /  50 fn  classes=['dScMgTrampoline_c']
ov006  0x21225ac..0x21245a8  16 bad /  55 fn  classes=['dScMgTrampoline2_c']
ov006  0x212a554..0x212b7f8  10 bad /  19 fn  classes=['dScMgFlower_c']
main   0x20049f0..0x20707a4   9 bad / 2984 fn
```

`dScMgCard_c` is a merged TU, so I resolved its one out-of-span function:

```
OUT OF SPAN: 0x20c0f0c func_ov006_020c0f0c
```

That is **precisely the function #2079's manifest excludes by hand** — my own review of #2079 recorded "the `func_ov006_020c0f0c` exclusion" among its four substantive notes, and that the manifest declared 34/34 symbols where tu_map's unit held 35. So: the landed TU is clean, the author caught the bad member by eye, and this PR turns that manual catch into a mechanical one. That is the strongest evidence available that the guard agrees with human judgement on a case decided before the guard existed.

Consumer audit, for completeness: `tubuild.py`, `tu_config.py`, `tu_names.py` and `sinit_owners.py` all read `build/tu_map.json`, which is a gitignored artifact; `check_dead_references.py` only mentions it in prose. No committed state goes stale on merge and no byte gate consumes it, so there is nothing to regenerate and no ROM risk. Two `tools/` files, no `config/`, no `src/`, no headers.

## The ask: the new tests never run in CI

`.github/workflows/tool-tests.yml` enumerates its unittest modules explicitly and deliberately — its own comment explains that a glob "silently absorbs a future toolchain-dependent test and turns the green hollow." **`tools.test_tu_map` is not in that list.** As submitted, the 10 tests I verified pass here would never execute on any future PR, and the defect they pin could come back green.

The fix is one line in the run list plus an entry in the workflow's per-module inventory comment block. The suite is toolchain-free — no ROM, no compiler, 0.000s — so it is exactly the kind that belongs there, and adding it is consistent with the no-glob convention rather than an exception to it.

**Verdict: approved, merging on that one addition** (or immediately as a follow-up if you would rather not re-trigger validation — say which and I will land it either way). Everything else here reproduces to the number.

andrewboudreau (commented):
Merged. The one ask is landed rather than left to you: **#2083** adds `tools.test_tu_map` to `.github/workflows/tool-tests.yml` (run list + inventory entry), so your 10 tests now actually execute on every `tools/**` PR.

While wiring it I ran the mutation check the wiring is for, and it is worth having on this PR rather than only on that one. With `tools/tu_map.py` reverted to `d1edc757b` and nothing else changed, the command that job runs reports `Ran 514 tests / FAILED (failures=6, skipped=3)`. On main before #2083 the same revert is silently green — so the guard you added is now defended by CI and not only by review.

Cost measured with the run step extracted from the YAML and executed: `Ran 504 / 53.990s` before, `Ran 514 / 55.583s` after, no new skips.

PR #2084, opened 2026-08-31
Promote ov070/daBrq_c as a genuine C++ translation unit

andrewboudreau:
Renames the RTTI-proven Amp class to daBrq_c and replaces 18 per-symbol sources with one compiler-spelled class TU. mwccarm now owns D1/D0, RTTI, the vtable, and SpawnInfo; the exact C factory remains separate because the natural placement-new form does not reproduce the ROM. Local and independent verification: 18/18 exact and relocation-clean, four owned data objects verified with zero differences, Amp_Spawn linkcheck VERIFIED blind: 0, 11,088/11,088 source functions, 106/106 modules, and a stock-identical ROM. Static merge-result gates pass. Attribution has the expected 18 structural losses for TU consolidation; please use attribution-override.


andrewboudreau (commented):
Merged. Reviewed and reproduced independently against the **merge tree**, not the branch — the branch is four commits behind `main` (`#2082`, `#2083` and two progress refreshes), so what I gated is `git merge-tree origin/main 7402d8be8434` = tree `b633153b19e4`, which is the 31 files this PR actually lands. Clean merge, no conflicts.

## The class name is ROM-proven, and I checked it in the cartridge rather than in config

The one claim a TU promotion cannot recover from is a wrong class name, so I went to the image instead of to `symbols.txt`:

```
overlay_0070.bin @ 0x423c:  37 64 61 42 72 71 5f 63 00
                            '7' d  a  B  r  q  _  c  \0
```

That is a length-prefixed mwcc RTTI type name — `7daBrq_c` — and `daBrq_c` occurs exactly once in the overlay. `Amp` occurs **zero** times. So the rename is not a coining, it is a correction, and the eight-way `_ZTS7daBrq_c` / `_ZTI7daBrq_c` / `_ZTV7daBrq_c` naming in `symbols.txt` was already sitting next to code symbols that said `Amp`. This PR closes that split.

It also drops the real defect that split created: `_ZTV3Amp` and `_ZTV7daBrq_c` were **both** declared at `0x02123278`, two names for one word. One of them is now gone.

## The `.data` claim is the strongest evidence here, and it is arithmetic

`106/106 exact` is blind to vtables — it compares module bytes, and a vtable that no source claims is simply never compared. What makes this one different is the new stanza:

```
src/actors/daBrq_c.cpp:
    complete
    .text start:0x02120570 end:0x021210ac
    .data start:0x0212323c end:0x021232f4
```

`0x021232f4 - 0x0212323c = 0xb8 = 184` bytes, covering `_ZTS7daBrq_c`, `_ZTI7daBrq_c`, `Amp_SpawnInfo` **and `_ZTV7daBrq_c`**. There are exactly two `.data start:` claims in the whole tree — this one and ov047's `0x02112258..0x02112320` = 200 bytes — and my own build reports

```
source-owned data claims: 2  (reproducing 2, mismatching 0)
... 384 source-owned data ...
```

`200 + 184 = 384`. So the vtable, the typeinfo and the type-name string were compiled by mwccarm and compared to the cartridge byte-for-byte. That is a real vtable gate, not an inference from module fidelity.

## The rest of the reproduction

Full `rombuild.py -j 16` on the merge tree in a wired worktree:

```
  reproducing: 11,088     mismatching: 0
ROM-build analysis: PASS
```

Static gates on the merge result, both sides:

```
source-coverage     pass     pass     ok   (entries 10830 -> 10814, -16)
```

The delink consolidation is exhaustive, which I checked by walking the ends: `570 -> 5d0 -> 644 -> 724 -> 8a4 -> 910 -> 9e4 -> bf8 -> cac -> ce4 -> d34 -> d70 -> da8 -> dc4 -> e20 -> e24 -> e8c -> eec -> 10ac`. Every deleted stanza's `end` is the next one's `start`, and the union is exactly the single new `.text` range. Nothing fell out of the enrolled set, which is what the validator's `+0` on enrolled ranges says independently.

Two things in the validator report that look like losses and are not:

- **Byte-verified `-17`.** This is the known undercount for a promoted TU: the report credits the first symbol of a multi-symbol source and none of the other seventeen. Enrolled ranges moved `+0` and my build compiled and compared all 18, so the `-17` is bookkeeping, not coverage.
- **`18 changed, 0 lost` attribution.** Attribution is a stated non-goal here; nothing is lost, and the consolidation is the cause.

`include/Amp.h` is deleted safely: all eight files that included it on `main` are deleted by this same PR.

## One finding, not a blocker — and I will land the fix

`include/decl_common.h:696` still carries

```c
extern int _ZTV3Amp[];
```

`_ZTV3Amp` no longer exists in any `symbols.txt` after this PR, and that declaration is now its **only** occurrence anywhere in the tree outside `build/` and the historical `notes/data/` tables. It cannot break the link — an unreferenced declaration emits nothing, which is why my build is green — but it is a name that resolves to nothing, sitting in the header every C file reaches. `decl_common.h:740` already declares the live `_ZTV7daBrq_c` right below it.

Not worth holding a fully verified TU promotion for one dead line, so this is merged and I will land the removal as a follow-up rather than hand it back.

## What I liked

The header and `actor_renames.tsv` both say plainly which spellings are proven and which are inferred — `SetState`'s signed parameter is called out as indistinguishable from other 32-bit forms, and `Amp_Spawn` / `Amp_SpawnInfo` are labelled evidence-bounded aliases in three separate places instead of being quietly renamed to match. The `1076` to `0x434` change in the factory makes it line up with the size assert by eye. That is the right shape for this work.

PR #2086, opened 2026-08-31
Declare dScMgBase_c slot 18 (OnYoshiTryEat) across the minigame family

andrewboudreau:
## What this is

The first of eighteen. `include/dScMgBase_c.h` said slots 18-35 *"are eighteen further
virtuals new at this class; their signatures are not reconstructed yet, so they stay
undeclared"* — and that one sentence is what caps every minigame vtable at 18 of the
cartridge's 36 slots. A byte-exact prefix, but not ownable. This declares slot 18,
`OnYoshiTryEat`, on the base and on all 30 descendants that override it.

## Why it is not one 158-change commit

My earlier note said this had to land atomically. That was wrong, and the correction is
the point of this PR.

mwcc emits a vtable only as long as the slots it has been told about. Declaring **all
eighteen** base slots at once widens every emitted table from 18 slots to 36 and writes
the *base's* body into every slot a descendant has not yet declared an override for —
PARTIAL becomes DIFFERS, real wrong bytes, across the whole family.

Declaring **one** base slot widens every table by exactly one word. After declaring
slots 18..k a table is k+1 slots long, and for every i ≤ k that word is right if the
class either declares its own override or genuinely inherits the base body. Slots above
k are simply not emitted — still a prefix, never a disagreement. So the atomic unit is
one base slot across the whole family, and there are eighteen of these, not one.

## The signature is measured, not copied from `dActor_c.h`

`include/dActor_c.h:131` declares `virtual int OnYoshiTryEat();`. `dActor_c` sits on the
*other* branch of `fBase_c` — it hangs off `dBase_c` directly, where this family reaches
`dBase_c` through `dScene_c`. The two are parallel extensions that share indices and
thirteen names, which makes that header an excellent **naming** hint and no kind of
signature authority. Here it is wrong:

- **Arity is `(int)`.** 13 of the 24 independently decompiled descendant overrides read
  r1 and branch on it — `dScMgTeresa_c` takes an entirely different path when it is 0.
  Twelve unrelated classes do not read a garbage register by coincidence. The base's own
  body ignores it, and that proves nothing either way: an unused argument is never read.
  Only an override that actually reads a parameter is evidence, and it gives a **lower**
  bound on the arity.
- **Return type is `int`, not `void`.** All 24 free-function bodies are written `void`,
  which is what makes this a trap. But `dScMgCoin_c::OnYoshiTryEat` is a real *member*
  definition ending `return 0;` — declaring `void` would have changed its bytes. The
  return type is not mangled, so the 24 free functions are unaffected either way.

`config/arm9/overlays/ov006/symbols.txt` already named Coin's override
`_ZN11dScMgCoin_c13OnYoshiTryEatEv` — **wrong arity**, corrected here to `...Ei` along
with its source file, its delinks entry and its member definition.

The *name* is inherited rather than proven. Unlike slots 19-30, slot 18 carries no
`recovered name:` comment on either side; it comes from `dActor_c.h` plus the
already-landed `symbols.txt` entry for Coin. Only the signature is measured, and the
header and the base body both say so in as many words.

## Declaration order IS the index

`dScMgSlot1_c` already declared `OnHitByMegaChar /* slot 27 */` and
`OnHitFromUnderneath /* slot 28 */`; `dScMgAmida_c` declared `Unk36 /* slot 36 */`. With
the base declaring nothing above 17, those comments were aspirational — the declarations
actually sat at indices 18/19 and 18. **That is exactly why those two were the last
DIFFERS vtables left after #2081.**

So each new override is inserted *before* any pre-existing slot ≥ 18 declaration, and
*after* the class's first declared virtual, which keeps the key function (and therefore
the emitting TU) where it is. Slot 18 moves Slot1's and Amida's wrongness up one index
without changing its size — 4 bytes and 8 bytes, same as before. They clear once the
intervening slots land.

## Measured

Against `origin/main@e39f3ebac`, both sides built and measured, not inferred:

```
                          base      HEAD     delta
source-built functions   11,088    11,088       +0    (reproducing, mismatching 0)
module fidelity          106/106   106/106            100.000000% of compared bytes
ROM-build analysis        PASS      PASS
ROM data verified           464       464       +0
ROM data partial            253       253       +0
ROM data differs              7         7       +0
verified bytes           35,288    35,288       +0
partial bytes            10,016    10,136     +120
```

+120 bytes is **30 emitted vtables each one word longer, every one byte-exact.** Diffing
the two `romdata_check --json` runs symbol by symbol: no DIFFERS introduced, none
cleared, and no change in the differing byte count of any of the seven that remain.
Those seven are the same seven as on base and are unrelated to this family, except for
Amida and Slot1 above.

(The absolute figures moved under me while this was in review — #2084 landed a `.data`
claim that lifted base `partialBytes` from 9,980 to 10,016. The **delta is unchanged**
at +120 / 30 tables, which is why it is quoted against a freshly built base rather than
an earlier run.)

One number is worth reading twice. Applying the descendants *without* the base
declaration also builds green, at +108 bytes / 27 tables — because with the base silent
above slot 17, each descendant's declaration introduces its own brand-new virtual that
lands at the same index, pointing at the same body. Byte-identical and semantically
wrong, and no gate in the tree can tell the difference. The three extra tables that
appear once the base declares the slot are the classes with **no** override of their
own; their bytes matching is the independent confirmation that they really do inherit.

## What it unlocks

Slots 19-35, one PR each, same recipe — and they are much smaller than this one.
Descendant overrides per slot, measured: 19→11, 20→4, 21→4, **22→0**, 23→3, 24→6,
25→7, 26→19, 27→6, 28→6, 29→6, 30→6, 31→7, 32→1, 33→19, 34→4, 35→1. Slot 18 was the
outlier at 30; the median slot touches six classes, and slot 22 touches none — base
declaration and base rename only. The full table is in the notes.

At the end of that the minigame family emits full
36-slot vtables and can move from the `deadstrip-data` licence to
`production_mode: "intact-object"` ownership of its `_ZTV`/`_ZTI`/`_ZTS`.

Notes updated in place: `notes/dScMgBase_c-slots-18-35.md` (the "cannot be staged"
section and the atomicity hazard were both superseded) and
`notes/minigame-provenance.md`.


andrewboudreau (commented):
Merged. Reviewed against the **merge tree** rather than the branch — `git merge-tree origin/main <head>` = tree `66e00b697379`, commit `da899e7fa`, which is what this PR actually lands. Clean merge.

This is a keystone: sixteen more slot PRs are meant to stack on it, so I did not take the green build as proof of anything and went to the cartridge for each claim separately.

## The failure mode this PR could have had, and does not

A green build cannot tell "the base declaration landed" from "only the descendants declared it". If each descendant independently declared `virtual int OnYoshiTryEat(int);`, every one would start a *brand-new* virtual at index 18 pointing at the same body — byte-identical output, semantically wrong, and invisible to `rombuild`, to `106/106`, and to every static gate.

The only thing that separates them is how many vtables widen. `romdata_check.py --json`:

```
main   partialBytes 10016
#2086  partialBytes 10136          +120 bytes = 30 tables
```

30 tables is the whole family. The base-less failure mode would have widened only the 27 tables that carry an override, for +108. So the base declaration is real and it took effect. Everything else in the report is byte-identical — `differs` stays at 7, the `differing` list is unchanged, `verifiedBytes` unchanged.

That is also the check every stacked slot PR needs, and the number to compare against is the class count for that slot, not a fixed one.

## Slot 18 is the right slot, verified in the image

`_ZTV11dScMgBase_c` is at `0x020bc0c0` in `overlay_0004.bin`:

```
+0x40  slot 16  0x020b2a84 = _ZN11dScMgBase_cD1Ev
+0x44  slot 17  0x020b2a18 = _ZN11dScMgBase_cD0Ev
+0x48  slot 18  0x020b299c   <- the address this PR names
+0x4c  slot 19  0x020b2994
```

The dtor pair landing on 16/17 pins the convention (`slot N` = the word at the `symbols.txt` address + `4*N`; the address in `symbols.txt` is the address point, so offset-to-top and typeinfo sit at `-8` and `-4`). It cross-checks against the eight slots the header already declared: `1, 2, 5, 6, 7, 9, 10, 12` all hold **ov004** addresses — real overrides at this class — while every undeclared slot holds an **arm9** address inherited from `dScene_c`. Slot 18 holds an ov004 address. Correct.

## The arity claim is right, and understated — but the evidence you used is the weak direction

The PR argues arity from the callee side: "13 of the 24 independently decompiled descendant overrides read r1". My own sweep of all 30 override bodies makes it **15 of 30** (`cmp r1, #0/#1/#3/#4/#5/#9/#0x10/#0x12`, `mov r4/r5/r8, r1`). Those agree — you counted only the free-function bodies and excluded the two already-promoted TU classes, `dScMgBSC_c` and `dScMgJump_c`. Correction upward, not a defect.

But a callee reading r1 is not proof by itself; a function can read a register that holds garbage. The **caller** side is the sound direction, and it is unanimous. Of the 22 virtual call sites through `+0x48` in ov004 and ov006:

```
minigame-family slot-18 call sites : 22
  set r1 before the call           : 21
  consume the returned r0          :  0
```

`mvn r1, #0` (that is, `-1`) at `0x020dbd04`, `0x020df254`, `0x020e9a38`, `0x020ea114`; `mov r1, #3` at `0x020e050c`. Callers deliberately set the argument. That is much better evidence for `OnYoshiTryEatEi` than the body-side count, and I would lead with it in the stacked PRs.

Independent confirmation from the symbol table, which I had not expected: the two hierarchies are spelled differently in the ROM's own mangling. `dActor_c` and its descendants carry `...13OnYoshiTryEatEv` — no parameter — on `Coin`, `Player`, `PowerStar`, `Fireball`, `BookShot`, `daObjMarioCap_c` and 30 more. The minigame scenes carry `...13OnYoshiTryEatEi`. So the header comment's central argument — parallel extensions that share slot indices without sharing signatures — is not just plausible, it is written into the mangled names. Worth quoting in the comment, because it settles the `dActor_c.h` discrepancy outright rather than noting it as unexplained.

## One measured overclaim, not a blocker

The PR states:

> `dScMgCoin_c::OnYoshiTryEat` is a real member definition ending `return 0;`, so declaring `void` would have changed its bytes.

It would not. The trailing `mov r0, #0` at `0x020de628` is required by the `self->unk_51c8 = 0;` store on the next instruction; r0 merely survives to `bx lr`. I ran it rather than arguing it — patched both headers to `virtual void OnYoshiTryEat(int arg);`, changed the definition to `void`, deleted the `return 0;`:

```
as shipped (int, return 0) : (True, '2004/b56')
void, return 0 removed     : (True, '2004/b56')
```

Both byte-match the cartridge. And the wider evidence agrees that the return type is simply not in the image:

- **0 of 31** `...EatEi` override bodies sets an explicit return value before returning.
- **0 of 22** call sites consumes the returned r0. Two return immediately after the call, which passes r0 through, but a void caller of a void callee looks identical.

So `int` is a *modelling choice*, not a measurement — same shape as "a veneer's bytes cannot evidence a name", except here it is the return type. Keep `int`: it matches the `dActor_c` twin, it leaves descendants free to return a value, and nothing contradicts it. But the header comment should say inferred, not proven, because sixteen stacked PRs are going to inherit that sentence. I will land the comment correction as a follow-up rather than hand back a fully verified PR for one sentence.

## Second finding: a twin divergence

`src/_ZN11dScMgBase_c13OnYoshiTryEatEi.c` declares the flat-C body as

```c
void _ZN11dScMgBase_c13OnYoshiTryEatEi(struct dScMgBase_c *self, int arg)
```

while `include/dScMgBase_c.h` declares the C++ side `int`. Byte-correct either way — the ROM body is a lone `bx lr` — but the two halves of the twin now disagree in text with nothing recording it. This is exactly the region `check_header_offsets` cannot reach yet (its flat-C `#else` half is unscoped; that is my open task). Not a blocker, and it resolves itself whichever way the sentence above is settled, but it should be settled deliberately rather than left as drift.

## The rest of the reproduction

```
tools/rombuild.py -j 16   ->  11,088 reproducing / 0 mismatching
                              source-owned data claims: 2 (reproducing 2)
                              ROM-build analysis: PASS
tools/port_refcheck.py    ->  405 references, 0 stale
```

Static gates on the merge result, both sides:

```
```

The 25 delink and symbol renames are consistent, including the arity fix `_ZN11dScMgCoin_c13OnYoshiTryEatEv.cpp` -> `...Ei.cpp`.

**One caution if you re-run my numbers.** Overlays overlap in address space — three overlays cover `0x020f8c68` — so resolving an address to bytes by scanning `overlays.yaml` for the first range that contains it silently returns the wrong overlay's code. I got three phantom "returns a value" hits that way before catching it. Attribution has to come from the `config/arm9/overlays/ovNNN/` path the symbol was read from.

## What I liked

Declaring one slot at a time, and saying plainly in the comment why: declaring all eighteen at once would write the base body into every undeclared descendant slot and turn PARTIAL into DIFFERS family-wide. That is the correct order of operations and it is the reason this PR is safely stackable. The comment also flags the name as inherited from `dActor_c.h` rather than independently proven, and flags that `dActor_c.h` declares it with no parameter — labelling the contradiction instead of quietly resolving it. That is the right instinct, and the mangled-name evidence above now closes it.

PR #2087, opened 2026-08-31
Declare dScMgBase_c slot 19 (OnTurnIntoEgg) across the minigame family

andrewboudreau:
Second of eighteen. Stacked on #2086 (slot 18) — review that one first; this
diff is only the slot-19 half.

`dScMgBase_c` still declares nothing above vtable slot 17, so every emitted
table in the minigame family stops short of the cartridge's. #2086 added slot
18. This adds **slot 19** on the base and on the **eleven** descendants that
override it, and renames the twelve ROM bodies to their mangled C++ names.

## The signature is measured, and `dActor_c.h` is wrong here too

`include/dActor_c.h:132` declares slot 19 as taking a `Player &`. It does not.
Two of the eleven overrides read `r1`, and both **compare** it against small
integer constants rather than dereferencing it:

- `dScMgJump_c` — `if (sel == 0)`
- `dScMgBSC_c` — `if (mode == 4) ... else if (mode == 5)`

Comparing a reference against 4 and 5 is meaningless, so the parameter is an
`int`. This is worth naming as a general rule, because it is the second time
`dActor_c.h` has been wrong at a slot in this range (it declared *no*
parameter at slot 18): **a word in `r1` that could be either an `int` or a
reference occupies one register either way, so codegen alone cannot separate
`Ei` from `ER6Player`. Dereference-versus-compare is the discriminator.**

Return type is `int`: the ov004 base body ends `return 1;` and all eleven
overrides return a value.

Unlike slot 18, the **name is independently recovered**. The ov004 base body,
`dScMgJump_c` and `dScMgBSC_c` each carry their own
`// recovered name: <class>_OnTurnIntoEgg` comment in their legacy source, so
`OnTurnIntoEgg` does not rest on `dActor_c.h` at all.

## Measured against the slot-18 parent

```
                                   #2086     this PR    delta
source-built functions            11,088     11,088       +0
  reproducing                     11,088     11,088
  mismatching                          0          0
module fidelity                  106/106    106/106       100.000000%
ROM-build analysis                  PASS       PASS
ROM data verified                    464        465        +1
ROM data partial                     253        252        -1
ROM data differs                       7          7        +0
verifiedBytes                     35,288     35,368       +80
partialBytes                      10,136     10,180       +44
```

No DIFFERS introduced, none cleared, none changed size. The seven that remain
are the same seven: `_ZTIN8daDemo_c10anmModel_cE`, `_ZTIN8daDemo_c13simpleModel_cE`,
`_ZTV10dCcAcPos_c`, `_ZTV12dScMgAmida_c` (4), `_ZTV12dScMgSlot1_c` (8),
`_ZTV18PoppingLavaBubbles`, `_ZTV7daKrb_c`.

## One vtable symbol is now complete (and what that does NOT mean)

`partialBytes` moves `+44`, not `+120`, and that difference is the point:

```
10,136  +  30x4  -  76  =  10,180
           ^^^^     ^^
           30 tables one    a 76-byte table that LEFT the
           word wider       partial pool entirely
```

and the departed 76 bytes reappear as **+80 verified**. The symbol is
`_ZTV14dScMgD3DBase_c` (ov006, `0x0213c62c`), and this is a direct per-symbol
readout on both commits, not an inference from the totals:

```
parent (#2086):  PARTIAL   _ZTV14dScMgD3DBase_c  ov006 0x0213c62c  76 bytes
this PR:         VERIFIED  _ZTV14dScMgD3DBase_c  ov006 0x0213c62c  80 bytes
```

`dScMgD3DBase_c`'s `_ZTV` symbol is 20 slots long, and slot 19 was the last
one it needed, so the build now reproduces **every byte the configuration asks
it to** for that symbol.

**That is not the same as the class's vtable being complete, and an earlier
revision of this description said it was.** Correcting it, because the
distinction is one a reviewer of the next sixteen PRs needs: `VERIFIED` from
`romdata_check` means the emitted object matches the ROM across the extent of
the symbol *as `symbols.txt` defines it*, and for a vtable that extent is
wherever dsd happened to cut. Read straight through the boundary and this table
plainly does not stop:

```
  slot 19  +0x4c  0x020b2994
  slot 20  +0x50  0x020b2990   <-- _ZTV14dScMgD3DBase_c ENDS (80 bytes)
                               <-- data_ov006_0213c67c begins (64 bytes)
  slot 21  +0x54  0x020b298c
  ...
  slot 35  +0x8c  0x020ad660
  slot 36  +0x90  0x6f6d6168   <-- "hamo", a string: the real end
```

The class's vtable is **36 slots / 144 bytes** at `0x0213c62c`, split by the
config into `_ZTV14dScMgD3DBase_c` (80 bytes, slots 0-19) plus
`data_ov006_0213c67c` (64 bytes, slots 20-35). Every one of those sixteen
addresses is a `dScMgBase_c` slot this campaign has yet to declare -- `0x020b298c`
is slot 21, `0x020ae198` slot 22, and so on down to `0x020ad660` at slot 35.
So the honest claim is the smaller one: the first minigame vtable **symbol**
reproduced in full. The first complete class vtable arrives at slot 35, and it
is the same table.

## Why one slot at a time

Repeated from #2086 because it is the thing that makes this campaign safe.
Declaring all eighteen base slots at once widens every emitted table from 18
slots to 36, and any descendant that does not simultaneously declare its own
override gets the **base's** body written where the cartridge holds its own —
PARTIAL becomes DIFFERS, real wrong bytes. Declaring one base slot widens the
tables by exactly one word, so slot *i* together with every descendant
override of slot *i* is a self-contained, byte-verifiable change. Eighteen
reviewable PRs, not one 158-file commit.

**Read the delta arithmetic twice** if you review a later slot in this series.
`+30 tables x 4 bytes` is the number *with* the base declaration. Without it,
each descendant's own `virtual int OnTurnIntoEgg(int);` introduces a brand-new
virtual at index 19 *in that class*, pointing at the same body — byte-identical
output, semantically wrong, and invisible to 106/106, to PASS, and to the
DIFFERS count. The only tell is the table **count**: 27 widened tables means the
base is silent, 30 means it is declared. The three-table difference is the
classes with no override of their own, and their bytes matching is the
independent proof they genuinely inherit.

## What is in the diff

- `include/dScMgBase_c.h` — slot 19 declared, after slot 18 and after the
  first declared virtual (`~dScMgBase_c`), so the key function does not move.
- `src/_ZN11dScMgBase_c13OnTurnIntoEggEi.c` — base body, renamed from
  `src/func_ov004_020b2994.c`.
- 7 standalone renames `src/func_ov006_*` → `src/_ZN<len><Class>13OnTurnIntoEggEi.*`
  (Jump2, Jump, Memory2, Memory, Roulette, Trampoline2, Trampoline).
- 4 in-place identifier renames inside already-intact TUs
  (`dScMgBSC_c`, `dScMgCard_c`, `dScMgMCarlo_c`, `dScMgMCarlo2_c`).
- 11 descendant headers gain the declaration immediately after their slot-18 one.
- `config/arm9/overlays/ov00{4,6}/{symbols,delinks}.txt`.

## Next

Slots 20-35, same recipe, census in `notes/dScMgBase_c-slots-18-35.md`.
**Slot 22 has zero descendant overrides** — a one-line base declaration and
nothing else — so it is the cheapest possible confirmation that the recipe is
sound if you want one before the larger slots.


andrewboudreau:
Correcting something I wrote in the description above, since it would have misled a reviewer of the next sixteen PRs.

I claimed `_ZTV14dScMgD3DBase_c` going PARTIAL -> VERIFIED made it "the first minigame vtable the build reproduces in full from source." It is the first vtable **symbol**. The class's vtable is not complete.

`VERIFIED` from `romdata_check` means the emitted object matches the ROM across the extent of the symbol *as `config/.../symbols.txt` defines it*. For a vtable that extent is wherever dsd happened to cut, and here it cut at slot 20. Reading straight through the boundary at `0x0213c62c`:

```
  slot 19  +0x4c  0x020b2994
  slot 20  +0x50  0x020b2990   <-- _ZTV14dScMgD3DBase_c ENDS (80 bytes)
                               <-- data_ov006_0213c67c begins (64 bytes)
  slot 21  +0x54  0x020b298c
  ...
  slot 35  +0x8c  0x020ad660
  slot 36  +0x90  0x6f6d6168   <-- "hamo", a string: the real end
```

So the real table is **36 slots / 144 bytes**, configured as `_ZTV14dScMgD3DBase_c` (slots 0-19) + `data_ov006_0213c67c` (slots 20-35). Every one of those sixteen addresses is a `dScMgBase_c` slot this campaign has yet to declare. The first *complete class* vtable arrives at slot 35, and it is this same table.

No code changed -- description only. The measurement and the diff are unaffected.

andrewboudreau (commented):
Merged. Reviewed against the **merge tree**, not the branch — `git merge-tree origin/main@ac19524c809c cbe7fb9442da`, which is what actually lands. Every ROM-level claim in the description was re-derived from the cartridge independently, and all of them hold. Two things came out of it that the PR did not have, and one config defect that this PR is the first to expose.

## The metric in your table stops working at this slot — use the bucket-invariant sum

This is the important part for the sixteen PRs behind this one, so it goes first.

On #2086 the discriminator was `partialBytes`: a green build cannot tell "the base declaration landed" from "only the descendants declared it", and the only thing that separates them is *how many* vtables widen. At slot 18 that read cleanly, +120 = 30 tables = the whole family.

At slot 19 it does not, and it fails in the most dangerous possible direction:

```
                     main       merge      delta
verified              464         465        +1
partial               253         252        -1
verifiedBytes       35288       35368       +80
partialBytes        10136       10180       +44   <- your table
verified+partial    45424       45548      +124   <- the real widening
```

`+44` is **numerically identical** to the descendants-only failure signature (11 overrides × 4 bytes). Had I quoted `partialBytes` alone, as #2086 taught me to, I would have posted a false CHANGES REQUESTED on a correct PR.

The cause is that one table reached full ROM width and graduated PARTIAL → VERIFIED, taking its *entire* byte count out of the partial bucket with it. `verifiedBytes + partialBytes` is invariant under a symbol moving between buckets, so that is the number to quote from here on. **#2088 and #2089 should report it, not `partialBytes`.**

## The decomposition closes exactly, per symbol

I dumped `romdata_check`'s best-verdict table per symbol on the merge tree rather than inferring from aggregates. The family is 33 classes, all 33 with a `_ZTV` in config:

```
_ZTV*dScMg*  in a byte bucket : 31   (30 PARTIAL + 1 VERIFIED)
             DIFFERS          :  2   (dScMgAmida_c, dScMgSlot1_c -- bytes in neither bucket)
             total            : 33
```

31 tables × 4 bytes = **+124**, which is the invariant delta to the byte. The graduating symbol is `_ZTV14dScMgD3DBase_c` at 80 bytes, so `verifiedBytes` +80 and `partialBytes` 124 − 80 = **+44**. Every number in your report is now individually accounted for.

Against the failure mode: descendants-only would have widened the 11 override-carrying tables, +44 on the invariant sum. Measured +124. The base declaration is real and it took effect.

## Return type: measured here, and this is the sharp contrast with slot 18

On #2086 I *disproved* the return-type claim by experiment — a `void` variant still byte-matched, 0 of 31 bodies set r0, 0 of 22 call sites consumed it, and I said the header comment should read inferred rather than proven.

Slot 19 is the opposite, and the caller side settles it:

```
minigame-family slot-19 call sites (ov004) : 14   (ov006 has none)
  set r1 before the call                   : 14
  do `cmp r0, #0` immediately after         : 14
```

Every caller sets the argument, and every caller consumes the result. Combined with the base body being literally `mov r0, #1 ; bx lr`, `int OnTurnIntoEgg(int)` is measured, not modelled. Worth saying so explicitly in the header comment, precisely because the slot-18 sentence beside it is not.

## Arity: your census is exactly right, and the method matters

Confirmed 2 of 11 — `dScMgBSC_c` (`mov r5, r1` then `cmp r5, #4` / `cmp r5, #5`) and `dScMgJump_c` (`cmp r1, #0`), the two classes you name, against the constants you quote. Comparing a reference against 4 and 5 is meaningless, so `Ei`, not `ER6Player`. `dActor_c.h:132` is wrong here, as you say.

One caution for anyone reproducing it: **a naive "does the body touch r1" scan gives a badly wrong answer.** r1 is scratch and gets reloaded mid-function; without def-use tracking I got a pile of false `ldr rX, [r1]` hits that would have contradicted the PR. The scan has to carry a live set seeded with `{1}`, follow `mov rD, rS` copies into it, kill a register on write, and clear r0–r3 on any `bl`. With that, the answer is exactly the eleven-way split you describe.

## Finding: this PR is the first to expose a config defect, and it inflates a ratchet metric

`_ZTV14dScMgD3DBase_c` graduating to VERIFIED is **premature**, and not because of anything you did.

Its recorded extent is `0x50` — 20 words — because the next symbol in `config/arm9/overlays/ov006/symbols.txt` is `data_ov006_0213c67c`, sitting `0x50` past it. The cartridge's table plainly continues past that point:

```
_ZTV14dScMgD3DBase_c @0x0213c62c      _ZTV11dScMgBase_c @0x020bc0c0 (ov004)
  slot 18  0x020b299c                   slot 18  0x020b299c
  slot 19  0x020b2994                   slot 19  0x020b2994
  slot 20  0x020b2990  <- "data"         slot 20  0x020b2990
  slot 21  0x020b298c                   slot 21  0x020b298c
  slot 22  0x020ae198                   slot 22  0x020ae198
  slot 23  0x020ae1a0                   slot 23  0x020ae1a0
```

Slots 20–23 are byte-identical to the base's, inherited unchanged, and `0x020b2990` / `0x020b298c` are `func_ov004_020b2990` and `func_ov004_020b298c` — named 4-byte function bodies adjacent to slot 19's. That is a vtable, not a data object. `data_ov006_0213c67c` is marked `ambiguous`, is delinked nowhere, and is referenced by nothing else in the tree; it is a dsd auto-boundary that landed *inside* a vtable.

So `romdata_check` will now grade that table complete at 20 of at least 36 slots. It cannot hide a mismatch — DIFFERS is computed over emitted bytes regardless of extent — but `verified` is what `validate_merge` ratchets, and the consequence is a forward trap: **if someone later fixes `data_ov006_0213c67c`, `verified` drops 465 → 464 and the coverage ratchet goes red on a correctness fix.** Whoever fixes it has to land the slot-20+ declarations in the same PR.

A heuristic sweep says this is not a singleton — 104 `_ZTV` symbols tree-wide are followed by an auto-named `data_*` whose first two words still look like slots, 93 of them marked `ambiguous`. That count is a candidate list, not 104 confirmed defects; each needs the identity check I did above. I will open it as a follow-up rather than hold this PR, since the queue behind you declares slots 20 and 21 anyway.

## Second finding, non-blocking: the twin you authored is the least informative one

`src/_ZN11dScMgBase_c13OnTurnIntoEggEi.c` declares the flat-C half as

```c
int _ZN11dScMgBase_c13OnTurnIntoEggEi(void)
```

with no `self` and no `int` — while `include/dScMgBase_c.h` says `virtual int OnTurnIntoEgg(int mode)`, and the sibling twin this PR sits next to says

```c
void _ZN11dScMgBase_c13OnYoshiTryEatEi(struct dScMgBase_c *self, int arg)
```

Byte-neutral either way. The renamed descendant bodies are heterogeneous already (`(void)`, `(char* c)`, `(char *self)`, `(char *thiz, int sel)`) but those are pre-existing legacy text that this PR only moved; the base body is new, and it is the one that sets the convention for the fifteen slots after it. Same finding as #2086's, same region `check_header_offsets` cannot reach — its flat-C `#else` half is unscoped, which is my open task. I will settle both twins together in the slot-18 comment follow-up I already owe.

## Reproduction

```
tools/rombuild.py -j 16   ->  source-built functions 11,088 / reproducing 11,088 / mismatching 0
                              source-owned data claims: 2 (reproducing 2)
                              ROM-build analysis: PASS
tools/port_refcheck.py    ->  405 references checked, 0 stale
```

Static gates on the merge result (`premerge_check.py 2087`, merge tree `7646fc0a0a5b`, base 27m old):

```
```

`differs` stays at 7 and the `differing` list is byte-identical between base and merge — same 16 rows, same symbols, same sizes. The config edit is exact: `func_ov004_020b2994 → _ZN11dScMgBase_c13OnTurnIntoEggEi`, `size=0x8 addr:0x020b2994`, `complete` preserved, and slot 19 sits at the *lower* address than slot 18's `0x020b299c`, which is what the vtable says. Declaration order in the header is 18 then 19, so slot assignment is right.

## Two cautions if you re-run my numbers

- **Do not compare line endings through process substitution.** `diff <(git show A:f) <(git show B:f)` applies `core.autocrlf` asymmetrically under MSYS and fabricated a lone-CRLF-line regression for me that does not exist. Counting `\r\n` against bare `\n` on raw bytes in Python says the descendant headers are wholly CRLF files gaining a correct CRLF line, and `dScMgBase_c.h` is pure LF and stays pure LF. No regression — I dropped that finding before it reached this comment.
- Still true from #2086: overlays overlap in address space, so resolving an address to bytes by scanning `overlays.yaml` for the first containing range silently returns the wrong overlay's code. Attribution has to come from the `config/arm9/overlays/ovNNN/` path the symbol was read from.

## What I liked

Naming the discriminator as a general rule — dereference-versus-compare separates `Ei` from `ER6Player` when codegen cannot — is the right kind of thing to write into a PR body, because it is reusable and it is the reason the `dActor_c.h` contradiction is now settled twice rather than noted twice. Continuing to do one slot per PR is still correct: declaring all eighteen at once would write the base body into every undeclared descendant slot and turn PARTIAL into DIFFERS family-wide.

PR #2088, opened 2026-08-31
Declare dScMgBase_c slot 20 (Virtual50) across the minigame family

andrewboudreau:
Third slot of the `dScMgBase_c` keystone campaign. Stacked on #2087 (slot 19),
which is stacked on #2086 (slot 18). **Review and land those first.**

The unit is one base slot plus every descendant override of that same slot.
That is the smallest change that is safe: it widens each already-emitted vtable
by exactly one word, and no descendant can end up with the base's body in a slot
where the cartridge holds its own. Declaring several slots at once loses that
property.

## Slot 20 has no name

`Virtual50` is a placeholder, and the header says so. It is the spelling
`include/dActor_c.h:133` already uses, derived from the byte offset
(slot 20 x 4 = 0x50).

All five bodies carry a `recovered name: <class>_Virtual50` comment, which looks
like corroboration and is not: every one of those is the same coined placeholder
propagated by the recovery pass, not a name read out of the ROM, a symbol table,
or a string. Five copies of a guess are one guess. Naming it would be inventing,
so it keeps the placeholder until something real turns up.

## Signature, and what is not established

**Arity: no explicit parameters.** The ov004 base body is a bare `bx lr`. All
four overrides are single tail calls that pass nothing on -- no argument
register is read or forwarded in any of the five.

**Return type: not determined by these bodies, and the header says that in
those words rather than implying otherwise.** A bare `bx lr` and four tail calls
cannot separate `int` from `void`; only an override with an early return can,
because that is where the two allocate registers differently
(`include/dActor_c.h`'s own comment records the technique). This takes
`dActor_c.h:133`'s `int` as the working answer. If a later override with an
early return shows otherwise, **that override is the evidence** and this
declaration changes.

## The five bodies

| class | address | file |
|---|---|---|
| `dScMgBase_c` | `0x020b2990` | `src/_ZN11dScMgBase_c9Virtual50Ev.c` |
| `dScMg3DEsp_c` | `0x020e9c10` | `src/_ZN12dScMg3DEsp_c9Virtual50Ev.c` |
| `dScMgCup_c` | `0x020dfed4` | `src/_ZN10dScMgCup_c9Virtual50Ev.c` |
| `dScMgSound_c` | `0x0211c5b8` | `src/_ZN12dScMgSound_c9Virtual50Ev.c` |
| `dScMgTeresa_c` | `0x02120238` | `src/_ZN13dScMgTeresa_c9Virtual50Ev.c` |

## `dScMgTeresa_c` is the one to look at

Its class comment said slot 20 was "a bare `FreeGfxSlotsById(8)` thunk with no
`this` at all" and left it as a raw `extern "C"` helper. That reasoning was
fine while nothing declared the slot; it stops being fine the moment
`dScMgBase_c` does. The comment is rewritten in this PR.

Leaving Teresa's declaration out would put the **base's** body into Teresa's
slot 20, where the cartridge holds Teresa's own -- a real DIFFERS, in a vtable,
which `rombuild` does not look at.

An earlier revision of the apply script did exactly that. Its "already declared"
guard was `\bVirtual50\b`, which matched Teresa's **class comment** and skipped
her silently. The guard is now `^\s*virtual[^;]*\bNAME\b` (multiline), and the
script asserts `declared-count == override-count` before it will write anything.
Recorded here because the next fifteen PRs run the same script.

## Verification, against the slot-19 parent

```
ROM-build analysis: PASS
```

`python tools/romdata_check.py`, slot 19 -> slot 20:

```
metric               slot19     slot20    delta
verifiedBytes         35368      35372       +4
partialBytes          10180      10300     +120
```

NEW differs: none. Cleared: none. Size changes: none.

**`partialBytes +120` is the load-bearing number, not `differs +0`.** 120 = 30
tables x 4 bytes, i.e. every emitted table in the family grew by exactly one
slot. That is the arithmetic that proves the *base* declaration is really doing
the work. Had the base stayed silent, each descendant's own `virtual` would have
introduced a **new** virtual in that class instead -- pointing at the same body,
byte-identical, PASS, no DIFFERS, invisible to 106/106 -- and only 27 tables
would have widened, for +108. The two outcomes are distinguishable only here.

Per-symbol (via `RD.check_object`, which recompiles from the tree and so needs
no rombuild), all four overriders' vtables read `PARTIAL 84 bytes` = 21 slots
with no DIFFERS. That is direct evidence each override landed at index 20 rather
than being appended, Teresa's included.

## Static gates

`python tools/premerge_check.py` on the merge tree: all 8 gates
(`converted-ratchet`, `dead-references`, `duplicate-sources`, `header-offsets`,
`langmode-ratchet`, `layout-check`, `src-tu-refs`, `source-coverage`) green on
base and green on the merge result.

## What is left

Slots 21-35, fifteen more, same shape. The emitted tables stop at slot 20 until
they land. Note that `_ZTV14dScMgD3DBase_c`'s configured 80-byte extent is not
the end of that class's table -- the real table is 36 slots at `0x0213c62c`,
continuing into `data_ov006_0213c67c`, and every address in it from `0x020b298c`
down to `0x020ad660` is a slot this campaign still owes. See the correction note
on #2087.

not a goal here.


andrewboudreau (commented):
Approving. Reviewed against the **merge tree** rather than the branch — `premerge_check.py 2088 --fetch`, merge tree `00b8a33593c7`, base `origin/main@2dbd0cb082d8`. Every ROM-level claim in the description was re-derived from the cartridge independently and all of them hold. The measurement came out sharper than the one on #2087, and the PR body understates how decisive it is.

## The widening, decomposed per symbol — 33 rows, and nothing else in the tree moved

I dumped `romdata_check`'s best-verdict table per symbol on the merge tree and diffed it against the same dump on main, rather than reading aggregates:

```
changed rows tree-wide : 33      all 33 are _ZTV*dScMg*
                          30 PARTIAL  80 -> 84   (+120)
                           1 VERIFIED 80 -> 84   (+4)     _ZTV14dScMgD3DBase_c
                           2 DIFFERS  84 -> 88, 88 -> 92  (+8, in neither bucket)
total emitted delta    : +132 over 33 rows
```

which lands on the aggregates exactly:

```
                     main       merge      delta
verifiedBytes       35368      35372         +4
partialBytes        10180      10300       +120
verified+partial    45548      45672       +124   <- the bucket-invariant number
verified/partial/differs/unnamed  465/252/7/522  unchanged
differing list      16 rows, byte-identical
```

**Every single dScMg table in the family widened by one word, and no non-dScMg symbol changed at all.** That is the full-family signature — the base declaration landed and took effect.

## Your stated failure-mode baseline is stale, and the truth is stronger

The body carries `+108 / 27 tables` as the descendants-only baseline. That is text carried down from slot 18. For slot 20 the descendants-only failure mode widens only the four classes that declare an override:

```
descendants-only would be : 4 tables x 4 =  +16
measured                  : 33 tables x 4 = +132  (+124 on the invariant)
```

An 8x separation, not the 1.15x the stale line implies. Worth fixing in the body so the next slot PR does not inherit it again — and please quote `verifiedBytes + partialBytes`, not `partialBytes` alone. That is the #2087 lesson: `partialBytes` broke there because a table graduated PARTIAL -> VERIFIED and took its whole byte count out of the bucket. Here it happens to read cleanly, but only by luck of no graduation this slot.

## The four-overrider census is right

Full 34-table census of the family at slot 20 (`word at _ZTV<class> + 0x50`, resolved through the `config/arm9/overlays/ovNNN/symbols.txt` the table was read from, never by scanning `overlays.yaml` — overlays overlap in address space):

```
own slot 20 : dScMgCup_c, dScMg3DEsp_c, dScMgSound_c, dScMgTeresa_c   (4)
inherit     : 29
not family  : _ZTVN12dScMgSlot1_c9betIcon_cE  (nested callback table, not a descendant)
```

Exactly the four you rename. No missed overrider. Slot 20 of `_ZTV11dScMgBase_c` is `0x020b2990` (ov004 `@0x020bc0c0 + 0x50`), the body there is a bare `bx lr` at `size=0x4`, and the header declares it after slot 19, so the slot assignment is correct by declaration order.

## Arity and return type — both confirmed, and one of them needs the careful method

- **"no explicit parameters"**: confirmed both ways. No body reads incoming r1 under def-use liveness. This is the scan that has to be done properly — `dScMgCup_c` and `dScMgSound_c` both *write* r1 early, so a naive "does the body touch r1" grep says they take an argument and contradicts you. With a live set seeded `{1}`, `mov` copies followed, registers killed on write and r0-r3 cleared on `bl`, the answer is zero readers. The single call site corroborates it: only `this` in r0.
- **"return type NOT determined"**: correct, and I can say a little more than you did. The one call site, `ov004 @0x020b6e80`, discards r0 — `ldr r0, [pc, #0x78]` lands immediately after the `blx r1`. So `void` is consistent with every observation, but nothing *proves* it, and the comment saying so is the right call. That call site really is a minigame scene: it dispatches on global `data_ov004_020beb68`, which `include/dScMgBase_c.h` itself already declares `extern "C" void *`.

## Finding, non-blocking: the header CRLF flip-flops once per slot PR

Raw blob-to-blob byte counts (not `diff`, and not through process substitution — `core.autocrlf` under MSYS fabricates differences there):

```
include/dScMgBase_c.h    main  [CRLF=0   LF=185]
                         merge [CRLF=203 LF=1  ]
```

The keystone header flips wholesale from LF to CRLF. Meanwhile each of the four descendant headers is already a CRLF file and gains exactly one LF line (`dScMg3DEsp_c.h` 75->76, `dScMgCup_c.h` 102->103, `dScMgSound_c.h` 62->63, `dScMgTeresa_c.h` 35->41). So each slot PR adds LF lines, the next one normalizes the file to CRLF and adds its own, and the cycle repeats. Content is unaffected and the bytes do not care, but it makes the keystone header's diff unreviewable — the file everyone in this campaign has to read is the one whose diff is 100% noise. Fifteen slots remain; this will happen fifteen more times.

`.gitattributes` carries only `merge=union` rules for the ledgers, no `text`/`eol` attribute reaching headers. A one-line attribute for `include/**.h` would end it. Not a blocker for this PR, and not yours to fix alone — flagging it because it compounds.

## Finding: this PR supplies the first *byte* evidence for the `dScMgD3DBase_c` extent defect

On #2087 I flagged that `_ZTV14dScMgD3DBase_c` grades VERIFIED on a truncated extent — `data_ov006_0213c67c` sits `0x50` past it, is marked `ambiguous`, is delinked nowhere, and is referenced by nothing. That was an inference from address patterns.

This PR turns it into a measurement. `romdata_check` compares `len(linked)` emitted bytes against `RV.rom_bytes(module, addr, len(linked))`, which ignores config extents entirely. That table now emits **84** bytes where the recorded extent is **80**, and all 84 compare equal. The 81st through 84th bytes are the first word of `data_ov006_0213c67c`'s range, and they reproduce from source.

So the splitter is provably a dsd auto-boundary that landed inside a vtable, not a data object. Nothing here is unsafe — the bytes match, `rombuild` is 106/106, and DIFFERS is computed over emitted bytes regardless of extent. But the overlap grows by one word per slot PR, and the forward trap from #2087 still stands: whoever fixes `data_ov006_0213c67c` drops `verified` 465 -> 464 and reddens the coverage ratchet on a correctness fix, so the slot-20+ declarations have to land in the same PR as the fix. Tracking it separately; it does not hold this one.

## Reproduction

```
tools/rombuild.py -j 16      source-built functions 11,088 / reproducing 11,088 / mismatching 0
                             source-owned data claims 2 (reproducing 2, mismatching 0)
                             ROM-build analysis: PASS
tools/port_refcheck.py       405 references checked, 0 stale
```

`premerge_check.py 2088 --fetch`, merge tree `00b8a33593c7`:

```
```

Config edits are exact and the `complete` markers are preserved: `func_ov004_020b2990 -> _ZN11dScMgBase_c9Virtual50Ev` at `size=0x4 addr:0x020b2990` in `ov004/symbols.txt`, the matching `delinks.txt` rename, and the four ov006 descendant renames. EOL-normalized content diff is the advertised change and nothing else (+22/-3 base, +1 each on three descendants, +10/-4 Teresa).

## What I liked

Naming the slot `Virtual50` and saying plainly in the comment that it is a placeholder spelled from the byte offset — not a recovery — is the right move, and it is the opposite of the failure mode where a coined name gets treated as evidence later. Keeping one slot per PR also remains correct: declaring the remaining slots at once would write the base body into every undeclared descendant slot and turn PARTIAL into DIFFERS family-wide.

PR #2089, opened 2026-08-31
Declare dScMgBase_c slot 21 (OnGroundPounded) across the minigame family

andrewboudreau:
Fourth slot of the `dScMgBase_c` keystone campaign. Stacked on #2088 (slot 20),
which is stacked on #2087 (slot 19), which is stacked on #2086 (slot 18).
**Review and land those first.**

One base slot plus every descendant override of that same slot. That is the
smallest safe unit: it widens each already-emitted vtable by exactly one word,
and no descendant can end up with the base's body in a slot where the cartridge
holds its own.

## The name is well-sourced; the signature is where the work is

All five bodies carry `recovered name: <class>_OnGroundPounded`, and
`include/dActor_c.h:138` declares the same name at the same index on the
parallel branch. Two sources -- but the second is a hierarchy that shares
indices, not signatures, so it corroborates the *name* and nothing else.

### Arity: no explicit parameters

The ov004 base body is empty. `dScMgBSC_c` and `dScMgCard_c` branch on
`this->mHudScore` alone. `dScMgMemory_c` and `dScMgMemory2_c` tail-call on one
field of `this`. No body in the set reads a second argument register.

`dActor_c.h:138` spells `OnGroundPounded(dActor_c &other)`. It has been wrong on
every parameter list this campaign has actually measured -- slot 18's arity,
slot 19's parameter type -- and an unread parameter leaves no trace either way,
so it is not carried.

### Return type: `void`, and this is a weaker claim than the last two slots

Worth being explicit, because the previous three PRs could each point at a body
that settled it and this one cannot.

All four overrides converge on a **single tail call**. `b callee` is the same
instruction whether the caller returns `int` or `void` -- the callee's `r0`
becomes the caller's `r0` either way -- so these bodies carry no information
about the return type at all.

Rather than assert that, this PR demonstrates it: it changes all four from

```c
int  f(void *thisPtr) { ...; return g(x); }
```

to

```c
void f(void *thisPtr) { ...; g(x); }
```

and **the ROM bytes do not move.** 11,088/11,088 still reproducing, 106/106
still exact. That null result *is* the measurement: it is what "these bodies
cannot discriminate" looks like when you check instead of assuming.

So the `void` comes from `dActor_c.h:138`, whose comment records that slots 21,
24 and 27 were *measured* to return void via an override with early returns --
in that hierarchy.

> **Corrected after this PR was opened.** The first version of this section, and
> of the header comment it describes, said `dActor_c.h` "has been right on every
> return type this campaign checked and wrong on every parameter list." That is
> false, and reading ahead into slot 24 is what falsified it. Commit
> `a792a0643` fixes the header; this section is the corrected version.

`dScMgBase_c` has six slots whose own body pins a return type. `dActor_c.h`
matches five of them and differs on one:

| slot | `dScMgBase_c` | `dActor_c.h` | | evidence in this tree |
|---|---|---|---|---|
| 18 | `int` | `int` | agree | sets `r0` on a constant-return path |
| 19 | `int` | `int` | agree | sets `r0` on a constant-return path |
| 22 | `int` | `int` | agree | `func_ov004_020ae198`, `return 1;` |
| 23 | `int` | `int` | agree | `func_ov004_020ae1a0`, `return 1;` |
| 24 | `int` | `void` | **DIFFER** | `func_ov004_020ae140.cpp` ends `return 1;` |
| 27 | `void` | `void` | agree | `func_ov004_020af27c`, early bare `return;` |

Slot 24 is the one that matters, and **it is not a defect in `dActor_c.h`.** That
header names 24 as one of its three *measured* voids and is right about its own
hierarchy -- `Stump` and `BigBrickBlock` proved it there, and
`include/Stump.h:44` and `include/QuestionBlock.h:63` record how.
`dScMgBase_c`'s slot-24 body sets `r0` to 1 and reproduces byte-exact, so *this*
hierarchy returns `int` at the same index. Both measurements stand.

The two hierarchies simply do not hold the same function at slot 24. They are
parallel branches that each began adding virtuals at 18 -- `dActor_c` off
`dBase_c` directly, `dScMgBase_c` off `dBase_c` through `dScene_c` -- and the
`recovered name:` comments that make the slots look paired **were assigned by
that index**, so they cannot also be evidence for it.

So the honest split is not "return types transfer, parameter lists do not." It
is five-for-six versus zero-for-two, on small samples, from a header that is a
hint in both columns. Five-for-six is the whole case for `void` here, and the
declaration stays `void` on that basis and says so in those words.

If a later override with an early return shows `int`, **that override is the
evidence** and this declaration changes -- and it would outrank the count above.

## The five bodies

| class | address | file |
|---|---|---|
| `dScMgBase_c` | `0x020b298c` | `src/_ZN11dScMgBase_c15OnGroundPoundedEv.c` |
| `dScMgBSC_c` | `0x02125248` | `src/actors/dScMgBSC_c.cpp` (intact TU) |
| `dScMgCard_c` | `0x020db6ec` | `src/actors/dScMgCard_c.cpp` (intact TU) |
| `dScMgMemory_c` | `0x020f523c` | `src/_ZN13dScMgMemory_c15OnGroundPoundedEv.c` |
| `dScMgMemory2_c` | `0x020f730c` | `src/_ZN14dScMgMemory2_c15OnGroundPoundedEv.c` |

The two intact TUs are renamed in place, identifier and `symbols.txt` entry
together; they cannot become standalone files without breaking their TU.

## Verification, against the slot-20 parent

```
ROM-build analysis: PASS
```

`python tools/romdata_check.py`, slot 20 -> slot 21:

```
metric               slot20     slot21    delta
verifiedBytes         35372      35376       +4
partialBytes          10300      10420     +120
```

NEW differs: none. Cleared: none. Size changes: none.

**`partialBytes +120` is the load-bearing number, not `differs +0`.** 120 = 30
tables x 4, i.e. every emitted table in the family grew by exactly one slot --
which is what proves the *base* declaration is doing the work. Had the base
stayed silent, each descendant's own `virtual` would have introduced a **new**
virtual in that class, pointing at the same body: byte-identical, PASS, no
DIFFERS, invisible to 106/106, and only 26 tables would have widened.

Per-symbol (`RD.check_object`, which recompiles from the tree and needs no
rombuild), all four overriders read `PARTIAL 88 bytes` = 22 slots with no
DIFFERS -- direct evidence each override landed at index 21 rather than being
appended at the end.

## Static gates

`python tools/premerge_check.py --repo <worktree> --base cpp/minigame-slot20 2089`
against merge tree `ee4625e9a4f8`:

```

```

All eight green. That run predates commit `a792a0643`, which is comment-only;
premerge is re-run against the new head below.

## Where this ends

`_ZTV11dScMgBase_c` (ov004, `0x020bc0c0`) is **144 bytes = 36 slots in
`symbols.txt`, with 36 distinct addresses and no shared stub** -- the base's own
table is configured end to end, unlike `_ZTV14dScMgD3DBase_c`, which dsd cut at
slot 20 (see the correction note on #2087). So when slot 35 lands, the base
vtable goes PARTIAL -> VERIFIED and that verdict will genuinely mean complete.
That is the campaign's finish line: fourteen more slots.

Slots 22 and 23 are already measured and both return `int` on hard evidence --
`return 1;` in each base body sets `r0`, which a `void` function would not, and
`dScMgTrampoline2_c`'s slot-23 override at `0x02122f24` has four early
`return 0;` paths and a final `return 1;`. Slot 22 has **zero** descendant
overrides, making it the smallest commit in the whole campaign.

not a goal here.


andrewboudreau:
## Correction: I published a false claim in this PR, and here is the retraction

The original body and the original `include/dScMgBase_c.h` comment both said:

> `dActor_c.h` has been right on every return type it states here and wrong on
> every parameter list.

**That is false.** Reading ahead into slot 24 falsified it about two hours after
I opened this PR. Commit `a792a0643` fixes the header comment; the body above is
the corrected version. Both the header and the body now name the exception
rather than claiming a perfect record.

### What actually falsified it

`dScMgBase_c` has six slots with a body of their own that pins a return type.
`dActor_c.h` matches five and differs on one:

| slot | `dScMgBase_c` | `dActor_c.h` | | evidence |
|---|---|---|---|---|
| 18 | `int` | `int` | agree | sets `r0` on a constant-return path |
| 19 | `int` | `int` | agree | sets `r0` on a constant-return path |
| 22 | `int` | `int` | agree | `func_ov004_020ae198`, `return 1;` |
| 23 | `int` | `int` | agree | `func_ov004_020ae1a0`, `return 1;` |
| 24 | `int` | `void` | **DIFFER** | `func_ov004_020ae140.cpp` ends `return 1;` |
| 27 | `void` | `void` | agree | `func_ov004_020af27c`, early bare `return;` |

`src/func_ov004_020ae140.cpp` is enrolled and delinked
(`ov004/delinks.txt:146`, `.text 0x020ae140..0x020ae198`), so it is inside the
11,088 reproducing set -- its `return 1;` is a measurement, not a guess.

### The part worth keeping

Slot 24 is **not** a defect in `dActor_c.h`. That header calls 24 one of its
three *measured* voids and it is right about its own hierarchy --
`include/Stump.h:44` and `include/QuestionBlock.h:63` record the `Stump` and
`BigBrickBlock` overrides that measured it, using the early-return register
allocation test. Both measurements stand.

The two hierarchies simply **do not hold the same function at slot 24**. They
are parallel branches that each began adding virtuals at 18 -- `dActor_c` off
`dBase_c` directly, `dScMgBase_c` off `dBase_c` through `dScene_c`. And the
`recovered name:` comments that make the slots look paired *were assigned by
that index*, so they cannot also be evidence for it. That circularity is the
reusable lesson, and it is the same trap as coined names blocking data
verification: a label derived from an assumption cannot then confirm it.

So `dActor_c.h` is a **hint in both columns** -- roughly five-for-six on return
types, zero-for-two on parameter lists (slot 18's arity, slot 19's parameter
type). Never a measurement for this hierarchy.

### What did not change

The declaration. `virtual void OnGroundPounded();` stays, because five-for-six
is still the best signal available and because flipping all four overrides
between `int` and `void` was already measured in this PR to move no ROM byte.
What changed is that the reason is now stated at its true strength.

`a792a0643` is comment-only -- no declaration, no body, no config:

```
ROM-build analysis: PASS
ROM data from source: 465 verified, 252 partial, 7 differ, 522 unnamed
```

Identical to the pre-correction numbers, including `verifiedBytes 35,376`.

### Static gates

`premerge_check.py --base cpp/minigame-slot20 2089`, merge tree `ee4625e9a4f8`
(run before `a792a0643`):

```

```

All eight green, and CI is green on the new head as well.

Flagging this in the comment rather than silently editing the body, because a
reviewer who read the first version took away a rule that does not hold.

andrewboudreau (commented):
**CHANGES REQUESTED** (posted as a comment: this repo refuses a formal request-changes review from the account that opened the PR).

Requesting changes — on **one comment, for a count, not for a byte**. Everything mechanical in this PR is correct and I verified it end to end against the merge tree, not the branch: `premerge_check.py 2089 --fetch`, merge tree `548977c4acfb`, base `origin/main@5c24b731940c`. The slot is right, the census is right, the arity proof is right, and the widening is the full-family signature. The problem is that the new header comment builds its `void` recommendation on a six-row evidence table, **two of whose rows are not evidence**, and it says so itself in the strongest possible terms — *"Five-for-six is the whole case for `void` here."* The real count is three-for-four. That text is the keystone the next fourteen slot PRs will cite, and #2093 (slot 24) is already open citing exactly the row I had to correct.

## First, the measurement: this is the full family, not the descendants

Per-symbol dump of `romdata_check`'s best-verdict table on the merge tree, diffed against the same dump on post-#2088 main — aggregates are not enough to tell the two outcomes apart:

```
changed rows tree-wide : 33      all 33 are _ZTV*dScMg*, each +4
                          30 PARTIAL
                           2 DIFFERS   (bytes in neither bucket)
                           1 VERIFIED  (_ZTV14dScMgD3DBase_c)
total emitted delta    : +132 over 33 rows
bucket-invariant       : verifiedBytes + partialBytes  +124
verifiedBytes          : 35372 -> 35376
verified/partial/differs/unnamed : 465/252/7/522 unchanged, differing list byte-identical
```

Against the failure mode this whole campaign is guarding: descendants-only would widen the **four** tables that declare an override, `4 x 4 = +16`. Measured 33 tables and +132. The base declaration landed and took effect — a 8x separation, unambiguous.

Quoting `verifiedBytes + partialBytes` here as #2087 asked: thank you. Please keep doing that. No table graduated buckets this slot so `partialBytes` alone would also have read cleanly, but that is luck, not a property.

## Claims I re-derived from the cartridge, all confirmed

- **slot 21 of `_ZTV11dScMgBase_c` is `0x020b298c`** — ov004 `@0x020bc0c0 + 0x54`, resolved through the `config/arm9/overlays/ov004/symbols.txt` the table was read from, never by scanning `overlays.yaml` (overlays overlap in address space; three of them contain some of these addresses). Base body is a bare `bx lr`, `size=0x4`. Header declares it after slot 20, so slot assignment is right by declaration order.
- **Exactly four descendants own slot 21.** Full 34-table family census: `dScMgBSC_c`, `dScMgCard_c`, `dScMgMemory_c`, `dScMgMemory2_c` own it; 29 inherit the base body; `_ZTVN12dScMgSlot1_c9betIcon_cE` holds `0x000000f0` at that offset and is the nested callback table, not a descendant. No missed overrider — exactly the four you rename.
- **"no explicit parameters"** — confirmed by def-use liveness with the live set seeded `{r1}`, `mov` copies followed into it, registers killed on write, r0–r3 cleared at every `bl`. **Zero live-r1 readers in all five bodies.** This is the scan that has to be done carefully; a naive "does the body touch r1" grep is worthless in this family and has contradicted a correct PR before.
- **"return type NOT determined by these five bodies"** — confirmed. Zero constant-return paths among the five.
- Config edits exact, `complete` markers preserved: base `func_ov004_020b298c -> _ZN11dScMgBase_c15OnGroundPoundedEv` `size=0x4 addr:0x020b298c` plus the `delinks.txt` rename, and four ov006 descendant renames (`dScMgCard_c` 0x34, `dScMgMemory_c` 0x14, `dScMgMemory2_c` 0x14, `dScMgBSC_c` 0x34). 14 files, only slot-21 content.

## The change I am asking for: two rows of the evidence table are not evidence

The comment offers this as the case for spelling slot 21 `void`:

```
    18  int  / int   agree   sets r0 on a constant-return path
    19  int  / int   agree   sets r0 on a constant-return path
    22  int  / int   agree   func_ov004_020ae198, `return 1;`
    23  int  / int   agree   func_ov004_020ae1a0, `return 1;`
    24  int  / void  DIFFER  func_ov004_020ae140.cpp ends `return 1;`
    27  void / void  agree   func_ov004_020af27c, early bare return
```

I scanned every body at every one of those slots for a genuine constant-return path — an immediate written to r0 that reaches a return **without being consumed as a source operand** along the way:

```
slot 18   0 of 31 bodies      <- the row says "sets r0 on a constant-return path"
slot 19  10 of 12 bodies      confirmed
slot 22   1 of 1              confirmed  (func_ov004_020ae198  mov r0,#1 ; bx lr)
slot 23   3 of 4              confirmed
slot 24   6 of 7              confirmed, and stronger than you claim -- below
slot 27   0 of 7 bodies       <- the row says "agree"
```

Re-run allowing conditional forms (`moveq`, `movne`, …) in case an exact-mnemonic match was hiding one: slot 18 has **60** r0-immediate sites across its 31 bodies and **none** of them is a return value; slot 27 has 1 site and it is not one either.

**Slot 18 pins nothing.** This is independently the same result I posted on #2086 — 0 of 31 bodies set r0 as a return, 0 of 22 call sites consumed it, and a `void` variant still byte-matched. That row is not just unsupported, its stated reason is the opposite of what the ROM does.

**Slot 27 pins nothing either.** `func_ov004_020af27c` has two exits leaving two unrelated values in r0: an early `popne/bxne` at `0x020af290` where r0 is a field just loaded from `r4+0x4000+0x630`, and the fall-through at `0x020af2f4` where r0 is the address `r4+0x4600`. Nothing about that is a return value. It is *consistent with* `void` — it is not agreement with `void`, and counting it as a fifth agreeing row treats absence of contrary evidence as evidence.

Beware the shape that produces a false positive here, because a straight-line scan that breaks only on an r0 *write*, a call, or a branch will report one at slot 18 — `_ZN11dScMgCoin_c13OnYoshiTryEatEi`, ov006 `0x020de5b0`:

```
  0x020de620  add   r1, r4, #0x5000
  0x020de624  str   r0, [r1, #0x1d4]
  0x020de628  mov   r0, #0
  0x020de62c  str   r0, [r1, #0x1c8]    <- CONSUMES the constant
  0x020de630  pop   {r4, lr}
  0x020de634  bx    lr
```

That is a stored zero, not a returned one. The scan has to break when r0 appears in an instruction's *read* set, not only its write set. I nearly posted a retraction of my own #2086 statement over it.

So the honest table, restricted to slots that actually pin a type:

```
    19  int  / int   agree   10 of 12 bodies end `mov r0,#1` (base 0x020b2994)
    22  int  / int   agree   func_ov004_020ae198, `return 1;`
    23  int  / int   agree   func_ov004_020ae1a0, `return 1;`
    24  int  / void  DIFFER  6 of 7 bodies pin int, incl. the base
    18  --   / int   pins nothing: 0 of 31 bodies has a constant-return path
    27  --   / void  pins nothing: 0 of 7; two exits leave unrelated r0 values
```

**Three-for-four, not five-for-six** — and both rows that drop out were counted as agreeing, so the correction moves in the direction of weakening the `void` case, not strengthening it.

## Slot 24's DIFFER is stronger than you wrote, and it matters for #2093

You cite the base alone. It is six of seven bodies. `func_ov004_020ae140` (size 0x58) converges every path on the same tail:

```
  0x020ae158  beq   #0x20ae18c
  ...
  0x020ae18c  mov   r0, #1
  0x020ae190  pop   {r4, lr}
  0x020ae194  bx    lr
```

and `dScMgJump_c`, `dScMgJump2_c`, `dScMgD3DBase_c` (all via `func_ov006_020e6e78`), `dScMgTrampoline_c` and `dScMgTrampoline2_c` all do the same `mov r0, #1`. `dActor_c.h` lists slot 24 among its *measured* voids. So on the one slot where the two hierarchies can both be measured against the cartridge, `dActor_c.h` is wrong, on six independent bodies. That is worth saying in the comment, because it is the third time this campaign has caught `dActor_c.h` (slot 18 arity, slot 19 parameter type, now slot 24 return type) and **#2093 is declaring slot 24 right now.**

## What I am *not* asking for

Not asking you to change the spelling. `void OnGroundPounded()` is byte-neutral, the comment is explicit that it is a hint and not a measurement, and it invites a later override to overturn it — that framing is exactly right and is the opposite of this project's recurring failure mode. I did not re-run your "flipping all four overrides between `int` and `void` moves no ROM byte" experiment; nothing in my request depends on it.

The ask is a single edit to one comment: drop rows 18 and 27 to a "pins nothing" line, restate the count as three-for-four, and strengthen slot 24 to six bodies. Byte-neutral, no config change, no rebuild needed on my side to re-verify — I will re-check the text and merge.

## Reproduction

```
tools/rombuild.py -j 16      source-built functions 11,088 / reproducing 11,088 / mismatching 0
                             source-owned data claims 2 (reproducing 2, mismatching 0)
                             ROM-build analysis: PASS
                             35,376 verified data bytes  (+4 over #2088)
tools/port_refcheck.py       405 references checked, 0 stale
```

`premerge_check.py 2089 --fetch`, merge tree `548977c4acfb`, base `origin/main@5c24b731940c` — all eight gates `pass -> pass`: converted-ratchet, dead-references, duplicate-sources, header-offsets, langmode-ratchet, layout-check, src-tu-refs, source-coverage.

One caution carried from #2088 and still unfixed: `include/dScMgBase_c.h` flips wholesale between LF and CRLF once per slot PR, so the keystone header's diff is mostly noise every time. Fourteen slots to go. A one-line `include/**.h` attribute would end it; not yours to fix alone and not a blocker here.

andrewboudreau (commented):
**Lifting my CHANGES REQUESTED.** The correction commit does what I asked and more — it found a falsifier I had not, and it reframed the count as a hint rather than an authority, which was the part that actually mattered. Merging.

Two rows in the new table are still overstated, though, and I measured them rather than arguing about them. This is a **non-blocking** note; I am taking the fix myself in a follow-up rather than sending this back a third time, because one of the two errors is already on main and this PR only cites it.

## Slot 24 is right, and it is a good catch

`func_ov004_020ae140.cpp` ends `return 1;` with no store consuming the constant, so the `mov r0, #1` at the join disappears under `void`. That is a real pin, `dActor_c.h` really does call 24 one of its measured voids, and "parallel branches that each began adding virtuals at 18" is the right reading. I had not looked at 24 at all.

## Slot 18 is not a pin — measured, byte-for-byte

The header's slot-18 block rests the whole `int` on one sentence: *"dScMgCoin_c::OnYoshiTryEat is a real member definition ending `return 0;`, so declaring void would have changed its bytes."*

It would not. The body's last two statements are

```c
    self->unk_51c8 = 0;
    return 0;
```

The `mov r0, #0` is emitted for the **store**, and r0 still holds 0 at `bx lr`. Dropping the `return 0;` changes nothing. I built it both ways in a wired worktree:

```python
import build_pin as B, pathlib
src = pathlib.Path('src/_ZN11dScMgCoin_c13OnYoshiTryEatEi.cpp')
B.verify(src, '_ZN11dScMgCoin_c13OnYoshiTryEatEi', 0x020de5b0, 0x8c, 'ov006')

  as declared, int          -> (True, '2004/b56')
  void in dScMgBase_c.h + dScMgCoin_c.h + the definition,
  with `return 0;` deleted  -> (True, '2004/b56')
```

Same address, same 0x8c, both exact. The base's own body is `_ZN11dScMgBase_c13OnYoshiTryEatEi`, `size=0x4` at `0x020b299c` — a lone `bx lr`, which the header already says sets nothing. So neither the base nor its named witness discriminates, and slot 18 belongs out of the tally, not in its "agree" column.

This is the false positive I flagged on #2095: a scan that breaks only on an r0 **write** calls `mov r0,#0` a constant return even when the very next instruction consumes it. The rule that survives is *break when r0 appears in an instruction's read set, not only its write set.*

## Slot 27 is not a pin either

`func_ov004_020af27c` is declared `void` and has an early return, but neither exit carries a `mov r0, #k`:

- the early exit leaves whatever `self->unk_4630` loaded;
- the fall-through leaves whatever the last store computed.

Declaring it `int` and simply not returning a value emits the same bytes. "Early bare return" is only visible in the C we wrote, not in the cartridge. It is *consistent with* `void` and inconsistent with returning a fixed constant, which is worth something — but it is not a measurement of the return type, and it should not sit in a column headed `agree`.

## So the count is three-for-four

| slot | pins a return type? | via | dActor_c.h |
|---|---|---|---|
| 18 | **no** | base is `bx lr`; the witness byte-matches as `void` | — |
| 19 | yes, `int` | base body `return 1;` | agree |
| 22 | yes, `int` | `func_ov004_020ae198`, `return 1;` | agree |
| 23 | yes, `int` | `func_ov004_020ae1a0`, `return 1;` | agree |
| 24 | yes, `int` | `func_ov004_020ae140`, `return 1;` | **DIFFER** |
| 27 | **no** | no `mov r0, #k` on either exit | — |

**Three-for-four**, not five-for-six.

**None of this changes the declaration.** `void` at slot 21 stands: the four overrides are single tail calls, the flip was measured to move no ROM byte, and a smaller majority of a smaller sample is still the best signal on offer. What changes is the number, and the fact that the strongest-sounding row in the table — the one whose supporting sentence claims a byte difference — is the one that is not there.

## The part that is already on main

The sentence *"so declaring void would have changed its bytes"* landed in `include/dScMgBase_c.h` with #2086, not here; this PR inherits it. Correcting it is on me — I opened that thread and I now have the measurement — so I will land the slot-18 sentence and these two rows together in a single follow-up rather than bouncing this one. If you restack #2090 and hit a conflict in that comment block, that will be why.

One ask for the stack, so the number does not fan out: **#2095 currently derives "seven-for-eight" from this table.** With 18 and 27 out it is five-for-six there. Please take the count from whatever `dScMgBase_c.h` says at the time you restack, rather than carrying a figure forward across PRs.

PR #2090, opened 2026-08-31
Declare dScMgBase_c slot 22 (OnAttacked1) across the minigame family

andrewboudreau:
Fifth slot of the `dScMgBase_c` keystone campaign. Stacked on #2089 (slot 21) ->
#2088 (slot 20) -> #2087 (slot 19) -> #2086 (slot 18). **Review and land those
first.**

This is the smallest change in the campaign, and the cleanest test of the thing
the campaign is actually asserting.

## Slot 22 has no overrides at all

All 32 descendants inherit the ov004 body. So the entire change is:

- one `virtual int OnAttacked1();` in `include/dScMgBase_c.h`, with its evidence
  block
- `src/func_ov004_020ae198.c` -> `src/_ZN11dScMgBase_c11OnAttacked1Ev.c`
- the two ov004 config lines that follow the rename

No descendant header, no second body, no in-place TU rename. Four files.

## Why zero overrides makes this the load-bearing check

Every previous slot in this campaign had to argue that `partialBytes +120` came
from the *base* declaring the slot, and not from N descendants each inventing a
new virtual of their own that happened to point at the same body -- the failure
mode that is byte-identical, passes 106/106, produces no DIFFERS, and is
invisible to every gate except the widened-table count.

Here there are no descendants declaring anything. There is exactly one
`virtual` in the diff. If it failed to take, nothing else in the tree could
have produced those 30 words. `+120` is therefore not corroborating evidence
this time -- it is the whole proof, unaliased.

```
metric               slot21     slot22    delta
verifiedBytes         35376      35380       +4
partialBytes          10420      10540     +120
NEW differs: NONE       cleared: none
```

120 = 30 emitted tables x one word each.

## Evidence for the signature

**Name** -- `OnAttacked1`, from the ov004 body's own `recovered name:` comment.
`include/dActor_c.h:139` spells the same name at the same index, but that is a
parallel hierarchy sharing slot numbers, so it is a hint that agrees rather
than a second reading. (See #2089 for why slot-index correspondence between
these two hierarchies is not evidence: they measurably disagree at slot 24.)

**Arity** -- no explicit parameters. With no override there is no second body to
check a second argument register against, and the base body reads none.
`dActor_c.h:139`'s `dActor_c &` is not carried: an unread parameter leaves no
trace either way, and that header's parameter lists have been measured wrong at
every slot this campaign has checked.

**Return type -- `int`, MEASURED.** Not the coin flip slots 20 and 21 were:

```c
int _ZN11dScMgBase_c11OnAttacked1Ev(void)
{
    return 1;
}
```

`return 1;` sets `r0`, which a `void` function would not, and the file
reproduces byte-exact.

## Verification, against the slot-21 parent

```
source-owned data claims: 2 (reproducing 2, mismatching 0)
ROM-build analysis: PASS
```

No size changes, no new DIFFERS, none cleared.

## Where this ends

`_ZTV11dScMgBase_c` (ov004, `0x020bc0c0`) is 144 bytes = **36 slots** in
`symbols.txt`, with 36 distinct addresses and no shared stub -- configured end
to end. When slot 35 lands, the base vtable goes PARTIAL -> VERIFIED and that
verdict will genuinely mean *complete*, unlike `_ZTV14dScMgD3DBase_c`, which dsd
cut at slot 20 (see the correction note on #2087).

Thirteen slots left. Slot 23 (`OnAttacked2`, `0x020ae1a0`, 3 overrides) is
already measured `int` on hard evidence: the base body is `return 1;`, and
`dScMgTrampoline2_c`'s override at `0x02122f24` has four early `return 0;` paths
and a final `return 1;` -- the early-return discriminator, not a tail call.

not a goal here.


andrewboudreau:
## Static gates

`premerge_check.py --repo <worktree> --base cpp/minigame-slot21 2090`, merge tree `b21eb4c3eaad`:

```

```

All eight green on the merge tree, not just on the branch.

Note the parent moved while this was open: `cpp/minigame-slot21` now carries `a792a0643`, a comment-only correction to #2089's return-type rationale (`dActor_c.h` is five-for-six on return types for this hierarchy, not six-for-six -- slot 24 falsifies it). This branch is rebased onto it. See https://github.com/tangosdev/sm64ds-decomp/pull/2089#issuecomment-5480213798 for the retraction; nothing about slot 22 depends on it, since slot 22's return type is measured from its own body rather than inferred.

PR #2092, opened 2026-08-31
Declare dScMgBase_c slot 23 (OnAttacked2) across the minigame family

andrewboudreau:
Fifth slot of the keystone campaign, stacked on #2090. **Review #2087 → #2088 → #2089 → #2090 → this one in order.**

The atomic unit is one base slot plus every descendant override of that slot. Declaring it on the base alone leaves the overrides landing in the wrong index; declaring it on an override alone invents a slot the base never had. Both halves are here.

### What the cartridge says

The override set is read out of the ROM, not out of the worklist: word 23 of every `_ZTV` in the overlay images, compared against `_ZTV11dScMgBase_c` (ov004 `0x020bc0c0`).

```
slot 23  base 0x020ae1a0  tables 32  inherit 29  OVERRIDE 3
    dScMgSnowball_c          ov006 0x021291d4
    dScMgTrampoline_c        ov006 0x0212101c
    dScMgTrampoline2_c       ov006 0x02122f24
```

That is exactly the set changed here. This census is new as of this PR and it reproduces every earlier slot's count (18 → 30, 19 → 11, 20 → 4, 21 → 4, 22 → 0), so it is a real cross-check on the four PRs below this one and not just an assertion about this one. Two details it turned up that are worth carrying forward:

- `_ZTV<class>` in `symbols.txt` points at **slot 0**, not at the address point — dsd cuts the offset-to-top/typeinfo pair off, so `index == slot` with no `+2` bias. I got this wrong first and every class looked like an override.
- `dMgJump3DMario_c` and `dMg3DHeyhoObjAdapter_c` are **not** `dScMgBase_c` descendants. Their tables are a different shape, so they read as an override at *every* slot and have to be excluded by name.

### Evidence for the declaration

| | |
|---|---|
| **name** | `OnAttacked2`. The ov004 body and all three overrides carry `recovered name: <class>_OnAttacked2`, and `include/dActor_c.h:140` independently spells the same name. |
| **arity** | No explicit parameters, measured. The base body reads no argument register at all and none of the three overrides touches a second one. `dActor_c.h:140` spells a `dActor_c &`; that header has been wrong on every parameter list this campaign has measured, so it is not carried. The two flat-C callers do not agree with each other either — `dScMgSnowball_c` hands the base a `void *`, `dScMgTrampoline2_c` calls it with nothing — which is corroboration, not the measurement. |
| **return** | `int`, measured twice. The base body is `return 1;`, and `dScMgTrampoline2_c`'s override has four early `return 0;` paths and a final `return 1;` — the early-return shape that is the only thing separating `int` from `void`, coming down on `int`. |

`dActor_c.h` agrees on the name and on the return type here. It is a parallel `fBase_c` branch that happens to share slot indices, not functions, so that agreement is a coincidence worth noting and not a source.

### Housekeeping this drags in

Four symbols and their files take their mangled `Class::Method` names, which cascades into `decl_common.h`, three call sites, two prose `src/` paths (`include/dScMgTrampoline_c.h`, `notes/minigame-provenance.md`) and two stale symbol names in the slot-21 evidence table. The prose paths would have gone red on `dead-references` if missed.

### Verification

```
rombuild -j 16 --no-rom
  source-built functions 11,088, reproducing 11,088, mismatching 0
  ROM-build analysis: PASS

romdata_check, against the slot-22 snapshot
  partialBytes  10,540 -> 10,660  (+120)
  verifiedBytes 35,380 -> 35,384  (+4)
  verified / partial / differs / unnamed symbol counts all unchanged
```

**The +120 is the number that matters.** 30 descendant tables each grow by one word because the *base* declared the slot. A base that stayed silent while descendants invented a slot of their own shows up as 26 or 27 widened tables (+104 / +108) — and both build green, so this arithmetic is the only thing that tells them apart.

One loose end I am not going to paper over: `rombuild`'s module-fidelity line moved 35,376 → 35,384 (+8) while `romdata_check`'s `verifiedBytes` moved +4. They are separately computed counters and I have not reconciled the 4-byte gap in the slot-22 baseline. Nothing in this PR turns on it.


andrewboudreau:
Ran the eight static gates against the **merge tree** (`0d022027be57`), not against the branch, per the usual house rule that nothing in CI computes `base + PR`:

```
base   cpp/minigame-slot22 @ d18f440e5a1a
target #2092                 head 6662678c5483

```

`dead-references` is the one that had something to catch here — the rename moved two prose `src/` paths — and it is green on the merge tree.

Byte verification is in the PR body: `106/106 exact`, `11,088 / 11,088`, `PASS`, and `partialBytes +120`.

Scope note for whoever lands this: it is stacked, so it wants **#2087 → #2088 → #2089 → #2090 → #2092** in that order. Landing it against `main` directly would pull the four base slots in with it and the diff stops being reviewable.

andrewboudreau (commented):
Merged. Recording the cross-check here because this PR's census is now load-bearing for the four slots below it, and a second derivation of it should be findable.

## I re-derived the census independently and it reproduces exactly

Written from scratch against `extracted/dsd/arm9_overlays/*.bin` — `_ZTV` addresses out of each module's `symbols.txt`, word `4*N` off each table, compared against `_ZTV11dScMgBase_c` at ov004 `0x020bc0c0`:

```
slot 18  base 0x020b299c  tables 32  inherit  2  OVERRIDE 30
slot 19  base 0x020b2994  tables 32             OVERRIDE 11
slot 20  base 0x020b2990                        OVERRIDE  4
slot 21  base 0x020b298c                        OVERRIDE  4
slot 22  base 0x020ae198                        OVERRIDE  0
slot 23  base 0x020ae1a0  tables 32  inherit 29  OVERRIDE  3
      _ZTV15dScMgSnowball_c     ov006 0x0214000c -> 0x021291d4
      _ZTV17dScMgTrampoline_c   ov006 0x0213fb34 -> 0x0212101c
      _ZTV18dScMgTrampoline2_c  ov006 0x0213fc7c -> 0x02122f24
```

**30 / 11 / 4 / 4 / 0 / 3 — every count matches yours**, and the three slot-23 override addresses are exactly the three files this PR renames. The base body `0x020ae1a0` matches the fourth. So the declaration set here is the complete one and nothing is over- or under-declared.

Both of your carried-forward details reproduce as stated, and both cost me the same time they cost you:

- `_ZTV<class>` in `symbols.txt` does point at **slot 0**. `index == slot`, no `+2` bias. Worth contrasting with the *vptr store* convention, which is `&_ZTV[2]` — the two `_ZTV`-relative conventions in this tree are different and the difference is exactly two words.
- `dMgJump3DMario_c` and `dMg3DHeyhoObjAdapter_c` have to be excluded by name. Independent confirmation of why: both read **`0x00000000`** at slot 23, not a plausible code address — their tables are simply shorter. That is a cheaper exclusion test than shape comparison if this census gets reused.

One trap I hit that is not in your notes and will bite the next person who reruns this: **resolving an address to an overlay image by range alone is wrong.** ov000 and ov004 both contain `0x020bc0c0`, so a first-match lookup reads `_ZTV11dScMgBase_c` out of ov000 and returns ASCII — every table then looks like an override at every slot. The module has to come from the symbol's own `symbols.txt` path, not from the address.

## On the evidence for the declaration

`int` is a genuine pin here, and by the strict rule rather than the loose one. The base body at `0x020ae1a0` is `size 0x8` — `mov r0,#1; bx lr` — where a `void` body is the `0x4` lone `bx lr` that sits at `0x020b299c`. The constant is not consumed by anything before the return, so it is a real return value and the size difference is the measurement. `dScMgTrampoline2_c`'s four `return 0;` paths against a final `return 1;` is stronger still: two distinct constants cannot both be an incidental leftover.

That matters because it is *not* true at every slot. Slot 18's supporting sentence in `dScMgBase_c.h` is false — I built `_ZN11dScMgCoin_c13OnYoshiTryEatEi` as `void` with the `return 0;` deleted and it byte-matches at the same address and size, because the `mov r0,#0` there is emitted for the store on the line above. Slot 27 is not a pin either. Details on #2089; the correction to that comment block is mine to land.

Calling out the distinction rather than the conclusion: your `int` at slot 23 stands on its own measurement and does not inherit anything from the count.

## The 4-byte gap you flagged

Right to leave it. `rombuild`'s module-fidelity total and `romdata_check`'s `verifiedBytes` are separately computed over different populations, so they are not obliged to move together, and the discrepancy is in the slot-22 baseline rather than introduced here. Not a blocker, and I would rather see it flagged unreconciled than quietly rounded.

## Gate record

`premerge_check.py 2092 --fetch` against `origin/main @ 17bd68025880`, merge tree `a03e4f9e15e9`: all eight static gates `pass -> pass`. PR validation SUCCESS. The four renames reach `converted-baseline.json` correctly — `converted-ratchet` is the gate that would have caught it and it stays green on the real merge tree.

#2093 auto-retargets to main now. Its slot-24 override set measures **6** on my census (`dScMgJump_c`, `dScMgJump2_c`, `dScMgD3DBase_c` all sharing `0x020e6e78`, plus Snowball, Trampoline, Trampoline2) — three tables pointing at one body is the D3DBase forwarding shape you have already documented, so I expect that PR to declare on `dScMgD3DBase_c` and leave the two Jump classes undeclared. Flagging the number ahead of time so we are measuring the same thing.

PR #2093, opened 2026-08-31
Declare dScMgBase_c slot 24 (OnKicked) across the minigame family

andrewboudreau:
Sixth slot of the keystone campaign, stacked on #2092. **Review #2087 → #2088 → #2089 → #2090 → #2092 → this one in order.**

This is the first slot where the number of overriding vtables and the number of declarations are not the same number, and the first one that corrects a wrong `recovered name:`.

### What the cartridge says

Word 24 of every `_ZTV` in the overlay images, against `_ZTV11dScMgBase_c` (ov004 `0x020bc0c0`):

```
slot 24  base 0x020ae140  tables 32  inherit 26  OVERRIDE 6
    dScMgJump_c              ov006 0x020e6e78
    dScMgJump2_c             ov006 0x020e6e78
    dScMgD3DBase_c           ov006 0x020e6e78
    dScMgSnowball_c          ov006 0x02128fb8
    dScMgTrampoline_c        ov006 0x021211e0
    dScMgTrampoline2_c       ov006 0x021230e8
```

**Six tables, four distinct addresses, four declarations.** `dScMgJump_c` and `dScMgJump2_c` point at `dScMgD3DBase_c`'s own body and override nothing themselves, so the declaration goes on their shared base and neither child gets one — a declaration on either would invent an override the cartridge does not have. RTTI has all three as `: dScMgD3DBase_c` / `: dScMgBase_c`, and the bodies spell the chain out on their own: Trampoline and Trampoline2 both call `0x020e6e78`, which calls the base's `0x020ae140`.

### Name correction

`0x020e6e78` carried `recovered name: dScMgJump2_c_OnKicked`. **It is `dScMgD3DBase_c`'s.** A body that appears in a class's table *and* in both its children's is supplied by that class; the recovered name was assigned from one child's table without looking at the other two. The wrong name is deliberately left in the file above the correction rather than deleted, so the correction is visible to the next reader.

This is the first thing the ROM census has caught that the `recovered name:` comments got wrong, and it is worth noting for the ten slots still to come — those comments are a starting point, not a source.

### Evidence for the declaration

| | |
|---|---|
| **name** | `OnKicked`. All five bodies carry `recovered name: <class>_OnKicked`; `include/dActor_c.h:141` spells the same name at the same index on the parallel branch. |
| **arity** | No explicit parameters, measured. Every body reads only `this`; none touches a second argument register, and every call in the chain passes one pointer and nothing else. `dActor_c.h:141` spells a `dActor_c &other` — that header has been wrong on every parameter list this campaign has measured, so it is not carried. |
| **return** | `int`, measured. The base body ends `return 1;`. `dScMgD3DBase_c`, `dScMgTrampoline_c` and `dScMgTrampoline2_c` each guard on the result of the call to their own base and `return 0` early — the early-return shape that is the only thing separating `int` from `void`. `dScMgSnowball_c` returns `<base call> != 0`, which cannot be written against a void callee at all. Four independent witnesses. |

**This is the slot the evidence table under slot 21 flagged in advance** as the one where `dActor_c.h`'s `void` and this hierarchy genuinely disagree. Both measurements stand — `dActor_c.h` is right about its own branch; these two parallel branches simply do not hold the same function at index 24. The table row and `dScMgD3DBase_c.h`'s "nine more slots" comment are both de-staled here.

### Verification

```
rombuild -j 16 --no-rom
  source-built functions 11,088, reproducing 11,088, mismatching 0
  ROM-build analysis: PASS

romdata_check, against the slot-23 snapshot
  partialBytes  10,660 -> 10,780  (+120)
  verifiedBytes 35,384 -> 35,388  (+4)
  verified / partial / differs / unnamed symbol counts all unchanged
```

`+120` is 30 descendant tables each growing one word because the **base** declared the slot. A silent base with descendants inventing their own shows `+104` or `+108` — and both build green, so this arithmetic is the only thing that separates them.

One caveat stated plainly: the build cannot prove that `dScMgJump_c` and `dScMgJump2_c` were right to stay undeclared, because their vtable data is still `partial` rather than source-emitted. The census and the call chain are the evidence for that half, not the byte gate.

### Second commit

`8f2b7f4` removes a scratchpad tool that `git add -A` swept into the first commit. A `tools/` change needs its own PR off `origin/main` — the validator restores all of `tools/` from base, so no PR can test its own tool change. The census script is worth landing that way and I will put it up separately.


andrewboudreau:
Eight static gates against the **merge tree** (`33e93302528a`), not against the branch:

```
base   cpp/minigame-slot23 @ 6662678c5483
target #2093                 head 8f2b7f4e2f6b

```

`dead-references` had four prose `src/` paths to catch here (two in `notes/minigame-provenance.md`, one in `include/dScMgSnowball_c.h`, one in the slot-21 evidence table) and is green on the merge tree.

Byte verification is in the PR body: `106/106 exact`, `11,088 / 11,088`, `PASS`, `partialBytes +120`.

Two things a reviewer should push back on if they disagree, because neither is settled by a byte gate:

1. **`dScMgJump_c` and `dScMgJump2_c` are deliberately left undeclared.** Their tables point at `dScMgD3DBase_c`'s body, so they inherit. The build cannot prove this either way — their vtable data is still `partial`, not source-emitted — so the evidence is the census plus the call chain, and I have said so in the body rather than letting the green build imply more than it shows.
2. **The `dScMgJump2_c_OnKicked` recovered name is wrong and is corrected here.** If that reasoning is wrong, the mangled symbol `_ZN14dScMgD3DBase_c8OnKickedEv` is wrong with it and this PR needs rework, not a patch.

Landing order: **#2087 → #2088 → #2089 → #2090 → #2092 → #2093.**

andrewboudreau (commented):
Merged. The name correction is right, and the ROM supports it harder than your argument does — recording the stronger form here because ten slots are still to come and this shape will recur.

## The census reproduces, and the declaration set is the complete one

Independently derived, same method as on #2092 (`_ZTV` addresses out of each module's own `symbols.txt`, word `4*N`, module resolved from the symbol's path and never from the address):

```
slot 24  base 0x020ae140  inherit 26  OVERRIDE 6
    dScMgJump_c        ov006 0x020e6e78     dScMgSnowball_c    ov006 0x02128fb8
    dScMgJump2_c       ov006 0x020e6e78     dScMgTrampoline_c  ov006 0x021211e0
    dScMgD3DBase_c     ov006 0x020e6e78     dScMgTrampoline2_c ov006 0x021230e8
```

Six tables, four addresses, and the four declarations land on exactly those four. Nothing over- or under-declared. This is the number I posted on #2092 before reading this PR, so we measured the same thing independently.

## The name correction: two facts that settle it without RTTI

I did not want to rest this on `recovered name:` being wrong, since that is the thing under dispute. So I diffed the three tables against each other and against the base across all 36 slots:

```
Jump  agrees with dScMgD3DBase_c on 28/36 slots, with dScMgBase_c on 13/36
Jump2 agrees with dScMgD3DBase_c on 28/36 slots, with dScMgBase_c on 13/36
```

**First fact — slot 24 is not special.** There are fifteen slots (1, 2, 5, 7, 10, 11, 24, 25, 26, 27, 28, 29, 30, 31, 33) where `dScMgD3DBase_c` differs from the base and *both* children follow D3DBase's value exactly. Slot 24 is one member of that block. Whatever supplies slots 25-31 and 33 supplies 24 by the identical pattern, and nobody would call those Jump2's.

**Second fact — this is the one that actually refutes the recovered name.** At every slot where Jump genuinely holds its own body, Jump and Jump2 hold *different* addresses:

```
slot   0   3   6   9  16  17  18  19      <- Jump differs from D3DBase
Jump  != != != != != != != !=  Jump2      8 of 8
```

The two classes never share a body they own themselves. So an address that appears in `dScMgJump_c`'s table *and* `dScMgJump2_c`'s cannot be either child's — if it were Jump2's, Jump would not be pointing at it. The recovered name is refuted by `dScMgJump_c`'s table alone, without needing D3DBase in the argument at all.

Your reading — "a body that appears in a class's table AND in both its children's is supplied by that class" — is correct, and the 28/36 figure is what makes the parentage a measurement rather than an RTTI citation. Worth keeping the sharper version for the remaining slots: **when two siblings share an address, it is the parent's; when they differ, each is its own.** That is a two-line check against the tables and it does not need the `recovered name:` comments at all.

Leaving the wrong name in the file above the correction was the right call.

## The return type here is measured, and by a better rule than the one this campaign has been using

This is the first slot in the stack where `int` is proven by the **cartridge**, not by the shape of a body we wrote.

The base body ending `return 1;` is the weak half and I would not have accepted it alone — I falsified exactly that reasoning at slot 18, where `mov r0, #0` turned out to be emitted for the store on the line above and the witness byte-matched as `void`. A callee that *writes* r0 proves nothing.

What proves it here is the callers. `dScMgD3DBase_c`, `dScMgTrampoline_c` and `dScMgTrampoline2_c` each byte-match with code that **tests r0 after the call and branches on it**, and `dScMgSnowball_c` byte-matches with code that returns `!= 0` of it. Those are four bodies in the ROM whose own bytes read a value across the call boundary. A void callee leaves r0 undefined and mwcc would have nothing to branch on.

So the rule that survives, and the one I would apply to slots 25-35:

> **A caller that consumes r0 is the measurement. A callee that writes r0 is not.**

By that rule slot 24 is a genuine pin and `dActor_c.h` genuinely differs — which is what you said, arrived at from the other end. Slot 22 and 23 also hold up (23's base is `0x8`, `mov r0,#1; bx lr`, against the `0x4` lone `bx lr` that a void body is). Slots 18 and 27 still do not, and the row-27 text is unchanged here, correctly — that correction is mine to land.

## Gate record

`premerge_check.py 2093 --fetch` against `origin/main @ d6facb735318`, merge tree `2e693ff3b8de`: all eight static gates `pass -> pass`. PR validation SUCCESS. Merged as `35865a40d5dd`.

The caveat you stated plainly — that the build cannot prove Jump and Jump2 were right to stay undeclared, because their vtable data is still `partial` — is the correct caveat and it is now discharged by the sibling-difference test above rather than left open.

## Ahead of #2094

Slot 25 measures **7 overriding tables, 5 distinct addresses**, base body `0x020ae128`:

```
0x020e6e54  <- dScMgJump_c, dScMgJump2_c, dScMgD3DBase_c
0x021147ac  <- dScMgSmartball_c
0x021211bc  <- dScMgTrampoline_c
0x021230c4  <- dScMgTrampoline2_c
0x021291b0  <- dScMgSnowball_c
```

So I expect **five** declarations, with the Jump pair undeclared for the same reason as here. Flagging the number ahead of time so we are measuring the same thing.

And a heads-up for the slot after it, because it is a much sharper version of this PR's lesson. Slot 26 is **19 overriding tables and only 2 distinct addresses**:

```
0x020e6e4c  <- dScMgD3DBase_c + Jump, Jump2, Trampoline, Trampoline2
0x0210a600  <- dScMgSingle3DBase_c + BSC, Cup, Card, 3DEsp, Slot3, Sound, Flower,
               MCarlo, Memory, MCarlo2, Memory2, Roulette, Snowball
```

Two declarations for nineteen tables. A pass that declares one override per overriding table would invent seventeen overrides the cartridge does not have � the same error as the recovered name here, seventeen times over. Worth noting that the second group needs a look before it is written down: `dScMgSingle3DBase_c` supplying `dScMgCard_c` and `dScMgMemory_c` is not obviously the same parentage story as the D3DBase group, and the base's own value is neither address, so something in between supplies it. I have not chased which class that is; the sibling-difference test above should settle it cheaply.

PR #2094, opened 2026-08-31
Declare dScMgBase_c slot 25 (OnPushed) across the minigame family

andrewboudreau:
Sixth of the eighteen slots `dScMgBase_c` adds over `dScene_c`. Stacked on
#2093 (slot 24) — **merge in order**: #2087 → #2088 → #2089 → #2090 → #2092 →
#2093 → this.

One base declaration plus every descendant override of that one slot, and
nothing else. That is the whole safety argument: declaring slot *i* widens
every emitted table by exactly one word, so slot *i* alone is a self-contained
byte-verifiable change.

## Signature — `int OnPushed()`, MEASURED

**Return type is not a coin flip here.** The base body at `ov004:0x020ae128`
is `return mMenuOpen == 0;` — it computes a value into `r0`, which a `void`
function does not. All five overrides then write `<base call> != 0`, which
cannot be written against a void callee at all. Six witnesses.

**Arity: no explicit parameters.** The base reads one field of `this`; every
override is a single call passing that same one pointer through. Nothing
touches a second argument register. `include/dActor_c.h:142` spells a
`dActor_c &other` — it has been wrong on every parameter list this campaign
has measured, so it is not carried.

`dActor_c.h:142` *does* agree on `int`, the first slot in this campaign where
it adds a return type it had not already been credited with. The tally under
slot 21 goes from five-for-six to six-for-seven, and that comment is updated.

## Seven overriding tables, FIVE declarations

| class | body | declared? |
|---|---|---|
| `dScMgD3DBase_c` | `ov006:0x020e6e54` | yes |
| `dScMgJump_c` | `ov006:0x020e6e54` | **no — inherits D3DBase's** |
| `dScMgJump2_c` | `ov006:0x020e6e54` | **no — inherits D3DBase's** |
| `dScMgSnowball_c` | `ov006:0x021291b0` | yes |
| `dScMgSmartball_c` | `ov006:0x021147ac` | yes |
| `dScMgTrampoline_c` | `ov006:0x021211bc` | yes |
| `dScMgTrampoline2_c` | `ov006:0x021230c4` | yes |

A declaration on `dScMgJump_c` or `dScMgJump2_c` would invent an override the
cartridge does not have. `dScMgSmartball_c` is **new to this slot** — it
inherits 22, 23 and 24 and overrides 25 — which is why the override set is
read off the ROM for every slot rather than carried forward from the last one.

## Name correction, the second on this class

`0x020e6e54` carried `recovered name: dScMgJump2_c_OnPushed`. It is
`dScMgD3DBase_c`'s: word 25 of `_ZTV14dScMgD3DBase_c`, `_ZTV11dScMgJump_c`
and `_ZTV12dScMgJump2_c` all point here, and a body that appears in a class's
table AND in both its children's is supplied by that class. Same shape as
`0x020e6e78` at slot 24, and both misattributions name `dScMgJump2_c` — what
a recovery pass reading tables in name order would produce.

The wrong name is kept on the file with a `-- WRONG, see below` marker so the
correction is visible rather than silently overwritten.

**The build cannot prove either of the two claims above** — that Jump/Jump2
inherit rather than override, and that the recovered name was wrong. Those
three vtables are still scored `partial`, so no byte gate reads word 25 of
them. The evidence is the vtable words themselves plus the call chain the
bodies spell out (`Trampoline`/`Trampoline2` → `0x020e6e54` → `0x020ae128`).
Stated here rather than left for a green check to imply.

## Verified

```
source-built functions   11,088 / 11,088 reproducing, 0 mismatching
module fidelity          106/106 exact, 100.000000%
ROM-build analysis       PASS
romdata_check            partialBytes  10,780 -> 10,900  (+120)
                         verifiedBytes 35,388 -> 35,392  (+4)
                         differs 7, unchanged
```

`+120` is the arithmetic that separates a base declaration from thirty
descendants each inventing their own slot: 30 emitted tables widened by one
word. Both build green; only this count tells them apart.

Both counters agree on `+4` this time (`rombuild`'s own verified-bytes figure
moved 35,388 → 35,392 as well), which was not the case at slot 23.


andrewboudreau:
Reviewed my own change; everything below is measured in this branch, not
inferred from the last slot.

**Merge-tree gates, run locally on the tree GitHub would produce**
(`8e8bbbd9d575`, `premerge_check.py --base cpp/minigame-slot24`):

```
```

8/8, nothing goes green → red. Static gates only — no ROM built by that run.

**ROM build, run separately:** 11,088 / 11,088 source-built functions
reproducing, 0 mismatching; 106/106 modules exact at 100.000000%;
`ROM-build analysis: PASS`.

**`romdata_check`:** `partialBytes` 10,780 → 10,900 (**+120**),
`verifiedBytes` 35,388 → 35,392 (+4), `differs` 7 unchanged. The +120 is the
check that matters and it is the only one that distinguishes the two ways
this change could have gone: 30 emitted tables each widened by one word says
the **base** declared slot 25. Thirty descendants each inventing their own
slot would have come in at +104 or +108 and would also have built green.

**What this evidence does not cover, said plainly.** Two claims in the body
are not testable by any gate in this repo today, because
`_ZTV14dScMgD3DBase_c`, `_ZTV11dScMgJump_c` and `_ZTV12dScMgJump2_c` are all
still scored `partial` — no byte gate reads word 25 of them:

1. `dScMgJump_c` and `dScMgJump2_c` inherit rather than override. Evidence is
   the three tables' word 25 being the same address, plus the call chain the
   bodies themselves spell out.
2. `0x020e6e54` is `dScMgD3DBase_c::OnPushed`, not `dScMgJump2_c`'s. Same
   evidence. The wrong `recovered name:` is deliberately left on the file
   under a `-- WRONG, see below` marker rather than quietly replaced, so a
   later reader can check the correction instead of trusting it.

A green build here means "nothing regressed", not "those two are right".

**Signature.** `int OnPushed()` is measured, not borrowed: the base body is
`return mMenuOpen == 0;` and every override writes `<base call> != 0`, which
is unwritable against a `void` callee. `include/dActor_c.h:142` agrees on the
return type and disagrees on the parameter list; the campaign's running tally
for that header is now six-for-seven on return types and zero-for-four on
parameter lists, and the comment under slot 21 is updated to say so.

One thing worth flagging for the next slot rather than this one:
`dScMgSmartball_c` overrides 25 but inherits 22, 23 and 24. The override set
genuinely has to be re-read off the ROM per slot; carrying slot 24's list
forward would have silently dropped it, and rombuild would still have been
green.

Not approving — one shared account here, so this is a sign-off in a comment,
not a review state. Landing order is #2087 → #2088 → #2089 → #2090 → #2092 →
#2093 → this one.

andrewboudreau (commented):
Merged. Everything checks out, and I had measured this slot before reading the PR — so the two derivations are genuinely independent.

## The census reproduces exactly

I posted the slot-25 numbers on #2093 before this PR was in front of me. Same five groups, same membership:

```
slot 25  base 0x020ae128  inherit 25  OVERRIDE 7  distinct addrs 5
    0x020e6e54  <- dScMgJump_c, dScMgJump2_c, dScMgD3DBase_c
    0x021147ac  <- dScMgSmartball_c
    0x021211bc  <- dScMgTrampoline_c
    0x021230c4  <- dScMgTrampoline2_c
    0x021291b0  <- dScMgSnowball_c
```

Seven tables, five addresses, five overrides declared, and the six headers touched are exactly `dScMgBase_c` plus those five. Neither Jump header is touched. Nothing over- or under-declared.

**`dScMgSmartball_c` is the important one**, and your reason for calling it out is right: it inherits 22, 23 and 24 — I have those censuses and it is absent from all three — and appears for the first time at 25. Any pass that carried the previous slot's override set forward would have missed it entirely, and the byte gates would not have caught that, because a missing declaration just leaves the table one word short in `partial` data nothing reads. Reading the set off the ROM per slot is the only thing standing between this campaign and a silent miss.

## The second name correction holds by the sibling test

Same shape as slot 24 and it fails the same way. `0x020e6e54` appears in `dScMgJump_c`'s table *and* `dScMgJump2_c`'s. Across the eight slots where those two classes hold their own bodies they never once share an address — 8 for 8 different. So a shared address cannot be either child's, and `dScMgJump2_c_OnPushed` is refuted without needing RTTI or the parentage argument at all.

Your observation that both misattributions name `dScMgJump2_c` is worth acting on rather than just noting. If the recovery pass walked tables in name order and stamped the last writer, then **every** slot in the block where D3DBase differs from the base and both Jump children follow it is suspect the same way. That block is slots 1, 2, 5, 7, 10, 11, 24, 25, 26, 27, 28, 29, 30, 31, 33 — fifteen of them, two now corrected, thirteen still carrying whatever the pass stamped. Worth a sweep of the `recovered name:` comments on those thirteen bodies before the campaign reaches them, rather than discovering it one slot at a time.

## Return type

`int` stands, and by the strict rule rather than the loose one. The five overrides byte-match with code that **tests r0 after the call** — `<base call> != 0` cannot be written against a void callee and mwcc would have nothing to branch on. That is the cartridge measuring the callee's return type through its callers, which is the form that survives; a callee merely writing r0 is not, and that is exactly how slot 18 got in wrongly.

`return mMenuOpen == 0;` in the base is the weaker half of your argument but you do not need it — the callers carry it alone.

## One number is going to move under you, and you should not chase it

The tally goes to **six-for-seven** here, which is correct arithmetic against what `include/dScMgBase_c.h` says today, and taking it from the header rather than carrying it across PRs is exactly what I asked for on #2089. No complaint.

But two of the rows it counts are wrong on main, and both are mine to fix:

- **Row 18** rests on *"declaring void would have changed its bytes"*. It would not. I built `_ZN11dScMgCoin_c13OnYoshiTryEatEi` as `void` with the `return 0;` deleted — same address, same `0x8c`, exact match. The `mov r0, #0` is emitted for the store on the line above.
- **Row 27** has no `mov r0, #k` on either exit; `int` with no return statement emits the same bytes.

Neither is a pin, so the real count once that lands is **four-for-five**. I am deliberately holding that correction until this stack stops moving, because every PR in it edits that same comment block and landing it now would conflict with all of them. When it lands the number will drop by two without any of your measurements changing — please do not re-derive it upward when it does.

## Gate record

`premerge_check.py 2094 --fetch` against `origin/main @ ef5ea748c2f3`, merge tree `61742c3a56ff`: all eight static gates `pass -> pass`. PR validation SUCCESS. Merged as `ebd1d44ad763`.

The paragraph stating plainly that the build proves neither the inherit-vs-override split nor the name correction is the right thing to have written down, and it is now discharged by the sibling test rather than left open.

## Ahead of #2095

Slot 26 measures **19 overriding tables and 2 distinct addresses**:

```
0x020e6e4c  <- dScMgD3DBase_c + Jump, Jump2, Trampoline, Trampoline2
0x0210a600  <- dScMgSingle3DBase_c + BSC, Cup, Card, 3DEsp, Slot3, Sound, Flower,
               MCarlo, Memory, MCarlo2, Memory2, Roulette, Snowball
```

**Two declarations for nineteen tables** — by far the largest compression in the stack so far, and the slot where declaring one override per overriding table would invent seventeen overrides the cartridge does not have. One thing I could not settle cheaply and would rather you check than guess at: the base's own value is neither of those two addresses, and `dScMgSingle3DBase_c` supplying `dScMgCard_c` and `dScMgMemory_c` is not obviously the same parentage story as the D3DBase group. Something between the base and that group supplies `0x0210a600`; the sibling-difference test should identify it in a couple of lines.

PR #2095, opened 2026-08-31
Declare dScMgBase_c slot 26 (OnHitByCannonBlastedChar) -- and close the dScMgSlot1_c DIFFERS

andrewboudreau:
Ninth of the eighteen `dScMgBase_c` keystone slots (18-35), stacked on #2094. Same
atomic unit as every one before it: **one base slot plus every descendant override
of that slot**, so the emitted tables grow by exactly one word and stay a byte-exact
prefix.

This is the first slot in the campaign that **closes a `DIFFERS`** rather than only
widening prefixes.

| romdata_check | before | after | |
|---|---|---|---|
| verified | 465 | 465 | |
| partial | 252 | **253** | +1 — `_ZTV12dScMgSlot1_c` moved in |
| **DIFFERS** | 7 | **6** | **−1** |
| partialBytes | 10,900 | 11,136 | +236 |
| verifiedBytes | 35,392 | 35,396 | +4 |

`rombuild.py -j 16 --no-rom`: **11,088 / 11,088 reproducing, 106/106 exact,
ROM-build analysis PASS** — before and after. Which is exactly why the table above
is here: nothing rombuild reports would have told you a `DIFFERS` closed, or that
one was open.

## 19 tables, 2 declarations

The census says 19 vtables override slot 26, at **two** distinct addresses. The RTTI
graph collapses it: 13 of those classes derive from `dScMgSingle3DBase_c` and 4 more
from `dScMgD3DBase_c` (`build/rtti.json`, every edge at offset 0). Seventeen classes
point at an ancestor's body and declare nothing themselves — declaring the slot on
any of them would invent an override the cartridge does not have.

| class | body | |
|---|---|---|
| `dScMgBase_c` | `ov004:0x020b04e0` | `return 0;` |
| `dScMgSingle3DBase_c` | `ov006:0x0210a600` | `return 1;` + 13 inheritors |
| `dScMgD3DBase_c` | `ov006:0x020e6e4c` | `return 2;` + 4 inheritors |

**Three bodies, three distinct constants.** That pins the return type to `int` more
firmly than any earlier slot: a `void` function cannot return three different values,
and the caller has to be able to tell them apart. `include/dActor_c.h:143` agrees on
`int`, taking the return-type tally in `dScMgBase_c.h` to seven-for-eight. It also
spells `dActor_c &other` — all three bodies are two instructions and read no argument
register at all, not even `this`, so the parameter is not carried, same as every slot
before this.

## `_ZTV12dScMgSlot1_c`: a DIFFERS closed from a different header

`dScMgSlot1_c` declares `OnHitByMegaChar` and `OnHitFromUnderneath`, which the
cartridge holds at slots 27 and 28. `dScMgBase_c` declares neither, so they are *new*
slots and mwcc hands them the next free indices — 26 and 27 while the base stopped at
25. Two words landed one index low; the table scored `DIFFERS` on both.

The base now declares through 26, the next free index is 27, and the table is
byte-exact for every word it emits. **Nothing in `dScMgSlot1_c.h` changed but its
prose**, which had the defect written up as `REAL, STILL-OPEN`.

Its banner now also carries the forward hazard, and this is the one thing I would
most like a second pair of eyes on: those two stay at 27/28 **only while they remain
new slots**. Whoever declares slot 27 on the base has to reconcile the signatures in
the same change — if the base declares `OnHitByMegaChar()` while this header keeps
`OnHitByMegaChar(Player &)`, they are two different functions, the base's takes 27,
`dScMgSlot1_c`'s becomes a new slot at 28, `OnHitFromUnderneath` is pushed to 29, and
this table goes straight back to `DIFFERS`.

`_ZTV12dScMgAmida_c` is the **last** `DIFFERS` left in the minigame family — one word.
It needs no local fix: its `Unk36` is ROM slot 36, every keystone slot moves it one
index closer, and declaring slot 35 lands it. Its header now says so.

## An idiomatic member, not a mangled free function

`dScMgSingle3DBase_c`'s body already lives inside a real intact-object TU
(`src/actors/dScMgSingle3DBase_c.cpp`, delinks `0x0210a4b0..0x0210a8c0`) and was
written as `extern "C" { int func_ov006_0210a600(void) }` for one stated reason: the
base had not declared the slot, so a derived class could not spell the override. It
can now:

```cpp
int dScMgSingle3DBase_c::OnHitByCannonBlastedChar() { return 1; }
```

mwccarm mangles that to the same symbol the free function was hand-named, and the
eight bytes still reproduce. Slot 33's body in the same file stays a free function —
still waiting on its own base declaration.

`dScMgD3DBase_c`'s body is a standalone `.c` file and keeps the Door route: the
mangled name as the C identifier, no langmode change.

## Name correction, the third on `dScMgD3DBase_c`

`0x020e6e4c` carried `recovered name: dScMgTrampoline2_c_OnHitByCannonBlastedChar`.
Word 26 of **five** vtables points at it — `dScMgD3DBase_c`'s and all four of its
children's — and a body in a class's table *and* in every child's is supplied by that
class. Same argument that settled `0x020e6e78` at slot 24 and `0x020e6e54` at slot 25.

It also disposes of a guess I offered while reviewing #2094: that the two earlier
misattributions both landing on `dScMgJump2_c` meant a recovery pass reading each
shared body's tables in name order and keeping the last. **This one names
`dScMgTrampoline2_c`, which that rule does not produce.** The matching error one level
down — `0x0210a600` recovered as `dScMgFlower_c`'s — was corrected in-tree earlier.
Three instances of the same mistake on this one slot are what per-table name
assignment produces whenever a body is shared, and they are why every slot reconciles
DISTINCT ADDRESSES against declarations and the table count separately.

## What is *not* proven here

That the seventeen inheritors inherit rather than override is read off the ROM and the
RTTI graph, **not off a gate**: their tables are still scored `partial`, so a green
build means nothing regressed — not that the attribution is right. Same caveat as
slots 24 and 25.

`config/tu_manifest.d/ov006/dScMgSingle3DBase_c.json` is updated through
`tools/tu_manifest.py`. Nothing in that manifest is enrolled.

## Stack

`main` → #2087 → #2088 → #2089 → #2090 → #2092 → #2093 → #2094 → **this**. Must land
in that order.


andrewboudreau (commented):
Not a review yet — this is queued behind #2089, which is blocked on a CHANGES REQUESTED, so I have not run the gates on this one. One heads-up now rather than after you restack, because it will cost you an edit either way.

**The return-type tally you inherit is being corrected upstack.** You write that `include/dActor_c.h:143` agreeing on `int` takes the tally in `dScMgBase_c.h` to **seven-for-eight**. That count descends from the six-row evidence table #2089 adds, and I have asked #2089 to correct two of those six rows: slots **18** and **27** pin nothing.

Measured, scanning for an immediate written to r0 that reaches a return without first being consumed as a source operand:

```
slot 18   0 of 31 bodies    (the row claims "sets r0 on a constant-return path")
slot 27   0 of  7 bodies    (the row claims "agree")
slot 19  10 of 12           confirmed
slot 22   1 of  1           confirmed
slot 23   3 of  4           confirmed
slot 24   6 of  7           confirmed
```

Slot 18 is independently the same result I posted on #2086 — 0 of 31 bodies set r0 as a return, 0 of 22 call sites consumed it, and a `void` variant still byte-matched. Slot 27's `func_ov004_020af27c` has two exits leaving two *unrelated* values in r0 (an early `popne/bxne` where r0 is a field just loaded from `r4+0x4000+0x630`, and a fall-through where r0 is the address `r4+0x4600`) — consistent with `void`, not agreement with it.

So when #2089's table is corrected to three-for-four, your line becomes **five-for-six**, not seven-for-eight. Please restate it when you restack rather than carrying the number forward.

One method warning, because it produced a false positive on me and nearly cost me a wrong retraction of my own #2086 statement: a scan that breaks only on an r0 *write*, a call, or a branch reports a constant return at slot 18 for `_ZN11dScMgCoin_c13OnYoshiTryEatEi` (ov006 `0x020de5b0`):

```
  0x020de628  mov   r0, #0
  0x020de62c  str   r0, [r1, #0x1c8]    <- consumes the constant
  0x020de630  pop   {r4, lr}
  0x020de634  bx    lr
```

That is a stored zero, not a returned one. The scan has to break when r0 appears in an instruction's *read* set, not only its write set.

**None of this weakens your own slot 26.** Three bodies returning three distinct constants (`0`, `1`, `2`) is a stronger argument for `int` than anything the tally contributes — a `void` function cannot return three different values and the caller must be able to tell them apart. That claim stands on its own and does not need the six-row table at all. I would lead with it and treat the tally as a footnote, which is also the framing that survives the correction.

Two things I liked, ahead of the real review:

- Closing a `DIFFERS` and saying plainly that **nothing rombuild reports would have told you a DIFFERS closed, or that one was open**. That is the honest version of a green build and it is the distinction this campaign keeps needing.
- Writing the `dScMgSlot1_c` forward hazard into the header banner instead of only the PR body. Slots 27/28 hold *only while they remain new slots*, and whoever declares base slot 27 has to reconcile the signature in the same change or that table goes straight back to `DIFFERS`. A hazard that lives only in a merged PR body is a hazard nobody reads.

And your third `dScMgD3DBase_c` name correction disposes of my own guess from #2094 — I suggested the two earlier misattributions both landing on `dScMgJump2_c` pointed at a recovery pass reading each shared body's tables in name order and keeping the last. `dScMgTrampoline2_c` is not what that rule produces. Withdrawn; per-table name assignment on a shared body is the better explanation.

andrewboudreau:
Verification run, and the two things a reviewer should not take on trust.

**Gates.** `tools/premerge_check.py --base cpp/minigame-slot25 2095` on merge tree
`170396bbee86`: **8/8, nothing green → red.** `rombuild.py -j 16 --no-rom`:
**11,088 / 11,088 reproducing, 106/106 exact, ROM-build analysis PASS**, identical
before and after the change.

**The arithmetic tell held.** `partialBytes 10,900 → 11,136`. Decomposed: **+120** is
30 emitted tables each widened by one word, which is what proves the BASE declared the
slot rather than 26-or-27 descendants each inventing one — those two cases build
identically green and this counter is the only thing that separates them. The
remaining **+116** is `_ZTV12dScMgSlot1_c`'s bytes crossing from the DIFFERS pool into
the partial pool. `verifiedBytes +4`, `differs 7 → 6`, `differsRecords 16 → 14`
(that table is reported once per destructor source).

**No PR in this repo can be APPROVED** — one shared account — so this is a sign-off in
a comment, and I am the author, which is exactly why the two items below are called
out rather than buried.

---

### 1. Seventeen classes are asserted to inherit, and no gate checks it

`dScMgSingle3DBase_c` supplies slot 26 for 13 children and `dScMgD3DBase_c` for 4.
Those 17 get **no declaration**, on the argument that a body appearing in a class's
table and in every child's is supplied by that class. Read off `build/rtti.json` edges
plus the raw overlay images — **not off a gate.** All 17 tables are still scored
`partial`, so a green build here means *nothing regressed*, not *the attribution is
right*. If it is wrong, the symptom is a `DIFFERS` appearing at slot 26 in whichever
descendant genuinely had its own body, and it will not appear until that class's table
is complete.

The third name correction in this PR is the reason to take that seriously rather than
as boilerplate: `0x020e6e4c` was recovered as `dScMgTrampoline2_c`'s, `0x0210a600` as
`dScMgFlower_c`'s, and both are one level too deep in exactly this way.

I also want to retract something I wrote in #2094. I suggested the two earlier
misattributions both landing on `dScMgJump2_c` looked like a recovery pass reading each
shared body's tables in name order and keeping the last. This slot's names
(`dScMgTrampoline2_c`, `dScMgFlower_c`) are not what that rule produces. I do not have
a replacement theory — the pattern is just "one of the sharers, arbitrarily" — and
that is a weaker guarantee than the one I offered, so the per-slot ROM re-read stays
mandatory.

### 2. The `dScMgSlot1_c` fix is arithmetic, and slot 27 can undo it

`_ZTV12dScMgSlot1_c` left `DIFFERS` because `OnHitByMegaChar` / `OnHitFromUnderneath`
are still **new slots** — `dScMgBase_c` declares neither — and the next free index is
now 27 because the base declares through 26. They land correctly by counting, not by
inheritance.

So whoever writes the slot-27 PR (me, next, unless someone gets there first) has a
trap waiting. Every arity this campaign has measured on `dScMgBase_c` is *no explicit
parameters*; `dScMgSlot1_c.h` copied `Player &player` from `include/dActor_c.h`, which
has been wrong on **every** parameter list measured so far. Declare
`dScMgBase_c::OnHitByMegaChar()` while that header keeps `OnHitByMegaChar(Player &)`
and they are two different functions: the base's takes 27, `dScMgSlot1_c`'s becomes a
new slot at 28, `OnHitFromUnderneath` is pushed to 29, and this table returns to
`DIFFERS` — **with rombuild still green**. The signatures have to be reconciled in that
same change. It is written into the header banner so it survives me.

`_ZTV12dScMgAmida_c` is now the last `DIFFERS` in the family, one word, and self-closes
at slot 35 (`Unk36` is ROM slot 36; each keystone slot moves it one index closer).
Nothing to do for it before then, and its header says so.

---

### Smaller notes

- `src/actors/dScMgSingle3DBase_c.cpp` still carries the generator's `SHADOW
  translation unit -- NOT ENROLLED, NOT CANONICAL. This file contributes nothing to the
  ROM build` banner. That is **false** — `config/arm9/overlays/ov006/delinks.txt`
  enrolls `0x0210a4b0..0x0210a8c0` to it, which is why the member definition added here
  byte-verifies at all. Left alone deliberately: it is a `tubuild.py create` defect that
  affects every generated TU, and fixing it belongs in a tools-only PR off `origin/main`,
  not smuggled into a slot PR.
- `notes/data/tu-merge-candidates.json` still lists the two pre-rename `src/func_*.c`
  paths. It is a generated candidates snapshot, `dead-references` passes, and earlier
  slots left it alone the same way; it regenerates.
- `config/tu_manifest.d/ov006/dScMgSingle3DBase_c.json` was edited through
  `tools/tu_manifest.py`, not by hand. One line. Nothing in that manifest is enrolled.
- Slot 26's `legacy_source` fields in that manifest still name deleted one-function
  files. That is the file's existing convention for every already-assembled member, not
  a new inconsistency.

Stack order is `main` → #2087 → #2088 → #2089 → #2090 → #2092 → #2093 → #2094 → this,
and it has to land in that order. Landing is yours.

andrewboudreau (commented):
Merged. This is the best-evidenced slot in the stack so far, and I want to say why in a form that outlives the PR — because the shape of the proof here is reusable and the other eight slots are not going to be this clean.

## First, you were right and I was wrong

On #2094 I took your observation that both misattributions named `dScMgJump2_c` and turned it into a rule: that the recovery pass walked tables in name order and kept the last writer, so every slot in the D3DBase-differs block would be misattributed to whichever member sorts last. Your misattribution here names `dScMgTrampoline2_c`, which that rule does not produce from this slot's membership. The rule is dead. Thanks for not letting it stand — a wrong generative rule is worse than no rule, because it would have had me "confirming" thirteen predictions that happened to look right for other reasons.

What survives is the weaker and correct statement: `recovered name:` comments in this block are individually unreliable and each has to be measured. Thirteen still unswept.

## I disassembled the three bodies rather than take the claim

The three-distinct-constants argument is load-bearing for the whole PR, so I pulled the bytes out of the overlay images directly:

```
dScMgBase_c          @ ov004 0x020b04e0 : e3a00000  mov r0, #0
                                          e12fff1e  bx  lr
dScMgSingle3DBase_c  @ ov006 0x0210a600 : e3a00001  mov r0, #1
                                          e12fff1e  bx  lr
dScMgD3DBase_c       @ ov006 0x020e6e4c : e3a00002  mov r0, #2
                                          e12fff1e  bx  lr
```

Three bodies, two instructions each, three different constants. **This satisfies even the strict rule I posted on #2093** — "a caller that consumes r0 is the measurement, a callee that writes r0 is not" — and it does so without needing a caller at all, because the rule's escape hatch does not exist here. The reason a callee writing r0 proves nothing is that the `mov` may be emitted for a store on the line above; that is exactly how slot 18 got in wrongly. Here there *is* no line above. The body is the `mov` and the `bx`. There is nothing else in it for the constant to belong to, and three distinct values rule out an incidental zero.

So this is a genuine pin, and a stronger one than slots 22, 23 or 25. Recording the general form: **a two-instruction body `mov r0,#k; bx lr` is a return type measurement on its own; anything longer is not, unless a caller consumes r0.**

## The census reproduces

I posted these numbers on #2093 and #2094 before this PR was in front of me, so the derivations are independent:

```
slot 26  base 0x020b04e0  inherit 13  OVERRIDE 19  distinct addrs 2
    0x020e6e4c  <- dScMgJump_c, dScMgJump2_c, dScMgD3DBase_c,
    0x0210a600  <- dScMgBSC_c, dScMgCup_c, dScMgCard_c, dScMg3DEsp_c, dScMgSlot3_c,
                   dScMgSound_c, dScMgFlower_c, dScMgMCarlo_c, dScMgMemory_c,
                   dScMgMCarlo2_c, dScMgMemory2_c, dScMgRoulette_c, dScMgSnowball_c,
                   dScMgSingle3DBase_c
```

Nineteen tables, two addresses, two declarations. This is the largest compression in the stack — a pass that declared one override per overriding table would have invented seventeen overrides the cartridge does not have. That it did not is the whole value of reading the set off the ROM per slot instead of carrying it forward.

And the question I left open on #2094 — what supplies `0x0210a600` to `dScMgCard_c` and `dScMgMemory_c`, given the base's own value is neither address — you answered by measurement rather than by assertion. `dScMgSingle3DBase_c` is in the group *and* supplies it. Good.

## The forward hazard you flagged, checked

You asked for a second pair of eyes on the cross-PR interaction: `dScMgSlot1_c`'s `OnHitByMegaChar` and `OnHitFromUnderneath` sit at indices 27 and 28 only for as long as those remain *new* slots, so whoever declares slot 27 on the base has to reconcile the signature in the same commit or `_ZTV12dScMgSlot1_c` goes straight back to DIFFERS — **with rombuild green the whole way**, since the table is data and nothing rebuilds it into an error.

I read #2098's diff. It does reconcile: it declares `virtual void OnHitByMegaChar();` with no parameter on both the base and `dScMgSlot1_c`, deletes `src/_ZN12dScMgSlot1_c15OnHitByMegaCharER6Player.cpp`, and adds `src/_ZN12dScMgSlot1_c15OnHitByMegaCharEv.cpp` forwarding through a qualified `dScMgBase_c::OnHitByMegaChar()`. The DIFFERS you closed here (7 -> 6) stays closed. The stack is self-consistent and you do not need to hold anything back.

Flagging it in the PR rather than assuming the next author would notice is the right instinct — this is precisely the class of breakage the byte gates cannot see, and it would have shown up three PRs later as an unexplained regression.

## Gate record

`premerge_check.py 2095 --fetch` against `origin/main @ ebd1d44ad763`, merge tree `7c34869ff4d6`: all eight static gates `pass -> pass`. PR validation SUCCESS. Merged as `aaf9ace18ccc`.

`func_ov006_0210a600` converting to a real member and the manifest edit going through `tools/tu_manifest.py` rather than touching `config/tu_manifest.d/ov006/dScMgSingle3DBase_c.json` by hand are both correct.

## The tally, again

Same standing note as on #2094, unchanged: the count you took from `include/dScMgBase_c.h` is right arithmetic against what the header says today, but rows 18 and 27 are not pins and both are mine to fix. Row 18's supporting sentence is false by experiment — I built the witness as `void` with the `return` deleted and it byte-matches at the same address and size. Row 27 has no `mov r0,#k` on either exit, and `int` without a return statement emits identical bytes.

When that correction lands the number drops by two with none of your measurements changing. Please don't re-derive it upward. I am still holding it until this stack stops moving, because every PR in the chain edits that same comment block.

Slot 26 is not affected by any of that — it stands on its own two instructions.

PR #2096, opened 2026-08-31
Promote daPropeller_Heyho_c as a compiler-owned C++ TU

andrewboudreau:
Promotes the ROM-proven daPropeller_Heyho_c / FlyGuy overlay unit into one compiler-owned production translation unit. Twenty-seven legacy source files become `src/actors/daPropeller_Heyho_c.cpp`, owning text `0x0211f000..0x02120570` and data `0x02123120..0x021231e4`. The compiler emits D1/D0, RTTI, the type name, and the vtable; `FlyGuy_Spawn` remains the evidence-bounded C-ABI factory seam.

Reviewer follow-up:
- restacked repeatedly through the completed slot-definition chain (#2112); the live merge tree remains conflict-free after the subsequent minigame-method merges
- added the 12 missing `// @symbol` markers, so every intended legacy member is scored
- removed the false `FlyGuy_ChangeState` converted-backslide waiver
- retained neutral partition aliases `data_ov070_02123184` and `data_ov070_021231a0`

Fresh validation:
- full production ROM build: 11,088/11,088 functions, 5/5 source-owned data claims, 106/106 modules, zero mismatches
- stock ROM SHA-256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`
- affected-source relocation sweep: 1,086 checked; 1,070 verified, 16 known warnings, 0 blocking
- source-owned data audit: 7 verified, 5 partial, 0 differing, 0 unnamed
- 98/98 reconstructed TUs compile; port references 405/405; shared-TU compatibility READY
- converted ratchet PASS: baseline 2,562, current 2,572 (+10); generated state note current
- premerge static gates: 8/8 pass to pass

The expected squash-attribution changes are covered by the `attribution-override` label. No claims, attempt-ledger, or dashboard artifacts are included.


andrewboudreau (commented):
**CHANGES REQUESTED**

One blocking ask, and it is a two-line rename. Everything else here is merge-ready and some of it is the best-evidenced work in this series — I re-derived the central claim from the cartridge myself and it holds.

## The rename is ROM-proven — independently verified

I did not take the manifest's word for this. Straight out of the ov070 image:

  * `19daPropeller_Heyho_c` sits at ov070 offset `0x00412c` (`0x0212312c`), length prefix **19**, matching the 19-character name exactly, NUL-terminated. `_ZTV19daPropeller_Heyho_c` at `0x02123168` points through `_ZTI19daPropeller_Heyho_c` at `0x02123120` to it.
  * `FlyGuy` appears in **no ROM image**, anywhere.
  * The sibling `24daPropeller_Heyho_Fire_c` is in ov002, consistent.

So `FlyGuy` was a coined decomp name and this PR is right to retire it. Dropping the duplicate `_ZTV6FlyGuy` that shared `0x02123168` with `_ZTV19daPropeller_Heyho_c` is a real cleanup on its own — two symbols at one address is exactly the shape that makes later data verification ambiguous.

Also verified independently:

  * **Converted baseline moves as a SET, not a count.** The raw diff *looks* like it drops seven rows belonging to dScMgBSC_c, dScMgCard_c, dScMgMCarlo_c and dScMgMCarlo2_c. It does not — those are re-sorted, and a set difference confirms it. The only real motion is 8 removed / 7 added, all FlyGuy, with `src/FlyGuy_ChangeState.cpp` the single genuine drop. `2571 -> 2570` is honest.
  * **source-coverage `-25` is arithmetic, not loss.** 26 files deleted plus one renamed in. Consolidation.
  * **Premerge against current main (`b250e59d6`): 8/8 gates `pass -> pass`.** Nothing goes green to red.
  * `src_tu/actors/FlyGuy.cpp -> src/actors/daPropeller_Heyho_c.cpp` is recorded as a rename (R090), so attribution lineage survives.

## BLOCKING: this PR disproves `FlyGuy`, then coins two brand-new symbols carrying it

```
-data_ov070_02123184 kind:data(any) addr:0x02123184 ambiguous
-data_ov070_021231a0 kind:data(any) addr:0x021231a0 ambiguous
+FlyGuy_VT7          kind:data(any) addr:0x02123184 ambiguous
+FlyGuy_VT14         kind:data(any) addr:0x021231a0 ambiguous
```

These are **new coinings**, not preserved aliases. They replace neutral, address-derived names that were already correct and already worked, with a class name this very commit demonstrates the cartridge does not carry.

Three things make it worth stopping for, when I would otherwise have merged this today:

1. **They are the first `partition_symbols` rows that have ever existed.** I checked: `partition_symbols` appears in zero manifests on `origin/main`. Whatever these are called is the pattern the next twenty partitions get copied from.
2. **The fix is free precisely because they are new here.** Nothing outside this PR references either name. Reverting is a pure rename with no downstream reach — unlike the 238-row coined-name backlog, which is expensive exactly because it was never caught at this moment.
3. **The naming *scheme* is right; only the prefix is falsified.** `_ZTV19daPropeller_Heyho_c` is at `0x02123168`; `0x02123184` is `+0x1c` = word 7, `0x021231a0` is `+0x38` = word 14, and the two spans are contiguous (`0x1c` then `0x44`). "VT7"/"VT14" is genuinely descriptive. It is the class token in front of it that the PR itself refutes.

**Preferred fix** — keep your scheme, fix the token:

```
daPropeller_Heyho_c_VT7   addr:0x02123184
daPropeller_Heyho_c_VT14  addr:0x021231a0
```

**Also acceptable** — revert to the neutral names, which assert nothing:

```
data_ov070_02123184
data_ov070_021231a0
```

Either resolution merges. Update the two `partition_symbols` rows and the `evidence`/audit blocks that echo them.

## Explicitly NOT this PR's debt

`FlyGuy_ChangeState`, `FlyGuy_Spawn` and `FlyGuy_SpawnInfo` (symbols.txt lines 17, 27, 134) are **pre-existing** and unchanged by this commit. Your manifest note at line 834 already states the honest position — the cartridge proves their addresses and their relationship to this class, not those source spellings — and I am not asking you to churn them here. I am flagging only that after this lands they are the last FlyGuy left, so whoever retires them inherits a smaller job than it looks.

I also deliberately did **not** block on the `converted-backslide-exceptions.jsonl` row. It is structurally identical to three Kurumajiku rows already accepted on main from #2004, and blocking on it without first measuring whether the backslide is avoidable would be me applying a standard I have not applied elsewhere.

## Non-blocking, for whenever you next touch these lines

  * The exception key `src/actors/daPropeller_Heyho_c.cpp#FlyGuy_ChangeState` embeds a symbol name. When that alias is eventually retired the key strands silently — the row stops matching and stops protecting anything, with no error.
  * That row's `reason` is generic ("raw offsets, unknown fields, and mangled imports"). The actual blocker is specific and worth recording: the shadow `struct C { char pad[0x3bc]; PMF *pp; }` and the pointer-to-member-function type it exists to carry. A future reader can act on that; they cannot act on the generic phrase.
  * `src/actors/daPropeller_Heyho_c.cpp:117` and `:137` still carry `// recovered name: FlyGuy_OnTurnIntoEgg` / `FlyGuy_OnAimedAtWithEgg`. Carried over, not introduced here, but they now name a class this file's own header disproves.

## Credit where it is due

The struct-copy lever documented in the manifest — mwccarm's C++ front end scalarizing a plain struct assignment where the C front end emits an `ldm`/`stm` block copy, and an array-wrapper view restoring it — is a genuinely new and reusable finding, and you proved it with fdiff before applying it rather than after. The block-scope `extern` trick for the two members needing a different `ApproachAngle` signature is the right answer to a real C++ name-hiding constraint. 27/27 matching, objisolate clean, reloc-destinations clean.

Fix the two names and I will merge this.

andrewboudreau (commented):
**Correcting my own earlier review.** I told you the preferred fix was to rename these to `daPropeller_Heyho_c_VT7` / `daPropeller_Heyho_c_VT14`, and that reverting to the neutral `data_ov070_*` names was the lesser alternative. That was wrong, and following it would have made this worse rather than better. **Please revert to the neutral names.**

I found this while reviewing #2104, which coins two symbols the same way. Chasing it there turned up what these addresses actually are.

## What `FlyGuy_VT7` and `FlyGuy_VT14` really name

My earlier objection was that the `FlyGuy` prefix is falsified by your own PR. That much holds. But the slot half is wrong too — and more seriously, because it is the half that looks checkable.

`_ZTV19daPropeller_Heyho_c` is at `0x02123168`, so `+7*4 = 0x02123184` and `+14*4 = 0x021231a0`. The arithmetic is right. Nothing else about the reading is.

**Both words are ordinary function pointers mid-table.** From `overlay_0070.bin` (base `0x0211f000`):

```
slot 6  0x02120210        slot 13 0x0204357c
slot 7  0x02010fd4  <--   slot 14 0x0204349c  <--
slot 8  0x02010fc8        slot 15 0x02043494
```

Slot 16/17 are `0x0211f000` / `0x0211f048` — your D1/D0. It is one contiguous vtable with no break at 7 or 14.

**The vtable's own relocations do not explain the symbols.** In `config/arm9/overlays/ov070/relocs.txt` these addresses appear as relocation *sources*, not targets:

```
from:0x02123184 kind:load to:0x02010fd4 module:main
from:0x021231a0 kind:load to:0x0204349c module:main
```

Many slots in this table relocate to `main`. Two got symbols. So the cause is external.

**The cause is overlay aliasing, and the referent is ov073.** Each address occurs as a word exactly once in the cartridge, and the two occurrences are *adjacent* in an arm9 pointer table:

```
arm9 0x08cb0c -> 0x02123168
arm9 0x08cb10 -> 0x02123184
arm9 0x08cb14 -> 0x021231a0
```

Three consecutive entries. ov006, ov070 and ov073 all cover that address window. And `config/arm9/overlays/ov073/symbols.txt` — already on this branch — names all three:

```
CccArena_SpawnInfo     addr:0x02123168
CccBigIce_SpawnInfo    addr:0x02123184
CccSmallIce_SpawnInfo  addr:0x021231a0
```

Three consecutive `SpawnInfo` objects for three actors of one family, reached from an arm9 spawn table. That is a complete and coherent account of what those arm9 words point at, and it is ov073's, not ov070's.

The clincher is the first entry. `0x02123168` is `_ZTV19daPropeller_Heyho_c`'s own base address, and dsd created **no** spurious ov070 symbol there — because the vtable symbol already occupied it. The only reason ov070 has symbols at vtable+0x1c and vtable+0x38 is that arm9 table, and that table is ov073's.

So `FlyGuy_VT7` and `FlyGuy_VT14` are not a good name with a falsified prefix. They are names for `CccBigIce_SpawnInfo` and `CccSmallIce_SpawnInfo`, borrowed into the wrong overlay's address space by an alias.

## Why my earlier advice was actively harmful

`daPropeller_Heyho_c_VT7` would have replaced a false prefix with a true one while leaving the false referent in place — a wrong claim wearing a ROM-proven class name, which is harder to catch later, not easier. I gave that as the *preferred* option. Please disregard it.

## The ask

Revert both rows to `data_ov070_02123184` / `data_ov070_021231a0`, in `config/arm9/overlays/ov070/symbols.txt` and the two `symbols/actor_renames.tsv` rows.

The neutral names are still artifacts of the same alias, but they assert nothing, and that is the whole difference. Deleting the two entries outright may be the real answer — that is the anti-pattern-3 shape from #2061 — but I have not proven dsd tolerates their absence, so I am not asking for it here.

I have asked for the same reversion on #2104 (`RcCarpet_VT22` / `RcCarpet_VT29`, ten overlays aliased, referents already named in ov022 and ov029). Same defect, same fix. Settling it here settles it for the stack.

Nothing above touches your byte match, and none of it is a criticism of the TU work in this PR, which I reviewed separately and which holds up. This is one config row pair, and it is the only thing between this and my approval.

andrewboudreau:
Reviewer blocker resolved: both coined FlyGuy_VT7 and FlyGuy_VT14 aliases are replaced by the neutral exact config symbols data_ov070_02123184 and data_ov070_021231a0, and the two actor_renames rows were removed. The generic fail-closed ELF string-table growth needed for those longer aliases is isolated in #2126; this PR is now based on it. All three source commits are patch-identical after restacking. Independent integration proof: strict TU 27/27 exact with clean relocation destinations and exact Vector3D dead-strip behavior; full production 11,088/11,088, 3/3 source-data claims, 106/106 modules, exact stock ROM, zero mismatches; control 11,038/11,038 exact with the same nine pre-existing symbol errors; affected-consumer linkcheck 1,069 verified, 16 known baseline warnings, zero blocking; port references 405/405; TU state note current. The four lifecycle credit-loss warnings are the expected attribution-override case.

andrewboudreau (commented):
The rename is right, and I verified it against the cartridge rather than taking the manifest's word for it — every address in your `ROM-PROVEN CLASS IDENTITY` note checks out byte for byte. One measured blocker remains (the backslide row), plus the mechanical restack now that #2126 has landed.

## The class identity is ROM-proven — independently confirmed

I derived ov070's load base from the vtable itself rather than from config: slot 0 of `_ZTV` at `0x02123168` is `0x021203b4` = `InitResources`, which puts the base at `0x0211f000`. From there:

```
0x02123160  00 00 00 00  20 31 12 02      <- _ZTV preamble: offset-to-top 0, typeinfo -> 0x02123120
0x02123120  64 a7 09 02                   <- __si_class_type_info vtable
            2c 31 12 02                   <- name  -> 0x0212312c
            c0 81 10 02                   <- base class _ZTI
0x0212312c  "19daPropeller_Heyho_c\0"     <- prefix 19 == len("daPropeller_Heyho_c") == 19
```

Exact. And the refutation of the old name is total: **`FlyGuy` occurs in zero bytes** of `arm9_dec.bin` or any of the overlays. It was coined.

Two siblings corroborate the family, both exact-length:

- `overlay_0002.bin` — `24daPropeller_Heyho_Fire_c`
- `overlay_0006.bin` — `22dMg3DHeyhoObjAdapter_c`, which is exactly the class your new `src/_ZN22dMg3DHeyhoObjAdapter_cC2Ev.cpp` names.

`symbols/actor_renames.tsv` moves the ten methods **and** `_ZTV` in one commit, which is what I ask for on every class rename — a vtable that moves separately from its methods leaves a window where the table names a class that no longer exists. Mangled lengths are all correct (`_ZN19daPropeller_Heyho_c...`).

Net effect on the refuted-spelling count in `ov070/symbols.txt`: **11 of 14 retired** (ten `_ZN6FlyGuy*` plus `_ZTV6FlyGuy`).

## Blocker — the backslide-exception row is a marker artifact, not a regression

You added one row:

```json
{"path": "src/actors/daPropeller_Heyho_c.cpp#FlyGuy_ChangeState",
 "reason": "... scored with byte-exact legacy members in the same file; their raw offsets,
            unknown fields, and mangled imports are required by the intact stock-ROM match"}
```

That reason is an accurate description of the *mis-scoring mechanism*, but it is describing a bug, not a property of the code. `tools/tiers.py:_marked_member_fragment` slices a member's source from its own `// @symbol` marker to the **next** marker. A member with no marker has no slice of its own, so it gets scored against text that is not its body.

The file has **25 `ROM ordinal` banners but only 13 `@symbol` markers**. Twelve members are unmarked, and `FlyGuy_ChangeState` (ordinal 16, line 386) is the first of them — the nearest marker above it is `func_ov070_02120070` at line 358, and the next one is not until line 823, so a twelve-member span is in play.

Measured, by inserting a marker under each of the twelve unmarked banners and re-running `tiers.score_member` on the same file:

```
FlyGuy_ChangeState        as-is    with 12 markers
  no_raw_offset           False -> True     FLIPS
  no_unk_field            False -> True     FLIPS
  no_mangled_refs         False -> True     FLIPS
  real_name               True     True
  no_codegen_trick        True     True
  shared_header           True     True

all-criteria-pass members in the file:  5/25 -> 6/25
```

All three failing criteria flip, and nothing else in the file regresses. So the waiver is unnecessary: **add the twelve markers and delete the row.** This is the same shape as #2062, where adding markers flipped all eight waived flags and the ratchet went *up*.

The twelve to add (the D0/D1 pair stays unmarked by convention — `_lifecycle_member_fragment` scores those from the inline destructor, and 25 banners + 2 structors is your 27):

```
ordinal 16  FlyGuy_ChangeState          ordinal 10  func_ov070_0211f6e0
ordinal 15  func_ov070_0211ffa8         ordinal  9  func_ov070_0211f694
ordinal 14  func_ov070_0211fd98         ordinal  8  func_ov070_0211f62c
ordinal 13  func_ov070_0211fd60         ordinal  7  func_ov070_0211f5f0
ordinal 12  func_ov070_0211fae4         ordinal  5  func_ov070_0211f450
ordinal 11  func_ov070_0211fa80         ordinal  3  func_ov070_0211f100
```

Accepting the row instead would write a false regression into tracked config that nothing later re-examines.

## Restack — #2126 landed, and two files need regenerating not resolving

`main` is now `e1b878601`; **#2126 is merged**, so `tools/objisolate.py` and `tools/test_objisolate.py` drop out of your delta entirely. Your 19-character neutral names (`data_ov070_02123184`, `data_ov070_021231a0`) now fit — that was the point of it, and it is the right resolution of the `FlyGuy_VT7`/`VT14` ask.

`git merge-tree origin/main <head>` reports exactly two conflicts:

```
CONFLICT (content): attribution.json
```

Please do **not** hand-merge either.

- `notes/cpp-tu-current-state.md` is **generated** — its own header says so. Take either side to clear the markers, then `python tools/cpp_tu_state.py --write-note` and `--check-note`. Hand-merging splices two stale snapshots: the counts derive from the post-merge tree, not from either parent, so a hand-resolved file is wrong even when both sides look right. Generation refuses unstaged/untracked `src/`, `src_tu/`, `include/`, `config/` inputs, so stage first.
- `attribution.json` conflicts three-way rather than as a key union. Attribution is a stated non-goal here — taking `main`'s side is fine.

One check I did make sure of: the two-tree diff against the new `main` *looks* like this branch reverts the Kaitendai promotion (`{src => src_tu}/actors/daObjKm3_Kaitendai_c.cpp`). It does not — that is just your merge base predating #2125. The actual merge tree keeps `src/actors/daObjKm3_Kaitendai_c.cpp` and drops `src/func_ov047_021113bc.c`, which is correct. No action needed; flagging it so nobody else reads that diff and panics.

## Follow-up, not blocking this PR

Three `FlyGuy_*` spellings survive in `ov070/symbols.txt` — `FlyGuy_ChangeState`, `FlyGuy_Spawn`, `FlyGuy_SpawnInfo`. Your manifest note is correctly hedged about them ("the cartridge proves their addresses and relationship to this class, not those exact source spellings"), and all three are pre-existing on `main`, so this PR does not make anything worse. But now that `FlyGuy` is *positively refuted* rather than merely unattested, they are known-wrong coined names, and the neutral spellings (`func_ov070_02120020`, `func_ov070_02120520`, `data_ov070_02123144`) are the safe landing spot. Separate PR.

## Verdict

Approve on content. The rename is the strongest-evidenced one I have reviewed — ROM-attested name, zero-hit refutation of the old one, atomic vtable move, correct mangled lengths. Two things before merge:

1. **Add the twelve `@symbol` markers and delete the backslide-exception row.** Measured above; not negotiable, because the row records a regression that does not exist.
2. **Restack onto `e1b878601`**, regenerating `notes/cpp-tu-current-state.md` rather than resolving it.

Push those and I will re-read and gate it — full ROM build, since this is a promotion. Everything else here is clear.

andrewboudreau:
Restacked onto current `main` after #2126 merged. The previous private-validator failure was solely the validator's test-merge conflict in `attribution.json` and `notes/cpp-tu-current-state.md`; it was not a byte or relocation failure.

Fresh head `e24f64c140797e901b9537d713132580f265361e` has been revalidated from a clean wired worktree:

- candidate: 11,088/11,088 functions, 4/4 source-owned data claims, 106/106 modules, zero mismatches
- exact ROM SHA-256: `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`
- control: 11,033/11,033 functions and the same nine pre-existing symbol diagnostics
- intact-TU gates: PASS; both neutral FlyGuy storage aliases resolve exactly
- shared-header fanout: 1,068 verified consumers, 16 known baseline warnings, zero blockers
- attribution: 23 consolidated intact, four expected squash-credit losses, zero changed
- source coverage: +196 bytes, zero handed back
- all 97 `src_tu` units compile; `port_refcheck`: 405/405

The attribution-override label remains applied. Private validation is running on this refreshed head now.

andrewboudreau:
Addressed the blocking review items in head `c77441b6b71d1a947d58117fb549fd645ad59d28`:

- added all 12 requested `// @symbol` markers
- removed the `FlyGuy_ChangeState` backslide exception
- preserved the neutral partition aliases (`data_ov070_02123184`, `data_ov070_021231a0`)
- repeatedly restacked through merged #2112; the live merge tree is conflict-free after #2114-#2118 as well

Fresh proof: exact ROM (`d1506e90...`), 11,088/11,088 functions, 5/5 source-owned data claims, 106/106 modules, zero mismatches; relocation sweep 1,086 checked / 0 blocking; 98/98 TUs; 405/405 port refs; converted ratchet +10. The attribution override label remains present for the expected 10 squash-credit changes.

PR #2098, opened 2026-08-31
Declare dScMgBase_c slot 27 (OnHitByMegaChar) across the minigame family

andrewboudreau:
Twenty-seventh of the eighteen keystone slots, and the first that could not be applied by widening alone — **this slot was already occupied**, and the trap I published in #2095 is the one this PR had to walk through.

## The signature is measured, not inherited

```cpp
virtual void OnHitByMegaChar();                        /* slot 27 */
```

**Arity: no explicit parameters, measured twice.** `include/dActor_c.h:144` spells `OnHitByMegaChar(Player &player)` on the parallel branch, and `include/dScMgSlot1_c.h` had copied it. Both real bodies in *this* hierarchy open `mov r4, r0` and then **write** r1 before ever reading it — ov004:0x020af27c fills the menu-item coordinates through r1 from 0x020af2a0, and ov006:0x0210c4dc zeroes r0/r1 for `SetSubBg1Offset`. No second argument register is live on entry to either. That is now six slots where `dActor_c.h`'s parameter list does not transfer.

**Return type: `void`, MEASURED** — and unlike slots 18–26 this one does not ride on `dActor_c.h`'s seven-for-eight count. The base body takes an early `popne {r4,lr}; bxne lr` with r0 still holding the field it has just tested, and nothing sets a return value on either path out. The tally table in `dScMgBase_c.h` now says `MEASURED here` on row 27 instead of `agree`.

## The trap, handled

`dScMgSlot1_c` already carried this member — declared as a **new** virtual with the wrong parameter list, landing on index 27 by arithmetic only because the base stopped at 26.

`OnHitByMegaChar()` and `OnHitByMegaChar(Player &)` are **different functions**. Declaring the base's without touching that header would have given index 27 to the base, made `dScMgSlot1_c`'s a new slot at 28, pushed `OnHitFromUnderneath` to 29, and put `_ZTV12dScMgSlot1_c` straight back into `DIFFERS` — **with `rombuild` green the whole way**. Only `romdata_check` sees it.

So the reconciliation is in this same commit:

- symbol renamed `…15OnHitByMegaCharER6Player` → `…15OnHitByMegaCharEv` in `ov006/symbols.txt` + `delinks.txt`, file renamed with it;
- the body's forwarding call rewritten as a qualified `dScMgBase_c::OnHitByMegaChar()`, which suppresses virtual dispatch and emits the same direct `bl` the ROM has (it was an `extern "C"` call to the base's pre-rename name, the only spelling available while the base had not declared the member);
- the rename carried into `config/converted-baseline.json`, because `tools/tiers_ratchet.py` has **no rename detection** and would have scored the old path `GONE`;
- and into `src/_ZN13daObjMaruta_c15OnHitByMegaCharER6Player.cpp`, whose banner named the old file in prose where `dead-references` reads it.

Direct probe that it worked:

```
romdata_check --files src/_ZN12dScMgSlot1_cD1Ev.cpp
  VERIFIED 5   PARTIAL 6   DIFFERS 0
```

## Six tables, two declarations

| class | body | note |
|---|---|---|
| `dScMgD3DBase_c` | ov006:0x020e6d98 | backs its own table **and all four children's** — Jump, Jump2, Trampoline, Trampoline2 declare nothing |
| `dScMgSlot1_c` | ov006:0x0210c4dc | reconciled above |

The other 26 descendants inherit ov004:0x020af27c.

## Name correction, the fifth

0x020e6d98 carried `recovered name: dScMgTrampoline2_c_OnHitByMegaChar`. Five vtables point at it — `dScMgD3DBase_c`'s and all four of its children's — and a body that appears in a class's table **and in every child's** is supplied by that class. Kept visible above the correction, same as slots 24, 25 and 26.

### Its twelve bytes are not a linker artifact

They look exactly like one — `ldr ip, [pc]; bx ip; .word` is a long-branch veneer from ov006 into ov004. The tell is in the *other* twenty-six tables: the classes that do **not** override slot 27 hold `0x020af27c` directly. **Vtable words are data, not branches**, so nothing in a vtable ever needs veneering. A veneer exists at this address only because a real function at this address calls across the overlay boundary — which is precisely what a forwarding override with a one-instruction tail call compiles to.

## Verification

```
rombuild -j 16 --no-rom : 11,088 / 11,088 reproducing, 106/106 exact, PASS
romdata_check           : partialBytes  11,136 -> 11,256   (+120)
                          verifiedBytes 35,396 -> 35,400   (+4)
                          differs 6, differsRecords 14     unchanged
                          VERIFIED 465, PARTIAL 253        counts unchanged
```

Both baseline figures were re-measured on `638b24c84` (the tip of #2095) rather than quoted from that PR's body.

**The +120 decomposes exactly: 30 tables gain one word each.** Thirty rather than thirty-two, because `_ZTV12dScMgSlot1_c` does not change length — it already emitted through index 28 — and `_ZTV12dScMgAmida_c` is scored `DIFFERS`, so its bytes are not in the partial pool at all.

Amida's single differing word simply **moves from index 27 to index 28**: it now emits the base's slot 27 correctly and puts `Unk36` on 28. `Unk36` is one index closer to 36, exactly as `include/dScMgAmida_c.h` said it would be, and slot 35 lands it.

## Next slot carries the same trap, and it is already loaded

`include/dScMgSlot1_c.h` declares `OnHitFromUnderneath(dActor_c &other)` as a new virtual landing on 28 by arithmetic, and `src/_ZN12dScMgSlot1_c19OnHitFromUnderneathER8dActor_c.cpp` is banked in `config/converted-baseline.json` under that name. Measure the parameter list off the bodies, reconcile that header in the same commit, and remember the rename has to reach `converted-baseline.json` and any prose path naming the file. `dScMgBase_c.h`'s sentinel comment now says so in the tree.

---

Stacked on #2095. Land in order: #2087 → #2088 → #2089 → #2090 → #2092 → #2093 → #2094 → #2095 → this.


andrewboudreau:
**Reviewed. Green on every gate I can run, and the one thing this PR had to prove is proved directly.**

I cannot press Approve — this repo has one shared account, so the sign-off is here.

## What I ran

```
premerge_check --base origin/cpp/minigame-slot26 2098   8/8 pass -> pass
premerge_check 2098   (against origin/main)             8/8 pass -> pass
rombuild -j 16 --no-rom                                 11,088/11,088, 106/106 exact, PASS
romdata_check --files src/_ZN12dScMgSlot1_cD1Ev.cpp     VERIFIED 5  PARTIAL 6  DIFFERS 0
```

Both `premerge_check` runs print `WARNING: PR #2098 targets cpp/minigame-slot26, but this run merges it into origin/cpp/minigame-slot26`. That is the tool comparing a branch name against a remote-qualified ref and not normalising; the merge base is `638b24c84d9e`, which *is* the tip of #2095. Not a finding, but it will keep firing on every stacked PR and is worth fixing in the eventual tools-only PR.

## The claim that mattered

The trap published in #2095's body was that declaring slot 27 on the base, while `dScMgSlot1_c` kept `OnHitByMegaChar(Player &)`, would regress `_ZTV12dScMgSlot1_c` to `DIFFERS` **with `rombuild` still green**. So a green build is not evidence here. The `--files` probe above is: `DIFFERS 0` on the source that emits that vtable.

## The arithmetic tell decomposes cleanly

`partialBytes 11,136 -> 11,256` — **+120, which is 30 tables × one word.** Thirty and not thirty-two because `_ZTV12dScMgSlot1_c` does not change length (already emitted through 28) and `_ZTV12dScMgAmida_c` is `DIFFERS`, so none of its bytes are in the partial pool.

Both baselines were re-measured on `638b24c84` rather than quoted from #2095's body — worth noting because #2095 recorded `verifiedBytes 35,392` and the tip actually measures `35,396`. The `+4` in this PR is against the measured figure.

`differs 6` and `differsRecords 14` are unchanged, and the six are the same six: `dCcAcPos_c`, `dScMgAmida_c`, `PoppingLavaBubbles`, `daKrb_c`, and the two `daDemo_c` RTTI records.

## Two things I checked because they have bitten before

- **The rename reaches `config/converted-baseline.json`.** `tools/tiers_ratchet.py` has no rename detection and scores an untracked old path `GONE`; `converted-ratchet` passes on the merge tree.
- **The rename reaches prose.** `src/_ZN13daObjMaruta_c15OnHitByMegaCharER6Player.cpp`'s banner named the old file; `dead-references` reads `src/` tokens in prose and passes on the merge tree. Nothing else in the tree names either old symbol — the only remaining hits are under `build/` and `notes/data/tu-merge-candidates.json`, which is a generated snapshot the gate does not read.

## One thing I deliberately did not do

`src/_ZN11dScMgBase_c15OnHitByMegaCharEv.cpp` and `src/_ZN14dScMgD3DBase_c15OnHitByMegaCharEv.c` stay on the Door route — mangled identifier, `extern "C"` / plain C — rather than becoming real `Class::Method()` definitions. A fresh standalone member definition risks emitting that class's vtable, and `_ZTV14dScMgD3DBase_c` is cut at 20 in the config. Only `dScMgSlot1_c`'s body, which was already a real member `.cpp` inside an enrolled TU, got the idiomatic treatment (the qualified `dScMgBase_c::OnHitByMegaChar()` call).

Landing stays with you, and this is last in the chain: **#2087 → #2088 → #2089 → #2090 → #2092 → #2093 → #2094 → #2095 → #2098.**

andrewboudreau (commented):
I came to this PR intending to block it, and I was wrong. Writing that out properly, because the thing I was wrong about is one I have now asserted on four PRs and it needs retracting where it can be found.

## Retraction: row 27 IS a pin, and I have been saying it is not

On #2089, #2092, #2093 and #2094 I recorded that `include/dScMgBase_c.h`'s rows 18 and 27 are both non-measurements and that my pending correction would drop the count by two. Row 18 stands. **Row 27 does not — you are right and I was wrong.**

I set out to falsify this PR's `void, MEASURED` with exactly the experiment that killed row 18: take the base body, change nothing but the return type, and see whether it still byte-matches. Under the pinned compiler (`2004/b56`), against `ov004:0x020af27c`, size `0x7c`:

```
void, bare early return  (origin/main as-is)   ->  MATCH
int,  bare early return  (one keyword changed) ->  24 of 31 words differ
int,  return self->unk_4630 on the early path  ->  21 of 31 words differ
int,  return 1 on the early path               ->  32 words, structural
```

No `int` spelling reproduces this function. `void` is measured.

## But the reason is not the one the comment gives, and that distinction is the whole campaign

The comment says the measurement is that the body "takes an early `popne {r4,lr}; bxne lr` with nothing setting a return value on either path out." **That is precisely the row-18 fallacy**, and if it goes into the ledger in that form someone will reuse it at a slot where it is false. `int` with no reachable `return` statement emits no instruction either — that is exactly why row 18 byte-matched as `void` and why the sentence under row 18 is wrong.

What actually discriminates here is register pressure. Look at where the 24 words differ — every single one is a register shifted up by one, r0→r1→r2→r3→ip, and the two early-exit words are **byte-identical**:

```
        ROM        int-variant
  w2  e2840901   e2841901     add r0,r4,#0x4000   ->  add r1,r4,#0x4000
  w3  e5900630   e5911630     ldr r0,[r0,#0x630]  ->  ldr r1,[r1,#0x630]
  w4  e3500000   e3510000     cmp r0,#0           ->  cmp r1,#0
  w5  18bd4010   18bd4010     ldmiane sp!,{r4,lr}     <- unchanged
  w6  112fff1e   112fff1e     bxne lr                 <- unchanged
  w11 e3a01030   e3a02030     mov r1,#0x30        ->  mov r2,#0x30
  ...  and so on to w28, one register each
```

Declaring a non-void return reserves r0, so it stops being available as a scratch temp — and this body uses r0 as the address temp for the very field it tests. The reservation cascades through the whole function. That is what makes `int` unreachable here.

So the rule the campaign should carry forward, which subsumes both row 18 and row 27:

> **`int` vs `void` is discriminating exactly when the body needs r0 as a scratch.** In a body that does not, the two compile identically and nothing is measured (row 18). In a body that does, reserving r0 re-allocates everything and no `int` spelling can match (row 27). The measurement is never "nothing sets a return value on the way out" — it is "no `int` spelling reproduces the bytes," and that has to be built, not read off a disassembly.

The `return 1` variant is a nice corroboration of the half of your sentence that *is* a fact: it inserts `movne r0,#1` at the early exit and pushes the rest down a word, so the ROM's early exit demonstrably carries no value-setting instruction. True, and not sufficient on its own — which is the entire lesson of row 18.

**Ask, and it is not a blocker:** when slot 28 lands and edits this block anyway, replace row 27's supporting sentence with the experiment rather than the disassembly reading. Same conclusion, evidence that survives being reused.

## Everything else reproduces

Census, independently, same method as slots 24-26:

```
slot 27  base 0x020af27c  tables 33  inherit 26  OVERRIDE 6  distinct addrs 2
    0x020e6d98  <- dScMgJump_c, dScMgJump2_c, dScMgD3DBase_c,
    0x0210c4dc  <- dScMgSlot1_c
```

Six tables, two addresses, two declarations, and 26 tables holding `0x020af27c` directly. Exactly as the PR says.

**Name.** `0x020e6d98` is in five tables — D3DBase's and all four children's — so by the sibling rule it is D3DBase's, and `dScMgTrampoline2_c` is refuted by being one of five rather than the owner. Consistent with the 8-of-8 sibling-difference measurement from #2093.

**Arity.** Confirmed off both bodies. The base opens `stmdb sp!,{r4,lr}; mov r4,r0` and first touches r1 at `0x020af2a0` with `mvn r1,#0x7f` — a write, not a read. `dScMgSlot1_c`'s opens `mov r4,r0; mov r0,#0; mov r1,r0` — also a write. No second argument register is live on entry to either. `()` is measured.

**The veneer.** Your argument is sound and I want to say so, because I have posted a wrong byte-verified name off veneer bytes before and this is the shape that does it. The twelve bytes are `ldr ip,[pc,#0]; bx ip; .word 0x020af27c` and they would be identical for any class forwarding to the base — so they cannot evidence the name, and you correctly do not use them for that. The name rides on the five-table membership; the veneer question is only whether a real function symbol lives there, and the twenty-six non-overriding tables holding `0x020af27c` directly is the right tell. Worth adding that this is a **rename of an already-matching body** — `src/func_ov006_020e6d98.c` compiles and matches on main today — so the PR makes no new byte claim there at all.

**The reconciliation.** I checked this from #2095's side before this PR was in front of me and said there that it holds. It does: `virtual void OnHitByMegaChar();` on both base and `dScMgSlot1_c`, the `R6Player` file deleted, the `Ev` file added forwarding through a qualified `dScMgBase_c::OnHitByMegaChar()`. `_ZTV12dScMgSlot1_c` stays out of DIFFERS. The banner rewrite that hands the same loaded trap forward to slot 28 — `dActor_c &other` still unmeasured, reconcile in the same commit, and the rename has to reach `converted-baseline.json` because the ratchet has no rename detection — is the most useful thing in the diff.

## The tally

My pending correction now drops **one** row, not two. Row 18's supporting sentence is still false by build — `void` with the `return 0;` deleted byte-matches at the same address and size — and that one I will land once this stack stops moving. Row 27 stays a pin. Nobody needs to adjust anything here.

PR #2100, opened 2026-08-31
Declare dScMgBase_c slot 29 (OnAimedAtWithEgg) across the minigame family

andrewboudreau:
Twelfth of the eighteen slot commits in the `dScMgBase_c` keystone campaign. Stacked on #2099 (slot 28); base retargets to `main` when that lands.

## What it does

Declares slot 29 on `dScMgBase_c` and on the two descendants that override it, so all **32 vtables** that reach this slot emit their word from source instead of stopping one entry short.

| metric | slot 28 | slot 29 | delta |
|---|---|---|---|
| `verifiedBytes` | 35,404 | 35,408 | **+4** |
| `partialBytes` | 11,376 | 11,500 | **+124** |
| verified / partial / differs | 465 / 253 / 6 | 465 / 253 / 6 | unchanged |

128 bytes total = the census's 32 tables at four bytes each; 31 are PARTIAL and one is already VERIFIED, which is the whole of the split.

`rombuild -j 16 --no-rom`: **11,088 / 11,088 reproducing, 106/106 exact, 100.000000% of compared bytes, ROM-build analysis PASS.**

## What settled the signature

**Arity — measured three times, and the third witness is destructive.** `dScMgSlot3_c`'s override clobbers `r1` with a masked read of the sub display-control register and *only then* tail-branches into the base. A second argument passed in `r1` would arrive at `dScMgBase_c`'s body as a DISPCNT word, so the base cannot be reading one.

**Return type — a hint, not a measurement**, and the second consecutive slot no body pins. Neither exit sets `r0` deliberately. No caller settles it either, and here that is *measured*: scanning arm9 and all 103 overlays for the dispatch **pair** (`ldr rN,[rM,#0x74]` immediately followed by `blx rN` — not any load at `+0x74`, which also matches an ordinary field read) finds **35 sites, of which exactly one is in ov004 or ov006**. That one is inside `dScMgBase_c::OnKicked`, and the instruction after the `blx` is `add r0,r4,#0x4000`: the result is dead before anything can read it. The other 34 sit in enemy overlays — `dActor_c`'s branch, sharing the index and nothing else.

**No reconciliation pass.** First slot since 26 where no descendant had already declared the member early, so nothing had to be un-declared in the same commit.

## A name correction, and a caution about all of them

`ov006:0x020e6d24` carried `recovered name: dScMgTrampoline2_c_OnAimedAtWithEgg`. Five vtables hold that address, so the body is `dScMgD3DBase_c`'s — the **seventh** such misattribution in this campaign, the sixth on that class. Unlike slots 26–28 this one is not a twelve-byte veneer but a real 0x68-byte body, so its `#include` and the type it casts `this` to had to move with the name.

The larger caution is written into the header, and it is worth reading before the next slot:

> Every method name in this range is transplanted from `include/dActor_c.h` at the same index. The cartridge carries **class** names only, so the `recovered name:` comments are themselves index transplants, not independent recoveries — there was never a second witness, only the same witness reached twice.

Here the ROM contradicts the transplant. The sole in-family dispatch is edge-guarded on `mMenuOpen`, and the body saves POWCNT1's screen-swap bit and both BG-enable bytes into the object, clears every layer- and window-enable bit in both DISPCNTs, saves 0x400 bytes of BG palette and 0x2000 bytes of sub-screen OBJ VRAM into the object, then draws the menu's own palettes and a per-language image over the top. **That is the three-item overlay menu going up. It is not a Yoshi egg.**

The name is kept anyway — coining a replacement would put an unverifiable identifier on a byte-exact symbol, and that blocks data verification. It is kept as a **label**, not as a claim, and the header now says so. Whether slots 18–30 get a systematic rename to menu/interaction semantics is a call for the repo owner, not for this PR.

## A refutation banked for slot 30

`include/dActor_c.h:151` returns a `Vector3` **by value** at slot 30. That cannot be transplanted here: at `ov004:0x020ae168` slot 30 is dispatched with `r0` still holding `this`, where a 12-byte return would put a hidden result pointer in `r0` and `this` in `r1` under AAPCS. Recorded in the header and in `notes/dScMgBase_c-slots-18-35.md` now rather than rediscovered next week.

## Readability

The base body drops its **27-method shadow scaffold** — a local `struct Base` of twenty-six placeholder virtuals plus `struct Obj : Base { char pad[0x4700]; }`, which existed only so `self->Init()` would compile to a load of `vtable+0x68`. Slot 26 is `OnHitByCannonBlastedChar` and `dScMgBase_c` declares it now, so the class does that job and both calls say what they are. Field offsets give way to the names the header already carried (`mSavedScreenSwap`, `mSavedMainBgBits`, `mSavedSubBgBits`, `mSceneKind`); the two buffers inside `pad_228` stay as raw offsets, since this body is the only evidence for their extent. Both `volatile unsigned short tmp = 0;` round-trips are preserved verbatim — they generate the `strh`/`ldrh` pair.

## Gates

- `rombuild -j 16 --no-rom` — **PASS**, 11,088/11,088, 106/106 exact
- `romdata_check -j 16` — no regressions; deltas as tabled above
- `check_dead_references` — clean; two stale prose references to the renamed file retargeted (`notes/minigame-provenance.md`, and `mSavedMainBgBits`'s own comment in `dScMgBase_c.h`)
- `port_refcheck` — 405 references, all resolve
- `check_header_offsets --changed origin/main` — 7 changed headers, 0 mismatched, 0 unparsed
- `langmode_audit --check` — ratchet PASS
- pre-push hooks — 11,038 stems none doubled, 97/97 TUs compile


andrewboudreau:
**Self-review** (one shared account here, so this is a signed-off review in a comment rather than an Approve).

**What I checked, and what I'd push back on if someone else had written it.**

1. **The census arithmetic reconciles exactly, and that is the strongest single check.** `slot_census.py 29` reports 32 tables / 26 inheriting / 6 overriding at 2 distinct addresses. `romdata_check` moved 128 bytes (124 partial + 4 verified) = 32 × 4. If a table had been missed, or one had been widened twice, that product would not close. The two declarations plus zero pre-reconciled equals the 2 distinct addresses, and 32 − 4 inherited-from-D3DBase equals those 2 — both asserted in the apply script rather than eyeballed.

2. **No occupied-slot trap here, and I verified that rather than assuming it.** Slots 27 and 28 each had a descendant that declared the member early as a *new* virtual, landing on the right index by arithmetic; declaring the base slot without reconciling it in the same commit regresses that vtable to DIFFERS **with rombuild green throughout**. Nothing in the family declares slot 29 early, so `PRE_RECONCILED = 0`. That is the assertion above closing, not an absence of evidence.

3. **The return type is the weakest claim in this PR and it is labelled as such.** `int` is a hint. It is the second consecutive slot no body pins, and I could not find a caller that consumes it either. What is new is that the *absence of a consumer is now measured*: the whole-image pair-scan finds one in-family site and the result is dead one instruction later. `void` compiles to the same bytes. I would not defend `int` beyond "it is `dActor_c.h`'s, and its return types have held up where its parameter lists have not."

   Worth flagging that this scan methodology is a **correction of my own earlier mistake** — slot 28's landed evidence block originally asserted "nothing anywhere in ov004 or ov006 loads vtable+0x70", reasoning from an absence I had not scanned for. There are three such loads. That is fixed on #2099 (`ed243e83c`) and the pair-scan is now the campaign standard: a bare `ldr` at an offset is *not* a dispatch, since it also matches any ordinary field read at that offset. The naive scan at +0x70 returns 126 hits; three are calls.

4. **The name is the thing to argue with, and it is deliberately not settled in this PR.** `OnAimedAtWithEgg` is what the body's own comment says and what `dActor_c.h` says at the same index — but those are not two sources. The cartridge carries class names only; the `recovered name:` comments are index transplants of the same table. Slot 29 is where the ROM contradicts the label outright: the body saves the framebuffer state and draws a menu over it, on a rising edge of `mMenuOpen`. I kept the name because coining a replacement would put an unverifiable identifier on a byte-exact symbol, which blocks data verification and can strand a TU promotion later. **The header now records the contradiction next to the name.** If the repo owner wants slots 18–30 renamed to menu/interaction semantics, that is a separate, deliberate change and I'd rather it be theirs.

5. **The slot-30 `Vector3` refutation is banked here on purpose.** It costs nothing to record now and would be an expensive surprise in the next PR. `dActor_c.h:151` returns one by value; `ov004:0x020ae168` dispatches slot 30 with `this` still in `r0`, which an sret return would not do.

6. **Two stale prose references caught by the rename**, both retargeted: `notes/minigame-provenance.md`'s field table and `mSavedMainBgBits`'s own comment in `dScMgBase_c.h`. `check_dead_references` went from 1 failure to clean. `audit/enrollment_report.md:329` still names the old path — that is a generated snapshot, pre-existing on `main`, outside the gate's prose scan, and I left it alone rather than hand-editing a generated file.

7. **`premerge_check --base origin/main 2100`**: merge tree `fa80d3f60662`, all eight gates `pass → pass`, nothing goes green → red. Its base-name WARNING is expected — this PR targets `cpp/minigame-slot28` because it is stacked on #2099; the run merges into `origin/main`, which is the tree that will actually exist once #2099 lands.

**What this does not prove.** No ROM was built by the static gates, and a local `rombuild` cannot clear module fidelity the way the PR validator does — that verdict is the validator's. Six vtables in the family still score DIFFERS and will until slot 35 closes `_ZTV12dScMgAmida_c`.

Ready as far as I can take it; landing is the repo owner's call, and #2099 goes first.

andrewboudreau (commented):
**CHANGES REQUESTED** — one ask, and it is the same one still open on #2099 rather than a new class of problem. Reviewed at `50604acdb`.

First, the part I went looking to disprove and could not: **your return-type framing is right, and I can now upgrade it from a disassembly reading to a build measurement.**

The header says slot 29's `int` is "A HINT, the second consecutive slot that no body pins." I tried to pin it. Built in a wired worktree detached at `50604acdb`, `2004/b56`, against ov004:0x020af094 size 0x1e8:

```
CONTROL  shipped void                          MATCH
         void + local (restructure control)    MATCH
         int, early exit `return r26;`         MATCH          <- int is spellable
         int, early exit `return 2;`           size 0x1ec  (+1 word)
         int, early exit `return 0;`           size 0x1ec  (+1 word)
         int, both exits valued `return 0;`    size 0x1f0  (+2 words)
d3dbase  int, tail call    ov006:0x020e6d24    MATCH
slot3    int, tail call    ov006:0x0210aa3c    MATCH
```

where `r26` is the slot-26 result hoisted into a local:

```c
int r26 = self->OnHitByCannonBlastedChar();
if (r26 == 2)
    return r26;
```

That is the spelling your own disassembly note describes — "returning the 2 it has just compared against slot 26's result" — and it costs nothing, because r0 already holds it at the `bxeq`. The constant spellings cost a word precisely because they force a `mov`. So the ROM is **neutral** on slot 29, in both directions, by build. `int` is not refuted and neither is `void`.

I am recording the negative rows too, because they are the trap: had I stopped at `return 0;` I would have posted "the ROM refutes `int`" with a real-looking measurement behind it. A return-type refutation has to test the *free* spelling — the one that returns the value already in r0 — not just the one that is easy to write. That generalizes past this slot and past this PR.

## The ask

The ROM being neutral is exactly why this has to be resolved by consistency, and right now four places disagree with one:

| where | spelling |
|---|---|
| `include/dScMgBase_c.h:503` | `virtual int OnAimedAtWithEgg();` |
| `include/dScMgD3DBase_c.h:98` | `virtual int` |
| `include/dScMgSlot3_c.h:59` | `virtual int` |
| `include/decl_common.h:2322` | `extern void _ZN11dScMgBase_c16OnAimedAtWithEggEv(void*);` |
| `src/_ZN11dScMgBase_c16OnAimedAtWithEggEv.cpp:47` | `extern "C" void` |
| `src/_ZN14dScMgD3DBase_c16OnAimedAtWithEggEv.cpp` | `void` |
| `src/_ZN12dScMgSlot3_c16OnAimedAtWithEggEv.c` | `void` |

`decl_common.h:2322` is the one worth naming explicitly: `_ZN11dScMgBase_c16OnAimedAtWithEggEv` **is** the mangled name of the member the class declares `virtual int`. The compiler never sees the collision — one is a class member, the other an `extern "C"` free function — but the linker resolves them to a single symbol carrying two declared return types. That is also why a naive flip of just the definition does not build (`illegal function overloading`); it cost me a run to find out, so: the decl has to move with it.

Either resolution merges, same as I said on #2099:

- **Honor the headers.** All three definitions plus `decl_common.h:2322` become `int`, the base gets the `r26` hoist above, and the two forwarders `return` their tail call. All four are byte-verified `MATCH` in the table above — this is a known-good edit, not a suggestion to go find one.
- **Or make the headers `void`,** matching what the four definitions already say and what `decl_common.h` already declares.

I have no preference between them. What I will not wave through is the slot shipping with the vtable's own declaration contradicting the symbol it points at, when the ROM has declined to break the tie.

## Carried from #2099

`src/_ZN12dScMgSlot1_c19OnHitFromUnderneathEv.cpp` still carries the third copy of the "nothing loads vtable+0x70" claim that the ov004/ov006 scan refuted. I confirmed this PR does not touch it. It is #2099's debt, not yours, and I am noting it here only so it does not fall through the gap between two stacked PRs.

## Sequencing

This is stacked on #2099 (`cpp/minigame-slot28`), which is blocked on the two asks above, so #2100 cannot merge ahead of it regardless of what happens in this thread. Resolve #2099 and this one collapses to a small edit.

## Credit where it is due

The caller measurement is the best work in this stack. Scanning for the dispatch *pair* — `ldr rN,[rM,#0x74]` immediately followed by `blx rN`, not any load at +0x74 — is the difference between a measurement and a coincidence, and finding 35 sites of which exactly one is in this class's two modules, then reading that one and showing its result is dead before anything can consume it, is what turned "no caller settles it" from an assumption into a fact. It is also what let me trust that the neutral result above is the real answer rather than an untested caller waiting to contradict it.

The SEMANTIC CAUTION block is the right call and I want it to stay. Every name on this class is an index transplant from `dActor_c.h`, there was never a second witness, and this is the slot where the body — save POWCNT1's swap bit, blank both DISPCNTs, stash the BG palette and 0x2000 bytes of sub-OBJ VRAM, decompress a per-language image over it — visibly is not a Yoshi egg. Keeping the name as a **label** rather than coining a replacement is right: a coined name on a byte-exact symbol would block TU promotion later and would read as evidence it is not.

Banking the slot-30 warning forward is the cheapest thing in this PR and possibly the most valuable. `dActor_c.h:151` declaring a by-value `Vector3` cannot survive a dispatch site that has `this` in r0 and the loaded pointer in r1; under AAPCS a 12-byte return needs the hidden result pointer in r0. Whoever writes slot 30 will save a day because you wrote that down here.

And the sixth name correction on `dScMgD3DBase_c` is handled correctly — five tables referencing 0x020e6d24 makes it D3DBase's, and noting that this one is a real 0x68-byte body rather than the twelve-byte veneers of slots 26–28 is the detail that justifies the include and cast changes riding along with it.

PR #2130, opened 2026-09-01
Rename ov002/Tree to its cartridge RTTI name daTree_c and promote it as src/actors/d_a_tree.cpp

andrewboudreau:
Second snake_case genuine-TU promotion, stacked on #2129.

Collapses the eight one-function sources for class `Tree` into a single
translation unit enrolled as one `complete` delinks span, named by the
snake_case scheme in `notes/tu-naming-and-swallowers.md` sec 1
(`tools/tu_names.py`). The class name is untouched.

### Why the destructors keep their shim form

A real `Tree::~Tree()` is this class's key function, so defining it would make
this TU emit the complete `_ZTV4Tree`/`_ZTI4Tree`/`_ZTS4Tree` group. Those
records cannot be word-compared against the cartridge while the class carries
the coined name `Tree` — the ROM's own RTTI string is `daTree_c`, and an
`_ZTI`/`_ZTS` record is a length-prefixed mangled string, so a coined name
misses on both the prefix and the body (see `tools/tubuild.py`
`apply_compiler_only_policy`). Keeping the mangled `extern "C"` shims leaves
`_ZTV4Tree` an UNDEFINED import, so this entry needs neither
`compiler_only_output` nor `externalized_output`.

The cost is that `_ZN4TreeD0Ev` and `_ZN4TreeD1Ev` leave the CONVERTED tier;
both are logged in `config/converted-backslide-exceptions.jsonl` with a reason.
Re-spelling the class to `daTree_c` would recover them and is the natural
follow-up, but it moves every mangled symbol in the delinks span and is
deliberately not bundled here.

### Verification

- `tools/rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0**
- `tools/eligible.py`: 11096 / 11211
- `tools/port_refcheck.py`: 405 checked, all resolve
- `tools/check_dead_references.py`: no new dead references, no broken links
- `tools/layout_check.py`: clean
- `tools/tiers_ratchet.py --update`: 2572 -> 2570 (the two dtor backslides above)


andrewboudreau:
Reworked after the validator rejection.

The original push here promoted `ov002/Tree` while the class kept its **coined** name, and it lost a ROM-data verification: the eight legacy per-function objects between them proved `_ZTV4Tree` against the cartridge, the merged TU proved nothing, and the validator failed with `ROM data verified from source fell from 566 to 565 symbol(s)`.

Two defects with one cause:

1. A coined class name reaches no symbol home, so `_ZTI`/`_ZTS`/`_ZTV` could never be word-compared against the ROM.
2. `InitResources` — the class's **first declared** virtual, which is what the key-function rule selects, not the first slot — was a hand-mangled `extern "C"` free function, so no TU anywhere defined the key function and mwccarm emitted no RTTI group at all. With the destructor moved inline, objisolate refused the whole TU: `_ZN8daTree_cD1Ev has 0 defined symbols, expected one`.

Measured in `extracted/overlays/overlay_0002.bin` (ov002 base `0x020ad660`): the vtable preamble at `0x0210abf8` is `[0, 0x0210abd0]`, and `_ZTI8daTree_c` at `0x0210abd0` reads `[0x0209a764, 0x0210abc4, 0x0208e390]` — `__si_class_type_info`, a typeinfo name whose bytes are `"8daTree_c"`, and `_ZTI8dActor_c`. `config/arm9/overlays/ov002/symbols.txt` already carried both records at those addresses under the coined name.

Fix: rename to `daTree_c`, snake stem `src/actors/d_a_tree.cpp`, `InitResources` becomes a real member, destructor inline, `&_ZTV8daTree_c[2]` at the spawn. All nine emitted records are `deadstrip-data` rows with a canonical module and address.

`tools/rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0**, and ROM data verified from source rises to **567** — one *above* the pre-promotion baseline. CONVERTED ratchet 2590/2590; eligible, port_refcheck, check_dead_references, layout_check clean.

`Tree_Spawn` keeps its name: `symbols.txt` names that ROM symbol unmangled and it is not RTTI-derived.

PR #2156, opened 2026-09-01
Promote daIDonketu_c (ChillBully) to one snake_case translation unit

andrewboudreau:
ov027's Bully-on-the-ice. Seven per-function sources collapsed into
`actors/d_a_i_donketu.cpp` (source-owned), one `complete` delinks span, and the class
renamed to the name the cartridge itself carries.

## The rename and the promotion are one edit

They cannot be split into two building commits. The promotion needs the destructor
inline; inlining it makes the two surviving per-function destructor sources redefine it:

```
object 'daIDonketu_c::~daIDonketu_c()' redefined
  src/_ZN12daIDonketu_cD0Ev.cpp:20
  src/_ZN12daIDonketu_cD1Ev.cpp:17
```

Measured, not assumed — that is why this is one commit.

## The name is the ROM's

Read out of `extracted/overlays/overlay_0027.bin`: the vtable object's preamble at
`0x02113928` is `[0, 0x021138b0]`; the typeinfo record there reads
`[_ZTVN3abi20__si_class_type_infoE, 0x021138c8, 0x021138bc]`; the name string at
`0x021138c8` is `"12daIDonketu_c"` and the base's at `0x021138a4` is `"7daOts_c"`, which
the tree already spells that way. The measurement is written out at the top of
`include/daIDonketu_c.h`.

The coined `ChillBully` reached no symbol home, so its `_ZTI`/`_ZTS`/`_ZTV` could never be
word-compared against the cartridge and this TU could not have declared a verifiable
`compiler_only_output`. The rename is what makes the promotion possible at all.

## Inline destructor, key function `Behavior`

Out of line, mwccarm emits D0 before D1 — the reverse of ROM order, which `objisolate`
refuses for the whole TU (this entry's own pre-promotion block recorded `ordinal pair(s)
not in ROM order: [(0, 1)]`) — plus a homeless D2. Inline gives the retail D1/D0 pair in
ROM order and no D2. The key function is then the first **declared** non-inline virtual,
`Behavior`, so this TU still owns the class's `_ZTV`/`_ZTI`/`_ZTS` group; all fourteen
vague-linkage outputs are licensed as `deadstrip-data` in the manifest.

The factory's second vptr store is `&_ZTV12daIDonketu_c[2]`: this TU now defines the
vtable, and mwccarm's symbol addresses the object start with its two-word preamble while
`symbols.txt`'s names the slot array. `_ZTV7daOts_c` is not defined here and takes no bias.

## Evidence

| check | result |
| --- | --- |
[...]

PR #2148, opened 2026-09-01
Rename IceSlideManager to daSldMng_c and consolidate it into one TU

andrewboudreau:
Another coined class moves to the spelling the cartridge's own RTTI carries, and its five
one-function sources plus a never-enrolled `src_tu/` shadow collapse into a single genuine
translation unit under the snake_case scheme in `notes/tu-naming-and-swallowers.md` sec 1
(`tools/tu_names.py`).

| coined name | cartridge name | new TU | module / `.text` |
|---|---|---|---|
| `IceSlideManager` | `daSldMng_c` | `src/actors/d_a_sld_mng.cpp` | ov019 `0x0211261c..0x0211277c` |

Two commits, deliberately separate: the rename alone, then the promotion.

## Why the rename is load-bearing, not cosmetic

A coined class name reaches no symbol home. Its `_ZTI`/`_ZTS`/`_ZTV` records are emitted
under a mangling the config carries nowhere, so `romdata_check` can never word-compare them
against the cartridge and the TU cannot declare a verifiable `compiler_only_output`
(`tools/tubuild.py apply_compiler_only_policy`).

Measured, not assumed, in `extracted/overlays/overlay_0019.bin` (ov019 base `0x021111a0`,
the `.text start:` on line 1 of `config/arm9/overlays/ov019/delinks.txt`): the vtable
object's preamble at `0x021133c4` is `[offset-to-top 0, 0x0211338c]`, and
`_ZTI10daSldMng_c` at `0x0211338c` reads `[0x0209a764, 0x02113398, 0x0208e390]` —
`_ZTVN3abi20__si_class_type_infoE` (`config/arm9/symbols.txt`), a typeinfo name at
`0x02113398` whose bytes are the string `"10daSldMng_c"`, and `_ZTI8dActor_c`, so the
cartridge states the direct base too.

`config/arm9/overlays/ov019/symbols.txt` already carried `_ZTI10daSldMng_c` and
`_ZTS10daSldMng_c` at those two addresses while every function symbol and the vtable were
still spelled `IceSlideManager` — the same half-configured state `daTree_c` was in.

## The key function, and what it proves

With the destructor inline in the header, the class's key function is its **first declared
non-inline virtual** — `InitResources`. The new TU defines it, so it is the key-function TU
and emits the whole RTTI group. All nine emitted records are licensed as `deadstrip-data`
with their cartridge addresses, and `tools/romdata_check.py` compares them, relocations
applied:

```
[...]

PR #2160, opened 2026-09-01
Point two pilot-table intake files at where those factories live now

andrewboudreau:
**`main` is red** on `tools/check_dead_references.py`, and has been since #2127 landed:

```
FAIL: 2 prose reference(s) name a path that does not exist:
  notes/tu-filename-reconstruction-pilot.md
      names `src/FlyGuy_Spawn.c`, which is not in the tree
  notes/tu-filename-reconstruction-pilot.md
      names `src/WaterfallMist_Spawn.cpp`, which is not in the tree
```

Reproduced on a clean checkout of `3d4377d61`, so this is not a branch artifact.

Neither PR was wrong on its own. #2127's pilot table was written against the tree as it
stood, and the TU promotions that deleted those two files were in flight beside it — two
green PRs, one red `main`. Every open PR now inherits the failure through its
merge-with-`main` check; it is what turned #2148 red, and #2130, #2151 and #2156 are
behind it too.

## The fix

The table's second column is headed **"Current intake file"**, so the honest fix is to
name the file each factory lives in today rather than to bank the dead paths in
`config/dead-reference-baseline.json`:

| profile | was | is now |
| --- | --- | --- |
| `WATERFALL` | `src/WaterfallMist_Spawn.cpp` | `src/actors/d_a_obj_waterfall.cpp` |
| `PROPELLER_HEYHO` | `src/FlyGuy_Spawn.c` | `src/actors/daPropeller_Heyho_c.cpp` |

Both were checked by grep, not assumed — `WaterfallMist_Spawn` is defined in the waterfall
TU, and `FlyGuy_Spawn` is ROM ordinal 26 in `daPropeller_Heyho_c.cpp`. Every other row in
the table still names a live file.

Prose only; no build input is touched. `check_dead_references` passes on the branch.


andrewboudreau:
Closing as an exact duplicate of #2159: same base, same file, and byte-for-byte identical two-line path repair. #2159 is older and already has all public checks green.

PR #2167, opened 2026-09-02
Promote ov002/daSoundObj_c to a genuine snake-case TU

andrewboudreau:
`ov002/SoundObject` becomes `daSoundObj_c`, and its seven per-function files become one real translation unit at `src/actors/d_a_sound_obj.cpp`.

Two commits, deliberately separate: the rename lands alone, then the promotion.

ov002 `0x0210c048` is a `__si_class_type_info`. Word [1] points at `0x0210c054`, which holds exactly `12daSoundObj_c`; word [2] points at arm9 `0x0208e390`, which is `_ZTI8dActor_c`. So the ROM spells the class `daSoundObj_c` and its one base, at subobject offset 0, is `dActor_c` — which is what `include/SoundObject.h` already described from the factory and the destructor, under a coined name.

That rename is not cosmetic. A coined name is a length-prefixed mangled string that matches nothing at any address, so none of the class's vague-linkage data could ever be word-compared against the cartridge — which is what blocks a class from being its own key-function TU and from declaring a verifiable `compiler_only_output`.

One orphan removed: `include/daSoundObj_c.h` already existed as a `tools/deepen_rtti.py` placeholder (`pad_000[0xd4]` plus four `unk_` fields, no base clause, no vtable) and was `#include`d by nothing. The real header takes that filename.

## The inline destructor is the load-bearing edit

Declared out of line, mwccarm 2004/b56 emits D0 before D1 — the reverse of the cartridge, which has D1 at `0x020f934c` and D0 at `0x020f9370` — and adds a third, homeless D2 that no ROM address claims. `tools/objisolate.py` then refuses the **whole TU** rather than one function. Defined in the header the pair comes out in ROM order with no D2 at all, and `tubuild`'s `functions_occur_in_expected_order` goes from `PARTIAL -- ordinal pair(s) not in ROM order: [(0, 1)]` to `PASS`.

**Safe only because the class is a leaf, and that is measured rather than assumed.** `_ZTI12daSoundObj_c` occurs exactly once as a word anywhere under `extracted/` — ov002 offset `0x5ea78`, which is `0x0210c0d8`, the typeinfo slot of its own vtable header at `_ZTV-4`. No other class's `__si_class_type_info` points at it, so no descendant exists to inline this body where the ROM has a `bl`.

## Nine vague-linkage records get ROM homes

Naming the class put its RTTI and vtable in this object, and this TU defines the key function. Per-symbol romdata verdicts:

| record | verdict |
| --- | --- |
| `_ZTV12daSoundObj_c` ov002 `0x0210c0dc` | VERIFIED, 124 bytes = 31 slots |
| the four `_ZTI` records | VERIFIED |
| the four `_ZTS` records | PARTIAL |
| anything | no DIFFERS |

The `_ZTS` PARTIALs are the known dsd extent shortfall — a range that stops short, not a disagreement about bytes.

**The vtable claim is deliberately scoped to the slots.** A `symbols.txt` `_ZTV` address is the address point, eight bytes past the table's real start, so the `{offset-to-top, _ZTI pointer}` header pair is emitted here and word-compared by nothing. The 31 slots are proved; those two words are not.

## The one pragma, resolved

The legacy sources carried a single directive, `#pragma opt_loop_invariants off` on `InitResources`. It is the **file-global last-wins** kind, not the positional kind (`long_calls` does not appear in this TU). Bracketing it around that one member had no effect whatever — the closing `on` set the whole file's state back — so it is left unbracketed, with its file-wide safety argued in place: no other member in this TU has a loop.

## Verification

- `tools/rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0**
- `tools/tubuild.py verify ov002/daSoundObj_c`: **7/7 MATCH**, objisolate clean, reloc destinations clean, emission order PASS → TEXT-VERIFIED
- CONVERTED holds at **2591**, ratchet **+0/−0** — the two moved entries are the rename into `path#symbol` form, not a backslide
- `eligible`, `port_refcheck`, `check_dead_references`, `layout_check`: all clean

## One commit that is not this branch's

`check_dead_references` FAILs on `origin/main` today, independently of any TU work: `notes/tu-filename-reconstruction-pilot.md` names `src/func_02023624.c` and `src/Stage_Spawn.cpp`, and neither is in the tree — both were renamed by an earlier class rename that did not carry into the prose. The third commit points those two rows at `src/dScBoot_c_classInit.c` and `src/dScStage_c_classInit.cpp`. Prose only, no build effect; carried here only because a red references gate would otherwise block this merge.


andrewboudreau:
## Review pass, and what it changed

An independent read rebuilt all seven functions itself with the canonical pin (`2004/b56`, reloc-destination check active) and got **7/7 MATCH**, including the compiler-synthesized `D0` — so the inline-destructor argument is confirmed empirically, not just argued. It also reproduced both load-bearing evidence claims:

- **Leaf.** Scanning every file under `extracted/` for the LE word `0x0210c048` gives 2 raw hits at the *same* file offset `0x5ea78` in two identical images (same md5), i.e. **1 logical hit**. ov002 base `0x020ad660 + 0x5ea78 = 0x0210c0d8` = `_ZTV-4`, the class's own vtable typeinfo slot. No compression hole in the scan: `arm9_dec.bin` is present and clean, and every `overlay_*.bin` length equals its `code_size`, so the images are decompressed.
- **Name and base.** `0x0210c048` reads `[__si_class_type_info+8, 0x0210c054, 0x0208e390]`; `0x0210c054` is `12daSoundObj_c\0`; `0x0208e390` is `_ZTI8dActor_c`, itself an `__si_class_type_info` whose `_ZTS` reads `8dActor_c`.
- **Vtable diff, run independently:** `_ZTV12daSoundObj_c` and `_ZTV8dActor_c` are 31 words each and differ in **exactly slots 0, 6, 16, 17** = `InitResources`, `Behavior`, D1, D0 — the four members the header declares. The `_ZTV… + 2` address-point spelling matches the ROM literal at `0x020f9758` and five landed siblings.

`port_refcheck` 405/0 stale, `check_dead_references` green.

### Fixed in 5a4ea6143

The review's substantive findings were all about the *prose*, and it was right about the worst one:

1. **The header's headline justification was false.** The "TWO WITNESSES" preamble said the factory initializes the members in order and the destructor destroys them in reverse. Neither does — `daSoundObj_c_Spawn` is `operator new(0xe4)` / base ctor / vptr store / return and touches no member, `~daSoundObj_c` is a vptr store and a tail call, and the destructor comment 25 lines below already said exactly that. Rewritten: the **size** has one witness (Spawn's `mov r0, #0xe4`), the **offsets** have a different one (`InitResources` and the two callbacks).
2. **Three field names were contradicted by this TU's own code.** `0xd4` and `0xd8` are handed straight to `Sound::PlaySub` as `soundID` and `vol`, and `0xd8`'s complement drives the distance argument as `(0x7f - vol) << 12`. So `mLevelID`/`mTimerThreshold` → **`mSoundID`/`mVolume`**, and `mTimer`/`mTimerReset` → **`mCounter`/`mCounterLimit`** (a counter and the limit it is compared against). That reads the `0xc`-stride table as `{soundID s32, volume s32, count u16, flag u8}` per row. The header now carries `dActor_c.h`'s caveat sentence, which it lacked.
3. **`Behavior`/`InitResources` are overrides spelled like plain members** — now `virtual` with `/* slot 6 */` and `/* slot 0 */`, matching every sibling header. Codegen-free.
4. **`SoundObjectCallbackOwner`** keeps its coined name but is now a `TODO:` rather than settled framing, with the evidence that it is probably `daSoundObj_c` itself and the reason it cannot be settled from the image (the table at `data_ov002_0211110c` is in `.bss`, populated at runtime). It is a verify-cycle question, not a merge-path change.
5. One redundant `extern void* data_020a0eac;` dropped — `include/dActor_c.h` already declares it.

Re-proved after those edits: **7/7 MATCH, objisolate clean, reloc-destinations clean, TEXT-VERIFIED**, and **106/106 exact, 100.000000%, mismatching 0**. CONVERTED holds at 2591, ratchet +0/−0.

### Deliberately not done

- **`kPoolDist`** — the review read it as invented. It isn't: `src/func_0200f7f0.c` and the two `Sound::Play*SecretSound` files give the same `0x8777` the same name. Left alone, with a note saying so and saying why its sibling `0xcb33` stays bare.
- **`docs/class-reference.html` is stale** on `SoundObject` — correct, and worth a follow-up. Its only generator (`tools/rtti_reference.py`) is in no workflow and the page has not been refreshed across ~94 commits of main, so regenerating it here would bury this diff in unrelated churn. Separate refresh.
- **`sizeof(daSoundObj_c)` in place of the factory's `228`**, and the duplicated `g == 0x21` and the four `symbols.txt` names for one table's four fields — pre-existing, byte-neutral, not this PR's.
- **The nine-record `VERIFIED`/`PARTIAL` table** was not independently re-derived by the reviewer; it comes from this branch's own `rombuild` run, and the `PARTIAL` rows are the known dsd `_ZTS` extent shortfall.

### Merge gate

`premerge_check.py --repo . --fetch 2167` on the merge tree: nothing goes green → red; `dead-references` goes **fail → pass (fixed)**; `source-coverage` errors identically on base and merge (`git ls-tree … fatal: not a tree object`) and is not this PR's to own.

Signing off here rather than approving — this repo's shared account cannot approve its own PRs.

PR #2168, opened 2026-09-02
Promote ov002/daSetSE_c to a genuine snake-case TU

andrewboudreau:
Promotes `ov002/AmbientSoundEffects` to a genuine single-file translation unit under the name the cartridge gives it, `daSetSE_c`, at `src/actors/d_a_set_se.cpp`. Same recipe as #2167, one class further along.

### The name is read from the ROM, not from a tool's summary

ov002 `0x0210b470` reads `[0x0209a764, 0x0210b464, 0x0208e390]`: word 0 is `_ZTVN3abi20__si_class_type_infoE+8` (single inheritance), word 1 points at `0x0210b464` whose bytes are exactly `9daSetSE_c`, and word 2 points at arm9 `0x0208e390` = `_ZTI8dActor_c`. The manifest had already recorded those two retail records as the ones it could not license, and marked its own three derived records unlicensed "while its RTTI/type-name identity differs from retail". A coined name is a length-prefixed mangled string that matches nothing at any address, so under `19AmbientSoundEffects` none of this class's vague-linkage data could ever be word-compared — which is exactly what blocks a class from owning a key-function TU.

### It is a leaf, and that is measured

The word `0x0210b470` occurs **once** anywhere under `extracted/` — offset `0x5de64` of ov002, i.e. `0x0210b4c4`, the typeinfo slot of its own vtable header at `_ZTV-4` (`symbols.txt` puts the address point at `0x0210b4c8`). The two raw hits are the same image under two paths, identical by md5. No other class's `__si_class_type_info` points at it.

That is what makes the inline destructor safe, and the inline destructor is what unlocked the promotion: this entry had been stuck at `functions_occur_in_expected_order: PARTIAL -- ordinal pair(s) not in ROM order: [(0, 1)]`. Out of line, mwccarm 2004/b56 emits D0 before D1 — the reverse of the cartridge's `0x020f198c`/`0x020f19b0` — and adds a homeless D2 no ROM address claims, whereupon `objisolate` refuses the whole TU. Moved into the header, the pair comes out in ROM order, the D2 is gone, and that criterion is **PASS**.

### The vtable was diffed, not assumed

`_ZTV9daSetSE_c` and `_ZTV8dActor_c` are 31 words each and differ in exactly **seven** places — slots 0, 3, 6, 9, 12, 16, 17 — and each differing word is the address of one of this entry's own functions:

| slot | word | member |
|---|---|---|
| 0 | `0x020f1ac4` | `InitResources` |
| 3 | `0x020f19e8` | `CleanupResources` |
| 6 | `0x020f19fc` | `Behavior` |
| 9 | `0x020f19f4` | `Render` |
| 12 | `0x020f19f0` | `OnPendingDestroy` |
| 16 | `0x020f198c` | D1 |
| 17 | `0x020f19b0` | D0 |

Those are precisely the five overrides plus the destructor pair the header declares; the slot numbers are now written beside them. Every other slot is inherited and deliberately not redeclared. The claim is scoped to the slots — a `symbols.txt` `_ZTV` address is the address point, eight bytes past the table's real start, so the `{offset-to-top, _ZTI pointer}` header pair is emitted here and word-compared by nothing.

### What the promotion does

Eight `complete` per-function `delinks.txt` entries collapse into one contiguous range, `.text 0x020f198c-0x020f1bc4`; eight legacy sources are deleted and the shadow TU moves to the production path (git detects it as a rename, R054). Nine vague-linkage records are declared `deadstrip-data` with the cartridge address each is compared against — the three derived at ov002 `0x0210b4c8`/`0x0210b470`/`0x0210b464`, the six inherited in arm9.

### Proof

- `tubuild verify ov002/daSetSE_c` — **8/8 MATCH**, objisolate clean, reloc-destinations clean, emission order PASS → TEXT-VERIFIED, **no unlicensed output**.
- baseline `linkcheck` control — modules PASS, ROM **identical** to the tree's stock build. (Its `dsd check symbols` FAIL is the tool's own pre-existing tree-level failure: it appears with no TU substitution at all, which is what the baseline control exists to show.)
- `rombuild.py -j16` — **106/106 exact, 100.000000%, mismatching 0**, zero new symbol errors, storage aliases exact. Run twice: once on the rename alone, once after promotion.
- All nine declared records now get a per-symbol verdict from this TU — **5 VERIFIED** (`_ZTV9daSetSE_c` 124 bytes, `_ZTI9daSetSE_c`, `_ZTI8dActor_c`, `_ZTI7dBase_c`, `_ZTI7fBase_c`) and **4 PARTIAL** (the `_ZTS` records, the known dsd extent shortfall). Under the coined name the three derived ones could not be compared at all.
- The tree-wide counters move **590 → 591 verified, 173 → 174 partial**. That is +2, not +9, and deliberately so: six of the nine are inherited arm9 records already counted from other TUs that emit them, and the totals are de-duplicated per symbol. Do not read the +2 as a one-to-one map onto the three derived records — the per-symbol verdicts above are the claim; the counter delta is a side effect of a global set.
- CONVERTED holds at 2591, ratchet **+0/−0, BACKSLID 0** — the five banked entries migrate from per-file paths to `src/actors/d_a_set_se.cpp#symbol` identities.
- `check_dead_references`, `port_refcheck` (405/0 stale), `check_header_offsets` (`struct spans 0xd8`, 0 mismatched) all green.

### Third commit

Deletes two rows from `config/dead-reference-baseline.json` that #2167 made resolve — deletions only, no new row banked.


andrewboudreau:
## Review pass, and what came of it

Reviewed adversarially against the ROM rather than against my own prose: the vtable diff, the leaf scan and the `delinks.txt` range were all recomputed from `extracted/` independently. Those three came back **exactly as claimed** — 31 words each differing at slots 0, 3, 6, 9, 12, 16, 17; one logical hit for `_ZTI9daSetSE_c` at `_ZTV-4`; the range covering the 8 functions contiguously with 0x238 bytes and nothing else. Field names, types, `pad_0d0[4]`, the 0xd8 size, the 9 `compiler_only_output` addresses, the `converted-baseline` migration and the two `dead-reference-baseline` deletions also verified clean.

Four things did not survive, and `e1fdebe67` fixes them.

**1. I counted vptr stores wrong, in both headers.** "a one-level chain, so two vptr stores come out of it" — disassembled, D1 is `push {r4,lr} / ldr r1,[pc] / mov r4,r0 / str r1,[r4] / bl _ZN8dActor_cD2Ev / mov r0,r4 / pop / bx lr`, **one** store of its own; the base destructor does two more and then calls `_ZN7fBase_cD2Ev`, so three across the chain. `src/actors/d_a_set_se.cpp:171` had the correct description all along, so the header contradicted its own TU. Same sentence was in `include/daSoundObj_c.h` from #2167 and is corrected here too — it is on `main` today and I would rather not leave it there. Both now quote the instructions instead of counting.

**2. "no member of either class has a destructor of its own" was false for `dActor_c`** — it owns `fLiNdBa_c mActorListNode` at 0x50 and destroys it at `0x020112e8`. The claim that actually matters, that *these two classes* contribute no member destructor call, is true and is now what the comment says.

**3. The manifest licensed and un-licensed the same nine records.** Promotion added `compiler_only_output`, but the pre-rename `unlicensed_output_observed` block came through verbatim: still "Ten definitions are present in the raw object", still listing `_ZN9daSetSE_cD2Ev` — which the inline destructor removes — and still saying the derived records were "not licensed while its RTTI/type-name identity differs from retail", the exact condition commit 1 eliminates. It sat directly beneath `declared_function_set_equals_defined_function_set: PASS`. Removed; 49 of the 51 promoted entries carry no such block. `verification.object` now follows the renamed object.

**4. The rename reached eight ledger rows it had no business touching.** `symbols/actor_renames.tsv` is a live `addr -> symbol` map (`tools/cpp_rename.py:51` loads it), and `class_rename.py` substituted into every row containing the old name. Seven function rows `0x020f1bc4..0x020f1eb4` and the vtable row `0x0210b584` belong to **MugenBgm** per this repo's own `symbols.txt`; `tools/evidence_history.py:802` records why (`_ZN19AmbientSoundEffects*` is today's MugenBgm). Reverted. The two rows that genuinely are this class's, `0x020f1b94` and `0x0210b47c`, keep the new name.

### Declined, with reason

- **`partial_isolation` left as-is**, including `delinksChanged: false` and the eight `substitutedObjectPaths` this branch deletes. It is a record of the pre-promotion partial round, not a claim about the tree now, and the already-merged `daSoundObj_c` entry carries the identical block — so this is the convention, not staleness.
- **The stale per-function paths** in `notes/data/tu-merge-candidates.json`, `symbols/profile_reconstruction_registry.*` etc. are the established norm — `src/daCamTag_c_Spawn.c` is stale the same way on `main` after #2167. Not introduced here.

### Corrected in the PR body

`R100` was wrong — git reports **R054** for the `src_tu` -> `src` move (still a rename, still carries credit). And the "590 -> 591 verified" line implied nine records mapping onto a +2 counter move; the honest statement, now in the body, is that this TU produces **5 VERIFIED / 4 PARTIAL** per-symbol verdicts while the tree-wide totals are a de-duplicated global set that six inherited arm9 records were already counted in.

### Flagged, deliberately not fixed here

- Those eight `actor_renames.tsv` rows still name a class that does not own those addresses — a **pre-existing** defect in a live lookup table, with its own evidence to gather. Reverting was this branch's job; correcting them to `MugenBgm` is not.
- `docs/class-reference.html` still renders the old names (carried over from #2167). Its only generator is in no workflow, so regenerating it inside a promotion PR would bury the diff in unrelated churn.

Re-proved after every edit: `tubuild verify` **8/8 MATCH**, objisolate clean, emission order PASS; `rombuild -j16` **106/106 exact, 100.000000%, mismatching 0**; ratchet +0/-0, BACKSLID 0; `check_dead_references`, `check_header_offsets` (both headers, 0 mismatched) green. Nothing above changed a byte.

Signing off on the change as it now stands.

PR #2171, opened 2026-09-02
Rename VirtualDoor to daChRoom_c and promote it to a single TU

andrewboudreau:
`ov002/VirtualDoor` was a coined name. The cartridge spells the class
`daChRoom_c`, and once it is named that, its RTTI can be word-compared against
the ROM and it can own a translation unit. Two commits: the rename, then the
promotion of eight per-function objects into `src/actors/d_a_ch_room.cpp`.

ov002 `0x021085b8` is a `__si_class_type_info`: word[0] is
`_ZTVN3abi20__si_class_type_infoE`+8, word[1] points at `0x021085c4` which holds
exactly `10daChRoom_c`, and word[2] points at arm9 `0x0208e390` =
`_ZTI8dActor_c`. So the ROM spells the class `daChRoom_c` and its one base, at
subobject offset 0, is `dActor_c`. `_ZTI10daChRoom_c` and `_ZTS10daChRoom_c`
were already correctly named in `symbols.txt` before this branch -- the class
name in the source tree was the thing that disagreed with them.

## It is a leaf, measured not assumed

The word `0x021085b8` occurs exactly once anywhere under `extracted/` -- ov002
file offset `0x5af94` = `0x021085f4`, the typeinfo slot of its own vtable header
at `_ZTV - 4`. (The two hits the scan reports are the same 394048-byte image
under `extracted/dsd/arm9_overlays/ov002.bin` and
`extracted/overlays/overlay_0002.bin`.) No other class's
`__si_class_type_info` points at it, so nothing derives from it, so the inline
destructor is safe.

## Why the destructor has to be inline

Out of line, mwccarm 2004/b56 emits D0 before D1 -- the reverse of the
cartridge, which has D1 at `0x020b07f8` and D0 at `0x020b081c` -- and adds a
homeless D2 no ROM address claims; `objisolate` then refuses the whole TU.
Inline, the pair comes out in ROM order with no D2.

The body really is empty: D1 is 0x24 bytes, `push {r4,lr} / ldr r1,[pc,#0x14] /
mov r4,r0 / str r1,[r4] / bl 0x020112c8 / mov r0,r4 / pop {r4,lr} / bx lr` --
one vptr store, then the tail into `~dActor_c`. No member destructor call,
because the class declares no member at all; every field it touches is an
inherited `dActor_c` one. Compare `daObjWakame_c`, whose inline `{}` compiles to
0x30 bytes because it owns a `ModelAnim`.

## The overrides are a vtable diff

`_ZTV10daChRoom_c` and `_ZTV8dActor_c` (arm9 `0x0208e3a4`) are 31 words each and
differ in exactly slots 0, 3, 6, 9, 12, 16, 17 -- `InitResources`,
`CleanupResources`, `Behavior`, `Render`, `OnPendingDestroy`, D1, D0. Those
seven and no others are declared in the header; the other 24 slots hold the
base's own word and are inherited.

## Nine records, plus one

| symbol | home | verdict |
| --- | --- | --- |
| `_ZTV10daChRoom_c` | ov002 `0x021085f8` | VERIFIED, 124 bytes = 31 slots |
| `_ZTI10daChRoom_c` | ov002 `0x021085b8` | VERIFIED |
| `_ZTI8dActor_c` / `_ZTI7dBase_c` / `_ZTI7fBase_c` | arm9 | VERIFIED |
| the four `_ZTS` records | | PARTIAL |
| nothing | | DIFFERS |

PARTIAL on `_ZTS` is the known dsd extent shortfall -- a range that stops short,
not a byte disagreement. A tenth symbol, `_ZN7Vector3D1Ev`, is licensed
`deadstrip-duplicate` rather than `deadstrip-data`: it is a function with a real
ROM home at arm9 `0x020072c0`, the same disposition `daObjCannonShutter_c` uses.

## Two things deliberately NOT renamed

`class_rename.py` proposed ten `symbols/actor_renames.tsv` rows; eight of them
(`0x020b09b0`, `0x020b09d4`, `0x020b0a70`, `0x020b0a78`, `0x020b0a7c`,
`0x020b0a84`, `0x020b0d88`, `0x021086b4`) are **Exit's** addresses carrying the
`VirtualDoor` label on main -- the whole region is shifted by one class, and the
rows above them labelled `daCamTag_c` hold daChRoom_c's real addresses. Only the
two rows that verify against `symbols.txt` (`0x020b0980` Spawn, `0x021085d4`
SpawnInfo) are renamed here. The mislabelled region needs its own evidence and
its own PR.

The tool also rewrote a `// recovered name: VirtualDoor_Kill` comment in
`src/func_ov002_020b2150.cpp`; that comment cannot be this class's -- the
function indexes `0x1ac`, `0x3a4` and a `dBgCh_Actr` subobject off `this`, and
`daChRoom_c` is 0xd4 bytes total. Reverted.

## Gates

- `tubuild.py verify ov002/daChRoom_c` -- 8/8 MATCH, objisolate clean, reloc-destinations clean, TEXT-VERIFIED
- `tubuild.py linkcheck --baseline --module ov002` -- modules PASS, ROM built, IDENTICAL to the stock build
- `rombuild.py -j16` -- **106/106 exact, 100.000000%**, 11,088 functions reproducing, mismatching 0
- `tiers_ratchet.py --update` -- 2592 -> 2592, +0/-0 (a rename within the set)
- `romdata_check.py` -- 594 VERIFIED / 175 PARTIAL / 5 DIFFERS (all 5 pre-existing, other classes), exit 0
- `layout_check.py` -- clean; `check_dead_references.py` -- no new dead references; `check_header_offsets.py include/daChRoom_c.h` -- struct spans 0xd4, 0 mismatched


andrewboudreau:
Reviewed, rebased onto `68154f947`, and re-proven: **106/106 exact, 100.000000%, mismatching 0** on the rebased tree; ratchet 2592 -> 2592, 0 backslid.

Three things the review turned up, and what happened to each.

**The branch was behind main.** Fixed by rebase — the three commits replay cleanly. The review also read this as the PR reverting `docs/index.html`, `docs/progress-treemap.svg`, `contributions.json` and `README.md`; it isn't. The branch's own diff never touches any of those files, so a merge takes main's copies. That reading came from comparing branch tip against main tip rather than against the merge base.

**A manifest note pointed at text that no longer exists.** Real, and fixed in a third commit. `notes[2]` of `config/tu_manifest.d/ov002/daChRoom_c.json` told the reader that "the stale text below" claimed a byte-length mismatch — but the promotion commit had deleted that sentence from `notes[1]` rather than leaving it. It now quotes the deleted sentence and states the miscount outright (10 characters read as 11).

**The phantom legacy paths are the tree's convention, not a defect.** `legacy_source` (x8) and `partial_isolation.substitutedObjectPaths` (x8) in the manifest, plus `current_factory_file` in `symbols/profile_reconstruction_registry.{json,tsv}` and the one row in `notes/data/c-cpp-classification.tsv`, all name `src/daChRoom_c_Spawn.c` / `src/_ZN10daChRoom_c*.cpp` — paths that existed at the rename commit and were deleted by the promotion. The landed `daObjWakame_c` PR on main has exactly this shape (`"legacy_source": "src/daObjWakame_c_Spawn.c"`, and the registry's `current_factory_file` the same), so this is what those historical-record fields mean here. Changing it would be a tree-wide convention change, not this PR's.

**One correction to my own commit message.** `55c85aad3` says the mislabelled `daCamTag_c` ledger rows "hold this class's real addresses 0x020b085c..0x020b0938". The shift actually spans seven rows starting at `0x020b07f8` — D1, D0 and CleanupResources are omitted from the range as written. It understates a region this PR deliberately does not touch; the eight Exit-owned rows left under the `VirtualDoor` spelling still need their own evidence and their own PR.

Everything else came back clean: every ROM claim independently re-derived (the RTTI triple at ov002 `0x021085b8`, the leaf byte-scan at file offset `0x5af94` = `_ZTV-4`, the 31-slot vtable diff differing in exactly slots 0/3/6/9/12/16/17, the 0x24-byte D1 whose `bl` resolves to `_ZN8dActor_cD2Ev`, the `mov r0,#0xd4` factory, and all ten `compiler_only_output` addresses); no cross-class rename contamination; `//cpp` first bytes; strict reverse-ROM function order; delinks span contiguous with its neighbours on both sides.

Signing off on it here rather than as an approval — this repo's PRs are all authored by the shared account.

PR #2173, opened 2026-09-02
Rename Warp to daWarpkun_c and promote it to the single TU src/actors/d_a_warpkun.cpp

andrewboudreau:
`ov002/Warp` was a coined name with eight per-function sources. It is now `daWarpkun_c`, spelled as the cartridge spells it, and one translation unit: **`src/actors/d_a_warpkun.cpp`**.

Two commits, deliberately separate — the rename has to land before the promotion, because a coined name can never be word-compared against the ROM and a class whose vague-linkage records cannot be compared cannot own a key-function TU.

ov002 `0x0210ac7c` is a `__si_class_type_info`: word[0] = `0x0209a764` (`_ZTVN3abi20__si_class_type_infoE`+8), word[1] → `0x0210ac88`, which holds exactly `11daWarpkun_c`, and word[2] → arm9 `0x0208e390` = `_ZTI8dActor_c`. So the ROM spells the class `daWarpkun_c` and its one base, at subobject offset 0, is `dActor_c`.

## It is a leaf, and that is measured

The word `0x0210ac7c` occurs at exactly one logical offset anywhere under `extracted/`: ov002 file offset `0x5d658` = `0x0210acb8` = `_ZTV − 4`, the typeinfo slot of its own vtable header. (Two raw hits, same 394048-byte image under two paths.) No other class's `__si_class_type_info` points at it — so inlining the destructor cannot strand a descendant.

## The destructor is the lever

Out of line, mwccarm 2004/b56 emits D0 before D1 — the reverse of the cartridge (D1 `0x020ec388`, D0 `0x020ec3b8`) — and adds a homeless D2; `objisolate` then refuses the whole TU. Inline, the pair comes out in ROM order with no D2, which is why the manifest's old `deadstrip` license for `_ZN11daWarpkun_cD2Ev` is gone rather than edited.

**The body is empty but the destructor is not.** D1 is `0x30` bytes, `0xc` more than a memberless class's `0x24`; the extra is `bl _ZN7dCcAc_cD1Ev` on `this+0xd4`, generated by the compiler from the `mdCcAc_c` member. Write that member as padding and you get the `0x24` shape and the function stops matching. The size closes independently: `include/dCcAc_c.h` asserts `sizeof(dCcAc_c) == 0x34`, and `0xd4 + 0x34 = 0x108` — the literal `mov r0, #0x108` the factory hands to `fBase_c::operator new`.

## The override set is the vtable diff

`_ZTV11daWarpkun_c` vs `_ZTV8dActor_c` (arm9 `0x0208e3a4`), 31 words each, differ in exactly slots **0, 3, 6, 9, 12, 16, 17** — `InitResources`, `CleanupResources`, `Behavior`, `Render`, `OnPendingDestroy`, D1, D0. Precisely the seven members the header declares; every other slot is inherited and deliberately not redeclared.

## The nine vague-linkage records

Naming the class puts them in this object. All nine are licensed `deadstrip-data` with the ROM home each is compared against. `romdata_check` per-symbol:

| record | home | verdict |
|---|---|---|
| `_ZTV11daWarpkun_c` | ov002 `0x0210acbc` | **VERIFIED**, 124 bytes = 31 slots |
| `_ZTI11daWarpkun_c` | ov002 `0x0210ac7c` | **VERIFIED** |
| `_ZTI8dActor_c` / `_ZTI7dBase_c` / `_ZTI7fBase_c` | arm9 | **VERIFIED** |
| the four `_ZTS` records | | PARTIAL |
| nothing | | DIFFERS |

The PARTIAL rows are the known dsd extent shortfall on `_ZTS` — a range that stops short, not a byte disagreement. The vtable claim is scoped to the slots on purpose: a `symbols.txt` `_ZTV` address is the *address point*, eight bytes past the table start, so the `{offset-to-top, _ZTI}` header pair is emitted here and word-compared by nothing.

## Proof

- `rombuild.py -j16` — **106/106 exact, 100.000000%, mismatching 0**, ROM-build analysis PASS.
- `tubuild.py verify ov002/daWarpkun_c` — 8/8 MATCH, objisolate clean, reloc destinations clean, **TEXT-VERIFIED**, no unlicensed records.
- `tiers_ratchet.py --update` — CONVERTED holds at **2592, +0/−0**. The moved entries are a rename within the set, not a backslide; no exception appended.
- `layout_check.py` clean; `check_header_offsets.py include/daWarpkun_c.h` — 0 mismatched, struct spans `0x108`.

The `ov002` baseline linkcheck control refuses `daObjRcCarpet_c` / `daPropeller_Heyho_c` on `trackedConfigArm9Sha256`. That is tree-owned, not this branch: those two are among the 8 intact TUs the strict control demotes to ROM gap bytes, and the fingerprint is a content hash of the whole tracked `config/arm9` tree, so *any* edit under it — including this rename's `ov002/symbols.txt` — invalidates the stored partition baselines. `rombuild` is unaffected and reports 106/106.

## Follow-ups, deliberately not in this PR

- `symbols/actor_renames.tsv` — only the 2 rows that are genuinely this class's addresses (`0x020ec534`, `0x0210ac98`) were touched. Other mislabelled rows in that ledger need their own evidence and their own PR.
- `SoundObjectCallbackOwner`-style shadow types elsewhere in ov002 are untouched.


andrewboudreau:
## Review sign-off

Reviewed adversarially (no PR in this repo can be formally APPROVED — shared account — so this comment is the sign-off). Every cartridge claim in the PR was **re-derived independently from the images under `extracted/`**, not taken from the commit message.

**Confirmed from the ROM:**

- **The vtable diff.** 31 words each from `_ZTV11daWarpkun_c` (ov002 `0x0210acbc`) and `_ZTV8dActor_c` (arm9 `0x0208e3a4`), symbol-resolved. Differing slots are exactly **{0, 3, 6, 9, 12, 16, 17}** — `InitResources`, `CleanupResources`, `Behavior`, `Render`, `OnPendingDestroy`, D1 `0x020ec388`, D0 `0x020ec3b8`. All 24 other slots hold the base's own word byte-for-byte. Header words confirm the RTTI: `{0x0, 0x0210ac7c}` and `{0x0, 0x0208e390}`.
- **The `0x108` layout.** `daWarpkun_c_Spawn` @ `0x020ec534` disassembles to `mov r0,#0x108 / bl _ZN7fBase_cnwEj / bl _ZN8dActor_cC2Ev / str vptr / add r0,r4,#0xd4 / bl _ZN7dCcAc_cC1Ev`. D1 @ `0x020ec388` matches the header's quoted disassembly instruction for instruction, `bl 0x20149a4` = `_ZN7dCcAc_cD1Ev`, `bl 0x20112c8` = `_ZN8dActor_cD2Ev`. `0xd0` + 4 pad + `0x34` = `0x108`.
- **Ledger hygiene.** Every address changed in `ov002/symbols.txt` is inside `.text 0x020ec388..0x020ec56c` or is one of the two data rows `0x0210ac98` / `0x0210acbc`. `actor_renames.tsv` changed **exactly 2** rows, both in span. The neighbouring `Tree_*` and `YoshiEgg_*` rows are untouched context — no repeat of the `class_rename.py` cross-class contamination.
- **No substring contamination.** `WarpPipe_Spawn.c`, `WarpPlayer.c`, `IsPlayerWarping.c`, `Player::CanWarp`, `include/Player.h` all have a zero-line diff. Tree-wide grep for `_ZN4Warp`, `_ZTV4Warp`, `Warp.h` returns nothing.
- **All 9 `compiler_only_output` rows are genuinely data** — every one an `_ZTI`/`_ZTS`/`_ZTV` record, none a function, so the past *function-given-`deadstrip-data`* defect is not present here. No stale `unlicensed_output_observed` block survives.

**One real finding, now fixed in `fe5ddc7e9`.** The manifest's `notes[1]` was left contradicting its own file on three points — including the identical self-comparing `class_rename.py` artifact that #2171 had to fix in `63c228ee4` ("ROM name daWarpkun_c vs decomp daWarpkun_c (4 chars)", a name compared to itself with the old name's character count). It also still claimed the RTTI records were "left unlicensed, deferred to Phase F" and that "no header fix needed", both false after this PR. A superseding `notes[2]` now records all three.

**Not this PR's, for the record:**

- `premerge_check.py --repo . --fetch 2173`: every gate `ok` on the merge tree; `source-coverage` errors identically on **both** sides with `git ls-tree ... fatal: not a tree object` — pre-existing and tree-owned.
- The `ov002` baseline linkcheck control refuses `daObjRcCarpet_c` / `daPropeller_Heyho_c` on `trackedConfigArm9Sha256`. Those two are among the 8 intact TUs the strict control demotes to ROM gap bytes, and that fingerprint hashes the whole tracked `config/arm9` tree — so *any* edit under it invalidates the stored partition baselines. `rombuild` is unaffected: **106/106 exact, 100.000000%, mismatching 0**.
- `substitutedObjectPaths` carries `.c` where the tree had `.cpp`; the merge-base manifest already did, and the rename propagated it faithfully. `symbols/profile_reconstruction_registry` still names the deleted `src/daWarpkun_c_Spawn.c` as `current_factory_file` — the same as merged `daChRoom_c` row 348. Both are tree-wide conditions wanting their own PR.

Good to merge once `PR validation` is green.

PR #2174, opened 2026-09-02
Rename ChainChompFence to daObjWanwanShutter_c and promote it to the single TU src/actors/d_a_obj_wanwan_shutter.cpp

andrewboudreau:
Renames `ChainChompFence` to the name the cartridge itself spells, then promotes it from eight per-function sources to one genuine translation unit. Two commits, each independently proven at 106/106.

## The name is the cartridge's, not coined

```
_ZTV15ChainChompFence   ov014 0x021148b0   (address point)
header pair             0x021148a8 = 0x00000000   offset-to-top
                        0x021148ac = 0x02114868   _ZTI
_ZTI at ov014 0x02114868 = { 0x0209a764, 0x02114874, 0x021089ec }
      word[0] -> _ZTVN3abi20__si_class_type_infoE + 8   single inheritance
      word[1] -> _ZTS at 0x02114874, which reads "20daObjWanwanShutter_c"
      word[2] -> _ZTI10dBgActor_c  (ov002 0x021089ec)
```

Every renamed row was audited against this class's own span before the edit. Eleven `symbols/actor_renames.tsv` rows mention the class; **seven** of them changed (`git show 1d8757792 --numstat` reports `7 7`), and all seven fall inside `.text 0x02112e0c..0x021130ac` or its data (`0x02114868` `_ZTI`, `0x02114874` `_ZTS`, `0x0211488c` profile, `0x021148b0` `_ZTV`). No unrelated ledger row was touched — worth stating explicitly, because `class_rename.py` has previously rewritten rows belonging to other classes.

## The leaf proof needed care here

The inline-destructor lever is only safe if nothing derives from the class. Scanning every file under `extracted/` for the LE word `0x02114868` returns **four** raw hits, not the usual two:

| image | offset | resolves to | |
|---|---|---|---|
| `ov014.bin` / `overlay_0014.bin` | `0x370c` | `0x021148ac` = `_ZTV − 4` | its own vtable header |
| `ov015.bin` / `overlay_0015.bin` | `0x3738` | `0x021148d8` | **not a descendant** |

ov015 loads at the same base as ov014 (both `0x021111a0`), so the two overlays are mutually exclusive alternatives and can never be resident together — a class in one cannot derive from a class in the other. At `0x02114868` ov015 holds `_ZTI20daObjBk_Fall_Block_c`, and the triple at `0x021148d0` is `{0, 0, 0x02114868}` — a vtable header pair pointing at ov015's *own* typeinfo, not an `_ZTI` pointing at ours. A true descendant's hit would decode as an `_ZTI` triple whose `word[0]` is `0x0209a764`.

Leaf confirmed, so `virtual ~daObjWanwanShutter_c() {}` is safe.

## The empty braces are not an empty function

D1 is `0x44` bytes because it inlines `dBgActor_c`'s own inline destructor, which destroys the `Model` at `0xd4` and the `dBgW_KcMbg` at `0x124`. Both stay declared as real members; writing either as padding would emit the smaller shape and stop matching.

## Derived from dBgActor_c, and the vague-linkage set shows it

Twelve records are licensed in `compiler_only_output`. The class's **own** RTTI is emitted because it has no key function — the first declared virtual is the now-inline destructor. The base chain contributes typeinfo and name records only; the bases' **vtables are absent**, because `dBgActor_c`'s key function is `Kill()` and that lives in another TU.

```
own:   _ZTV20daObjWanwanShutter_c ov014 0x021148b0
       _ZTI20daObjWanwanShutter_c ov014 0x02114868
       _ZTS20daObjWanwanShutter_c ov014 0x02114874
bases: _ZTI10dBgActor_c ov002 0x021089ec   _ZTS10dBgActor_c ov002 0x02108a04
       _ZTI8dActor_c    arm9  0x0208e390   _ZTS8dActor_c    arm9  0x0208e384
       _ZTI7dBase_c     arm9  0x02086e78   _ZTS7dBase_c     arm9  0x02086e54
       _ZTI7fBase_c     arm9  0x02086d70   _ZTS7fBase_c     arm9  0x02086e60
dup:   _ZN7Vector3D1Ev  arm9  0x020072c0
```

`_ZN7Vector3D1Ev` is `deadstrip-duplicate`, not `deadstrip-data` — an enrolled source already owns that copy. `func_ov014_02112ea8` declares a `Vector3[2]` local and passes `v[1]` to `dActor_c::PoofDustAt`, which is what re-emits it. Same shape `ov002/daChRoom_c` already licenses.

**One deliberate divergence to flag for review.** The five merged siblings all give the reason "because this TU defines the class's key function". That sentence is false for every one of them: an inline destructor means the class has *no* key function. The rows here say what actually happens instead, rather than inheriting the wording.

## The vtable

32 slots, diffed word by word against `_ZTV10dBgActor_c` (ov002 `0x0210ae38`). Six differ:

| slot | | |
|---|---|---|
| 0 | `InitResources` | `0x02112ffc` |
| 3 | `CleanupResources` | `0x02112f3c` |
| 6 | `Behavior` | `0x02112fc0` |
| 9 | `Render` | `0x02112f80` |
| 16 | D1 | `0x02112e0c` |
| 17 | D0 | `0x02112e50` |

Slot 31 is `dBgActor_c::Kill` at `0x020ee55c` in **both** tables — inherited, not overridden.

## Proof

| check | result |
|---|---|
| `tubuild.py verify` | 8/8 MATCH, objisolate clean, reloc-destinations clean, emission order as expected → TEXT-VERIFIED, no refusals; compiler-only exact deadstrip |
| `rombuild.py -j16` | **106/106 exact, 100.000000%, mismatching 0**; source-owned data claims 8, reproducing 8, mismatching 0 |
| linkcheck baseline | the 9 dsd symbol errors are the tree's pre-existing set (`overlay_100/102`, `data_020ad524/560`, four ITCM, `_deq`); no new one, control ROM stock-identical |
| `romdata_check` | exit 0; the 5 differs are pre-existing and none is this class's. Per-symbol: `_ZTV20daObjWanwanShutter_c` **VERIFIED at 128 bytes = all 32 slots** from the address point — neither cut short nor overrun |
| `layout_check` | clean |
| `check_header_offsets` | 0 mismatched, struct spans `0x31f`, `sizeof` asserted `0x320` |
| `tiers_ratchet` | 2592 → 2592 (+0 / −0) — consolidation, not a loss |

Enrolled sources 10652 → 10645. ROM data from source: 595 → 597 verified, the two extra being the vtable and typeinfo the rename made nameable.

## Also in here

- Dropped the header's flat-C `#else` branch. Its stated reason was that the D0 file is a C translation unit reading these fields; no `.c` file includes this header any more, and all five merged predecessors are flat C++ headers.
- Repointed `config_tu/arm9/overlays/ov014/delinks.txt` at the renamed `src_tu` file, which `class_rename.py` does not cover.
- The two historical-alias rows (`ChainChompFence_Spawn` at `0x0211307c`, `ChainChompFence_SpawnInfo` at `0x0211488c`) are deliberately left as spelled: they are the "old" column of rows whose "new" column already carries the reconstructed `daObjWanwanShutter_c_classInit` and `g_profile_WANWAN_SHUTTER`.

---

## Third commit: adversarial review findings, applied

A review of the first two commits found four claims that were wrong or unsafe. Each was verified against the tree before being changed, and `rombuild.py -j16` was re-run afterwards — **106/106 exact, 100.000000%, mismatching 0**, with mwccarm recompiling the TU, so the header change is byte-proven rather than assumed.

1. **The four overrides were declared non-virtual, out of vtable order, unannotated.** An override that stops being declared `virtual` silently stops overriding *and still links*, so this was a live hazard, not a style nit. Now `virtual s32` in ROM vtable order with slot/address comments, matching the five predecessors. `s32` is `typedef signed int` (`include/types.h:11`), identical to the `int` the definitions already use.
2. **"the first entry in this campaign" to derive from `dBgActor_c` was false** — `daObjCannonShutter_c`, `daObjHatenaSwitch_c`, `daObjPushblock_c`, `daObjC1_Trap_c`, `daObjC0_Switch_c`, `daObjFm_Battan_c` and `daObjCtMecha05_c` all do. Replaced with the sibling list; heading above corrected too.
3. **The header's leaf proof said "the only word in the cartridge"** while the scan returns four raw hits. The ov015 dismissal that this PR body already carries is now in the header comment, so a reader repeating the scan does not misread the extra pair as a descendant and back out the inline destructor.
4. **Manifest note 2's "Spawn stores `&_ZTV…[2]`"** names a symbol that does not exist; the store is emitted by `daObjWanwanShutter_c_classInit` and by the destructor. Corrected by appending a superseding note rather than editing history (precedent: `fe5ddc7e9`, `63c228ee4`).

Also drops the unused `#include "dBgW_KcMbg.h"` — the type appears in this header only inside comments, `dBgActor_c.h` already pulls it in, and it sat *ahead* of `dBgActor_c.h`, whose own comment requires `common.h` to precede `Model.h`.

**Two findings deliberately not acted on.** The stale `current_factory_file` rows in `symbols/profile_reconstruction_registry.json` point at a `src/ChainChompFence_Spawn.c` that does not exist — but neither does the merged `daWarpkun_c` row's `src/daWarpkun_c_Spawn.c`, so this is tree-wide and pre-existing, not this PR's. And the dangling `config_tu/arm9/overlays/ov014/delinks.txt` line matches existing precedent (`src_tu/actors/ShutterBob.cpp` on line 6 is likewise absent) — convention, not damage.


andrewboudreau:
Signing off here rather than approving — a PR in this repo cannot be APPROVED, since the account that opened it is the account that would review it.

**What is proven, and by what.**

- `rombuild.py -j16` at the third commit: **106/106 exact, 100.000000%, mismatching 0**, and mwccarm recompiled the TU on that run rather than reusing cache — so the `virtual s32` redeclarations are byte-proven, not assumed. Re-run after the review fixes, not just before them.
- `premerge_check.py --repo . --fetch 2174` at head `674902e68`, merge tree `a597654486d7`: **nothing goes green -> red**. `source-coverage` errors identically on base and merge (`git ls-tree ... -- config: fatal: not a tree object`) and is pre-existing tree damage, not this PR's.
- All 8 repository checks and the external `PR validation` are SUCCESS.
- The class name is read out of the cartridge, not coined: `_ZTS` at ov014 `0x02114874` reads `20daObjWanwanShutter_c`, and the `_ZTI` triple's base word is ov002 `0x021089ec` = `_ZTI10dBgActor_c`.
- Leaf proof for the inline-destructor lever: the four raw hits on `0x02114868` are two images x two paths, and the ov015 pair is a same-base collision (ov015 also loads at `0x021111a0`), not a descendant.

**What I deliberately did not do.** The stale `current_factory_file` rows in `symbols/profile_reconstruction_registry.json` and the dangling `config_tu/.../ov014/delinks.txt` line both match existing tree-wide precedent, so neither is this PR's to fix; both are stated in the body so they are not silently inherited.

**One thing a later reader should know.** Review of my own first two commits found four wrong or unsafe claims, including a non-virtual override declaration that would have silently stopped overriding while still linking. Those are the third commit. The review was worth more than the original write.

PR #2175, opened 2026-09-02
Correct 360 rename-ledger rows that name the next class along

andrewboudreau:
`symbols/actor_renames.tsv` column 4 is a present-tense claim about what the
symbol at an address is called, and `tools/cpp_index.py:86` and
`tools/cpp_rename.py:51` read it back as a live address -> symbol map. 360 of
its 1,635 mangled/vtable rows named a symbol that lives at a different address
-- 22%, across 24 modules, worst in ov002 (81) and ov015 (28).

The shift is uniform: each class's `_Spawn`/`_SpawnInfo` rows are right while
its D1/D0/method/vtable rows carry the NEXT class's addresses.

  ov002 0x020b07f8   ledger _ZN10daCamTag_cD1Ev   symbols.txt _ZN10daChRoom_cD1Ev
  ov002 0x021086b4   ledger _ZTV11VirtualDoor     symbols.txt _ZTV4Exit
  ov002 0x020f15fc   ledger _ZN14BlueCoinSwitchD1Ev  symbols.txt _ZN14EnemySwitchTagD1Ev

WHERE IT CAME FROM. `tools/actor_names.py` writes this file (line 282, on every
run including the dry run) and never reads it back. Its `parse_spawnfunc` used
to scan past the factory's own body into the next class's D1 and take that
class's vtable -- "shifted every vtable and method name one class late", as its
own docstring now says. That was fixed on main in #1418-#1423; the checked-in
ledger is still the pre-fix output because it was never regenerated.

WHY NOT JUST REGENERATE. Because it no longer reproduces the file. `propose()`
refuses any address that already carries a real name, and by now nearly all of
them do: a fresh run in this worktree emits 1 row where the tracked file has
2,573. The ledger has stopped being a derived artifact and is maintained in
place, so the correction has to be made here.

WHY THE NAME AND NOT THE ADDRESS. Every one of the 360 rows embeds its own
address in column 3 (`func_ov002_020b07f8` on 0x020b07f8) -- 0 exceptions, and 0
across all 1,635 -- so address and placeholder are machine-derived and agree;
only the assigned name is wrong. Moving addresses instead would be
unrecoverable: `AmbientSoundEffects` is in no symbols.txt in the tree, so there
is no address to move it to. And symbols.txt is the side to trust: it is what
the ROM build consumes and what `dsd check symbols` validates. It is not
read-only -- class_rename.py:209, import_symbols.py:166 and actor_names.py:313
(--apply) all rewrite it -- but each of those writes exists to make the build
emit that symbol, so symbols.txt and the linked ROM agree by construction while
the ledger does not.

TWO JUDGEMENT CALLS.

  * 26 addresses carry both a legacy coined vtable name and the cartridge RTTI
    one (`_ZTV11PowerFlower` / `_ZTV18daObjPowerUpItem_c` at ov002 0x02109800).
    All 26 are exactly that shape -- same address, same vtable, same class, none
    with a third name. This takes the legacy spelling, matching the method rows
    around it; adopting the cartridge name is a separate claim for a separate PR.

  * 49 corrected vtable rows also carried an `alloc=0xNNN` instance size, and
    those figures tracked the NAME, not the address: of the 32 checkable against
    an `include/*.h` size assert for the row's PRE-correction class, 31 match
    exactly. Left in place they would re-attach to a class they are not evidence
    about, so they are withdrawn to `alloc=?`, already this file's marker for
    unknown on 28 rows.

The retired claim is kept in column 5 as `(was <symbol>)`. Five names --
AmbientSoundEffects, Seaweed, VirtualDoor, daObjC1_Trap_c, daObjPushblock_c --
have no other row in this file, and AmbientSoundEffects has no other
authoritative mention anywhere in the tree, so overwriting column 4 silently was
the one way this correction could have lost something.

`tools/check_rename_ledger.py` is the gate, `tools/test_check_rename_ledger.py`
its tests, and `.github/workflows/rename-ledger.yml` runs both on any PR
touching the ledger or a symbols.txt. `class_rename.py` rewrites ledger rows by
NAME, so this drift is reintroducible by an ordinary rename; nothing was
watching because the ledger is not compiled into the ROM and a 106/106 build
says nothing about it.

KNOWN LIMITS, deliberately not fixed here.

  * The check only covers `_ZN`/`_ZTV` rows, because only those have a
    symbols.txt counterpart. 938 coined rows are unchecked and at least a few
    carry the same shift (ov009 `DockPole_Spawn`/`DockPole_SpawnInfo` sit on
    MetalNet's addresses). The check prints the unchecked count on every run.
  * 8 vtable rows this PR did NOT correct still carry an `alloc=` figure their
    class's own size assert contradicts, `daObjCannonShutter_c`'s `alloc=0x1`
    most obviously. Their names are right, so this pass leaves them; nothing
    gates `alloc=` at all.

Proof: `rombuild.py -j16` 106/106 exact, 100.000000%, mismatching 0.
`check_rename_ledger.py` clean and idempotent. 14 new tests pass.
`test_cpp_rename_verify` 16 passed; both consumers still import and load.
`check_python_names` PASS, 0 new advisories.

## What an adversarial review changed

The first push of this branch made three structural claims that were wrong, and
the review caught all three. They are corrected above and in the tool's own
docstring:

| Claimed then | True |
| --- | --- |
| Three tools *read* the ledger, incl. `actor_names.py` | `actor_names.py` **writes** it and never reads it; the readers are `cpp_index.py` and `cpp_rename.py` |
| "No production tool generates `config/**/symbols.txt`" | `class_rename.py:209`, `import_symbols.py:166`, `actor_names.py:313` all rewrite it |
| Root cause unstated | `parse_spawnfunc` over-scan, already fixed in #1418-#1423 on a file never regenerated since |

The review's proposed remedy -- regenerate rather than hand-correct -- I tested
and it does not work: `python tools/actor_names.py` in this worktree emits
`renames=1`, because `propose()` skips every address that now carries a real
name. Regenerating would truncate the ledger from 2,573 rows to 1, so the
in-place correction stands. That is now stated in the tool docstring rather than
left for the next person to rediscover.

Also from the review, and fixed in this push:

- retired names are kept as `(was <symbol>)` instead of being overwritten -- five
  of them had no other row in the file
- `is_cartridge_spelling` used `da[A-Z_]`, which misread `_ZTV7da1up_c` as legacy
  and made `pick()` refuse ov002 `0x021083c8`
- `--fix` exited 0 even over rows it had refused
- a module with no `symbols.txt` was skipped in silence
- a stray LF row among CRLF rows was merged into the previous row and went
  unreported
- the tool is wired to a workflow and has tests; the first push had neither

Held up under the review unchanged: every numeric claim (1,635 / 360 / 22% / 49
/ 26 / the per-module table), the repair direction (re-proved two independent
ways), the `alloc=` reasoning (31 of 32 checkable figures match the
pre-correction class), and all 26 alias picks being genuine same-class pairs.

### DIFF
--- a/.github/workflows/rename-ledger.yml
+++ b/.github/workflows/rename-ledger.yml
@@ -1,0 +1,69 @@
+# Keeps symbols/actor_renames.tsv honest about what lives at each address.
+#
+# The ledger's fourth column is a present-tense claim, and `tools/cpp_index.py`
+# and `tools/cpp_rename.py` read it back as a live address -> symbol map. When
+# 360 of its 1,635 mangled/vtable rows named the NEXT class along -- the residue
+# of the `parse_spawnfunc` over-scan fixed in #1418-#1423, on a file that was
+# never regenerated afterwards and can no longer be regenerated at all -- every
+# one of those rows was a false lookup result sitting in a tracked file with
+# nothing watching it.
+#
+# Nothing was watching it because the ledger is not compiled into the ROM. A
+# byte-perfect 106/106 build says nothing about this file, which is exactly the
+# shape of gate this repo keeps having to add after the fact: the failure branch
+# a green run never reaches. `tools/class_rename.py` rewrites rows by NAME, so
+# the drift is reintroducible by an ordinary rename on any future PR.
+#
+# Costs nothing: two text files and no compiler, so it runs here rather than on
+# the private build box pr-validate.yml relays to. About a second over 2,574
+# ledger rows and 25 symbols.txt files.
+#
+# NARROW ON PURPOSE. Only `_ZN...` and `_ZTV...` rows are checked, because only
+# those have a symbols.txt counterpart to check against; the coined rows
+# (Foo_Spawn, g_profile_BAR) carry the same shift in at least a few places and
+# this gate will not catch them. The check prints its own unchecked count on
+# every run so the green cannot be read as covering the whole file.
+name: rename ledger
+
+on:
+  pull_request:
+    paths:
+      - "symbols/actor_renames.tsv"
+      - "config/**/symbols.txt"
+      - "tools/check_rename_ledger.py"
+      - "tools/test_check_rename_ledger.py"
+      - ".github/workflows/rename-ledger.yml"
+  push:
+    branches: [main]
+    paths:
+      - "symbols/actor_renames.tsv"
+      - "config/**/symbols.txt"
+      - "tools/check_rename_ledger.py"
+      - "tools/test_check_rename_ledger.py"
+      - ".github/workflows/rename-ledger.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: rename-ledger-${{ github.event.pull_request.head.sha || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  ledger:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        # No history needed: the check reads two tracked text files.
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Ledger rows agree with symbols.txt
+        run: python tools/check_rename_ledger.py --repo .
+      - name: The gate's own tests
+        # Every case is a defect found reviewing the first draft of the checker
+        # -- the digit in da1up_c, the mixed line endings, --fix exiting 0 over
+        # rows it refused, a silently missing symbols.txt, and the retired name
+        # that had nowhere else to live. stdlib unittest, no dependencies.
+        run: python -m unittest tools.test_check_rename_ledger -v
--- a/symbols/actor_renames.tsv
+++ b/symbols/actor_renames.tsv
@@ -1,6 +1,6 @@
 module	addr	old	new	why
 arm9	0x02086d78	data_02086d78	Camera_SpawnInfo	spawninfo
-arm9	0x02092680	data_02092680	_ZTV5Stage	vtable alloc=?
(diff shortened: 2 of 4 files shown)


andrewboudreau:
Sign-off (no PR in this repo can be APPROVED -- shared account, so this comment is the record).

Verified at head `201431f41`, worktree `C:\tmp\sm64ds-sm64ds-soundobj`:

- `rombuild.py -j16` -> **106/106 exact, 100.000000% of compared bytes, mismatching 0**
- `check_rename_ledger.py --repo .` clean; `--fix` idempotent (second run rewrites nothing)
- `python -m unittest tools.test_check_rename_ledger` -> 14 tests, OK
- `pytest tools/test_cpp_rename_verify.py` -> 16 passed; `cpp_index` and `actor_names` still import and load the ledger
- `check_python_names.py` PASS, 0 new advisories
- `premerge_check.py --repo . --fetch 2175` on merge tree `b06bbdd19011`: converted-ratchet, dead-references, duplicate-sources, header-offsets, langmode-ratchet, layout-check, src-tu-refs all pass on both sides; source-coverage errors identically on base and merge (`git ls-tree ... -- config: fatal: not a tree object`), pre-existing and not this PR's. Nothing goes green -> red.
- all 8 GitHub checks SUCCESS, including the new `ledger` job and the external `PR validation`

This branch was rewritten after an adversarial review that refuted three of its structural claims; the diff of what changed and what held up is in the PR body. The review's proposed alternative -- regenerate instead of hand-correcting -- was tested and rejected on evidence: `python tools/actor_names.py` emits `renames=1` against a tracked file of 2,573 rows.

Merging with `--merge`, no `--delete-branch`.

PR #2176, opened 2026-09-02
Promote ov002/daObjBC_Switch_c to a single-file TU, renamed from BlueCoinSwitch

andrewboudreau:
Renames the coined `BlueCoinSwitch` to the name the cartridge's own RTTI spells, then promotes it to a genuine single-file translation unit.

## The rename (145719519)

ov002 `0x0210b228` is a `__si_class_type_info` whose `_ZTS` at `0x0210b234` reads exactly `16daObjBC_Switch_c`, and whose one base at subobject offset 0 is `0x021089ec` = `_ZTI10dBgActor_c`. `BlueCoinSwitch` matched nothing at any address. Proved at 106/106 on its own.

Six ledger rows were deliberately **not** renamed: `symbols/actor_renames.tsv` rows 218/219/222/224/227/353 are `EnemySwitchTag`'s addresses, and their column-5 `(was _ZN14BlueCoinSwitch...)` strings are #2175's record of a retired false claim. `class_rename.py` substitutes by NAME and rewrote them; renaming a historical claim would falsify it, so they were reverted and `check_rename_ledger.py` re-run.

## The promotion (b8ae6fe0f)

One file, `src/actors/d_a_obj_bc_switch.cpp`, holding the whole contiguous linker run `.text 0x020f11b0..0x020f15fc` — nine functions, nothing else — replacing nine per-function objects in the production build.

**One source change was needed, and it is the interesting part:**

```
include/daObjBC_Switch_c.h
  virtual ~daObjBC_Switch_c();  ->  virtual ~daObjBC_Switch_c() {}
```

Out of line, this TU passed everything that looks at one object — 9/9 byte MATCH, `objisolate` clean, relocation destinations clean — and still failed `tubuild.py linkcheck`. mwcc emitted D0 before D1 plus a homeless D2; the ROM has D1 at `0x020f11b0` then D0 at `0x020f11f4` and no D2. Production isolation lays `.text` into the spanning delink in **emission order**, so the order decides whether the range links, and `objisolate` never sees it.

Safe only because the class is a leaf: a byte scan of everything under `extracted/` for the LE word `0x0210b228` finds exactly one logical hit, at file offset `0x5dc08` in both `ov002.bin` and `overlay_0002.bin` — this class's own `_ZTV - 4`. No `_ZTI` in the ROM names it as a base. The header now carries that argument as a do-not-move-it warning, matching `include/dBgActor_c.h`'s.

Also reconciled four shadow declarations away, re-verifying after each: the two bss handles are typed as the real `SharedFilePtr` and released through `SharedFilePtr::Release()`; `Model::LoadFile`, `dBgW_Kc::LoadFile` and `Model::Render` are real method calls now. `include/SharedFilePtr.h` explains why that matters — the mangled-name extern works by accident and double-mangles in a C++ TU. `Render`'s local `struct Base` vtable shadow is gone, and so is the header's flat-C `#else` branch, which existed only for the three `.c` members now folded in.

## Proof

- `tools/rombuild.py -j16` — **106/106 exact, 100.000000%, mismatching 0**, stock-identical ROM
- `tubuild verify` — 9/9 MATCH, objisolate clean, all nine in ROM-ascending emission order
- `linkcheck --baseline --module ov002` PASS; `tiers_ratchet` 0 backslid
- `check_header_offsets`, `check_rename_ledger`, `check_src_tu`, `check_src_tu_compiles` (98/98), `check_dead_references`, `layout_check` — clean

The branch fast-forwards `origin/main` (`666d80757`), so the tree gated above **is** the merge tree.

## What review caught

The prose had been written against the out-of-line shape and never re-read after the flip. The masthead asserted the destructor *stayed* out of line, and all 11 manifest `reason` strings said the vtable is emitted here "because this TU defines the key function" — exactly backwards. With the only virtual inline the class has **no** key function, which is precisely why the vague-linkage data lands here. Also corrected: the size arithmetic (it closes through `dBgActor_c`'s tail padding — `pad_31e` sits at `0x31e`, inside the base's rounding slack, not after `0x320`), the arm9 record count (six, not five), a stale langmode euphemism the tool no longer requires, and the `partial_isolation` block, now marked historical because it verified the shape that fails linkcheck.

## Left deliberately

`include/daObjBC_Switch_c.h` pulls `Model.h` before `dBgActor_c.h` — the order `dBgActor_c.h` flags as hazardous for `Matrix4x3`. Harmless here (this TU *calls* `UpdateClsnPosAndRot`, it does not compile it), and the file now says so instead of shipping the generator's "watch for new compile errors" disclaimer as an open question.

The three rows this leaves in `notes/data/c-cpp-classification.tsv` and the nine in `notes/data/tu-merge-candidates.json` are untouched on purpose: both are dated 2026-08-25 snapshots that already carry rows for five previously-promoted classes.


andrewboudreau:
Signing off on this myself (no PR in this repo can be APPROVED -- shared account).

I ran the promotion and I ran the review, so the sign-off is worth exactly what the artifacts are. What I actually checked, and what I did not:

**Checked mechanically, green:** `rombuild.py -j16` at 106/106 exact / mismatching 0 with a stock-identical ROM, re-run after every prose and manifest edit, not just once at the start. `tubuild verify` 9/9 MATCH with objisolate clean and all nine functions in ROM-ascending emission order. `linkcheck --baseline --module ov002` PASS. `tiers_ratchet` 0 backslid. `check_header_offsets` 7 fields 0 mismatched. `check_rename_ledger` all 1,635 mangled/vtable rows agree with symbols.txt. `check_src_tu_compiles` 98/98. `check_dead_references` clean after the promotion was carried into `notes/bgobject-provenance.md`.

**Checked by reading the image, independently re-derived during review:** the `_ZTS` string at `0x0210b234`, the `__si_class_type_info` vptr and its single base word, the leaf scan (the LE word `0x0210b228` appears at file offset `0x5dc08` in exactly two files, which are the same overlay twice), both vtables dumped at 32 words each and differing in exactly slots 0/3/6/9/16/17, and all 11 `compiler_only_output` canonical addresses against `config/arm9/symbols.txt` and `ov002/symbols.txt`.

**The one thing this PR got wrong and fixed:** the prose was written before the destructor was flipped inline and never re-read afterwards. The masthead said the destructor stayed out of line -- the opposite of what ships -- and 11 manifest `reason` strings gave a key-function rationale that is backwards for an inline-dtor leaf. A byte-perfect build says nothing about either. Worth noting as a shape: the failure was entirely in the layer no gate covers.

**Not covered by any of the above:** `romdata_check` still reports 5 DIFFERS and `dsd check symbols --fail` still reports 9 errors. Both reproduce identically on `666d80757` with none of this branch applied -- they are the tree's, not this PR's.

The branch fast-forwards `origin/main`, so the merge tree is the branch tree and the numbers above are the merge-tree numbers.

PR #2177, opened 2026-09-02
Rename Exit to daChScene_c and promote it to a genuine TU

andrewboudreau:
Renames the coined class `Exit` to the name the cartridge itself carries, `daChScene_c`, then promotes its nine per-function sources into one genuine translation unit at `src/actors/d_a_ch_scene.cpp`.

## The name is read from the ROM, not coined

ov002 `_ZTV` at `0x021086b4` is preceded by offset-to-top `0` and `_ZTI 0x02108674`, whose three words are `0x0209a764` (`__si_class_type_info` vtable + 8), `0x02108680` — its `_ZTS`, reading `11daChScene_c` — and `0x0208e390` (`_ZTI8dActor_c`, arm9). `config/arm9/overlays/ov002/symbols.txt` already spelled the `_ZTI`/`_ZTS` pair `daChScene_c` while naming the vtable four bytes later `_ZTV4Exit`, so the rename closes an inconsistency that was already in the tree.

All ten `symbols/actor_renames.tsv` rows the rename touched were checked against ov002's `symbols.txt` **before** the rename, per the systematic one-class shift corrected tree-wide by #2175. Every one is genuinely this class's.

Out of line, mwcc emits `D2, D0, D1`. The ROM has `D1` at `0x020b09b0`, then `D0` at `0x020b09d4`, and no `D2` at all. Production isolation lays `.text` into the spanning delink in **emission order**, so emission order decides whether the address range links — and the out-of-line spelling is refused even though all nine functions byte-match and `objisolate` calls the object clean. `objisolate` checks one object's relocations; it never sees the order the linker will see.

The manifest said so in advance: `functions_occur_in_expected_order` was `PARTIAL -- ordinal pair(s) not in ROM order: [(0, 1)]`, the two destructor variants. That criterion now reads PASS as shipped.

Moving a destructor inline changes the codegen of every class deriving from it, so it is only safe on a leaf. `daChScene_c`'s `_ZTI` is ov002 `0x02108674`, and a scan of all 977 `.bin` files under `extracted/` finds exactly one word anywhere pointing at it: `0x021086b0`, the `_ZTI` slot of this class's own vtable. Nobody names it as a base.

## Naming the class puts its vague-linkage data in this object

The inline destructor does not leave the class key-function-less — the key function is the first **declared** non-inline virtual, `InitResources`, and it is defined here. So mwcc emits `daChScene_c`'s own `_ZTV`/`_ZTI`/`_ZTS` plus the six inherited base-chain records: 3 + 6 = 9, all licensed as `deadstrip-data` in `compiler_only_output`. The stale `_ZN11daChScene_cD2Ev` deadstrip row went with the `D2` itself.

## Proof

- `tools/rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes, mismatching 0**, 11,088 reproducing functions, ROM byte-identical to the tree's stock build.
- Ratchet: 2592 -> 2592, **+0/-0**.
- `romdata_check` over the new file: 9 data symbols — VERIFIED 5 (168 bytes), PARTIAL 4, DIFFERS 0, UNNAMED 0.
- `check_dead_references`, `layout_check`, `check_src_tu` (41 TUs), `check_src_tu_compiles` 98/98, `check_header_offsets` (struct spans 0x104), `eligible.py` 11095/11211, `port_refcheck` 405, `check_rename_ledger` over 1,635 rows: all green.
- `premerge_check.py` against the merge tree: nothing goes green -> red. `source-coverage` is red on both sides (`fatal: not a tree object`), pre-existing and not this PR's to own.

## Third commit: prose the rename falsified

A review pass found no defect in the shipped code, and four sentences that had stopped being true. Most instructive one: `class_rename.py` substituted the new name into **both halves** of a sentence whose subject was the Exit-vs-daChScene_c mismatch, leaving the manifest claiming `daChScene_c` was "a different string length than the decomp's `daChScene_c` (4 chars)". A blind rename can turn a correct note into a nonsensical one without touching a line of code. Also corrected: a `.c` extension on a file that was `.cpp`, an includes comment that described an order the file does not use, and `Camera::LookAtExit` listed among shadow externs that have a real declaration to swap to — it has none. One dead `extern void* data_020a0eac;` removed, and the ROM re-proved at 106/106 afterwards.

## Known leftovers, deliberately not in scope

`dActor_c::FindWithActorID`, `Camera::LookAtExit` and `Player::SetNoControlState` remain mangled-name shadow externs. Swapping one for a real method call changes what mwcc emits at the call site, so each needs its own byte verification.


andrewboudreau:
Sign-off (no PR in this repo can be APPROVED — shared account, so this comment is the review record).

Reviewed by a separate agent against the tree, read-only. It independently re-derived the RTTI read from `extracted/overlays/overlay_0002.bin` (`[0x021086b0] = 0x02108674`, `_ZTI` = `{0x0209a764, 0x02108680, 0x0208e390}`, `_ZTS` = `"11daChScene_c"`), re-ran the leaf scan and found the same single hit, confirmed all nine `// @symbol` markers resolve in ov002 `symbols.txt`, confirmed the functions are in strictly reverse ROM order (`0x020b0f24` down to `0x020b09b0`, and `0x020b0f24 + 0x30 = 0x020b0f54` closes the delink run exactly against `_ZN4CoinD1Ev`), confirmed the destructor has no surviving out-of-line definition, and confirmed every touched ledger row is genuinely this class's.

It found no defect in the shipped code. Its four blocking/should-fix findings were all prose, and all are fixed in `4a356f681`.

Still open, tracked but out of scope here:
- `symbols/profile_reconstruction_registry.{json,tsv}` still point `current_factory_file` at the deleted `src/daChScene_c_Spawn.c`. Pre-existing pattern — `daChRoom_c` and `daSetSE_c` have the same stale rows from their own promotions.
- `notes/data/c-cpp-classification.tsv` and `notes/data/tu-merge-candidates.json` carry the wrong `.c` extension for the D0 source; that is where the one in the masthead came from.
- `include/daChScene_c.h`'s `#else` C branch is now unreachable (this header has exactly one includer tree-wide and it is `//cpp`) and still spells the inherited `param1` as `mParam`. Twin-header drift, worth its own sweep.
- `config/converted-baseline.json` gains 6 rows and loses 6; `Behavior`, the largest function in the TU, is outside the CONVERTED set on both sides, so the ratchet is flat.

PR #2178, opened 2026-09-02
Promote ov029/daObjWc_Obj05_c to a single-file TU (fold twin header, rename from CageLift)

andrewboudreau:
Three commits, each byte-neutral against the ROM on its own, converting ov029's `CageLift` into a genuine single-file translation unit under the class's own cartridge name.

### 1. Fold the twin header (`2b3db34aa`)

This class was declared **twice**, and both spellings were compiled into the ROM at once:

* `include/CageLift.h` — the real header, `struct CageLift : dBgActor_c`, with the vtable overrides and a `pad_320[0x10]` tail.
* `include/daObjWc_Obj05_c.h` — a flat, non-inheriting auto-generated struct of the *same* class, `#include`d by the live sources and cast to from `this` inside `InitResources` and `Behavior`.

The flat header's four own fields (`0x320..0x32c`) replace the pad in the real header; the stub is deleted, the two `this`-to-flat-struct casts go away, and `self->` becomes `this->`. Its `mPosY` (0x060) and `mAngleY` (0x08e) are deliberately **not** carried over — those are `dActor_c`'s own fields at those offsets, reached by inheritance, and re-declaring them was the defect.

`check_header_offsets`: 6 commented fields, 0 mismatched, struct spans 0x330.

### 2. Rename to the cartridge's own name (`0a3363828`)

ov029 `_ZTV` at `0x02113e74` is preceded by offset-to-top `0` and `_ZTI 0x02113e30`, whose three words are `0x0209a764` (`__si_class_type_info` vtable + 8), `0x02113e3c` (its `_ZTS`, reading `15daObjWc_Obj05_c`) and `0x021089ec` (`_ZTI10dBgActor_c`, ov002). ov029's `symbols.txt` already spelled the `_ZTI`/`_ZTS` pair `daObjWc_Obj05_c` while naming the adjacent vtable `_ZTV8CageLift`, so this closes an existing inconsistency rather than making a new claim.

Naming the vtable also gave `romdata_check` something it can compare: **600 -> 601** verified symbols, **43,008 -> 43,020** verified data bytes.

The ledger was audited before and after per the `class_rename.py` hazard; only the two rows at `0x02111e74` and `0x02113e50` are this class's, both verified against ov029's `symbols.txt`, and the four rewritten coined rows carry `(was CageLift_Spawn)` / `(was CageLift_SpawnInfo)` provenance in column 5.

### 3. Promote (`f5d6760a3`)

Nine functions at `0x02111ac4..0x02111ea4` — the destructor pair, four vtable overrides, an arg-shifting veneer, its tail-callee and the factory — move from nine legacy one-function sources into `src/actors/d_a_obj_wc_obj05.cpp`.

**The inline-destructor lever.** `virtual ~daObjWc_Obj05_c()` became `virtual ~daObjWc_Obj05_c() {}`. Out of line mwccarm emits `D2, D0, D1`; the ROM has `D1` at `0x02111ac4` then `D0` at `0x02111b08` and no `D2` anywhere in ov029. Production isolation places `.text` in emission order, so the out-of-line spelling was refused at link even though the object itself matched — the manifest's `functions_occur_in_expected_order: PARTIAL -- [(0, 1)]` predicted it, and now reads `PASS`.

**The lever is only safe on a leaf**, so that was proved first: scanning all 977 `.bin` files under `extracted/` for the word `0x02113e30` finds exactly two hits, both at file offset `0x2cd0` in the same overlay reached by two paths, which is `0x02113e70` = its own `_ZTV - 4`. Nothing names it as a base.

With the destructor inline the key function is `InitResources` — the first *declared* non-inline virtual — and it lives here, so mwcc emits the class's own `_ZTV`/`_ZTI`/`_ZTS` alongside the eight inherited base-chain records. All eleven are licensed in `compiler_only_output` as `deadstrip-data` with the cartridge address of their canonical copy. `romdata_check` word-compares them: **6 VERIFIED, 5 PARTIAL, 0 DIFFERS**.

### Proof

`tools/rombuild.py -j16` after each commit: **106/106 exact, 100.000000%, mismatching 0**, 11,088 reproducing functions.

Gates: CONVERTED ratchet PASS 2592/2592 - `check_src_tu` every reference resolves - `check_src_tu_compiles` 98/98 - `check_header_offsets` clean - `check_rename_ledger` clean over 1,635 rows - `check_dead_references` and `layout_check` clean - `port_refcheck` 405 references resolve.


andrewboudreau:
Reviewed with an agent against the tree. Six prose findings and one real dangling reference, all fixed in `a26613901`; the code and the ROM are untouched, and `rombuild.py -j16` re-ran at **106/106 exact, 100.000000%, mismatching 0** after.

**Three were `class_rename.py` substituting the new name into COMMENTS.** It rewrites prose as well as code, so any sentence whose subject was the contrast between the two names ends up naming the same class twice and saying nothing:

* the manifest claimed `include/daObjWc_Obj05_c.h` "used to exist TWICE", then named one of the two copies as a *different* header — what existed twice was the class;
* the header said its fields were folded in "from the flat `include/daObjWc_Obj05_c.h`", naming itself as its own source;
* the factory's "historical alias `daObjWc_Obj05_c_Spawn`" was rewritten from `CageLift_Spawn` — offering as the historical name one that never existed.

This is the third PR in this campaign where a rename has quietly falsified a comment. Worth treating as a standing post-rename step, not a per-PR nicety.

**Three plain errors:** "four own fields" where there are six; `0x0209a764` described as "`__si_class_type_info` vtable **+ 8**" when `config/arm9/symbols.txt:4499` names that exact address `_ZTVN3abi20__si_class_type_infoE` (which is how the `d_a_obj_wc_obj01` and `d_a_cam_tag` mastheads spell it too); and the 600 -> 601 `romdata_check` figure attributed to the rename without saying why a *rename* adds a record. It does, and the mechanism is now written down: the out-of-line destructor was this class's key function, so the legacy `_ZN15daObjWc_Obj05_cD1Ev.cpp` was already emitting a record spelled `_ZTV15daObjWc_Obj05_c`, and nothing could compare it while `symbols.txt` called that address `_ZTV8CageLift`.

**The one non-prose finding:** `config_tu/arm9/overlays/ov029/delinks.txt` still gave the `0x02111ac4..0x02111ea4` run to `src_tu/actors/CageLift.cpp`, a path this branch renames and then deletes. Retargeted at `src/actors/d_a_obj_wc_obj05.cpp`, matching how line 33 of that same file was already updated for the sibling obj06 TU.

Also recorded rather than hidden: `_ZTV15daObjWc_Obj05_c` is one of the five `romdata_check` PARTIALs because `config/arm9/overlays/ov029/symbols.txt:203` puts the phantom `data_ov029_02113e88` **inside** the table, at vtable slot 5, so dsd cuts the symbol's extent there and only 5 of the >=18 words can ever be compared. Every byte that is compared is equal. Pre-existing and pervasive in ov029 (obj06's `_ZTV` is cut at +8); this entry is simply the first to lean on it.

### Follow-ups this review surfaced, deliberately NOT in this PR

Each needs its own evidence, and none blocks the merge:

1. `notes/data/c-cpp-classification.tsv:1001` and `notes/data/tu-merge-candidates.json` name `src/_ZN15daObjWc_Obj05_cD0Ev.c` — wrong extension (it was `.cpp`) *and* a file that no longer exists. Generated datasets; the `.c`/`.cpp` drift is pre-existing and has now been flagged on four PRs.
2. The same TSV row's tree-name/ROM-name column has collapsed to `daObjWc_Obj05_c/daObjWc_Obj05_c`, carrying no information — same root cause as above.
3. `notes/plan-tu-merge-queue.md:271` column alignment broken by the longer name.
4. `symbols/actor_renames.tsv` lines 963/964 and 996/997 now record a rename *to* `daObjWc_Obj05_c_Spawn`/`_SpawnInfo`, names that never existed in the tree; the `(was CageLift_Spawn)` provenance is in column 5, but the ledger's historical column itself was rewritten. Both addresses are verified as this class's, so this is a wording call for a maintainer, not an address error.
5. `symbols/actor_renames_report.txt:166` reads `NOVTABLE actor 99 daObjWc_Obj05_c` — odd, since the class's vtable is named.
6. `tools/tu_promote.py:246` inserts promoted entries in place, producing the only sort inversion in the otherwise-sorted 2592-entry `config/converted-baseline.json`. No functional impact (the ratchet is a set), but it churns on the next regeneration.

### Sign-off

Merge tree gated with `tools/premerge_check.py 2178 --repo .`: nothing goes green -> red. `source-coverage` is red on both sides with `fatal: not a tree object` — pre-existing, not this PR's. All other CI checks green; PR validation passed on `f5d6760a3` and is re-running over the prose fixup.

I have reviewed this and consider it ready to merge. (No PR in this repo can carry a GitHub approval — shared account — so this comment is the sign-off.)

PR #2179, opened 2026-09-02
Promote ov014/daWanwan_c (Chain Chomp) to a single-file TU

andrewboudreau:
The Chain Chomp becomes a single-file translation unit. Three commits: the
cartridge-name rename, one tool-policy fix, then the promotion.

## `603cacfb9` — rename `ChainChomp` → `daWanwan_c`

ov014 `_ZTI 0x02114718` holds `_ZTS 0x02114740`, which reads `10daWanwan_c`;
its base word `0x021081c0` is `_ZTI12dEnemyBase_c`. Byte-neutral: rombuild
106/106 exact, mismatching 0.

## `a03031a8a` — objisolate: a duplicate-licensed function may be referenced

`deadstrip_plan` refused to externalise a requested symbol when a surviving
content section still referenced it — calling a homeless symbol proves the body
is live compiler output, not dead weight. That rule predates the
`deadstrip-duplicate` disposition (#1878), and the two disagree: a
duplicate-licensed *function* is not homeless. It has a configured ROM home, and
`expect` refuses the license unless this object's copy is byte-identical to the
cartridge at that address, so a surviving reference to one is exactly what the
cartridge itself does.

This blocked any TU whose kept ROM function destroys an array of a
vague-linkage element type. `daWanwan_c` holds fourteen `Vector3` (two
`Vector3[7]`), so its ROM destructor at `0x02111308` names `_ZN7Vector3D1Ev`
from its literal pool at `+0xa0` — and the cartridge's own D1 has `0x020072c0`
at that same offset.

The relaxation is keyed on `expect` and nothing else. Without it the refusal is
unchanged (existing test unchanged and still passing); an unnamed section symbol
stays refused either way, because an anonymous reference cannot be redirected to
a home by name; a wrong body still refuses, at the byte comparison. The new test
covers all three. 39/39 `test_objisolate` pass.

## `42dc63e80` — promote to `src/actors/d_a_wanwan.cpp`

29 one-function files become one TU owning ov014's whole `.text` run
`0x02111308..0x02112e0c`.

**Destructor inline.** Out of line mwccarm emits D2, D0, D1; the ROM has D1 at
`0x02111308` then D0 at `0x021113bc` and no D2 anywhere in ov014. The manifest's
`functions_occur_in_expected_order` read `PARTIAL -- pair (0, 1)` for exactly
that reason and now reads PASS.

**Twelve licensed records.** With no out-of-line virtual left — the four
[...]

PR #2181, opened 2026-09-02
Let a licensed duplicate keep its surviving reference, on both paths

andrewboudreau:
Splits the tool policy out of #2179 so it can land ahead of the source that needs it.

## Why it is split

The validator restores all of `tools/` from the base commit, so a PR that carries
both a tool policy and the source that policy licenses is compiled on the box with
**main's** tools. #2179 failed exactly that way: `intact TU link control failed`,
`baseline command exited 1`, while the identical command exits 0 here in every
configuration — standalone, on a fresh worktree, with a cold object cache, after
generating the base commit's control first, and from inside `rombuild.py` with
`build/tu/_baseline` deleted so the bootstrap actually fires.

Landing a *tightening* early is how main went red on 2026-08-30 (#1987 before
#1988). This is a **relaxation with no caller in the tree**, so main stays green:
name neither `expect` nor `duplicates` and the rule is unchanged.

## What it does

A promoted C++ TU that inlines a base's destructor can define a function whose
bytes duplicate a home the ROM already has. The deadstrip rule refused any
surviving reference to a compiler-only function, with no way to say "this one
resolves to a configured home after the link, and the cartridge makes the same
reference". Two callers now hold that licence, because they hold different evidence:

- `expect` carries the cartridge's bytes — it licenses the reference **and** demands
  the body match, as far as an unlinked object allows: every relocated word is
  masked out, and `rombuild._duplicate_body_reasons` links the body and compares
  exactly those masked words. Neither half proves byte-identity alone.
- `duplicates` licenses the reference alone, for `tubuild.apply_compiler_only_policy`'s
  scratch link path, which has no cartridge bytes and leaves the body proof to the
  production build. That path had been left on the un-relaxed rule.

Neither may name anything but a requested compiler-only function.

Also: the strict-stock-control bootstrap now captures its subprocess output and
quotes the last twelve non-blank lines into its refusal. `baseline command exited 1`
narrowed an eight-phase command that returns 1 from thirty-odd places to "one of
those"; a remote gate reports that exception and nothing else.

## Verification

- `tools/test_objisolate.py` 40 pass (was 39). The new test closes a gap: mwcc emits
  no `STT_SECTION` symbol for the fixture's section, so the anonymous-section
  refusal was never exercised. It repurposes a spare `$`-prefixed mapping symbol in
  the victim's own section and checks the refusal under both licensing paths.
- `tools/test_tu_production.py` 22 pass (was 20).
- Full rebuild on main with these tools and no source change: **11,088 reproducing,
  mismatching 0, module fidelity 106/106 exact, 100.000000%**.
- `check_dead_references`: no new dead references.

#2179 will be rebased onto this once it lands.

### DIFF
--- a/tools/objisolate.py
+++ b/tools/objisolate.py
@@ -686,7 +686,7 @@ def derive(raw, keep_symbol):
     return _apply(raw, p, keep_symbol), p
 
 
-def deadstrip_plan(raw, symbol_names, expect=None):
+def deadstrip_plan(raw, symbol_names, expect=None, duplicates=None):
     """Plan removal of exact, explicitly named compiler-only output sections.
 
     This is intentionally narrower than a linker dead-strip pass.  TU reconstruction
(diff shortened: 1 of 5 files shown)


andrewboudreau:
Signing off (no PR in this repo can be APPROVED — shared account).

All checks green. Merge-tree gate (`premerge_check.py --repo . --fetch 2181`): nothing goes green → red; `source-coverage` errors identically on both sides (`not a tree object`, pre-existing and not this PR's).

Byte proof, on main with these tools and no source change: **11,088 source-built functions reproducing, mismatching 0, module fidelity 106/106 exact, 100.000000%**. `test_objisolate` 40/40, `test_tu_production` 22/22, `check_dead_references` no new dead references. No source in the tree reaches the new licence, which is what makes landing it ahead of #2179 safe.

PR #2182, opened 2026-09-02
Promote ov029/ArrowLift to daObjWc_Obj02_c, a single-file TU

andrewboudreau:
Converts `ov029/ArrowLift` — the arrow-shaped rising platform — into a genuine single-file translation unit spelled as the cartridge spells it. Two commits, in the order that matters: the rename first, then the delink it unblocks.

## 1. Rename `ArrowLift` → `daObjWc_Obj02_c`

`ArrowLift` was a coined name. ov029's vtable at `0x02113cf0` is preceded by its `_ZTI` at `0x02113cac`, whose `_ZTS` at `0x02113cb8` reads `15daObjWc_Obj02_c`. The class already carried both spellings at that one address, so the coined alias `_ZTV9ArrowLift` was deleted from `config/arm9/overlays/ov029/symbols.txt` before `tools/class_rename.py` would run.

**Hand-repaired after the tool ran.** `symbols/actor_renames.tsv`'s `why` column records what an address *used to* be called. The tool rewrote 14 rows; exactly 7 of them are **WaterDiamond's** addresses whose history merely mentions ArrowLift (`0x02111728`, `0x02111760`, `0x02111908`, `0x0211192c`, `0x02111954`, `0x02111a04`, `0x02113db4`). Those 7 are restored to `_ZN9ArrowLift…` by hand — this is [the known `class_rename.py` failure mode](https://github.com/tangosdev/sm64ds-decomp/pull/2176). The 7 rows on this class's own addresses keep the new spelling.

Proved on its own: rombuild 106/106 exact, mismatching 0.

## 2. Delink the nine one-function sources into `src/actors/d_a_obj_wc_obj02.cpp`

The whole `.text` run `0x0211137c..0x02111728` becomes one `complete` delinks entry.

**The rename is what made this possible**, and the manifest said so before the fact: promotion was refused because the compatibility class emitted `_ZTI9ArrowLift`/`_ZTS9ArrowLift`/`_ZTV9ArrowLift` while the ROM's vtable preamble points at `_ZTI15daObjWc_Obj02_c` — records that cannot be dispositioned against a cartridge address they disagree with. Spelling the class correctly makes them agree, and they become licensable for the first time.

Eleven `compiler_only_output` entries, disposition `deadstrip-data`: this class's own RTTI trio plus the eight inherited base-chain records (`fBase_c`, `dBase_c`, `dActor_c`, `dBgActor_c`). Each names a canonical ROM home *outside* this entry's licensed `.text`, which dsd supplies independently. The class has no key function — the first virtual it declares is its destructor, which `include/daObjWc_Obj02_c.h` defines inline, so every user of the class emits its own vague-linkage copy rather than one TU homing it. That inline destructor is load-bearing: out of line, mwccarm emits D2/D0/D1, while the ROM has D1 then D0 and no D2.

`tools/romdata_check.py` word-compares all eleven against the cartridge with relocations applied: **6 VERIFIED (184 bytes), 5 PARTIAL (52 bytes), 0 DIFFERS**. The PARTIALs are extent shortfalls, not disagreements — every byte compared is equal. Same split the `ov029/daObjWc_Obj05_c` sibling reports.

The one legacy `#pragma create` worth keeping, `long_calls on` around the tail-call veneer at `0x021116e4`, is restored bracketed at its own site so it stays function-scoped rather than becoming TU-scoped.

## Proof

- `tubuild.py linkcheck --baseline --module ov029 -j16 --clean` → modules PASS, ROM built **byte-identical to stock**
- `rombuild.py -j16` on this branch rebased onto `88a46bfae` → **module fidelity 106/106 exact, 100.000000%**, 11,088 reproducing, mismatching 0, zero new symbol errors
- ratchet: baseline 2592, current 2592, backslid 0

Pre-existing and unrelated: `dsd check symbols --fail` FAIL and the 5 tree-wide `romdata_check` DIFFERS both reproduce on the base.

## Follow-ups, not fixed here

- `symbols/actor_renames_report.txt` still has 7 `already claimed by ArrowLift` lines. It is **generated** by `tools/actor_names.py`, whose `main()` rewrites the entire `actor_renames.tsv` from the coined census (reading `extracted/arm9_dec.bin` + `reference/DynamicAllocationDecomp`, never `symbols.txt`). Running it to "refresh" the report would destroy the curated ledger, so it is left alone deliberately.
- `symbols/profile_reconstruction_registry.{json,tsv}` still name `ArrowLift_SpawnInfo` / `ArrowLift_Spawn` / `src/ArrowLift_Spawn.c` — and that last file does not exist, which was already true before this change. Same class of staleness as the #2177 follow-ups.
- Two `dBgActor_c` extern CONFLICT warnings are recorded in the manifest notes: `_ZN10dBgActor_c21UpdateModelPosAndRotYEv` (kept) and `_ZN10dBgActor_c19UpdateClsnPosAndRotEv` (commented out for review).
- Key-function wording disagrees in-tree: `d_a_obj_wc_obj05.cpp`'s header says the key function is `InitResources` (first *declared* non-inline virtual); the correction landed in #2179 says an inline first-declared destructor means **no** key function at all. This file states only what is empirically verified and does not re-litigate it.


andrewboudreau:
## Review pass

I had an agent audit the diff against the claims in the description. Nine claims checked out as written; three findings were real and two are now fixed in `0f6efa9bc`.

**Fixed**

1. `config_tu/arm9/overlays/ov029/delinks.txt:14` still named `src_tu/actors/daObjWc_Obj02_c.cpp` — a path the promotion had moved into `src/`. The three already-promoted neighbours in that same file (obj01, obj05, obj06) all point at their `src/actors/` path; obj02 now matches. This is the same "tenth dangling `src_tu` entry" shape I flagged on #2179, caught before it landed this time.
2. The manifest's fourth note still opened *"Promotion remains refused"*, and the rename had made its explanation self-refuting: it read that the TU emits `_ZTI15daObjWc_Obj02_c` *"while the ROM vtable preamble points at `_ZTI15daObjWc_Obj02_c`"*, because the blanket rename rewrote the coined half of a sentence whose entire point was that the two spellings disagreed. Now records what happened, and carries the `romdata_check` verdicts (6 VERIFIED / 5 PARTIAL / 0 DIFFERS) so the figure lives in a committed artifact rather than only in a source comment.

Re-proved after both edits: `linkcheck --baseline --module ov029 --clean` builds the control ROM byte-identical to stock, `rombuild -j16` reports **106/106 exact, 100.000000%, mismatching 0**.

**Examined and deliberately not changed**

- The review flagged the header's provenance list (`src/_ZN15daObjWc_Obj02_cD1Ev.cpp` etc.) as naming files that "never existed", since the branch's first commit renamed them before the second deleted them. That is the established convention, not a defect: `src/actors/d_a_obj_wc_obj05.cpp:31-39` lists its provenance the same way, under the ROM spelling, and so does that manifest's `functions[].legacy_source`. The files did exist under these names at `d4388be5e`.
- `notes/data/c-cpp-classification.tsv:1044-1045` had its class column collapse from `ArrowLift/daObjWc_Obj02_c` to `daObjWc_Obj02_c/daObjWc_Obj02_c`, destroying the coined→ROM mapping the column exists to record. Generated file; identical damage already exists for obj05 (flagged on #2178) and for wanwan. Worth one pass over the generator rather than a hand-patch here.
- `config/converted-baseline.json` gains a third sort inversion, matching the two the wanwan and obj05 promotions already left. Root cause is `tools/tu_promote.py:246` inserting in place; already on the follow-up list.
- `docs/class-reference.html`, `docs/index.html`, `docs/progress-treemap.svg` now name a class that no longer exists. CI-regenerated on the next progress refresh.

**Sign-off:** I have reviewed this branch and consider it ready to merge. (No PR in this repo can be formally approved — shared account — so this comment is the sign-off.)

PR #2183, opened 2026-09-03
Rename SwitchActivatedPlank to daObjWc_Obj04_c and promote it to one TU

andrewboudreau:
ov029's `SwitchActivatedPlank` takes the name the cartridge spells for it and becomes a single translation unit, `src/actors/d_a_obj_wc_obj04.cpp`, owning the whole `.text` run `0x02112630..0x0211299c`. Nine one-function sources are deleted.

## The name

ov029's `_ZTV` at `0x021141a0` is preceded by its `_ZTI` at `0x0211415c`, whose `_ZTS` at `0x02114168` reads `15daObjWc_Obj04_c`; the spawn profile just before it at `0x0211417c` is `g_profile_WC_OBJ04`, so the ROM's own actor id agrees with the RTTI string. The coined alias `_ZTV20SwitchActivatedPlank`, which named that same address, is gone.

## Why renaming first is what makes it promotable

The TU emits the class's `_ZTV`/`_ZTI`/`_ZTS` plus the eight inherited base-chain records (`fBase_c`, `dBase_c`, `dActor_c`, `dBgActor_c`) as vague linkage. While the class was spelled `SwitchActivatedPlank` those records contradicted the cartridge's spelling at the same address, so they could not be dispositioned and all eleven sat unlicensed. They are now `deadstrip-data` in the manifest's `compiler_only_output`, each with a canonical ROM home outside this entry's licensed `.text`.

`tools/romdata_check.py --files src/actors/d_a_obj_wc_obj04.cpp` word-compares all eleven against the cartridge with relocations applied: **6 VERIFIED (184 bytes), 5 PARTIAL (52 bytes), 0 DIFFERS, 0 UNNAMED** — the same split `daObjWc_Obj02_c` and `daObjWc_Obj05_c` report. The PARTIALs are extent shortfalls, not disagreements: every byte compared is equal.

## Why the class's own trio is vague linkage here

Not for obj02's reason, and the manifest says so rather than copying it. obj02 defines its destructor inline in the class body, so it has no key function. obj04's header *declares* `~daObjWc_Obj04_c()` out of line — the first virtual it declares, hence the key function — and nothing DEFINES it as a C++ member.

That is not a shortcut taken to avoid the work. Defining it as a real member was tried on this tree and the build refused it: mwccarm emits the D2/D1/D0 triple as one group in the order **D0-then-D1**, while the cartridge keeps D1 (`0x02112630`) *below* D0 (`0x0211267c`), so `objisolate` reports `requested functions are not emitted in ROM order`, and the group also brings an unhomed `_ZN15daObjWc_Obj04_cD2Ev`. Placing the two at their ROM addresses means writing them as separate bodies under their mangled names.

## Two ratchet findings

- **D1 and D0 backslide, and the exception is banked.** The mangled bodies cost both the CONVERTED tier's no-raw-offset and no-mangled-refs criteria. Recorded in `config/converted-backslide-exceptions.jsonl` with the emission-order reason above. Byte-match over readability, stated out loud — which is what that mechanism is for. The clean one-line member versions this promotion deletes remain in git history if the placement rule ever changes.
- **`CleanupResources` backslid for a fixable reason.** `tiers` scores a member from one `// @symbol` marker to the next, and the two `func_ov029_*` helpers below it carried none, so its fragment swallowed their raw offset arithmetic. Both are marked now and it scores clean again. Worth knowing generally: an unmarked body silently poisons its predecessor's score.

## Other changes

- The `#pragma optimize_for_size on` carried in from the legacy `Behavior` source is bracketed with an explicit `off` at the tail, so its extent (ROM ordinals 6 down to 0) is stated rather than inferred from where the file ends. Byte-neutral, re-proved.
- `config_tu/arm9/overlays/ov029/delinks.txt` repointed from `src_tu/actors/daObjWc_Obj04_c.cpp` to the promoted path — the dangling-path defect #2179 landed with.

## Proof

- `python tools/rombuild.py -j16` → **106/106 exact, 100.000000%**, 11,088 functions reproducing, **mismatching 0**, ROM-build analysis PASS.
- `python tools/tubuild.py linkcheck --baseline --module ov029 -j16 --clean` → modules PASS, `sm64ds-tu.nds` sha256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`, **identical to the stock ROM**. (`dsd check symbols --fail` FAILs in the baseline control too — pre-existing, tree-owned.)
- `python tools/tiers_ratchet.py --check` → PASS, 2590/2590.

## Follow-ups, not fixed here

- `notes/archive/plan-scalar-markers.md` names `src/_ZN15daObjWc_Obj04_c8BehaviorEv.cpp:35`, a file this PR deletes.
- `symbols/profile_reconstruction_registry.{json,tsv}` still carry `SwitchActivatedPlank_SpawnInfo` / `SwitchActivatedPlank_Spawn` / `src/SwitchActivatedPlank_Spawn.c` (the last of which does not exist).
- `symbols/actor_renames_report.txt` is generated from the coined census, so its `already claimed by SwitchActivatedPlank` lines are correct as they stand and were deliberately reverted.


andrewboudreau:
Signing off (no PR in this repo can be APPROVED -- shared account, so this comment is the sign-off).

**Gates.** All ten rows green, including PR validation. `names` and `tools` show no row because this PR touches no `**.py` and nothing under `tools/` -- a path-filter skip, not a silent pass.

**Merge tree.** `origin/main` moved to `b5a32e07e` after this branch was cut, but that commit is the `[skip ci]` progress refresh and touches only `contributions.json`, `docs/index.html` and `docs/progress-treemap.svg` -- no overlap with anything here. `git merge-tree --write-tree origin/main HEAD` produces a tree with no conflicts, and the three generated files are regenerated by CI on main anyway.

**Review.** A review pass over the diff verified, independently of the build:

- Every ROM address quoted in the commit messages, the PR body, the TU header and the manifest resolves in `config/arm9/overlays/ov029/symbols.txt`: `_ZTI` 0x0211415c, `_ZTS` 0x02114168, `g_profile_WC_OBJ04` 0x0211417c, `_ZTV` 0x021141a0, D1 0x02112630, D0 0x0211267c. The `.text` run 0x02112630..0x0211299c is exactly the span the nine replaced delinks entries covered, with no gap or overlap.
- All 11 `compiler_only_output` rows carry all five fields, and all 11 canonical addresses match that symbol's home in the owning module's symbols.txt. Zero discrepancies. The row set matches `daObjWc_Obj02_c.json`'s shape; the one deliberate divergence is the own-trio reason, which is the declared-but-undefined key function rather than obj02's inline destructor.
- The rename touched only obj04's own addresses -- 7 rows in `actor_renames.tsv`, all at 0x02112630 / 0x0211267c / 0x02112750 / 0x02112794 / 0x021127cc / 0x021128b0 / 0x021141a0. No collateral. (A prior session found `class_rename.py` rewriting unrelated ledger rows, so this was checked row by row.)
- `converted-backslide-exceptions.jsonl` appends two lines and modifies nothing above them.
- `converted-baseline.json`'s textual diff looks like it moves 11 unrelated rows; that is a re-sort of an unsorted region. The set delta is exactly -4 coined paths, +2 member identities, net -2, matching 2592 -> 2590 and the two banked exceptions.

Three prose defects it found are fixed in `20d108f59`, none of which any gate reaches: a `_ZTV... == _ZTV...` "two names, one address" sentence `class_rename.py` had made self-referential, a destructor comment citing a path that never existed (`check_dead_references.py` scans prose targets, not `include/` or `src/` bodies), and a historical citation the rename had rewritten into a fabricated filename.

**One correction to my own rename commit message**, left in place rather than rewritten into landed history: it says `tools/actor_names.py` reads `extracted/arm9_dec.bin` and `reference/DynamicAllocationDecomp` "never symbols.txt". It does read symbols.txt -- `load_modules()` loads every module's, feeding the `name already exists` skip. The conclusion is unaffected: the `already claimed by` lines come from the coined census, so regeneration reproduces the coined spelling and reverting `actor_renames_report.txt` was right. Only the stated reason was wrong.

**Follow-ups, not fixed here.** `symbols/profile_reconstruction_registry.{json,tsv}` still carry `SwitchActivatedPlank_SpawnInfo` / `SwitchActivatedPlank_Spawn` / `src/SwitchActivatedPlank_Spawn.c`, the last of which is not a file. `notes/data/c-cpp-classification.tsv:840-841` collapses its coined->ROM column to `daObjWc_Obj04_c/daObjWc_Obj04_c`, losing the lineage -- the same shape obj02's rows at 1044-1045 already have, so it is a convention worth revisiting across both rather than a defect of this PR.

PR #2184, opened 2026-09-03
Rename WDW_Water to daObjWc_Mizu_c and promote it to a single TU

andrewboudreau:
Renames `WDW_Water` to `daObjWc_Mizu_c` -- the name the cartridge itself spells, in `_ZTS14daObjWc_Mizu_c` at ov029:0x021140a4 -- and then promotes the class to one production translation unit, `src/actors/d_a_obj_wc_mizu.cpp`, licensing the contiguous ov029 `.text` run `0x021121a4..0x02112630`.

Two commits, deliberately separate: the rename is proved at 106/106 on its own, so the promotion diff is only the promotion.

## What had to be settled

**D1 and D0 are mangled bodies, not a real destructor.** The cartridge keeps D1 at `0x021121a4`, *below* D0 at `0x021121f0`. A real `daObjWc_Mizu_c::~daObjWc_Mizu_c()` member makes mwccarm emit the D2/D1/D0 triple as one group in the order D0-then-D1, which `objisolate` refuses with `requested functions are not emitted in ROM order`; the group also carries an unhomed D2. Separate `// @symbol` bodies under the mangled names are the only placement that reaches both ROM addresses.

The manifest had flagged exactly this in advance as `functions_occur_in_expected_order: PARTIAL -- ordinal pair(s) not in ROM order: [(0, 1)]`. That criterion now reads PASS, as do the other four.

Both bodies were read off the cartridge (`tools/disasm.py`, ov029 base `0x021111a0`) rather than assumed: two vptr stores and **four** destructor calls -- this class's own `TextureTransformer` at +0x320, then `dBgActor_c`'s `dBgW_KcMbg` at +0x124 and `Model` at +0xd4, then `dActor_c` -- plus, in D0, `Memory::Deallocate` against `data_020a0eac`. The inherited comment claiming "three destructor calls ... this class adds no member with a destructor of its own" was wrong and is corrected.

Cost: the CONVERTED tier's no-raw-offset and no-mangled-refs criteria for those two functions, banked in `config/converted-backslide-exceptions.jsonl` with that reason; baseline 2590 -> 2588. Byte-match over readability, the same wall `ov029/daObjWc_Obj04_c` hit in #2183.

**The vptr stores carry no bias.** `classInit` stored `&_ZTV14daObjWc_Mizu_c[2]`, which production isolation refuses: `intact multi-symbol object has undefined RTTI addend 8; text-only production isolation cannot rewrite it`. `symbols.txt` records `_ZTV14daObjWc_Mizu_c` at `0x021140dc` and that *is* the address point -- the cartridge's own literal pool word at `0x021121e8` is exactly `0x021140dc`. Addend 0 is correct and byte-identical.

**A missing tiers marker.** ROM ordinal 3 (`func_ov029_021122b4`) had no `// @symbol` line. `tiers._marked_member_fragment` runs from one marker to the next, so an unmarked body scores its raw offset arithmetic against its *predecessor* -- the poison diagnosed on obj04. Marker added; all 9 functions are marked.

**No `compiler_only_output` is needed.** This TU references the RTTI records but defines none, so there is nothing for the manifest to license -- unlike obj04, which needed 11 entries. Adding them was tried and correctly refused with `compiler-only symbol(s) not defined`.

## Proof

- `tools/rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes**, 11,088 functions reproducing, **mismatching 0**, source-owned data 10/10.
- `tools/tubuild.py linkcheck --baseline --module ov029 -j16 --clean`: modules PASS, ROM **identical** to the tree's stock build. (`dsd check symbols --fail` FAIL is pre-existing and tree-owned; it fails the same way in the control.)
- `tools/tubuild.py verify` before promotion: 9/9 MATCH, objisolate clean, reloc destinations clean, emission order clean.
- `tools/tiers_ratchet.py --check` PASS at 2588.
- Gates run locally: dead references, references, src_tu, rename ledger, header offsets, data definitions, duplicate sources, tubuild conflicts -- all pass.

## Known follow-ups, not fixed here

- `notes/data/c-cpp-classification.tsv:547,6012` and `notes/data/tu-merge-candidates.json:2461-2469` still list the absorbed one-symbol paths. Same generated-snapshot staleness noted on #2178, #2179 and #2183; worth one sweep rather than N patches.
- `romdata_check` totals move 605 -> 603 verified with DIFFERS unchanged at the tree-wide 5. That is the record-count consolidation effect of nine objects becoming one, not a regression, but it is the third promotion to show it and the metric's arithmetic deserves the same documented treatment `byte-verified` already has.


andrewboudreau:
Reviewed by a second agent against the cartridge bytes, not just the diff.

**ROM-level claims: all verified.** Every address, size, call target and offset in the header, manifest and commit messages was re-derived from `overlay_0029.bin` at base `0x021111a0` and cross-checked against `config/arm9/symbols.txt` and `config/arm9/overlays/ov029/symbols.txt` — D1 `0x021121a4`/`0x4c`, D0 `0x021121f0`/`0x60`, the two vptr stores in each, all four destructor calls and their `add r0,r4,#N` offsets, `Memory::Deallocate` with `data_020a0eac` in D0 only, the literal pool word `0x021121e8 = 0x021140dc`, and independently the same addend-0 store in `classInit` (`mov r0,#0x348`, literal at `0x0211262c`). The rename commit's seven ledger rows are exactly the seven rows the diff moves. All 9 `// @symbol` markers match their blocks in the correct 8→0 reverse-ROM sequence, manifest addresses/sizes agree 9/9 with symbols.txt, and `converted-baseline.json` loses exactly the two banked identities and nothing else. `attribution.json` gained 9 keys and lost none.

**Three prose defects found and fixed in f36899cf**, none of them reachable by any gate:

1. The manifest's shadow-stage note still asserted `duplicate real destructor collapsed; +8 at the Spawn` — both undone by the promotion shipping in the same file. It is now marked superseded, and two new notes carry the promotion's own reasoning.
2. `include/daObjWc_Mizu_c.h` justified its flat `#else` branch with "the D0 file is a C translation unit … D0 is compiler-generated so it can never be migrated." Wrong three ways: there is no D0 file, it was never C (`src/_ZN9WDW_WaterD0Ev.cpp` was `//cpp`), and D0 is now hand-written. Nothing compiles that branch — every remaining includer is `//cpp`. Kept with its real reason stated (the only flat record of measured offsets).
3. The size assert was still spelled `WDW_Water_size_must_be_0x348`, the last non-generated place the coined name survived.

**Two corrections to my own claims:**

- The rename commit message says the rename "makes two more ROM data records resolvable — romData goes from 604 verified / 468 unnamed to 605 / 466." The arithmetic doesn't support the sentence: two records left `unnamed` but only one arrived at `verified`. Left in landed history rather than rewritten; the numbers are right, the inference from them was sloppy.
- The reviewer flagged that this TU escaping obj04's 11-symbol `compiler_only_output` was asserted without explanation, when obj04's own note warns against reasoning by analogy between classes. Fair. The manifest now says the honest thing: this TU emits no RTTI, observed rather than derived — adding obj04's entries was tried and the build refused them with `compiler-only deadstrip refused: compiler-only symbol(s) not defined`, and removing them is what let it link.

Rebuilt after the corrections: **106/106 exact, 100.000000%, mismatching 0**, ratchet PASS at 2588, `check_header_offsets` clean on the changed header (6 commented fields, 0 mismatched, struct spans 0x348).

Signing off on this as reviewed and correct. (No PR in this repo can be formally APPROVED — shared account — so this comment is the sign-off.)

andrewboudreau:
**Validator fix pushed** (`56ee21130`) — the one red check was the romdata ratchet, not the build:

> Validation failed: ROM data verified from source fell from 604 to 603 symbol(s)

**Diagnosis.** The promotion swapped the class's single real destructor member for two `// @symbol` mangled bodies, so nothing defines the key function. That alone was expected. What was not: with `virtual ~daObjWc_Mizu_c();` declared **first** in the class, mwccarm emitted no vtable and no RTTI at all, so `_ZTV/_ZTI/_ZTS14daObjWc_Mizu_c` left the set of ROM data any source verifies. Compiling the deleted `src/_ZN14daObjWc_Mizu_cD1Ev.cpp` proves it used to emit 11 data symbols; the promoted TU emitted 0.

**The lever is declaration position.** Measured on this class, everything else held:

| `virtual ~daObjWc_Mizu_c();` declared | data symbols emitted |
| --- | --- |
| first, before the non-virtual methods | 0 |
| last, after them | 11 (6 VERIFIED, 5 PARTIAL vs the cartridge) |

`include/daObjWc_Obj04_c.h` already had the destructor last, which is exactly why that sibling licenses 11 vague-linkage RTTI symbols and this class appeared not to need any — my earlier manifest note drew the wrong conclusion from that and has been replaced, not amended. There is only one virtual in the class, so moving the declaration cannot move a vtable slot.

The 11 symbols are now licensed as `deadstrip-data` with each one's canonical cartridge address; production isolation discards them after `romdata_check` word-compares them to the ROM.

**Re-proved locally:** `tools/rombuild.py -j16` → 106/106 exact, 100.000000%, mismatching 0; ROM data from source **605** verified (main is 604), 184 partial, 5 differ. `check_header_offsets` clean (6 commented fields, 0 mismatched, spans 0x348); tiers ratchet 2588/2588, BACKSLID 0.

Also correcting an arithmetic overstatement in `6008d9656`'s message: the rename moves two records out of `unnamed`, but only one of them becomes VERIFIED (`_ZTS…` lands in PARTIAL), so it is 604 → 605, not 604 → 605 *and* two more verified as that message reads.

andrewboudreau:
Scoping the claim above, after checking it against the rest of the tree: **declaration position is the lever only for a destructor that is DECLARED and left undefined**, which is the shape `daObjWc_Mizu_c` and `daObjWc_Obj04_c` share.

A class whose header gives the destructor an inline body — `virtual ~daWanwan_c() {}` — emits the same 11 symbols with the dtor declared first (`src/actors/d_a_wanwan.cpp`, checked: 11 emitted). That is the separate inline-dtor route, and it is why most already-promoted TUs carry `compiler_only_output` while declaring the destructor first. So this is not a defect class to sweep the tree for; nothing else here has the out-of-line-and-undefined shape without already licensing its RTTI.

PR #2190, opened 2026-09-03
Promote daObjWc_Obj03_c to a single-file TU

andrewboudreau:
Renames `WaterDiamond` to `daObjWc_Obj03_c` -- the name the ROM's RTTI gives it -- and consolidates the ten one-symbol `src/_ZN15daObjWc_Obj03_c*.cpp` shards plus `src/daObjWc_Obj03_c_classInit.c` into `src/actors/d_a_obj_wc_obj03.cpp`, written in reverse ROM order.

The previous `src_tu/` shadow carried its own **"THIS TU MUST NOT BE PROMOTED"** banner -- ordinals [2], [3] and [4] had been degraded to `func_<addr>` raw-pointer-arithmetic placeholders -- so the TU was reconstructed from the real named-member sources instead.

### The two conditions, same pair the sibling `daObjWc_Mizu_c` settled

**Emission order.** The cartridge keeps `D1` at `0x02111728` *below* `D0` at `0x02111760`, while a real destructor member makes mwccarm emit the D2/D1/D0 triple as one D0-then-D1 group (and bring an unhomed D2). Both variants are carried as `// @symbol` mangled bodies read off the cartridge with `tools/disasm.py`. That costs the CONVERTED tier's no-raw-offset and no-mangled-refs criteria for those two functions, banked in `config/converted-backslide-exceptions.jsonl`.

**No vptr bias.** `symbols.txt` records `_ZTV15daObjWc_Obj03_c` at `0x02113db4` and that IS the address point -- D1's literal pool word at `0x0211175c` is exactly `0x02113db4` -- so every vptr store names the symbol with addend 0. The `&_ZTV15daObjWc_Obj03_c[2]` spelling is refused by production isolation.

### Why the destructor is declared last

Nothing defines `~daObjWc_Obj03_c()` as a C++ member, so the vtable and RTTI have no key function to home them. Whether mwccarm emits them here turns on *where* the destructor is declared: measured on `daObjWc_Mizu_c`, declared first the TU emits 0 data symbols; declared after the other members it emits the full set. Declared first is not harmless -- it drops `_ZTI/_ZTS/_ZTV15daObjWc_Obj03_c` out of the set of ROM data any source verifies, and `tools/validate_merge.py` ratchets on that count with no escape hatch. The nine resulting symbols (this class's three plus the `dActor_c`/`fBase_c`/`dBase_c` vague-linkage passengers) are licensed as `compiler_only_output`.

Also drops a duplicate `_ZTV15daObjWc_Obj03_c` extern that the class rename left in `decl_common.h`.

### On the `tubuild verify` DIFFs

`tubuild verify` reports 7/10 MATCH with `objisolate: _ZTV15daObjWc_Obj03_c: unexpected reloc type=2 addend=0` on D1, D0 and classInit. That is a verify-path artifact: running the same command against the already-landed, CI-green `src/actors/d_a_obj_wc_mizu.cpp` produces the identical shape (6/9, same message, same three ordinals), and main builds 106/106 exact with that file. `tools/objisolate.py` demands `addend >= VTABLE_PREAMBLE` only for a `_ZTV` it is *externalising*; production drops it via the deadstrip-data disposition, where addend 0 is correct.

### Proof

`python tools/rombuild.py -j16`:

```
  reproducing: 11,088
  mismatching: 0
ROM-build analysis: PASS
ROM data from source: 606 symbol(s) verified, 185 partial, 5 differ
```

Gates run locally: `check_header_offsets --changed origin/main` (0 mismatched, struct spans 0x160), `check_references` (no new unresolvable references), `check_dead_references` (no new dead references, no broken links).


andrewboudreau:
Signing off (this repo's shared account cannot APPROVE, so this comment is the sign-off).

All ten CI checks green, and `tools/premerge_check.py --repo . 2190` reports `nothing goes green -> red` against the merge tree, with `entries: 10580 -> 10571 (-9)` correctly classified as consolidation rather than loss.

An agent review of the full diff verified independently: all ten ROM ordinals present in reverse ROM order; both `// @symbol` destructor bodies faithful to the cartridge (D1's literal at `0x0211175c` is `0x02113db4`, D0's `Memory::Deallocate` double indirection through `data_020a0eac` correct); all nine `compiler_only_output` canonical addresses matching `symbols.txt`; the header's field arithmetic closing at exactly the 0x160 that `classInit` passes to `fBase_c::operator new`; and the `complete` delinks entry abutting obj02's end and obj05's start.

It found five things worth fixing, all fixed in `7e231841c` and none of them able to move a ROM byte:

1. The class banner in `include/daObjWc_Obj03_c.h` survived the blanket rename and became self-refuting -- it still claimed "the ROM exposes both names at the same vtable address" after the rename left exactly one name.
2. `config_tu/arm9/overlays/ov029/delinks.txt` still pointed at `src_tu/actors/WaterDiamond.cpp`, which this branch deletes. Both sibling promotions (`4d09a5484` mizu, `f6889add6` obj04) carried this repoint and this one skipped it. No gate catches it -- `check_dead_references` only scans prose files -- and it would have bitten the next `config_tu`-rooted build.
3. `symbols/actor_renames.tsv` had its history column overwritten in place across all 14 rows in the block, asserting renames that never happened (`func_ov029_02111728` straight to the RTTI spelling). Restored as `(was _ZN9ArrowLiftD1Ev, then _ZN12WaterDiamondD1Ev)`.
4. `src/actors/d_a_obj_wc_obj03.cpp`'s "Historical alias" line named the current spelling as its own history; the reconstruction-provenance paragraph the promotion dropped is back with it.
5. Recorded in the manifest why this one reads `7/10` / `FAIL` while the equally healthy mizu sibling reads `9/9` / `PASS` -- mizu's numbers were machine-written from its pre-promotion shadow, which still spelled the vptr `&_ZTV[2]`.

Follow-ups NOT fixed here, all generated-snapshot staleness that self-heals or that deserves one sweep rather than N patches:

* `notes/data/tu-merge-candidates.json:8188-8197` and `notes/data/c-cpp-classification.tsv:691-692` still list the absorbed shards -- the same staleness noted on #2178, #2179, #2183 and #2184. Six mizu shards are still listed there too, long after mizu landed.
* `docs/class-reference.html`, `docs/index.html`, `docs/progress-treemap.svg` still say `WaterDiamond`; the "Refresh progress" job regenerates them.
* `symbols/actor_renames_report.txt:157-163` had `SKIP WaterDiamond:` rewritten to `SKIP daObjWc_Obj03_c:`, falsifying the log of a run that happened under the old name. Generated and self-healing, but rewriting logs in place is a habit worth not forming -- worth teaching `class_rename` to leave `*_report.txt` alone.
* `tools/tu_promote.py:246` still writes `converted-baseline.json` with the sort inversion first noted on #2182, and with LF endings in a CRLF tree.

PR #2195, opened 2026-09-03
Promote ov009/daMcFlag_c to a genuine single-file TU

andrewboudreau:
## What this does

Promotes `ov009/daMcFlag_c` (the class the project used to call `Flag`) from seven per-function delink shards to one genuine translation unit, `src/actors/d_a_mc_flag.cpp`, compiled once with the pinned toolchain and byte-verified **7/7** against the cartridge with relocations applied.

Two commits, in the order the recipe requires:

| | |
|---|---|
| `12b28c25b` | rename `Flag` -> `daMcFlag_c`, the cartridge's own RTTI spelling, alone |
| `2477e80d4` | the promotion |

### The name comes from the ROM

`_ZTS10daMcFlag_c` at ov009 `0x02113b6c` reads `10daMcFlag_c`, and the `__si_class_type_info` record at `0x02113b60` is `0209a764 02113b6c 0208e390` — its own type-name string and `_ZTI8dActor_c`, so the cartridge states the direct and only base too. `Flag` was a coined name; `daMcFlag_c_Spawn` and `daMcFlag_c_SpawnInfo` are convention spellings for two derived C symbols the ROM says nothing about (historical aliases `Flag_Spawn`, `Flag_SpawnInfo`, both kept in `symbols/actor_renames.tsv`).

The rename was an ALIAS case — `_ZTV10daMcFlag_c` already existed at the same address as `_ZTV4Flag` — so it needed `--allow-collision` and a hand dedupe of the doubled `symbols.txt` line. `class_rename` also fired on nine unrelated sites where "Flag" is an English word (`include/Camera.h`, four `enum Flag`/`typedef ... Flag` C shards, four notes files); all nine were reverted. `include/ModelAnim.h:42` is a real reference and was kept.

### Absorbed shards

| ROM | symbol | legacy file |
|---|---|---|
| `0x02112078` | `_ZN10daMcFlag_cD1Ev` | `src/_ZN10daMcFlag_cD1Ev.cpp` |
| `0x021120a8` | `_ZN10daMcFlag_cD0Ev` | `src/_ZN10daMcFlag_cD0Ev.cpp` |
| `0x021120ec` | `_ZN10daMcFlag_c16CleanupResourcesEv` | `src/_ZN10daMcFlag_c16CleanupResourcesEv.cpp` |
| `0x0211211c` | `_ZN10daMcFlag_c6RenderEv` | `src/_ZN10daMcFlag_c6RenderEv.cpp` |
| `0x02112144` | `_ZN10daMcFlag_c8BehaviorEv` | `src/_ZN10daMcFlag_c8BehaviorEv.cpp` |
| `0x02112190` | `_ZN10daMcFlag_c13InitResourcesEv` | `src/_ZN10daMcFlag_c13InitResourcesEv.cpp` |
| `0x021121f0` | `daMcFlag_c_Spawn` | `src/daMcFlag_c_Spawn.c` |

### This TU homes its own vtable

Unlike the recent mangled-body siblings (`ov029/daObjWc_Mizu_c`, `ov036/daObjRc_Dorifu_c`), both destructor variants here are **real C++ members**. `InitResources` is the key function and `~daMcFlag_c()` is declared inline in `include/daMcFlag_c.h`, which makes mwccarm emit D1 then D0 in the cartridge's order — declared out of line it emits `D2, D0, D1`, and production isolation, which places `.text` sections into the spanning delink in emission order, refuses the range.

That also means the two consequences run the other way from those siblings:

* the vptr store in `daMcFlag_c_Spawn` can spell `&_ZTV10daMcFlag_c[2]`, because the relocation is internal and mwccarm resolves the 8-byte address-point bias itself — no bare-symbol/addend-0 workaround;
* the class's own `_ZTI`/`_ZTS`/`_ZTV` are genuine key-function output rather than homeless vague linkage.

An earlier attempt that reused the shadow TU's TU-local classes verified 5/7 with `MISSING _ZN10daMcFlag_cD1Ev` / `D0Ev`: the shadow declared `extern int _ZTV10daMcFlag_c[]` and never emitted the vtable, so nothing odr-used the destructor. That is the documented second condition of the inline-dtor lever, not a property of the class.

### Wall 6az, measured

`InitResources` calls `ModelAnim::SetAnim` through its mangled free-function spelling with a **scalar** speed argument. Its real signature takes `Fix12<int>` by value (`include/ModelAnim.h:89`), and a by-value class argument makes mwccarm home the constant and reload it:

```
ldr r3, [pc, #0x1c]
ldm r3, {r3}
```

where retail has a plain `mov r3, #0x1000` — two instructions and a literal-pool word, `0x68` against the cartridge's `0x60`. Every aggregate form produces the same reload: a brace-initialised local, a `const` local, a returning inline helper, a gccext compound literal `(Fix12<int>){0x1000}`, and a `Fix12` given a real raw-bits constructor (that last one was tried in `include/math/Fix12.h` and reverted — it is not in this PR). The mangled free-function spelling with a scalar parameter is what the landed `src/actors/d_a_obj_bc_switch.cpp` already does for `Sound::ChangeMusicVolume` and `dBgActor_c::IsClsnInRange`. The reasoning is written out at the declaration in the source.

It cost nothing: that member did not hold the CONVERTED tier before the promotion either, so the ratchet moved **2584 -> 2584, +0/-0**, and no backslide exception was appended.

### Licensed compiler-only output

Nine vague-linkage data records, all `deadstrip-data` with canonical module and address so `romdata_check` word-compares each against the cartridge before production isolation discards it:

| symbol | canonical |
|---|---|
| `_ZTI7dBase_c` / `_ZTS7dBase_c` | arm9 `0x02086e78` / `0x02086e54` |
| `_ZTI7fBase_c` / `_ZTS7fBase_c` | arm9 `0x02086d70` / `0x02086e60` |
| `_ZTI8dActor_c` / `_ZTS8dActor_c` | arm9 `0x0208e390` / `0x0208e384` |
| `_ZTI10daMcFlag_c` / `_ZTS10daMcFlag_c` / `_ZTV10daMcFlag_c` | ov009 `0x02113b60` / `0x02113b6c` / `0x02113ba0` |

`config_tu/arm9/overlays/ov009/delinks.txt:28` is repointed off the now-deleted `src_tu/actors/Flag.cpp`.

## Proof

```
tools/tubuild.py verify ov009/daMcFlag_c
  byte comparison : 7/7 MATCH
  objisolate check: clean
  emission order  : all 7 function(s) in the expected ROM-ascending section order
  Result: 7/7 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tools/rombuild.py -j16
  ROM-build analysis: PASS
  ROM data from source: 609 symbol(s) verified
```

Gates: `eligible` (11082, baseline 11044; unresolved 42 against baseline 45), `check_header_offsets --changed origin/main`, `port_refcheck` (405 checked, all resolve), `layout_check` clean, `check_dead_references` no new dead references, `check_references` OK.


andrewboudreau:
Sign-off (no PR in this repo can be APPROVED -- shared account -- so this comment is the sign-off).

**Proof**
- `tools/rombuild.py -j16` on the final tree: `11,088` source-built functions, `reproducing 11,088`, `mismatching 0`; module fidelity **106/106 exact, 100.000000%**; ROM-build analysis PASS; 612 ROM data symbols verified.
- `tools/premerge_check.py --repo . 2195` against merge tree `ad44fdf8d264`: every gate `ok`. `source-coverage` errors identically on base and merge (`git ls-tree ... fatal: not a tree object`) -- a pre-existing local artifact, not this PR's.
- All GitHub checks pass; "PR validation" (tangos.dev) passed in 20m26s.

**The finding this PR settles.** For a key-function TU, production isolation empties the emitted `_ZTV/_ZTI/_ZTS` sections and turns the symbols into imports -- in `build/src/actors/d_a_mc_flag.o` all three are `SHN_UNDEF` with `st_size 0`. The vptr relocation is therefore *external*, not internal. It still lands because mwccarm folds the 8-byte address-point bias at compile time: all three `R_ARM_ABS32` relocations naming `_ZTV10daMcFlag_c` carry **addend 0** over a zero content word, resolving to `symbols.txt`'s `0x02113ba0`, which *is* the address point (`0x02113b98` offset-to-top, `0x02113b9c` `_ZTI`, `0x02113ba0` slot 0 = `InitResources`). Both source spellings reach the same relocation.

**Two review findings fixed in `74cd97131`** (both were confident prose the artifacts did not support):
1. `d_a_mc_flag.cpp` claimed both shared resource handles go through `SharedFilePtr::Construct`. Only `0x02113eb0` does; `0x02113eb8` goes through `func_02017acc`, still unnamed.
2. The manifest's PROMOTION note blamed a 5/7 MISSING-destructor state on the committed shadow TU. The shadow wrote `&_ZTV10daMcFlag_c[2]`, did emit the vtable, and verified 7/7. The 5/7 state belonged to an intermediate form of this rewrite that declared the vtable `extern` and so emitted none.

**Two known-wrong commit messages, left as-is rather than rewriting pushed history:**
- `927dfd706` says "ten factual defects"; the review found 12.
- `db00b72f5` credits itself with the `attribution.json` union merge; that union is in `146fbb0bf`.

PR #2197, opened 2026-09-03
Promote ov072/daBgSnwmn_c to a single production TU

andrewboudreau:
Eight one-function sources for the Big Snowman (`BIG_SNOWMAN`, ov072) become one file, `src/actors/d_a_bg_snwmn.cpp`, licensing the contiguous `.text` run `0x02120824..0x02120c00`. The ROM build now takes the whole class from a single object, the way the cartridge's own build did. It is done the way #2184 did `daObjWc_Mizu_c`.

The class needed no rename — the tree already spells it as the ROM RTTI does (`ov072:0x02122938`).

### Three things the class forced

**The destructors are `// @symbol` mangled bodies, not a real `~daBgSnwmn_c()`.** The cartridge keeps D1 (`0x02120824`) *below* D0 (`0x02120874`). A real C++ member makes mwccarm emit the D2/D1/D0 triple as one group ordered D0-then-D1, and the whole-range link refuses with `licensed .text functions are not emitted in ROM address order`. The two functions therefore lose the CONVERTED tier's no-mangled-refs and no-raw-offset criteria; banked in `config/converted-backslide-exceptions.jsonl` with that reason. Byte-match outranks readability.

**The vptr stores are spelled `&_ZTV11daBgSnwmn_c[2]`, not the bare symbol.** This is the one genuinely new finding here, and it qualifies #2184's banner, which says the stores "carry no bias". Both statements are right, for different reasons:

- `symbols.txt` records `_ZTV11daBgSnwmn_c` at `ov072:0x02122978`, which *is* the vtable's address point — the literal pool word at `0x02120870` holds exactly that.
- But this TU **defines** the vtable, and mwccarm's own `_ZTV` symbol addresses the vtable *object's* start, 8 bytes lower (offset-to-top, then `_ZTI`). `tools/objisolate.py:440-490` externalises the locally-defined symbol and subtracts that `VTABLE_PREAMBLE`, so it requires an addend of at least 8. With `(int)_ZTV11daBgSnwmn_c` it refuses **both** destructors: `_ZTV11daBgSnwmn_c: unexpected reloc type=2 addend=0`, and verify drops to 6/8.
- `&_ZTV11daBgSnwmn_c[2]` gives addend 8, which the subtraction turns back into `0x02122978`.

**The spelling follows from who defines the vtable, not from the class.** A TU that only *imports* one takes addend 0.

**The destructor is declared LAST in `include/daBgSnwmn_c.h`.** Nothing defines it as a real member, so the class has no key function and its vtable and RTTI are vague linkage. Declared first, mwccarm emits no RTTI at all and `_ZTV/_ZTI/_ZTS11daBgSnwmn_c` go unverified by any source.

### Also

Unified the two spellings of `data_ov072_02122c70` the legacy files carried — `extern const Vector3` beat `extern int[]`, which the callee's own mangled name (`RK7Vector3`) already claimed. This was a hard compile error the moment the two files shared a TU, not a style call.

### Proof

- `tubuild verify` 8/8 MATCH, objisolate clean, reloc destinations clean, emission order ROM-ascending.
- `tubuild linkcheck --baseline --module ov072` **SCRATCH-LINK-VERIFIED**; the scratch ROM was byte-identical to `build/sm64ds.nds`. `dsd check symbols --fail` fails on 9 pre-existing baseline errors, 0 new.
- `rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0**, 11,088 reproducing / 0 mismatching.
- Gates: `eligible`, `port_refcheck` (405 refs), `check_dead_references` (no new), `layout_check` clean, `check_header_offsets` (`daBgSnwmn_c.h`: 6 commented fields, 0 mismatched, spans `0x1f0`), `check_references` **OK: no new unresolvable references** (its `--update` offer is unrelated drift and is deliberately not banked here).

### Review pass

An agent re-checked every address, offset, tool claim and behaviour comment against the ROM and the tree. It confirmed the vtable, `objisolate`, struct-offset, `compiler_only_output` and `delinks.txt` claims verbatim, and found five wrong statements — all prose, all fixed in `e8fcf407f`:

- an inverted condition in the `InitResources` comment (the 0x111 spawn runs when star 5 is **not** collected, and it is not a bail-out — there is no early return);
- a false "its first 0x50 bytes are D1's, verbatim" (19 of those 80 bytes differ; the first 0x40 are the same *sequence*, not the same bytes);
- a cited manifest verdict that does not exist (the manifest says `functions_occur_in_expected_order: PASS` — the ROM order is fine, it is mwccarm's emission order that is not, and that check does not predict it);
- "declared LAST ... after the non-virtual methods", for a class that has no non-virtual member functions;
- an extern block claiming dCcAcPos_c, TextureSequence and SharedFilePtr "have no project header yet" when all three headers exist and two are already included.

It also caught two pre-existing errors in `include/daBgSnwmn_c.h`, fixed in the same commit: the destructors were named as `func_ov072_02120824 / _02120874`, symbols that do not exist, and `0x0d0` was listed among the fields that "really belong to dActor_c" when `dActor_c` is exactly `0xd0` bytes and `0x0d0` is this class's own `unk_0d0`.

One finding this PR does **not** fix: several externs here duplicate declarations that `include/dCcAcPos_c.h`, `include/TextureSequence.h` and `include/SharedFilePtr.h` already carry. Each swap has to be re-proved against the ROM, so it is left as a follow-up and recorded in the comment rather than done blind.


andrewboudreau:
Signing off (this repo's PRs can't be APPROVED — shared account — so the sign-off is this comment).

**Proof on the head commit `e8fcf407f`:**
- `rombuild.py -j16`: 106/106 exact, 100.000000%, 11,088 reproducing / **0 mismatching**. Re-run after the prose-fix commit, not just before it.
- `tubuild verify` 8/8 MATCH, objisolate clean, emission order ROM-ascending; `linkcheck --baseline --module ov072` SCRATCH-LINK-VERIFIED with the scratch ROM byte-identical to `build/sm64ds.nds`.
- "PR validation" (tangos.dev) passed on `e8fcf407f` itself — confirmed via the check-runs API, not just the branch view, since the branch was force-pushed after the first run.

**`premerge_check.py --repo . 2197` on the merge tree `bcc349c8e69a`:** every gate `ok` — converted-ratchet, dead-references, duplicate-sources, header-offsets, langmode-ratchet, layout-check, src-tu-refs. `source-coverage` errors on **both** base and merge with the identical, long-standing `git ls-tree ... fatal: not a tree object` artifact; it is red on the base too, so it is not this PR's to own. Nothing goes green → red.

Its informational line is worth reading rather than skipping: `entries: 10555 -> 10548 (-7)`. That is the eight one-function entries collapsing into one `complete` entry — consolidation, not lost coverage. The bytes are flat.

**Two things a reader should not take on trust from the original PR body**, both since corrected:
- The reference-baseline commit that was here has been **removed**, and its claim was false. It said the consolidation fixed three unresolvable references; the three removals are `_ZN5FaderD0Ev`, an ov004 `table$6` and ov007 `overlay_100/102`, none in ov072, and no `daBgSnwmn_c` reference was ever banked. That was unrelated drift. `check_references.py` reports OK without it.
- Five prose statements were wrong and are fixed in `e8fcf407f`; the "Review pass" section of the description lists them. The one that mattered most was an inverted condition in the `InitResources` comment.

**Known and deliberately not fixed here:** several externs in `d_a_bg_snwmn.cpp` duplicate declarations `include/dCcAcPos_c.h`, `include/TextureSequence.h` and `include/SharedFilePtr.h` already carry. Each swap changes what the compiler sees and has to be re-proved against the ROM, so it is a follow-up, recorded in the file's own comment.

andrewboudreau:
Re-signing off after the restack: `origin/main` had moved to #2196 (`daObjRc_Dorifu_c`'s destructor), so this branch now carries a clean merge of it (`48907e929`).

Re-proved on the merge, not just on the pre-merge head:
- `rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0**. ROM data from source went 612 → **614** verified, picking up #2196's two.
- "PR validation" passed on `48907e929` itself.
- `premerge_check.py --repo . 2197` on the new merge tree: every gate `ok`; `source-coverage` still errors identically on base and merge with the pre-existing `git ls-tree ... not a tree object` artifact, and its `entries: 10556 -> 10549 (-7)` is the eight-into-one consolidation, bytes flat.

PR #2198, opened 2026-09-03
ov032/daBakubaku_c: promote the whole class to one translation unit (23/23)

andrewboudreau:
## ov032/daBakubaku_c becomes one translation unit

Twenty-three separate one-function sources for the contiguous ov032 `.text` run
`0x021111a0..0x0211244c` collapse into a single production file,
`src/actors/d_a_bakubaku.cpp`, and `config/arm9/overlays/ov032/delinks.txt` goes
from twenty-three entries to one. The ROM build now takes the whole class from a
single object, the way the cartridge's own build made it.

### Proof

- `tubuild verify`: **23/23 MATCH**, objisolate clean, reloc destinations clean, emission order as expected.
- `tubuild partial`: 23/23 contribution-equivalent — one compile reduced 23 ways reproduces, byte for byte, the 23 objects it replaces.
- `tubuild linkcheck --module ov032`: **SCRATCH-LINK-VERIFIED**, ROM byte-identical to the stock build, `dsd check symbols` unchanged (9 pre-existing errors, 0 new).
- `tools/rombuild.py -j16`: **106/106 exact, 100.000000%, mismatching 0.**

### Two forced spellings, both explained in the file

**The destructors are mangled bodies, not a `~daBakubaku_c()` member.** The
cartridge puts D1 (`0x021111a0`) *below* D0 (`0x021111f0`), but mwccarm emits the
D2/D1/D0 group D0-first, so a real member definition makes the whole-range link
refuse with `licensed .text functions are not emitted in ROM address order`. Both
variants are therefore `// @symbol` bodies inside `extern "C"` blocks. That costs
those two functions the CONVERTED tier's no-mangled-refs and no-raw-offset
criteria, banked in `config/converted-backslide-exceptions.jsonl` with that reason;
The other 21 functions are unaffected.

`config/converted-baseline.json` moves five paths out and five in, and the count
reads flat at 2582 only because two of the additions are unrelated to this PR. Stated
honestly, this promotion **removes two members** (D0/D1) from the CONVERTED set and
**renames three** (`OnAimedAtWithEgg`, `OnPendingDestroy`, `Render`) from
`src/_ZN12daBakubaku_c*.cpp` to `src/actors/d_a_bakubaku.cpp#<symbol>`: net −2. The
other two additions are `src/actors/d_a_obj_rc_dorifu.cpp#_ZN16daObjRc_Dorifu_cD0Ev`
and `#…D1Ev`, gains `--update` banked opportunistically because the baseline on main
predates them. They are legitimate — that class now carries a real inline
`~daObjRc_Dorifu_c()` in its header and both members pass all five criteria today —
and the rows for them in `converted-backslide-exceptions.jsonl` are the historical
record of an earlier departure, not a standing exclusion. They are disclosed here
rather than stripped because additions never gate anything and hand-editing a set
ratchet is worse than carrying two extra true rows.

**The vptr stores use `&_ZTV12daBakubaku_c[2]`** — the *opposite* spelling to
#2184's `daObjWc_Mizu_c`, which needed the bare symbol. objisolate refused the bare
symbol here with `unexpected reloc type=2 addend=0`; `&_ZTV12daBakubaku_c[2]` cleared
it and both bodies matched.

The rule is in `tools/objisolate.py:472`: a `_ZTV` on the **externalise** list must
carry `addend >= VTABLE_PREAMBLE` (8), because that path converts mwccarm's
object-start `_ZTV` into the ROM's address point by subtracting 8. Mizu's object took
the *intact multi-symbol* path instead, which cannot rewrite an addend and so demanded
0 — a different refusal message. The spelling follows the isolation path, not the
class, and `daObjWc_Mizu_c` is not a relative of this one (`dBgActor_c` in ov029 vs
`dEnemyBase_c` in ov032), so nothing here should be read as a convention.

This does not contradict `config/arm9/overlays/ov032/relocs.txt` recording both
cartridge vptr loads as `to:0x02113824` addend 0, nor the legacy
`src/func_ov032_0211244c.c` storing that word with the bare symbol: those describe the
*cartridge's* relocation against a symbol that already **is** the address point
(`0x0211381c` = offset-to-top, `0x02113820` = `&_ZTI`, `0x02113824` = slot 0, verified
against the overlay image). The +8 is mwccarm's object-start bias, folded at compile
time and undone by isolation; both spellings land the same addend-0 word in the image.

### One byte-level correction worth noting

`func_ov032_02111e24` held `int ang;` where the ROM passes `Vec3_VertAngle`'s
result straight into `ApproachLinear` (`mov r1, r0`). Widening to `int` made mwcc
emit the `lsl`/`asr` sign-extension pair and the function grew by 4 bytes, which
also reordered its literal pool. `short ang;` is the ROM's shape.

### Header

`include/daBakubaku_c.h` keeps the class name and layout it already had (the tree
already spells this class as the ROM RTTI does, so no rename was needed); the
virtual destructor moves to be **declared last**, which is what makes mwccarm emit
`_ZTV`/`_ZTI`/`_ZTS` at all. `check_header_offsets`: 10 commented fields, 0
mismatched, struct spans `0x438`.

### Gates

`eligible.py`, `port_refcheck.py`, `check_dead_references.py`, `layout_check.py`,
`check_header_offsets.py --changed origin/main`, `check_references.py` (no new
unresolvable references), `tiers_ratchet.py --update --reason`,
`cpp_tu_state.py --write-note` — all run, all clean.

`check_references.py` reports "3 fixed since the baseline". Those three
(`_ZN5FaderD0Ev`, ov004 `table$6`, ov007 `overlay_100/102`) are **not** this PR's
doing — none is in ov032, and they were already fixed before this branch. The
baseline is deliberately left untouched.

### Follow-up, not done here

Several shadow declarations in `src/actors/d_a_bakubaku.cpp` duplicate things
`include/*.h` already carries. Each swap changes what the compiler knows about a
type and has to be re-proved against the ROM one at a time, so they are left as
the legacy files spelled them.


andrewboudreau:
Held in draft for the class-profile promotion queue. The current head handwrites D1/D0, stores the class vptr as _ZTV + 8, and stops before daBakubaku_c_classInit/profile ownership. This queue requires the destructor declared last with compiler-generated D1/D0, addend-0 public _ZTV relocations, and the named factory/profile folded into the TU. I am landing the older Snowman correction first, then this branch can be restacked and corrected.

andrewboudreau:
I checked new head 8d8b6fa79. It changes comments only and still has the held defects: handwritten D1/D0, source-level _ZTV + 8, no classInit/profile ownership, and two destructor backslide rows. The ov032 span is now claimed as clm_367dd0d49ea1 and a disjoint worktree is rebuilding the PR from current main with rename first, inline-last compiler-owned lifecycle, addend-0 production relocations, daBakubaku_c_classInit, and g_profile_BAKUBAKU. Please pause further pushes to this branch until that exact correction is ready, so we do not race on the same overlay.

andrewboudreau:
Sign-off (this repo's PRs cannot be APPROVED — shared account, so the review lives here).

A review pass found six issues on the earlier head. Four were real and are fixed in
`8d8b6fa79`; two were misreadings and are answered in the source rather than in a
comment, so the next reader does not have to re-derive them.

Fixed:

* The D1 comment claimed *"each member's size closes exactly on the next one's
  offset"*. Four of the five do. `ModelAnim` at `0x34c` is `0x64` bytes and closes at
  `0x3b0`, but `ShadowModel` is at `0x3b4`; the word between is `void *mState`, a
  field with no destructor — precisely the member a destructor cannot witness. The
  comment now says that and points at the header.
* `daObjWc_Mizu_c` was called a "sibling". It is not related: `dBgActor_c` in ov029
  against `dEnemyBase_c` in ov032.
* The vtable-spelling mechanism was asserted without its evidence. Now named:
  `tools/objisolate.py:472` requires `addend >= VTABLE_PREAMBLE` (8) for a `_ZTV` on
  the externalise list, which is the path this object takes; Mizu's intact
  multi-symbol path cannot rewrite an addend and demanded 0 — a different refusal
  message.
* The manifest's `_ZTV12daBakubaku_c` reason overstated `romdata_check`'s coverage.
  `0x02113824` is the address point; the emitted table starts two words earlier and
  those two words lie outside the symbol's extent. Recorded in the reason.

Answered, not changed:

* **`relocs.txt` addend 0 does not contradict `&_ZTV12daBakubaku_c[2]`.** That file
  records the *cartridge's* relocation against a ROM symbol that already **is** the
  address point. Verified in `extracted/overlays/overlay_0032.bin`: `0x0211381c` = 0
  (offset-to-top), `0x02113820` = `0x021137e4` (`&_ZTI`), `0x02113824` = slot 0. The
  `+8` in source is mwccarm's object-start bias, folded at compile time and undone by
  isolation; both spellings land the same addend-0 word in the image.
* **The two `d_a_obj_rc_dorifu.cpp` rows in `converted-baseline.json` are real gains,
  not a contradiction.** That class now carries a real inline `~daObjRc_Dorifu_c()`
  in its header and both members pass all five criteria today; the rows for them in
  `converted-backslide-exceptions.jsonl` are the historical record of an earlier
  departure, not a standing exclusion. `--update` banked them because the baseline
  predated them. They are disclosed in the PR body rather than stripped — additions
  never gate anything, and hand-editing a set ratchet is worse than carrying two
  extra true rows. The PR body no longer claims the count is "flat"; it states the
  real change (2 members leave CONVERTED, 3 are renamed).

The review confirmed everything else independently: all 23 banner addresses, every
struct offset, both destructor bodies call-for-call against the ROM, the factory's
`operator new(0x438)` + base ctor + vptr store + five forwards, `_ZTV`/`_ZTI`
addresses, all six override slots as the exact diff against `_ZTV12dEnemyBase_c`, all
twelve canonical homes, and the `short ang` byte-level finding.

Re-proved after every edit and again after merging main: `rombuild -j16` →
**106/106 exact, 100.000000%, mismatching 0**. `premerge_check.py 2198` → all eight
gates `pass`/`pass`, "nothing goes green -> red", with the informational
`entries: 10549 -> 10527 (-22)` that is the 23-into-1 consolidation. CI green
including PR validation (17m4s).

PR #2200, opened 2026-09-03
Promote ov052/daObjEmmLog_c to a single translation unit

andrewboudreau:
## What this does

Promotes `ov052/daObjEmmLog_c` — the rolling log of Tiny-Huge Island — to a single translation unit. `src/actors/d_a_obj_emm_log.cpp` replaces seven one-function legacy files and owns the whole `.text` range `0x021111a0..0x02111440`: the destructor pair, `CleanupResources`, `Render`, `Behavior`, `InitResources`, and the Spawn factory.

## The destructor is in the class body, and that is the whole promotion

The cartridge keeps D1 (`0x021111a0`) **below** D0 (`0x021111e4`). With an out-of-line `daObjEmmLog_c::~daObjEmmLog_c() {}` in the .cpp, mwccarm emits the D2/D1/D0 group in the order D0-then-D1 plus a homeless D2, and the whole-range link refuses:

```
emitted .text order of the licensed functions: NOT ROM-ascending
REFUSED -- licensed .text functions are not emitted in ROM address order
```

Written as `virtual ~daObjEmmLog_c() {}` in the class body it emits D1 then D0, the ROM's own order, and no D2 at all. Same lever, same reason, as `daObjRc_Dorifu_c`. It also makes the destructor this class's key function, so mwccarm homes `_ZTI/_ZTS/_ZTV13daObjEmmLog_c` here.

Both variants are compiler-generated; nothing in the file defines D1 or D0. The source body is genuinely empty — this class adds no member with a destructor (`mSpinAngle`/`mBasePosY`/`mBobAmplitude` are plain integers), and the two vptr stores and three destructor calls in the ROM are all base work, `dBgActor_c`'s own destructor being inline per `include/dBgActor_c.h`.

## The vptr spelling

`&_ZTV13daObjEmmLog_c[2]`. The bare symbol is refused with `unexpected reloc type=2 addend=0`, which is `tools/objisolate.py:472` saying this object took the **externalise** path — that path turns mwccarm's object-start `_ZTV` into the ROM's address point by subtracting `VTABLE_PREAMBLE` (8), so it requires `addend >= 8`. Other classes take the intact-multi-symbol path and demand the bare symbol instead (#2184's `daObjWc_Mizu_c`, refused with a different message). Neither spelling is a convention; the two refusal texts are the tool saying which path your object took.

Either way the emitted relocation is addend 0, and `ov052:0x02112520` already **is** the address point. Read out of `extracted/overlays/overlay_0052.bin` at load base `0x021111a0`:

| address | word | meaning |
| --- | --- | --- |
| `0x02112518` | `0x00000000` | offset-to-top |
| `0x0211251c` | `0x021124e0` | `&_ZTI13daObjEmmLog_c` |
| `0x02112520` | `0x02111348` | slot 0 — `InitResources` |

## Render is no longer a shadow struct

`tubuild create` carried the legacy file's `struct Base`/`struct Derived` pair verbatim. The ROM's call is the inherited `Model` at `+0xd4` dispatched through its own vtable slot 5, which `include/Model.h` records as `Render(const Vector3 *)`, so the body is now `mModel.Render(0)` and both shadow structs are gone. The only declaration this file still owns is the three-pointer resource block at `ov052:0x021124d4`, which no header names.

## Manifest

Eleven vague-linkage data passengers are licensed as `deadstrip-data`, each with the cartridge address dsd supplies independently: the `_ZTI`/`_ZTS` pairs for `fBase_c`, `dBase_c`, `dActor_c` and `dBgActor_c`, plus this class's own `_ZTI` (`ov052:0x021124e0`), `_ZTS` (`0x021124ec`) and `_ZTV` (`0x02112520`).

The `_ZTV` entry records the scope of what `romdata_check.py` actually compares: the symbol is the address point, so the two words *before* it — the offset-to-top and the `&_ZTI` — lie outside its extent and are not covered by that comparison. The function slots from `0x02112520` up are.

No `compiler_only_output` entry for a `D2`: the in-class destructor emits none.

## Proof

- `tubuild verify ov052/daObjEmmLog_c` — **7/7 MATCH**, objisolate clean, reloc destinations clean, emission order ROM-ascending.
- `tubuild linkcheck ov052/daObjEmmLog_c` — **SCRATCH-LINK-VERIFIED**. Whole range reproduces from one object, every module byte-exact, ROM built and **identical** to the stock build (`d1506e90…`). `dsd check symbols --fail` still fails on the same 9 pre-existing errors the baseline control has: 0 new, 0 resolved — not green, and not attributable to this TU.
- `rombuild.py -j16` — **106/106 exact, 100.000000%, mismatching 0**, run again after the Render and banner edits.

`config/converted-baseline.json` moved 2585 → 2586 through `tools/tiers_ratchet.py --update`. The set delta is **+4 / −3**, not +1 / −0: the three legacy whole-file rows (`src/_ZN13daObjEmmLog_c6RenderEv.cpp`, `…D0Ev.cpp`, `…D1Ev.cpp`) leave, and four `src/actors/d_a_obj_emm_log.cpp#…` member rows (CleanupResources, Render, D0, D1) arrive. Net count +1. No exception rows.

## Review corrections

An adversarial review pass found eight prose defects, all now fixed and all re-proved with a fresh `rombuild -j16` 106/106:

1. **`Render` was labelled vtable slot 4; it is slot 9.** Read back from `overlay_0052.bin`: `0x02112544` holds `0x02111284`. Slot 4 (`0x02112530`) is `0x02011220`, an arm9 method this class does not override. Corroborated by `include/dActor_c.h` (`BeforeCleanupResources` slot 4, `BeforeRender` slot 10) and by ~14 sibling headers that all spell `virtual Render(); /* slot 9 */`. The neighbouring slot claims — InitResources 0, CleanupResources 3, Behavior 6, D1/D0 16/17 — were checked and are correct.
2. **A leftover `tubuild create` block said the definition is "written at the bottom of this file".** It is not in this file at all; it is in the header, pulled in at the top. Rewritten.
3. **The same block said "declaring `~daObjEmmLog_c()` is enough" and "mwcc emits D2, D0 and D1".** Both contradict this PR's own thesis — a bare declaration is exactly what the link refuses, and the in-class definition emits no D2. It also claimed objisolate keeps "the one this file is bound to"; this file is licensed for both variants.
4. **`include/daObjEmmLog_c.h` still called the destructor "out-of-line"** in a sentence directly above the declaration this PR changed to in-class.
5. **"The only declaration this file still owns is the resource block"** ignored the `extern "C"` block below it, which also declares the arm9 sine table `data_02082214`, `func_020393a4`, the two allocator entry points, and `dBgActor_c::IsClsnInRange` (which `include/dBgActor_c.h` mentions in a comment but does not declare). Reworded to enumerate them.
6. **The D1 comment had the member destruction order backwards.** The ROM destroys `dBgW_KcMbg` at `+0x124` first (`0x021111b4: add r0, r4, #0x124`) and `Model` at `+0xd4` second (`0x021111c0`). The header stated it correctly, so the two files disagreed.
7. **The header cited a symbol name `func_ov052_02111410` that does not exist**; `symbols.txt` names it `daObjEmmLog_c_Spawn`. Pre-existing, fixed in passing.
8. **The baseline delta in this body** said +1 / −0; the set delta is +4 / −3 (corrected above).

The review independently reproduced and confirmed the technical core: all seven addresses and sizes, all eleven canonical RTTI addresses and modules, the vtable address point read-out, the `Model::Render` slot-5 / `+0xd4` dispatch, the `romdata_check.py` extent reasoning (`linked = linked[preamble:]` at :250 really does leave the two preamble words uncompared while comparing the full emitted slot array), and every layout claim (0x31e / 0x320 / 0x324 / 0x328) against the ROM bodies.


andrewboudreau:
Sign-off (in a comment, since no PR in this repo can carry an APPROVED review on the shared account).

**Local proof, this branch at `2411d9c626fe`:**
- `tubuild verify ov052/daObjEmmLog_c` — 7/7 MATCH, objisolate clean, reloc destinations clean, emission order ROM-ascending.
- `tubuild linkcheck ov052/daObjEmmLog_c` — SCRATCH-LINK-VERIFIED, ROM identical to the stock build, `dsd check symbols --fail` at the baseline's own 9 pre-existing errors (0 new, 0 resolved; not green, and not this TU's).
- `rombuild.py -j16` — 106/106 exact, 100.000000%, mismatching 0. Run three times: after promotion, after the `Render` reconciliation and banner rewrite, and again after the review corrections.
- `premerge_check.py 2200 --repo … --fetch` against `origin/main@0687255f33f5` — all eight gates pass/pass, nothing goes green → red.

**CI:** all nine checks pass, including PR validation (6m10s).

**Review:** an adversarial pass found eight prose defects and reproduced the technical core independently. All eight are fixed in `2411d9c6`. The one that mattered was a wrong vtable slot newly written into the source — `Render` labelled slot 4 when `overlay_0052.bin` puts `0x02111284` at slot 9 — plus a leftover `tubuild create` block that asserted a D2 group and a bottom-of-file definition the rest of the PR explicitly denies. The review confirmed all seven addresses and sizes, all eleven canonical RTTI addresses and modules, the vtable address-point read-out, the `Model::Render` slot-5 / `+0xd4` dispatch, the `romdata_check.py` extent reasoning, and every layout claim against the ROM bodies.

**Known-imperfect, disclosed:** `dsd check symbols --fail` is not green anywhere in this tree; the `_ZTV` romdata comparison does not cover the two words before the address point (recorded in the manifest reason); and the deleted legacy paths survive in `attribution.json`, `config/match_attempts.jsonl`, two `notes/data/*` shards, the profile registry and the generated `docs/*.html` — the same shape every previously promoted TU leaves, `check_dead_references.py` clean.

PR #2201, opened 2026-09-03
Promote ov018/daPgMthr_c to a single TU (src/actors/d_a_pg_mthr.cpp)

andrewboudreau:
Consolidates the 22 legacy per-function sources for the Mother Penguin actor
into one translation unit, `src/actors/d_a_pg_mthr.cpp`, covering ov018 `.text`
0x02111848-0x0211267c.

## Identity comes from the cartridge

The class already carries the ROM's own RTTI name, so this promotion has no
rename step:

| record | address | contents |
| --- | --- | --- |
| `_ZTS10daPgMthr_c` | 0x02113988 | `10daPgMthr_c\0` |
| `_ZTI10daPgMthr_c` | 0x0211397c | base word 0x0208e390 = `_ZTI8dActor_c` |
| `_ZTV10daPgMthr_c` | 0x021139bc | address point; slots 0..30 |

## Destructor defined in the class body

The cartridge keeps D1 (0x02111848) **below** D0 (0x02111898). An out-of-line
member definition makes mwccarm emit the group D0-then-D1 plus a homeless D2,
and the whole-range link refuses that with `licensed .text functions are not
emitted in ROM address order`. Written `virtual ~daPgMthr_c() {}` in the class
body, the group comes out ROM-ascending and no D2 is emitted at all, while the
destructor stays the first virtual declared and so remains the key function.

## Matrix4x3 spelling

`common.h` is included **first**, deliberately. `Matrix4x3` has two 0x30-byte
spellings under one guard. `func_ov018_02111d28` copies a whole matrix, and the
ROM copies it as twelve uniform words (`ldm`/`stm` x3) -- the flat
`s32 m[12]` spelling. The nested `{Matrix3x3 r; Vector3 t;}` spelling in
`include/math/Matrix.h` scalarizes the last three words into `ldr`/`str` pairs
and costs an extra live register, which is what the byte diff showed.

## Corrections made along the way

- `func_ov018_021122ec` was labelled `daObjSm_Lift_c::AfterClsn`. Every offset
  it touches (+0xd4, +0x130, +0x138, +0x98, +0x374, +0x37c) is a `daPgMthr_c`
  offset; re-attributed here.
- `include/decl_common.h`: `func_ov018_02111bf0` returns `void`, not `int` --
  ov018 0x02111d18..0x02111d24 pops straight to `bx lr` with no r0 set. All
  three files that referenced it are absorbed by this promotion.
- `_ZN8dActor_c13ClosestPlayerEv` takes `this`; the mangled `Ev` describes the
  explicit parameter list only. `func_ov018_02112234` shows `mov r0, r4` before
  the `bl`.

## Proof

| gate | result |
| --- | --- |
| `tubuild verify` | 22/22 MATCH, objisolate clean, reloc destinations clean, all 22 functions in ROM-ascending section order |
| `tubuild linkcheck` | SCRATCH-LINK-VERIFIED; linked ROM byte-identical to the stock build; 0 new `dsd check symbols` errors (9 pre-existing, same 9 as the baseline control) |
| `rombuild.py -j16` | 106/106 exact, 100.000000%, mismatching 0 |

## Known-imperfect, disclosed

The ten vague-linkage records this TU emits (six base-class RTTI records, this
class's `_ZTI`/`_ZTS`/`_ZTV`, and `Vector3`'s D1) are licensed in the manifest
as `compiler_only_output` with their cartridge homes. The `_ZTV` entry records
that `romdata_check`'s comparison is **twice** limited:

1. it is anchored at the address point 0x021139bc, so the two preamble words
   (0x021139b4 offset-to-top, 0x021139b8 `&_ZTI`) are never compared; and
2. `config/arm9/overlays/ov018/symbols.txt` records `data_ov018_021139f4`
   *inside* the table, cutting the symbol's extent short, so only slots 0-13
   are word-compared. Slots 14-30, up to the table's real end at 0x02113a38,
   are not covered.

`dsd check symbols --fail` is **not** green: it reports the same 9 pre-existing
errors as the baseline control (0 new, 0 resolved).


andrewboudreau:
## Sign-off

No PR in this repo can be APPROVED (shared account), so this comment is the review record.

### What was proved

| gate | result |
| --- | --- |
| `tubuild verify ov018/daPgMthr_c` | **22/22 MATCH**, objisolate clean, reloc destinations clean, all 22 functions in the expected ROM-ascending section order |
| `tubuild linkcheck ov018/daPgMthr_c` | **SCRATCH-LINK-VERIFIED**; the whole `.text` run 0x02111848..0x0211267c reproduces from one object; linked ROM **byte-identical** to the stock build |
| `rombuild.py -j16` | **106/106 exact, 100.000000%, mismatching 0** -- run twice, once after promotion and again after the banner rewrite |
| `romdata_check.py --files src/actors/d_a_pg_mthr.cpp` | 9 data records: 5 VERIFIED (168 bytes equal), 4 PARTIAL (36 bytes equal), **0 DIFFERS** |
| `premerge_check.py 2201 --repo . --fetch` | 7 gates pass/pass; nothing goes green -> red |
| CI | 10 checks green + `PR validation` **pass** (24m44s) |

### Two byte-level findings, both worth carrying forward

1. **`_ZN8dActor_c13ClosestPlayerEv` takes `this`.** The `Ev` suffix describes the *explicit* parameter list only; the implicit `this` is never in an Itanium mangling. `func_ov018_02112234` shows `mov r0, r4` immediately before the `bl`, and the call in `func_ov018_02111b3c` needs no such `mov` only because r0 already holds the actor. Declaring it `(void)` cost one instruction and shifted the whole tail.
2. **`Matrix4x3`'s two spellings are not interchangeable in codegen.** The ROM copies a whole matrix as twelve uniform words (`ldm`/`stm` x3). The nested `{Matrix3x3 r; Vector3 t;}` spelling in `include/math/Matrix.h` scalarizes the last three words into `ldr`/`str` pairs and needs an extra live register -- which is why the object pushed `{r4,r5,r6,lr}` where the cartridge pushes `{r4,r5,lr}`. `common.h` is now included **first** in this TU, deliberately, to get the flat `s32 m[12]` spelling. Both spellings are 0x30 bytes, so nothing else in the tree moves.

### One misattribution corrected

`func_ov018_021122ec` was labelled `daObjSm_Lift_c::AfterClsn`. Every offset it touches (+0xd4, +0x130, +0x138, +0x98, +0x374, +0x37c) is a `daPgMthr_c` offset; the body sets this actor's model animation, texture sequence and two state words. Re-attributed here and rewritten with raw offsets.

### Known-imperfect, disclosed rather than papered over

- **`_ZTV10daPgMthr_c` is only partly compared.** `romdata_check`'s comparison is anchored at the address point 0x021139bc, so the two preamble words (0x021139b4 offset-to-top, 0x021139b8 `&_ZTI10daPgMthr_c`) are never compared. On top of that, `config/arm9/overlays/ov018/symbols.txt` records `data_ov018_021139f4` *inside* the table, cutting the symbol's extent short, so only slots 0-13 are word-compared. Slots 14-30, up to the table's real end at 0x02113a38, are not covered by any check. This is recorded in the manifest entry's `compiler_only_output` reason, not just here.
- **`dsd check symbols --fail` is NOT green.** It reports 9 errors. The baseline control reports the same 9 (0 new, 0 resolved), so they are the tree's, not this PR's -- but the check is red and should not be reported as passing.
- **`source-coverage` errors on both base and merge** (`git ls-tree ... fatal: not a tree object`). Pre-existing, not this PR's to own, and worth someone's attention.
- **18 `TUBUILD CONFLICT` markers remain** in the promoted source, recorded as open conflicts in the manifest and matched by `check_tubuild_conflicts.py`. They are alternate declarations from the legacy files that were not applied; each is a small future reconciliation, not a defect. The two `struct dActor_c` shadow conflicts *were* reconciled (the real `include/dActor_c.h` wins; the shadow spelled the actor id field `kind`, the header calls it `actorID` at 0x00c) and their notes are rewritten as RESOLVED.

### Note on the goal's recipe

Two of the four steps in the standing recipe did not apply here, for reasons the cartridge decides:

- **"rename first"** is a no-op: 0x02113988 already holds `10daPgMthr_c`, so the class carries the ROM's own name.
- **"destructor declared last"** would break this class. D1 (0x02111848) sits *below* D0 (0x02111898), so an out-of-line member definition makes mwccarm emit the group D0-then-D1 plus a homeless D2 and the whole-range link refuses it. `virtual ~daPgMthr_c() {}` in the class body emits the group ROM-ascending with no D2 -- and, being the first virtual declared, keeps the destructor as the key function.
- **"vptr via addend-0 `_ZTV`"** does not arise: there is no explicit `_ZTV` reference in this source at all. The vptr stores are the ones the compiler generates for that in-class destructor.

PR #2203, opened 2026-09-03
Promote ov006/dScMgJump2_c to a single translation unit

andrewboudreau:
Consolidates the 29 legacy per-function sources for `ov006/dScMgJump2_c` -- the Jump minigame scene class, named as the cartridge's own RTTI spells it (`0x0213ccbc` holds the string `12dScMgJump2_c`) -- into one translation unit at `src/actors/d_sc_mg_jump2.cpp`, covering the whole `.text` run `0x020eebe8..0x020efc08`.

## What was interesting

**The destructor had to move into the class body.** `D1` sits at `0x020eebe8`, *below* `D0` at `0x020eec9c`. With an out-of-line member definition mwccarm emits the group as D0-then-D1 plus a homeless D2, and a single TU spanning the run cannot be linked -- the functions are not in ROM address order. Defining it in the class body emits D1-then-D0 and emits no D2 at all. This is the second class where that lever was needed (the first was `daPgMthr_c`, #2201).

Moving it in-class gives up the destructor as the key function, so the key function becomes the first virtual *declared but not defined inline* -- `OnYoshiTryEat` at `ov006:0x020efaa8`, which lives only in this TU. `_ZTV12dScMgJump2_c` is therefore still emitted by exactly one object.

**The vptr store spells `&_ZTV12dScMgJump2_c[2]`.** The bare symbol was refused with `_ZTV12dScMgJump2_c: unexpected reloc type=2 addend=0` (`tools/objisolate.py:472`). Both spellings land addend 0 in the image; only which isolation path the object takes differs.

**Three DIFFs were a struct-copy scheduling mismatch.** A two-word struct written with two named int fields compiles to interleaved `ldr/str/ldr/str`; the ROM emits `ldr/ldr/str/str`, which is what a struct whose member is `int w[2]` produces. Switching the type fixed all three at once.

## Manifest

A 13-entry `compiler_only_output` block licenses the class's own `_ZTI`/`_ZTS`/`_ZTV` plus the ten inherited base-RTTI `deadstrip-data` records (`fBase_c`, `dBase_c`, `dScene_c`, `dScMgBase_c`, `dScMgD3DBase_c`). Unlike `ov018/_ZTV10daPgMthr_c`, this vtable's extent is not truncated: the next symbol is `data_ov006_0213cd8c`, so `0x0213ccfc..0x0213cd8c` is `0x90` bytes -- all 36 slots are compared. The two preamble words (`0x0213ccf4` offset-to-top 0, `0x0213ccf8` `&_ZTI`) fall inside `data_ov006_0213cce0` and are not word-compared; that limitation is stated in the entry.

Eight of the twelve `tubuild create` CONFLICT notes are marked `(RESOLVED)` and their markers removed -- the TU now carries one declaration per symbol, typed to what the ROM bytes require, with every call site casting to it. Four remain open.

## Gates

| gate | result |
| --- | --- |
| `tubuild verify` | 29/29 MATCH, objisolate clean, reloc destinations clean, all 29 in ROM-ascending order |
| `tubuild linkcheck` | SCRATCH-LINK-VERIFIED; ROM byte-identical to the stock control |
| `rombuild.py -j16` | 106/106 exact, 100.000000%, mismatching 0 -- run again after `tu_promote` |
| `romdata_check.py --files` | 7 VERIFIED / 6 PARTIAL / 0 DIFFERS |
| `check_tubuild_conflicts.py` | 4 open notes, 4 markers, 0 problems |
| `check_dead_references.py` | no new dead references, no broken links |

**Known imperfect, disclosed:** `dsd check symbols --fail` is **not** green -- it still reports the same 9 pre-existing baseline errors the control has. Zero are new, but the command does not pass. The `_ZTV` comparison is limited to the slot words as described above.


andrewboudreau:
Sign-off (this account cannot use GitHub's Approve, so the review is here).

**Proved on the merge tree, not just the branch.** `tools/premerge_check.py 2203` against `origin/main @ 3b0a6aa03c61`, merge tree `3c9ea742e075`: converted-ratchet, dead-references, duplicate-sources, header-offsets, langmode-ratchet, layout-check, src-tu-refs and source-coverage all pass on both sides; nothing goes green -> red. `source-coverage` reports `entries: 10500 -> 10472 (-28)` -- that is the 29 legacy sources collapsing into one, bytes flat.

**Byte proof.** `tubuild verify` 29/29 MATCH with objisolate and reloc destinations clean and all 29 functions in ROM-ascending section order; `tubuild linkcheck` SCRATCH-LINK-VERIFIED with the built ROM byte-identical to the stock control; `rombuild.py -j16` 106/106 exact, 100.000000%, mismatching 0 -- run after `tu_promote`, again after the CONFLICT-note edits, again after the merge with main, and again after the header brace fix. `romdata_check.py --files` 7 VERIFIED / 6 PARTIAL / 0 DIFFERS.

**Two findings worth carrying forward.**

1. The D1-below-D0 emission-order wall is now confirmed on a second class. `dScMgJump2_c::~dScMgJump2_c` at `0x020eebe8` sits below `D0` at `0x020eec9c`; out-of-line, mwccarm emits D0-then-D1 plus a homeless D2 and the whole-range link refuses it. In-class emits D1-then-D0 and no D2. The key function moves to `OnYoshiTryEat` (`ov006:0x020efaa8`), still defined in exactly one object.
2. A two-word struct's *field spelling* changes scheduling in a whole-struct copy. Two named `int` fields compile to interleaved `ldr/str/ldr/str`; `int w[2]` produces the ROM's `ldr/ldr/str/str`. That one type change fixed three DIFFs at once.

**One self-caught regression, disclosed.** The first push took `header-offsets` green -> red: the parser cannot read a member whose opening brace is on its own line, so it reported `UNPARSED include/dScMgJump2_c.h:55: {` and computed the struct span as `0x5004` instead of `0x5a78`. Fixed by putting the brace on the signature line (`f1c734f04`), and re-proved -- 9 commented fields, 0 mismatched, 0 unparsed, rombuild still 106/106.

**Known imperfect, not hidden.** `dsd check symbols --fail` is **not** green: it reports the same 9 pre-existing baseline errors the control has. Zero new, but the command does not pass, and I am not claiming it does. The `_ZTV12dScMgJump2_c` comparison covers all 36 slots (`0x0213ccfc..0x0213cd8c`, `0x90` bytes) but not the two preamble words, which fall inside `data_ov006_0213cce0`. Four `tubuild create` CONFLICT notes remain open (`GetGameLanguage`, `Matrix4x3_FromTranslation`, `data_ov006_021421b8`, `func_ov004_020adb1c`) -- the alternates differ only in spelling, but I did not prove that, so they stay recorded.

PR #2209, opened 2026-09-03
ov015/daObjBk_Ukisima_c: rename off the coined name and promote to one TU

andrewboudreau:
`RotatingPlatformWf` was a coined name. The cartridge spells the class
**`daObjBk_Ukisima_c`**, read out of the ov015 image (base `0x021111a0`):

| addr | record |
|---|---|
| `0x021147b0` | `_ZTS17daObjBk_Ukisima_c` -- `"17daObjBk_Ukisima_c\0"`, 20 bytes |
| `0x02114798` | `_ZTI17daObjBk_Ukisima_c` -- `__si_class_type_info`, base pointer `0x021091ac` = `_ZTI16daObjKaitendai_c` (ov002), so the cartridge itself states the direct base |
| `0x021147e8` | `_ZTV17daObjBk_Ukisima_c` -- the address point: `V-8` = 0 offset-to-top, `V-4` = `&_ZTI`, `V+0` = `0x02112c98` (InitResources, slot 0), `V+12` = `0x02112c84` (CleanupResources, slot 3) |

Renamed first, then promoted the four per-function sources -- D1, D0,
`CleanupResources` (slot 3), `InitResources` (slot 0) -- into one TU,
`src/actors/d_a_obj_bk_ukisima.cpp`. This is the **immediate ROM neighbour** of
`ov015/daObjBk_Fall_Block_c` (#2206, `.text` starts at `0x02112cf4` where this
one ends) and has the identical four-function shape: both non-destructor methods
are one-line delegations to `daObjKaitendai_c`'s shared ov002 helpers over a
per-level descriptor at `ov015:0x021147a4`.

**The destructor is inline in the header on purpose.** Out of line, mwcc emits D0
ahead of D1 and the ROM has D1 first -- rombuild refuses the object outright. The
inline body gives the cartridge's order and drops the D2 variant the cartridge
never carried. The consequence is that **`linkcheck --partial` cannot run**: the
two legacy destructor sources fail to compile with
`object 'daObjBk_Ukisima_c::~daObjBk_Ukisima_c()' redefined`. The proof used
instead is `linkcheck --baseline` after promotion, which is the stronger form.

Only `CleanupResources` carries `#pragma long_calls` (bracketed on/off, positional),
exactly as its legacy file had it; `InitResources` deliberately carries none, as
its legacy file had none. All 13 vague-linkage records are banked as
`compiler_only_output` with measured ROM homes.

### Verification

[...]

PR #2210, opened 2026-09-03
ov036/daObjRc_Kaitendai_c: rename off the coined name and promote to one TU

andrewboudreau:
`RotatingPlatformRr` was a coined name. The cartridge spells the class
`daObjRc_Kaitendai_c` -- Rainbow Ride's spinning disc. Renamed off the coined
name first, then promoted its four per-function sources into a single TU.

## The name comes from the image, not from a guess

ov036 loads at base 0x021111a0 (`extracted/dsd/arm9_overlays/overlays.yaml`),
image `extracted/overlays/overlay_0036.bin`:

| address | what |
|---|---|
| 0x02113b38 | `_ZTS` -- the bytes spell `19daObjRc_Kaitendai_c\0` |
| 0x02113b20 | `_ZTI` -- `__si_class_type_info`, base pointer 0x021091ac = `_ZTI16daObjKaitendai_c` (ov002) |
| 0x02113b74 | `_ZTV` -- the address point, addend 0 |

Reading out from the address point: `V-8` = 0 (offset-to-top), `V-4` = `&_ZTI`,
`V+0` = 0x0211150c (`InitResources`, slot 0), `V+12` = 0x021114f8
(`CleanupResources`, slot 3). So the base clause in the header is what the
cartridge states, not an import.

## Shape

The same four-function shape as the already-landed sibling
`ov015/daObjBk_Ukisima_c` (#2209): D1, D0, `CleanupResources` (slot 3),
`InitResources` (slot 0), where both non-destructor methods delegate to
`daObjKaitendai_c`'s shared ov002 helpers (`func_ov002_020b676c` setup,
`func_ov002_020b66a8` teardown) with a per-level descriptor. The one difference
is that this disc picks between two start angles on `param1`.

`#pragma long_calls` brackets `CleanupResources` only, on/off exactly as its
legacy file had it; `InitResources` deliberately carries none. The externs are
kept at the legacy `void *` spelling rather than being tightened, because that
spelling is what reproduces.

Out of line, mwcc emits D0 ahead of D1 and the ROM has D1 first -- rombuild
refuses the object outright. The inline body gives the cartridge's order and
drops the D2 variant the cartridge never carried. The header banner says so in
place so nobody "fixes" it later.

A consequence worth stating: with the destructor inline, `linkcheck --partial`
cannot run here at all -- the legacy per-function destructor sources fail with
`object 'daObjRc_Kaitendai_c::~daObjRc_Kaitendai_c()' redefined`. The proof
cited below is `linkcheck --baseline` run *after* promotion, which is the
stronger of the two forms.

## Verification

- `tubuild verify`: **4/4 MATCH**, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
- `tubuild linkcheck --baseline`: ok
- `tools/rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes, mismatching 0**
- ratchet: neutral (+4 / -4 -- the four are a move into a CONVERTED file, not a removal)
- 13 `compiler_only_output` rows, every one with a configured ROM home (so none is a plain deadstrip claim); no `unlicensed_output_observed`
- static gates: references, dead-references, duplicate-sources, rename-ledger, data-definitions, src_tu, tubuild-conflicts, header-offsets (struct spans 0x320) all clean

## Plain English

Rainbow Ride has a big spinning disc you ride across the level. Its code was
sitting in the repo under a made-up name and split into four separate files,
one per function. The game cartridge itself actually stores the real class name
as text, so this reads that name straight out of the cartridge image, renames
everything to match, and glues the four files back into the one file the
original programmers would have written. The rebuilt game is still byte-for-byte
identical to the cartridge.


andrewboudreau:
Sign-off (no APPROVE available on this shared account, so it goes here).

Checked:
- The name is image-proved. ov036 base 0x021111a0; `_ZTS` at 0x02113b38 spells
  `19daObjRc_Kaitendai_c`, `_ZTI` at 0x02113b20 carries base pointer 0x021091ac
  = `_ZTI16daObjKaitendai_c` (ov002), and `_ZTV` at 0x02113b74 is the address
  point with addend 0: V-8 = 0, V-4 = &_ZTI, V+0 = InitResources, V+12 =
  CleanupResources. The header's base clause is the cartridge's, not an import.
- The `.text` extent is `tubuild`'s (0x02111444..0x02111544); `<class>_classInit`
  is a separate C file and stays out.
- Inline destructor is deliberate, banner in place. `linkcheck --partial` is
  therefore impossible here, and the cited proof is `linkcheck --baseline` after
  promotion, which is stronger.
- The externs are left at the legacy `void *` spelling rather than tightened.
  That is the spelling that reproduces, and tightening it is a separate change.
- All 13 `compiler_only_output` rows name a configured ROM home, so none is a
  plain deadstrip claim; no `unlicensed_output_observed`.
- The rebase over #2209 was resolved the right way round: the two conflicting
  comment blocks in include/daObjKaitendai_c.h and
  src/_ZN16daObjKaitendai_cD1Ev.cpp took BOTH renames (daObjBk_Ukisima_c and
  daObjRc_Kaitendai_c), not one side's; the generated files were restored from
  main and regenerated rather than hand-merged.
- Gates: tubuild verify 4/4 MATCH (objisolate clean, reloc-destinations clean),
  rombuild 106/106 exact / 100.000000% / mismatching 0 re-run AFTER the rebase,
  ratchet neutral, premerge_check on the merge tree all-ok, CI all green.

No blockers. Merging.

PR #2211, opened 2026-09-03
ov021/daObjCvShutter_c: rename off the coined name and promote to one TU

andrewboudreau:
`ShutterHmc` was a coined name. The cartridge spells the class
`daObjCvShutter_c` -- Hazy Maze Cave's switch-operated shutter, `cv` being the
ROM's own tag for that stage. Renamed off the coined name first, then promoted
its five per-function sources into a single TU.

## The name and the base class both come from the image

ov021 loads at base 0x021111a0 (`extracted/dsd/arm9_overlays/overlays.yaml`),
image `extracted/overlays/overlay_0021.bin`:

| address | what |
|---|---|
| 0x021148dc | `_ZTS` -- the bytes spell `16daObjCvShutter_c\0` |
| 0x021148c4 | `_ZTI` -- `__si_class_type_info`, base pointer 0x021099c0 = `_ZTI13daObjSwdoor_c` (ov002) |
| 0x02114914 | `_ZTV` -- the address point, addend 0 |

Reading out from the address point: `V-8` = 0 (offset-to-top), `V-4` = `&_ZTI`,
slot 0 = 0x02112ec0 (`InitResources`), slot 3 = 0x02112e68
(`CleanupResources`), slot 6 = 0x02112e7c (`Behavior`), slots 16/17 =
0x02112db4 / 0x02112e04 (D1 then D0).

The base clause is worth stating plainly: **this is not a direct `dBgActor_c`.**
The cartridge names `daObjSwdoor_c` as the direct base, which is exactly why the
destructor stores *three* vptrs where a one-level chain stores two, and why the
three overridden slots are precisely the three `daObjSwdoor_c` leaves null.

## The TU

Five functions, `.text 0x02112db4..0x02112ed4`, written in the reverse of ROM
order because mwccarm 2004/b56 emits sections in the reverse of source order.
`daObjCvShutter_c_classInit` is a separate C file and stays out of the TU --
the extent is `tubuild`'s, not the wider span the candidate table lists.
`InitResources` and `CleanupResources` hand the actor and this level's
descriptor to `daObjSwdoor_c`'s shared ov002 helpers; `Behavior` runs the shared
shutter step and then re-syncs the collision mesh when it is in range. Externs
are kept at their legacy spelling.

Out of line, mwcc emits D0 ahead of D1 and the ROM has D1 first -- rombuild
refuses the object outright. The inline body gives the cartridge's order and
drops the D2 variant the cartridge never carried. The header banner says so in
place so nobody "fixes" it later.

A consequence worth stating: with the destructor inline, `linkcheck --partial`
cannot run here at all -- the legacy per-function destructor sources fail with
`object 'daObjCvShutter_c::~daObjCvShutter_c()' redefined`. The proof cited
below is `linkcheck --baseline` run *after* promotion, which is the stronger of
the two forms.

## Verification

- `tubuild verify`: **5/5 MATCH**, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
- `tubuild linkcheck --baseline --module ov021 -j16 --clean`: ok, and the linked ROM is byte-identical to the stock build
- `tools/rombuild.py -j16`: **106/106 exact, 100.000000% of compared bytes, mismatching 0**
- ratchet: neutral (+4 / -4 -- a move into a CONVERTED file, not a removal)
- 13 `compiler_only_output` rows, every one with a configured ROM home (so none is a plain deadstrip claim); no `unlicensed_output_observed`
- static gates: references, dead-references, duplicate-sources, rename-ledger, data-definitions, src_tu, tubuild-conflicts, header-offsets (struct spans 0x324) all clean

## Plain English

Hazy Maze Cave has a shutter door that a switch opens. Its code was in the repo
under a made-up name, split across five separate files, and the notes had it
descending from the wrong parent class. The game cartridge stores the real class
name and the real parent as data, so this reads both straight out of the
cartridge, renames everything to match, and glues the five files back into the
one file the original programmers would have written. The rebuilt game is still
byte-for-byte identical to the cartridge.


andrewboudreau:
Sign-off (this repo cannot record a formal APPROVE -- shared account -- so it goes in a comment).

Re-verified after the rebase onto post-#2218 main:

- `rombuild.py -j16` -> module fidelity **106/106 exact, 100.000000%**, **mismatching 0**, ROM-build analysis PASS.
- `tubuild linkcheck --baseline --module ov021 --clean` -> modules PASS; the only baseline failure is the pre-existing tree-wide `dsd check symbols --fail`, which the control reproduces with no TU substitution.
- `premerge_check.py 2211 --repo --fetch` on the MERGE TREE (not the branch): every gate `pass -> pass`; converted entries `10391 -> 10387 (-4)`, which is the four consolidated members moving from per-function sources into the one TU, not a coverage loss.
- Ratchet regenerated with `tiers_ratchet.py --update` from an origin/main baseline, never hand-edited: `2592 -> 2592 (+4 / -4)`.
- The rebase's one conflict was `attribution.json`, resolved by UNION -- old `path#symbol` keys kept alongside the new ones, nothing pruned, nothing reordered.
- CI: all checks green, including `PR validation` (16m14s).

Merging.

PR #2215, opened 2026-09-03
ov015: promote daObjBkBillboard_c into a single C++ TU

andrewboudreau:
Renames the coined `PoleBillboard` to its ROM RTTI name `daObjBkBillboard_c` and folds its six per-function sources into one translation unit, `src/actors/d_a_obj_bk_billboard.cpp`, the way #2184 did `daObjWc_Mizu_c`.

## The name and the base come from the cartridge

Read out of `extracted/overlays/overlay_0015.bin` at ov015 base `0x021111a0`:

| address | contents |
| --- | --- |
| `_ZTS` `0x02114324` | `18daObjBkBillboard_c` |
| `_ZTI` `0x02114318` +8 | `0x0208e390` = `_ZTI8dActor_c` (arm9) — the **direct** base is `dActor_c` |
| `_ZTV` `0x02114360` | address point (addend 0); `V-8` = 0 offset-to-top, `V-4` = `&_ZTI` |

Vtable slots, against the addresses `config/arm9/overlays/ov015/symbols.txt` already carried:

| slot | address | method |
| --- | --- | --- |
| 0 | `0x021112a0` | `InitResources` |
| 3 | `0x02111254` | `CleanupResources` |
| 9 | `0x02111278` | `Render` |
| 16 / 17 | `0x021111a0` / `0x021111d0` | `~daObjBkBillboard_c` D1 / D0 |

## Destructor declared inline, on purpose

With an out-of-line body mwcc emits D0 ahead of D1, and the ROM has D1 first — `rombuild` refuses the object outright. The inline body also drops the D2 variant the cartridge never carried. The header carries the `MEASURED -- INLINE ON PURPOSE` banner, so nobody "fixes" it later.

Because the legacy per-function destructor sources are gone, `linkcheck --partial` is not meaningful here; the proof below is `linkcheck --baseline`, which is the stronger one.

## Two things worth reviewing

**Mixed-language TU.** `func_ov015_02111214` is a `.c`-derived member whose address falls inside this class's contiguous `.text` run (`0x021111a0..0x021112dc`), so it cannot be split out of the TU. `tubuild` carries it in its own `extern "C" { … }` block, which keeps C linkage and the unmangled ROM name. This is the first TU in the tree to do that; it byte-matches.

**A declaration conflict the two legacy files disagreed on.** `CleanupResources` declared the ROM object locally as `extern int data_ov015_02114960[]`; `InitResources` took `include/decl_common.h`'s `extern int data_ov015_02114960`. Both cannot stand in one TU. `decl_common.h`'s spelling wins — it is the project-wide declaration — and the call site takes `&data_ov015_02114960` explicitly, which is the same address the array spelling decayed to. The resolution is written into the source as a comment and recorded in the manifest, so `check_tubuild_conflicts` sees it. 6/6 MATCH says codegen is unchanged.

## Not renamed here

`PoleBillboard_ClsnFile` / `PoleBillboard_ModelFile` (ov015 bss `0x02114974` / `0x0211497c`) keep their coined names: they are shared with `KnockDownPlank` (`src/_ZN14KnockDownPlank16CleanupResourcesEv.cpp`) and renaming them is a separate decision about a separate class.

## Verification

```
tubuild verify ov015/daObjBkBillboard_c
  6/6 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tubuild linkcheck --baseline --module ov015 -j16 --clean
  modules PASS; ROM identical to this tree's stock build
rombuild -j16
  source-built functions 11,088  reproducing 11,088  mismatching 0
```

Static gates all clean: `check_references` (no new unresolvable references), `check_dead_references`, `check_duplicate_sources`, `check_rename_ledger`, `check_data_definitions`, `check_src_tu`, `check_tubuild_conflicts`, `check_header_offsets include/daObjBkBillboard_c.h` (struct spans `0x124`, 0 mismatched).

`config/converted-baseline.json` was restored from `origin/main` and regenerated with `tools/tiers_ratchet.py --update`; the four per-function paths are logged in `converted-backslide-exceptions.jsonl` as moves into a CONVERTED file, not removals.


andrewboudreau:
Sign-off (shared account, so this is a comment rather than an approval).

**Verified locally before pushing**
- `tools/rombuild.py -j16` -> **106/106 exact, 100.000000%**, mismatching **0**.
- `tubuild.py verify ov015/daObjBkBillboard_c` -> all members MATCH; no unlicensed sections after the 9 `compiler_only_output` deadstrip-data rows were banked with measured ROM homes.
- `tubuild linkcheck --baseline --module ov015 -j16 --clean` clean (the inline destructor rules out `--partial`, by design).
- `tools/premerge_check.py 2215 --repo ... --fetch`: every gate `pass -> pass`, `RESULT: nothing goes green -> red`. `source-coverage` info `entries: 10412 -> 10407 (-5)` is the expected TU consolidation, bytes flat.

**CI**: all static checks green; `PR validation` **pass** (23m30s).

The vtable claim rests on the cartridge, not on a guess: `_ZTS 0x02114324` spells `18daObjBkBillboard_c`, `_ZTI+8` points at `_ZTI8dActor_c`, and the `_ZTV 0x02114360` address point carries offset-to-top 0 with `&_ZTI` behind it. Only slots 0/3/9 and the two destructor slots differ from `dActor_c`, and only those are redeclared.

Merging.

PR #2216, opened 2026-09-03
ov039: promote daObjKumo_c into a single C++ TU

andrewboudreau:
Drops the coined `Cloud` spelling in favour of the ROM's own `daObjKumo_c` and folds the class's seven `.text` functions into one translation unit, `src/actors/d_a_obj_kumo.cpp`.

## The name and the base come from the cartridge

Read out of `extracted/overlays/overlay_0039.bin` at ov039 base `0x021111a0`:

| address | contents |
| --- | --- |
| `_ZTS` `0x02111824` | `11daObjKumo_c` |
| `_ZTI` `0x02111818` +8 | `0x0208e390` = `_ZTI8dActor_c` (arm9) — the **direct** base is `dActor_c` |
| `_ZTV` `0x02111858` | address point (addend 0); `V-8` = 0 offset-to-top, `V-4` = `&_ZTI` |

| slot | address | method |
| --- | --- | --- |
| 0 | `0x0211132c` | `InitResources` |
| 3 | `0x02111254` | `CleanupResources` |
| 6 | `0x021112a0` | `Behavior` |
| 9 | `0x02111278` | `Render` |
| 16 / 17 | `0x021111a0` / `0x021111d0` | `~daObjKumo_c` D1 / D0 |

## This one was the alias case

ov039's `symbols.txt` already carried **both** `_ZTV5Cloud` and `_ZTV11daObjKumo_c` at `0x02111858`. That is exactly the case `tools/class_rename.py`'s docstring calls out: the right operation is to **delete the coined alias**, not rename it onto a name that is already there (a duplicate at `size=0x0` fakes "lost N matched functions"). So the coined row is deleted here and the rename runs over the remaining coined identifiers. `include/decl_common.h` declared both spellings as well — line 705 was the coined one, and it is removed rather than renamed into a duplicate of line 477.

`symbols/actor_renames.tsv`'s vtable row now names `_ZTV11daObjKumo_c`; `check_rename_ledger` agrees with `symbols.txt` on all 1,652 mangled/vtable rows.

## Destructor declared inline, on purpose

With an out-of-line body mwcc emits D0 ahead of D1, and the ROM has D1 first — `rombuild` refuses the object outright. The inline body also drops the D2 variant the cartridge never carried. The header carries the `MEASURED -- INLINE ON PURPOSE` banner.

## Two things worth reviewing

**Mixed-language TU.** `func_ov039_02111214` is a `.c`-derived helper sitting between D0 and `CleanupResources` in the ROM's own `.text` order (`0x021111a0..0x0211137c`), so it belongs to this TU and cannot be split out. It keeps C linkage and its unmangled ROM name in its own `extern "C"` block — the same shape #2215 introduced for the ov015 billboard, whose helper at `0x02111214` is the same routine.

**A pragma that had to move.** The legacy `Behavior` source carried `#pragma opt_propagation off` immediately above its body. That pragma is **file-global last-wins** in mwccarm 2004/b56, not positional, so in the merged TU it has to sit at the top of the file. Measured both ways: with it there, all seven functions match; without it, `Behavior` alone diffs by 999 words and the other six are unaffected. The manifest records the measurement so nobody re-litigates it.

## Verification

```
tubuild verify ov039/daObjKumo_c
  7/7 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tubuild linkcheck --baseline --module ov039 -j16 --clean
  modules PASS; ROM identical to this tree's stock build
rombuild -j16
  source-built functions 11,088  reproducing 11,088  mismatching 0
```

Static gates clean: `check_references` (no new unresolvable references), `check_dead_references`, `check_duplicate_sources`, `check_rename_ledger`, `check_data_definitions`, `check_src_tu`, `check_tubuild_conflicts`, `check_header_offsets include/daObjKumo_c.h` (struct spans `0x124`, 0 mismatched).

`config/converted-baseline.json` was restored from `origin/main` and regenerated with `tools/tiers_ratchet.py --update`; the five per-function paths are logged in `converted-backslide-exceptions.jsonl` as moves into a CONVERTED file, not removals.


andrewboudreau:
Sign-off (this account cannot use GitHub's Approve, so this comment is the review record).

Re-verified after the rebase onto `origin/main` @ `5cc5efd30`:

- `tools/rombuild.py -j16` -> **module fidelity 106/106 exact, 100.000000%, mismatching 0**; 11,088 source-built functions reproducing, 0 mismatching.
- `tools/tubuild.py verify ov039/daObjKumo_c` -> 6/6 MATCH, objisolate clean, reloc-destinations clean, emission order as expected.
- `tools/premerge_check.py 2216 --repo <worktree> --fetch` -> every gate `pass -> pass`, nothing goes green -> red.
- Converted ratchet regenerated with `tools/tiers_ratchet.py --update` (never hand-edited): `2592 -> 2591 (+4 / -5)`.
- The rebase conflicted only in `attribution.json`; resolved by UNION (no key pruned, old promoted-path keys kept alongside the new ones).
- CI: all 10 checks green, including PR validation (25m49s).

The vtable claim is the addend-0 `_ZTV` address point measured in `extracted/overlays/overlay_0039.bin`, and the destructor is declared first and inline so mwcc emits D1 then D0 in the cartridge's own order.

PR #2219, opened 2026-09-03
ov036: promote daObjRc_Guruguru_c into a single C++ TU

andrewboudreau:
The RC_GURUGURU profile's powered turntable: a platform that spins about
its own yaw axis for as long as it is alive, carrying whatever stands on
it, and switches its collision mesh off once the actor is flagged.

The class was carried in the tree under the coined name
ArmedRotatingPlatform. The cartridge disagrees, and says so in its own
bytes: overlay_0036.bin at 0x02113d90 spells "18daObjRc_Guruguru_c", so
the rename comes first and everything downstream follows the ROM.

WHAT THE IMAGE PROVES

  _ZTS  ov036 0x02113d90   "18daObjRc_Guruguru_c" -- the class name.
  _ZTI  ov036 0x02113d84   __si_class_type_info; +8 points at
                           _ZTI10dBgActor_c, so the DIRECT base is
                           dBgActor_c and nothing else.
  _ZTV  ov036 0x02113dcc   the address point itself: V-8 is a zero
                           offset-to-top, V-4 is &_ZTI, V+0 is slot 0.
                           The vptr store is addend-0 against this
                           symbol.
  size  0x320              the factory's own literal, handed to
                           fBase_c::operator new.

Six functions, .text 0x02111d14..0x02111f5c, all six MATCH.

THE ONE MEMBER. dBgActor_c's last field ends at 0x31e and its size
rounds to 0x320, so the Itanium ABI places the first derived byte at
0x31e and sizeof stays 0x320. mAngVelY is an s16 there. InitResources
seeds it with -0x80, or with the spawn's mAngleZ when the level author
set one; Behavior adds it to mAngleY every frame -- that is the whole
spin.

TWO CODEGEN NOTES, both measured, both written into the source:

  data_ov036_02113d78 is a REAL three-word data object, not a literal
  pool. InitResources loads its address ONCE and then indexes it
  (ldr r0,[r1]; ldr r0,[r0,#4]; ldr r3,[r2,#8]), and CleanupResources
  loads that same one literal twice for [0] and [1]. Naming the three
  targets directly builds a three-entry pool instead and moves every
  call site -- 4/6 MATCH with two reloc destinations reported wrong.
  Keeping the indirection gives 6/6.

  mAngleY += mAngVelY, not the long-hand spelling: the compound form
  CSEs the field address into a register and reuses it for the load and
  the store, which is what the ROM does.

THE DESTRUCTOR is declared first and inline. Out of line, mwcc emits D0
ahead of D1 and the cartridge has D1 first; inline also drops the D2
variant the ROM never carried. The vtable and the eleven RTTI symbols
that come with it are banked as compiler_only_output deadstrip-data rows
with their measured cartridge homes.

Also corrects a stale cross-reference: daObjKm1_Ukishima_c.h cited the
old coined name for the 0x31e tail-padding precedent.

PROOF

  tubuild verify ov036/daObjRc_Guruguru_c   6/6 MATCH
  tubuild linkcheck --baseline --module ov036 --clean   modules PASS
  rombuild.py -j16   module fidelity 106/106 exact, 100.000000%,
                     mismatching 0, 11,088 source-built functions all
                     reproducing


andrewboudreau:
Re-verified after the rebase onto `f8eec70ea` (main had moved twice while this sat in CI).

**Rebase.** Two conflicts, both resolved by union rather than by picking a side:

- `attribution.json` — union-merged, 12 keys added, 1579 overrides total. No key pruned.
- `notes/platform-provenance.md` — main carried #2211's `ShutterHmc` -> `daObjCvShutter_c`, this branch carried `ArmedRotatingPlatform` -> `daObjRc_Guruguru_c`. The one line now carries **both** renames; neither side was discarded.

`config/converted-baseline.json` and `config/converted-backslide-exceptions.jsonl` were restored from `origin/main` and regenerated with `tools/tiers_ratchet.py --update`, never hand-merged: `2592 -> 2592 (+4 / -4)`, 4 removals logged with a reason.

**Proof on the rebased commit `efd3da2eb`:**

```
reproducing: 11,088
mismatching: 0
source-owned data claims: 16  (reproducing 16, mismatching 0)
ROM-build analysis: PASS
```

**Merge tree** (`tools/premerge_check.py 2219 --repo ... --fetch`, base `f8eec70ea` @ merge tree `173be992d833`) — every gate `pass -> pass`:

```
converted-ratchet   pass -> pass
dead-references     pass -> pass
duplicate-sources   pass -> pass
header-offsets      pass -> pass
langmode-ratchet    pass -> pass
layout-check        pass -> pass
src-tu-refs         pass -> pass
source-coverage     pass -> pass   entries: 10387 -> 10382 (-5), consolidation, bytes flat
```

Signing off in a comment rather than approving — this repo shares one account. Merging with `--squash`.

PR #2220, opened 2026-09-03
ov031: rename SlideDecorationSilverStar to its ROM name and promote it to one TU

andrewboudreau:
`SlideDecorationSilverStar` was a coined name, and it named the wrong thing twice over. The cartridge's own RTTI calls this class **`daObjHsBillboard_c`**, and the class is not the silver star — it is all four of the slide's decoration billboards.

## What the image says

Read from `extracted/dsd/arm9_overlays/ov031.bin`, ov031 base `0x021111a0`:

| record | address | what it holds |
|---|---|---|
| `_ZTS18daObjHsBillboard_c` | `0x021118f4` | the length-prefixed bytes `18daObjHsBillboard_c` |
| `_ZTI18daObjHsBillboard_c` | `0x021118e8` | `+4` → `_ZTS`, `+8` → `0x0208e390` = `_ZTI8dActor_c` |
| `_ZTV18daObjHsBillboard_c` | `0x02111984` | `V-8` = 0, `V-4` = `&_ZTI` — the address point |

Two corrections follow.

**The direct base is `dActor_c`, not `dBgActor_c`.** The RTTI base pointer states it, and the `0x128` allocation agrees.

**Four spawn profiles share one class.** `SlideDecorationBlueSmiley`, `SlideDecorationOrangeSmiley`, `SlideDecorationYellowStar` and `SlideDecorationSilverStar` — actor ids `0x12e..0x131` — have one implementation and one vtable between them. `InitResources` switches on the actor id to choose which of the four `SharedFilePtr`s at `data_ov031_02111424` to load. Naming the class after the silver star was picking one of the four arbitrarily.

## The vtable

It is `dActor_c`'s full 31 slots and ends at `0x02111a00`, the end of the overlay. Exactly five slots hold words inside ov031:

```
slot  0  0x021112a8  InitResources
slot  3  0x02111254  CleanupResources
slot  9  0x02111280  Render
slot 16  0x021111a0  D1
slot 17  0x021111d0  D0
```

Every other slot — **slot 6 included** — still points into arm9. There is no `Behavior` override; these billboards never tick, and the header declares none.

## The alias case

[...]

PR #2221, opened 2026-09-03
ov032: promote daObjTdWater_c into a single C++ TU

andrewboudreau:
The `TD_WATER` profile's water surface: a large animated water plane with a collision mesh under it, whose texture scrolls for as long as the actor is alive, and which is only spawned when the level's event bit is clear.

## The first alias case in this queue

This class already carried **both** names on one address — `_ZTV9HugeWater` and `_ZTV14daObjTdWater_c`, two symbols at ov032 `0x021139a4`. The right operation there is to *delete the coined alias*, not to rename it onto a name that is already present; `class_rename.py` refuses the second and says so. So the duplicate ROM-name row was dropped first and the rename then ran normally, which carries the mangled lengths, the ledger row and the filenames across in one idempotent pass.

## What the image proves

| record | address | what it says |
|---|---|---|
| `_ZTS14daObjTdWater_c` | ov032 `0x0211396c` | the bytes spell `14daObjTdWater_c`. It had **no symbol at all** before this; the row is added here, with the ledger row that records it. |
| `_ZTI14daObjTdWater_c` | ov032 `0x02113960` | `__si_class_type_info`; `+8` is `0x021089ec` = `&_ZTI10dBgActor_c`, so the direct base is `dBgActor_c` and nothing else. |
| `_ZTV14daObjTdWater_c` | ov032 `0x021139a4` | the address point itself — `V-8` is a zero offset-to-top, `V-4` is `&_ZTI`, `V+0` is slot 0. The vptr store is addend-0 against this symbol. |
| size | `0x334` | the factory's own literal, handed to `fBase_c::operator new`. |

Slots 0/3/6/9 and 16/17 were read out of the overlay and each one is the function the header declares; every other slot holds `dBgActor_c`'s word and is inherited.

Six functions, `.text 0x02112698..0x021128b8`, all six MATCH.

## The one member

A `TextureTransformer` at `0x320`. `dBgActor_c` ends at `0x31e` and rounds to `0x320`, and this member is 4-byte aligned — so unlike `daObjRc_Guruguru_c::mAngVelY` it does **not** sit in the base's tail padding. Those two bytes stay padding, and `0x320 + 0x14` closes exactly on `0x334`. `InitResources` hands it the BTA at `0x02112f64`; `Behavior` sets its speed to 1.0 and advances it every frame; `Render` folds it into the model's components before drawing.

## What the curation replaced

The six legacy files carried three shadow structs (`Sub`, `Base`, `D113afc`), a shadow `Fix12i` and a wall of `char*` arithmetic. All of it is gone: `mModel`, `mMeshCollider`, `mClsnMat` and `mAngleY` are `dBgActor_c`'s real members, `mModel.data` is the embedded `ModelComponents` at `+0x8`, and `TextureTransformer`'s own header supplies `Prepare`/`Update`/`Advance`.

Two `SetFile` calls keep their mangled free spelling — both carry `Fix12<int>`, wall `6az` — and one `SharedFilePtr` field read stays an explicit `+4`, because `include/SharedFilePtr.h` deliberately recovers no layout.

The destructor is declared first and inline. Out of line, mwcc emits D0 ahead of D1 and the cartridge has D1 first; inline also drops the D2 variant the ROM never carried. The vtable and the eleven RTTI symbols that come with it are banked as `compiler_only_output` deadstrip-data rows with their measured cartridge homes.

## Proof

Rebased onto `a814b992f` (after #2219 landed) and re-run, not carried over from before:

```
tubuild verify   6/6 MATCH, objisolate clean, reloc destinations clean
rombuild -j16    reproducing 11,088, mismatching 0
                 ROM-build analysis: PASS
```

The rebase's one conflict was `attribution.json`, resolved by union — 12 keys added, 0 changed, 1591 total, nothing pruned. `converted-baseline.json` and `converted-backslide-exceptions.jsonl` were restored from `origin/main` and regenerated with `tools/tiers_ratchet.py --update`: `2592 -> 2594 (+5 / -3)`, the three removals logged with a reason.

Static gates: `check_header_offsets --changed origin/main` — `daObjTdFuta_c.h` spans `0x31e`, `daObjTdWater_c.h` 1 commented field / 0 mismatched / spans `0x334`, `decl_common.h` clean; `check_rename_ledger` — every mangled/vtable row agrees with its module's `symbols.txt`; `check_dead_references` — no new dead references, no broken links; `check_decl_return_types` — the same three pre-existing `dScMgBase_c` disagreements this branch does not touch.

## Known follow-up

`class_rename.py` reported "no rule fired" on `symbols/profile_reconstruction_registry.{json,tsv}`, which still carry `HugeWater_SpawnInfo` and `HugeWater_Spawn` rows, and on `docs/class-reference.html`. The docs are generated — they want a regeneration, not a hand-edit.


andrewboudreau:
Sign-off (this account cannot APPROVE its own PRs, so the record goes in a comment).

**Local proof, on the rebased head `07fbefae3`:**

- `tools/tubuild.py verify ov032/daObjTdWater_c` -- all functions MATCH, objisolate
  clean, reloc-destinations clean, emission order as expected -> TEXT-VERIFIED, no
  promotion refusal.
- `tools/tubuild.py linkcheck --baseline --module ov032 -j16 --clean` -- the control
  run reproduces the tree's own pre-existing `dsd check symbols` failures and nothing
  else; ROM builds.
- `tools/rombuild.py -j16` -- **11,088 reproducing, mismatching 0**, module fidelity
  **106/106 exact, 100.000000%**, ROM-build analysis PASS.

**Merge-tree proof (`tools/premerge_check.py 2221 --repo ... --fetch`), re-run against
the current `origin/main` @ `a414ce6cd48a` now that #2220 and #2222 have landed
underneath:**

- merge tree `0e39cc05301c`; every gate pass -> pass, nothing goes green -> red.
- `source-coverage` info: `entries: 10372 -> 10367 (-5)` -- the five per-function
  legacy sources this TU consolidates, not a loss of coverage; bytes are flat.

**CI:** all 10 checks green on `07fbefae3`.

The two rebases this branch took were both `attribution.json` conflicts, resolved by
UNION in both directions (no `--ours`/`--theirs`, no pruning): the overrides map went
1591 -> 1603 -> 1615 keys, with the old per-function `path#symbol` keys kept alongside
the promoted file's.

Merging.

PR #2223, opened 2026-09-04
ov036: promote daObjRcBuranko_c into a single C++ TU

andrewboudreau:
The Rainbow Cruise swinging platform, `profile RC_BURANKO`: a mesh collider hung like a pendulum, tipped `0x2000` to one side at spawn and swinging about its Z axis forever. Seven per-function legacy sources become one hand-curated translation unit, `src/actors/d_a_obj_rc_buranko.cpp`.

### What the cartridge proves

Read out of `extracted/overlays/overlay_0036.bin`, base `0x021111a0`:

| record | address | what it says |
| --- | --- | --- |
| `_ZTS` | `0x02113a60` | bytes spell `16daObjRcBuranko_c` |
| `_ZTI` | `0x02113a54` | `__si_class_type_info`; `+8` -> `0x021089ec` = `_ZTI10dBgActor_c`, so the **direct base is `dBgActor_c`** and nothing else |
| `_ZTV` | `0x02113a98` | the address point itself: `V-8` is a zero offset-to-top, `V-4` is `&_ZTI`, `V+0` is slot 0. The vptr store is **addend-0** against this symbol. |
| size | `0x320` | `daObjRcBuranko_c_classInit`'s own literal (800) |

Only slots 0, 3, 6, 9, 16 and 17 of the 32-word table point inside ov036; every other slot still holds `dBgActor_c`'s arm9 word, slot 31 (`Kill`) included. The one field this class adds -- an `s16` swing velocity -- lives in the two tail bytes `dBgActor_c` already rounds `0x31e` up to, which is why the derived class costs no extra bytes and `sizeof` stays `0x320`.

### The pendulum

The whole behaviour is four lines: a constant 4-unit acceleration always pointed back at zero, integrated into `mAngleZ`. No damping term, so the swing never decays and never needs re-kicking.

### One RTTI deferral cleared

The coined `SwingingPlatform` alias is gone; the class name now comes from the ROM's own type string. That type-name record at `0x02113a60` was still spelled `data_ov036_02113a60` in `symbols.txt` -- one of the ten deferrals listed in `notes/rtti-reconciliation.md`. Nothing in `src/` referenced the coined spelling and its `include/decl_common.h` declaration was dead, so it is renamed here to `_ZTS16daObjRcBuranko_c`: the promoted TU's manifest needs the cartridge's own name to bank the record as `deadstrip-data` rather than a plain `deadstrip`, which is never compared against the cartridge. The note's list and counts are updated.

### The seventh function

The span carries `func_ov036_0211123c`, an unnamed in-span function that is really this class's private helper: it rebuilds `mModel`'s `Matrix4x3` from the actor's three angles, then writes the actor position `>> 3` into that matrix's translation row. It keeps C linkage and its existing name -- the cartridge gives it a global symbol and other config still names it -- so it stays at raw offsets, with comments saying what each one is. Promoting it to a real method is a separate change with its own codegen risk.

The destructor is declared **first** and inline in the header on purpose. Out of line, mwcc emits D0 ahead of D1 while the cartridge has D1 first, and it also emits the D2 base-object variant the ROM never carried. Declaring it first is what makes this TU the vtable's home, so `_ZTV` and the RTTI pair land here.

### Proof

```
tubuild.py verify ov036/daObjRcBuranko_c
    7/7 MATCH, objisolate clean, reloc-destinations clean,
    emission order as expected -> TEXT-VERIFIED, no promotion refusal

tubuild.py linkcheck --baseline --module ov036 -j16 --clean
    control reproduces only the tree's pre-existing dsd check symbols
    failures; ROM builds, sha d1506e90...c478e8

rombuild.py -j16   (re-run after the rebase onto current main)
    11,088 reproducing, mismatching 0
    ROM data from source: 640 verified (up from 637 -- the newly named _ZTS)
```

Static gates all clean: `check_dead_references`, `check_duplicate_sources`, `layout_check`, `check_src_tu`, `tiers_ratchet`, `check_tubuild_conflicts`, `check_header_offsets --changed origin/main`, `langmode_audit`.

The one `tubuild create` CONFLICT (two legacy spellings of the `func_ov036_0211123c` extern) is reconciled: the production TU defines the function itself, and the manifest note is rewritten to the `(RESOLVED) ... -- RECONCILED:` form. `converted-baseline.json` is regenerated by `tiers_ratchet.py --update`, never hand-edited; `attribution.json` overrides are unioned with the old keys kept.


andrewboudreau:
Signing off. Every check is green, and `premerge_check 2223 --repo` reports all eight static gates pass -> pass on the merge tree `85f85c56a5ac` after the rebase and ratchet regeneration. `rombuild -j16` was re-proved at 106/106 exact, 100.000000%, mismatching 0 on the rebased head `03cd4c4f18bb`. Merging.

(No PR in this repo can be formally APPROVED -- shared account -- so this comment is the sign-off.)

PR #2225, opened 2026-09-04
ov009: promote daObjMc_Metalnet_c into a single C++ TU

andrewboudreau:
The MC_METALNET profile's climbable chain-link net becomes one production
source, `src/actors/d_a_obj_mc_metalnet.cpp`, absorbing the eight legacy
one-function shards that used to carry it. The ROM build takes all eight
functions from a single object, the way the cartridge's own build did.

### What the cartridge proves about the name

| record | address | value |
| --- | --- | --- |
| `_ZTS` | ov009 `0x02113aa4` | `"18daObjMc_Metalnet_c"` |
| `_ZTI` | ov009 `0x02113a98` | `__si_class_type_info`; +8 → `_ZTI10dBgActor_c` (ov002 `0x021089ec`) |
| `_ZTV` | ov009 `0x02113ae0` | the address point; the emitted object starts two words earlier at `0x02113ad8` |
| size | `0x320` | the factory's own literal (800) — also `dBgActor_c`'s, because this class adds no field |

The class was carried under the coined name `MetalNet`, whose `_ZTV` alias sat
on this same address. That alias is deleted and the ROM's own type string is now
where the class name comes from. `MetalNetLift` (ov064) is a different class
with its own RTTI and is untouched. The factory and profile take the house
spelling: `MetalNet_Spawn` → `daObjMc_Metalnet_c_classInit`,
`MetalNet_SpawnInfo` → `g_profile_MC_METALNET` (ambiguous — `WL_POLELIFT`
points at the same descriptor).

### The vtable

Diffed slot by slot against `_ZTV10dBgActor_c` (ov002 `0x0210ae38`). Six slots
of the 32-word table point inside ov009 — 0 `InitResources`, 3
`CleanupResources`, 6 `Behavior`, 9 `Render`, 12 `OnPendingDestroy`, and the
16/17 destructor pair. Every other slot still holds the base's own word,
including slot 31 (`Kill`, `0x020ee55c`), so nothing else is overridden and
nothing else is redeclared in the header. Slot 12 is the unusual one for this
family: `fBase_c::OnPendingDestroy` is replaced by a four-byte empty body at
`0x02111ea4`.

### Two placement decisions worth reading

**The destructor is inline and declared first.** This TU defines every virtual
the class has, so there is no other candidate key function and no declaration
order suppresses the vtable emission — moving the declaration only changes
which destructor variants are emitted and in what order. Out of line, mwccarm
emits the D2/D1/D0 triple in D0-before-D1 order while the cartridge has D1 at
`0x02111dc4` *below* D0 at `0x02111e08`, which `objisolate` refuses outright,
and it emits a D2 the ROM never carried. Inline and first emits exactly the
retail D1/D0 pair in retail order.

**The vptr store carries the preamble bias.** Because the vtable is emitted
here, mwcc's `_ZTV18daObjMc_Metalnet_c` names the emitted *object* at
`0x02113ad8`, not the address point, so the factory stores
`(int)&_ZTV18daObjMc_Metalnet_c[2]` and production isolation subtracts the
two-word preamble back off. A TU that only *imports* someone else's vtable
writes the bare symbol instead; this one does not.

The eleven vague-linkage RTTI records the TU emits — this class's own trio plus
the `dBase_c`/`fBase_c`/`dActor_c`/`dBgActor_c` chain — are licensed as
`deadstrip-data` in the manifest with their measured ROM homes.
`tools/romdata_check.py` reports **6 VERIFIED, 5 PARTIAL, 0 DIFFERS** for this
source; the PARTIALs are the known dsd extent shortfall on the base-class
records, not a byte disagreement.

### Proof

```
tools/tubuild.py verify ov009/daObjMc_Metalnet_c
  8/8 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tools/rombuild.py -j16
  106/106 exact, 100.000000% of compared bytes, mismatching 0
  source-built functions 11,088, reproducing 11,088
```

Static gates — header-offsets, dead-references, duplicate-sources,
source-coverage — all clean. `config/converted-baseline.json` was regenerated
with `tools/tiers_ratchet.py --update` (2594 → 2594, +5 / −5); the five removals
are the consolidation itself and are logged in
`converted-backslide-exceptions.jsonl` with that reason.


andrewboudreau:
Signing off. All CI checks pass (PR validation included, 20m11s) on head 43c7632eedd6. `premerge_check 2225` against origin/main 2827785db77a: merge tree 6eb0067f798c, all eight static gates pass -> pass, entries 10360 -> 10353 (-7, the seven legacy one-function sources consolidated into one TU). Locally proved at 106/106 exact, 100.000000%, mismatching 0. Merging.

PR #2226, opened 2026-09-04
ov018: promote daObjIceBoard_c into a single C++ TU

andrewboudreau:
The shatterable ice sheet (profile `ICE_BOARD`, ov018) becomes one hand-curated translation unit, `src/actors/d_a_obj_ice_board.cpp`, replacing ten one-function sources. Same recipe as #2184 / #2223 / #2225.

## The name comes from the cartridge

The coined `IceSheet` spelling is gone. What the ROM proves:

| record | address | content |
| --- | --- | --- |
| `_ZTS` | ov018 `0x02113afc` | `"15daObjIceBoard_c"` |
| `_ZTI` | ov018 `0x02113af0` | `__si_class_type_info`; +8 → `_ZTI10dBgActor_c` (ov002 `0x021089ec`) |
| `_ZTV` | ov018 `0x02113b34` | the **address point**: V−8 zero offset-to-top, V−4 `&_ZTI` |

So the direct base is `dBgActor_c` and nothing else. Nine of the 32 slots point inside ov018 — 0, 3, 6, 9, 16, 17, 21, 27 and 31 — and every other slot still holds `dBgActor_c`'s arm9 word, so nothing else is overridden. The factory's own `0x320` literal and a destructor that touches only the inherited `Model` and `dBgW_KcMbg` independently pin an empty derived layout.

The derived spellings follow the class: `IceSheet_Spawn` → `daObjIceBoard_c_classInit` (house convention), `IceSheet_SpawnInfo` → `g_profile_ICE_BOARD`, and the two bss file pointers → `daObjIceBoard_c_ClsnFile` / `daObjIceBoard_c_ModelFile`.

## Why the destructor is inline and first

Declared out of line, mwccarm emits D0 ahead of D1 and the cartridge has D1 first — `rombuild` refuses that outright — and it also materialises the D2 base-object variant the ROM never carried. Declaring it **first and inline** is also what makes this TU the vtable's home.

That has a consequence: the TU then defines *every* virtual of the class, so mwccarm emits `_ZTV`/`_ZTI`/`_ZTS` for the class **and its whole base chain** — eleven records. Two things follow.

- The vptr store is `(int)&_ZTV15daObjIceBoard_c[2]`, not the bare symbol. When the TU emits the vtable, `_ZTV` names the emitted *object* — the two-word ABI preamble at `0x02113b2c` — and `[2]` steps onto the address point the cartridge actually stores, `0x02113b34`. A TU that only *imports* someone else's vtable writes the bare symbol instead.
- All eleven records are licensed in the manifest as `deadstrip-data` against their canonical cartridge addresses (the class's own trio in ov018, plus `dBase_c`/`fBase_c`/`dActor_c` in arm9 and `dBgActor_c` in ov002). `romdata_check` word-compares each emitted copy against the ROM, relocations applied, before production isolation discards it.

## Proof

```
tubuild verify   10/10 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
rombuild -j16    106/106 exact, 100.000000% of compared bytes, mismatching 0
romdata_check    6 VERIFIED, 5 PARTIAL, 0 DIFFERS
```

Static gates (`premerge_check --repo`), plus `check_tubuild_conflicts` and `check_header_offsets`, all pass → pass.


andrewboudreau:
Signing off (no APPROVE is possible on this shared account).

Verified before merge:
- `tubuild verify` MATCH, objisolate clean, TEXT-VERIFIED
- `tubuild linkcheck --baseline` and `rombuild -j16` both 106/106 exact, 100.000000%, mismatching 0
- `premerge_check 2226` merge tree `17f7d065d72b`: all eight static gates pass -> pass, entries 10353 -> 10344 (-9), bytes flat
- CI green on every check, mergeable/clean

Merging.

PR #2255, opened 2026-09-04
Adopt class-anchored actor TU names (wave 3)

andrewboudreau:
Third validator-sized rename-only wave. Moves 21 promoted spawn-table actor TUs into src/game/actors/ while replacing legacy CamelCase/alias basenames with tools/tu_names.py's NSMBW-lineage class-anchored d_a_* spellings. Every source move is Git 100% / zero inserted or deleted source lines.

The class identities come from ROM RTTI; the filenames are adopted lineage conventions, not recovered historical filenames. Numeric/acronym forms (for example d_a_obj_ct_mecha05.cpp and d_a_kp_fr.cpp) are the documented mechanical convention rather than guessed expansions. Battan's filename-derived private initializer is correspondingly licensed as __sinit_d_a_obj_fm_battan.cpp in its exact manifest/symbol metadata.

Validation: 176 focused repository tests; port_refcheck 405/405; cpp_tu_compat READY; dead-reference, duplicate-source, TU-conflict, and layout checks clean; attribution 21/21 intact with 0 lost; prepush_linkcheck 0 blocking; strict control and production full builds both 106/106 modules with zero mismatches; production 11,088/11,088 source-built functions and 16/16 data claims reproduce; pre-push compiled 124/124 shadow TUs.


andrewboudreau:
Superseded by #2270, which combines this work on current main, passed the complete public suite and private ROM validator, and is now merged.

PR #2259, opened 2026-09-04
Campaign tail: KM3_KURUMA, EYEKUN, OBJ_KUMO, PENGUIN_MOTHER, and a stem-collision fold pilot

andrewboudreau:
The campaign's tail: four rows of descriptor/factory work, one stem-collision fold piloted end to end, and the write-up of how one of the four went missing in the first place.

| commit | what |
| --- | --- |
| 1 | ov071 EYEKUN + ov047 KM3_KURUMA (the reverted rename, restored) |
| 2 | ov039 OBJ_KUMO **descriptor** (the factory half landed in #2263 — see section 3) |
| 3 | ov036 RC_KAITEN factory folded into its promoted TU — the fold pilot |
| 4 | ov018 PENGUIN_MOTHER **descriptor** (the factory half landed in #2263 — see section 3) |

> **Rebased onto `693947eb8`.** Sections 3 and 4 were rewritten after #2263 landed and took work this PR originally claimed. Read them as they now stand, not as they were first written. Conflicts and resolutions are in the rebase note at the end.

## 1. ov071 `0x02122cf0` — `MrI_SpawnInfo` -> `g_profile_EYEKUN`

Plain omission. Wave 31 renames `daEykn_c`'s two factories (`MrI_Spawn` -> `daEykn_c_classInit_EYEKUN`, `BigMrI_Spawn` -> `daEykn_c_classInit_EYEKUN_BOSS`) and the **EYEKUN_BOSS** descriptor at `0x02122d0c`, but skips the **EYEKUN** descriptor one line above it. Four of the family's five symbols were complete; this is the fifth.

RTTI adjacency confirms both descriptors belong to `daEykn_c`:

```
_ZTS8daEykn_c        0x02122cd8
_ZTI8daEykn_c        0x02122ce4
<descriptor>         0x02122cf0   <- this row (EYEKUN)
BigMrI_SpawnInfo     0x02122d0c   <- wave 31 (EYEKUN_BOSS)
_ZTV8daEykn_c        0x02122d30
```

No `delinks.txt` owner and no manifest entry for this address, so the edit is `symbols.txt` plus the ledger row.

**Merge note:** wave 31 edits line 132 of `config/arm9/overlays/ov071/symbols.txt`; this PR edits line 131. Adjacent lines, so git will report a conflict when the second of the two lands. The resolution is to keep both changes — they are different addresses.

## 2. ov047 `0x02112428` — restoring `g_profile_KM3_KURUMA`

This one was **not** an omission. It was a landed campaign rename that got silently overwritten with a *different class's* name, and the ledger row rewritten to match.

### Verified from the ROM before restoring

The brief was to confirm the identity from the bytes rather than from the addresses agreeing. Four independent lines of evidence, all pointing at `daObjKm3_Kuruma_c`:

**RTTI adjacency** (the method in the coined-class-name recipe — walk back from a `_ZTV` to the nearest `_ZTS`/`_ZTI`; a data symbol inside that span belongs to the class):

```
_ZTI17daObjKm3_Kuruma_c   0x021123fc
_ZTS17daObjKm3_Kuruma_c   0x02112414
<descriptor>              0x02112428   <- this row, INSIDE the Kuruma block
_ZTV17daObjKm3_Kuruma_c   0x0211244c
...
_ZTI17daObjKm3_Dorifu_c   0x021124cc   <- Dorifu's block is separate and later
_ZTS17daObjKm3_Dorifu_c   0x021124d8
StairsBs_SpawnInfo        0x021124ec   <- Dorifu's OWN descriptor
_ZTV17daObjKm3_Dorifu_c   0x0211254c
```

**Function layout.** Kuruma's `D1`/`D0`/`CleanupResources`/`InitResources` occupy `0x021113f8..0x021114c0`; the factory sits at `0x021114d4`; `_ZN17daObjKm3_Dorifu_cD1Ev` starts at `0x02111510`. A factory follows its own class's methods and precedes the next class's — the same shape as ov045 PoleLift and ov047 Kurumajiku.

**delinks.txt** gives `src/actors/daObjKm3_Kuruma_c.cpp` as the owner of `0x021114d4`.

**The TU source itself** is decisive. The factory allocates `0x320`, installs `&_ZTV17daObjKm3_Kuruma_c[2]`, and returns `daObjKm3_Kuruma_c *`; the descriptor is typed `KurumaSpawnInfo`, whose `spawn` member is declared `daObjKm3_Kuruma_c *(*)()`.

**Corroboration from the open waves.** Wave 30 (#2250) renames `0x021114d4` -> `daObjKm3_Kuruma_c_classInit`, `0x0211164c` `StairsBs_Spawn` -> `daObjKm3_Dorifu_c_classInit`, and `0x021124ec` -> `g_profile_KM3_DORIFU`. So Dorifu's real factory is `0x0211164c` and its real descriptor `0x021124ec`; wave 30 does not touch `0x02112428`. This PR completes the correction rather than competing with it.

The ROM says KM3_KURUMA. Restored.

### What actually happened

The class map's row `RickshawPlatformBs -> daObjKm3_Dorifu_c` is a correct **class** mapping. #2162 (an ov029 promotion) applied it, and the rename swept up two ov047 symbols whose coined `RickshawPlatformBs_*` spellings were themselves products of the old `parse_spawnfunc` over-scan defect — the off-by-one that attributed a class's factory and descriptor to the *previous* class. So a correct class rename, applied to two mis-attributed symbol names, produced a wrong result at `0x02112428`, in an overlay the promotion had no business editing.

Two properties made it invisible:

1. **`class_rename.py` rewrites the ledger row by name.** Any rename tool that keeps `symbols.txt` and `symbols/actor_renames.tsv` in sync keeps them *consistent with each other* while both drift from the cartridge. This is the general form of the defect, and it is why "the ledger agrees with the tree" is not evidence of correctness.
2. **Scan 2 of `check_profile_campaign.py` is structurally blind to it.** That scan asserts `symbols.txt` carries what the ledger claims. When both move together, it passes. The scan that *would* have caught this is scan 1 (registry proposed name vs. `symbols.txt`), and it did — this row was one of the two "unexplained" pending rows in the campaign census.

This is the **fourth** instance of a landed campaign name being silently reverted.

### Blast radius of the edit

`0x02112428` lives inside a `production_mode: intact-object` TU, so the rename lands in four places, not two: `config/arm9/overlays/ov047/symbols.txt`, `src/actors/daObjKm3_Kuruma_c.cpp` (the `extern "C" KurumaSpawnInfo` definition, with a provenance comment recording the historical alias), `config/tu_manifest.d/ov047/daObjKm3_Kuruma_c.json` (the data entry plus two prose notes), and the ledger.

**Merge note:** wave 30 also touches this manifest and this TU. Different symbols, so any conflict is textual and resolves by keeping both.

## 3. OBJ_KUMO and PENGUIN_MOTHER — the descriptor halves only (rewritten after #2263)

**This section originally claimed more than the PR now does. #2263 landed first and took the other half.**

`OBJ_KUMO` was listed in the pending census under "stem already owned by a promoted TU". It is not a filename-only case: it is a whole unstarted row that *also* happens to collide on its stem. Checking the other twelve the same way turned up a second instance of the same mis-filing, `PENGUIN_MOTHER`. Both were originally done here as four renames — factory and descriptor, for each of the two.

Since this PR was cut, **#2263 folded both factories into their promoted TUs** — `src/Cloud_Spawn.c` into `src/actors/d_a_obj_kumo.cpp`, `src/MotherPenguin_Spawn.c` into `src/actors/d_a_pg_mthr.cpp` — and applied the **factory** renames itself, in `symbols.txt`, in the ledger, and in each fold TU's `// @symbol` marker and `extern "C"` signature. On current `main` those TUs already emit `daObjKumo_c_classInit` and `daPgMthr_c_classInit`; `Cloud_Spawn` and `MotherPenguin_Spawn` survive there only as historical-alias prose.

So on rebase this PR's factory hunks **dropped out as redundant**, and both commits reduced to the **descriptor** rename alone, two files each:

```
ov039 0x02111834  Cloud_SpawnInfo          -> g_profile_OBJ_KUMO         (+ its ledger row)
ov018 0x02113998  MotherPenguin_SpawnInfo  -> g_profile_PENGUIN_MOTHER   (+ its ledger row)
```

#2263 did not touch either descriptor, so this is the half still outstanding and there is no overlap with it.

RTTI adjacency confirms both classes. `daObjKumo_c`: `_ZTI` 0x02111818, `_ZTS` 0x02111824, descriptor 0x02111834, `_ZTV` 0x02111858, factory allocating 0x124. `daPgMthr_c`: `_ZTI` 0x0211397c, `_ZTS` 0x02113988, descriptor 0x02113998, `_ZTV` 0x021139bc, factory allocating 0x38c.

The **filename checks are no longer outstanding** — #2263's folds closed them. With the descriptor renames here, `check_profile_campaign --list` reports **zero** outstanding checks for both OBJ_KUMO and PENGUIN_MOTHER.

The modify/delete conflict this produced, and how it was resolved, is in the rebase note at the end.

## 4. Stem-collision fold pilot — ov036 RC_KAITEN

Piloted the operation that resolves the filename-only rows, to price it rather than to clear the queue.

`daObjRc_Kaitendai_c_classInit` at ov036 `0x02111544` sits immediately after `InitResources` in the ROM's own `.text` order, inside the TU boundary `tools/tu_map.py` already recovered — the registry's own `tu_end` for RC_KAITEN is `0x2111580`, this function's end. It was outside the promoted TU only because the promotion predated the campaign.

The edit is four things:

- **`src/actors/d_a_obj_rc_kaitendai.cpp`** — factory added **first** (highest ROM address, and mwccarm 2004/b56 emits `.text` sections in reverse source order), keeping C linkage. Its four extern declarations are restated in the TU's existing `extern "C"` block rather than pulled in through `decl_ActorBase.h` / `decl_Platform.h` / `decl_common.h`, per the skill's declare-in-place rule.
- **`config/arm9/overlays/ov036/delinks.txt`** — the TU's single `.text` range grows `0x02111444..0x02111544` -> `..0x02111580`, and the legacy entry's identical range is removed. The partition is preserved exactly: no gap, no overlap, nothing unclaimed.
- **`config/tu_manifest.d/ov036/daObjRc_Kaitendai_c.json`** — fifth function entry, section end, `functions_matched`/`functions_declared` 4 -> 5, fold note.
- **`src/daObjRc_Kaitendai_c_classInit.c`** deleted, plus the two now-stale references to it in `notes/data/`.

### What it cost

The TU recompiled once and **all five members matched on the first try** — no member-by-member re-admission was needed, and no existing match was perturbed. What moved and what did not:

| | |
| --- | --- |
| `rombuild` | 106/106 exact, 100.000000%; 11,088 functions reproducing, 0 mismatching |
| `production_mode` | untouched — this manifest has none, and neither does any of the other twelve |
| delinks ranges | one number changed; the entry removed had the identical range |
| CONVERTED ratchet | 2597 -> 2597, does not move |
| `source_coverage --check` | net +0 B, nothing handed back |
| `eligible` | 11085 -> 11084, expected: the function is now inside a promoted TU rather than an eligible one-function file |

### Is it the same for the other twelve?

Structurally, yes. Every one of the thirteen has the identical shape:

- the factory's delinks range is **immediately adjacent** to the owning TU's last range in all 13 — the fold is always a single end-address change, never a re-partition;
- every owning TU has exactly **one** `.text` section;
- **no** owning manifest is in `production_mode`;
- **no** factory source carries a `#pragma` — so the file-global pragma hazard that has cost days elsewhere does not arise here;
- **no** factory source is wrapped in an `extern "C" { }` block, the shape `tubuild create` refuses;
- all 13 are `factory_shape=new_plus_inlined_ctor`, 9-22 code lines.

So I would call these **mechanical, with a per-case compile check that is cheap and fails loudly**. The two things that genuinely vary per case are (a) reconciling the factory's extern declarations against what the TU already declares, and (b) whether adding a member perturbs an existing member's codegen. Both are caught immediately by `rombuild` and neither needs judgement to *detect* — only to fix, if it happens. It did not happen here.

The recommendation as originally written flagged one case for separate handling: **`PENGUIN_MOTHER`** — a 22-function owning TU and a 22-line factory naming five constructors, versus 4-7 functions and a 9-line factory everywhere else — and proposed folding the uniform ones as a batch.

**That forward-look is spent.** #2263 has since folded both `OBJ_KUMO` and `PENGUIN_MOTHER`, including the case this section singled out as the risky one, and it landed green. The pilot's finding held: the outlier folded like the rest. What remains of this section is the pilot's own evidence, not a plan — the queue it was pricing has been worked in parallel.

## Census corrections

Three fixes to the pending-row grouping reported earlier, recorded here so the categories stay honest:

- **`OBJ_KUMO` and `PENGUIN_MOTHER` are not filename-only cases** — see section 3. Both were whole unstarted rows that also collide. (Both are now complete: #2263 took the factory half and the folds, this PR takes the descriptor half.)
- **`BK_BILLBOARD` does not carry `production_mode` at all.** I reported earlier that it "carries `production_mode` without being bound by it". That was wrong in a stronger direction than stated: there is **no** `production_mode` manifest anywhere in ov015, and no manifest owns its descriptor. What misled the earlier census was the registry's `current_profile_name` for that row (`PoleBillboard_SpawnInfo`), which is stale — `symbols.txt` at `0x0211433c` already carries `g_profile_BK_BILLBOARD`. It is a genuine filename-only row.
- **`STAR_CAMERA` is structurally excluded, not luckily excluded.** Its registry row has `factory_rename_recommended=no` and an empty `proposed_factory_name`, so the factory check has nothing to assert and the row can never report a factory as pending. Worth stating as a property: that category cannot silently absorb a real row later, because a real row would carry a non-empty proposed name.

With KM3_KURUMA and EYEKUN landed, the **unexplained** category of the pending census is empty.

## A finding about the gate itself

**Resolved since this was first written — kept for the record.** When this PR was cut, `tools/check_profile_campaign.py` (PR #2253) reported scan 1 as dark:

```
check_profile_campaign: registry scan SKIPPED -- symbols/profile_reconstruction_registry.tsv
lacks the column(s) overlay_resolution, factory_filename.
```

The tool degraded honestly — it said so rather than passing vacuously — but the dataset on `main` named those columns `overlay` and `current_factory_file`, so the registry scan had never actually run in CI. That has since been fixed: on the rebased branch scan 1 runs and reports `401 registry row(s), 391 in scope`. The general lesson stands — a schema mismatch between a gate and its dataset is a silent no-op, and the only thing that surfaced it was reading the tool's own output rather than its exit code.

Separately, and relevant to section 2 above: the registry's `current_*` columns are a snapshot from before the waves landed, so they are stale for any row a wave has since renamed. Anything reading them as current will mis-classify rows — that is what produced the BK_BILLBOARD error.

## Rebase note (onto `693947eb8`)

Four conflicts, all from work that landed while this PR sat open.

- **`config/arm9/overlays/ov071/symbols.txt`** — as predicted in section 1. Wave 31 renamed `0x02122d0c`, this PR renames `0x02122cf0`; adjacent lines, different addresses. Kept both.
- **`src/Cloud_Spawn.c` and `src/MotherPenguin_Spawn.c`** — `modify/delete`. #2263 deleted both by folding them into their promoted TUs. Took the deletion; the factory renames this PR carried for them were already applied by #2263 inside the fold TUs, so those hunks dropped out. See the rewritten section 3.
- **`include/daPgMthr_c.h`, `include/daTBasket_c.h`** — prose only, took `main`'s side. This matters for `daTBasket_c.h`: this PR's text would have **reintroduced** the stale `BooCage_Spawn` / `BooCage_SpawnInfo` spellings that `main` has since renamed to `daTBasket_c_classInit` / `g_profile_T_BASKET`. Both headers are now byte-identical to `main`.
- **`src/d_a_eykn_eyekun.c`** — a **false rename pairing**. Git matched the deleted `src/MotherPenguin_Spawn.c` to this unrelated file by content similarity (both are one-factory files with near-identical `decl_*` include blocks) and presented a content conflict inside it. The commit never touched that file. Restored from `main` and verified byte-identical. Rename detection is a heuristic; the file's own history is the authority.

`git diff --stat origin/main...HEAD` is **13 files** — this PR's own work only, no landed commit replayed.

## Gates (re-run on the rebased branch)

| gate | result |
| --- | --- |
| `rombuild.py -j 16 --no-rom` | `module fidelity: 106/106 exact, 100.000000%` — ROM-build analysis: PASS; 11,088 functions reproducing, 0 mismatching |
| `romdata_check.py` | exit 0; 648 verified / 204 partial / 5 differ / 414 unnamed (unchanged) |
| `check_rename_ledger.py` | exit 0; 1652 mangled/vtable rows agree |
| `check_dead_references.py` | exit 0; no new dead references, no broken links |
| `check_profile_campaign.py` | rows: 0 diverged. The coined-ledger scan reports 12 divergences that are **inherited from `main`** — verified by running the same check against a pristine `origin/main` export, which reports the identical 12. They are ov006 rename-chain-order damage, unrelated to this PR, and #2266 repairs them |
| `port_refcheck.py` | 405 checked, all resolve |
| `layout_check.py` | `layout-check: clean` |
| `tiers_ratchet.py --check` | `CONVERTED ratchet PASS baseline 2597 current 2597` |
| `source_coverage.py --check` | measurement only; the fold hands nothing back |
| `eligible.py` | 11071 / 11211 (the absolute moved with `main`; the fold's own effect is -1, the factory now living inside a promoted TU rather than an eligible one-function file) |
| `check_references.py` | 42 unresolved (baseline 45); no new unresolvable references |
| `unittest tools.test_srcpath tools.test_bytegate` | Ran 69 — OK (the suite `.github/workflows/tree-shape.yml` now runs on `src/**` and `config/**` changes) |
| `pytest tools/test_tubuild.py -q` | 68 passed |

No dataset changes: this branch does not touch `symbols/profile_reconstruction_registry.tsv`.


andrewboudreau:
Reviewed and signing off. (The review gate refuses `--approve` on my own PR, so this comment is the sign-off.)

The campaign tail. Rebased onto the repaired `main` with no conflicts; `campaign` went 12 → 0 by picking up #2266 alone, with no content change needed.

**Two things a reviewer should check, because both were wrong in an earlier revision of this PR and have since been corrected in the description:**

1. **#2263 already landed the OBJ_KUMO and PENGUIN_MOTHER folds**, including the factory renames in `symbols.txt`, the ledger, and each fold TU's `// @symbol` marker and `extern "C"` signature. This PR's factory hunks therefore dropped out as redundant and both commits reduced to two files each. What survives is only the **descriptor** half, which #2263 did not do: `Cloud_SpawnInfo → g_profile_OBJ_KUMO` and `MotherPenguin_SpawnInfo → g_profile_PENGUIN_MOTHER`. Sections 3 and 4 originally claimed the filename checks were still outstanding pending a fold; `check_profile_campaign --list` now shows **zero** outstanding checks for both rows. The description has been rewritten and carries a banner saying so.

2. **The ov036 `.text` partition is contiguous** — `0x021111a0 → 0x02111444 → 0x02111580 → 0x0211193c`, no gap, no overlap, nothing unclaimed, running unbroken to the section end. #2262 and #2259 do not fight; `src/daObjRc_Kaitendai_c_classInit.c` is deleted and its delinks entry removed. Re-verified on the final head after rebase, not carried over from the pre-rebase run.

**One resolution worth flagging to anyone reviewing a rebase in this tree.** Git's `-M` rename detection paired the deleted `src/MotherPenguin_Spawn.c` with the untouched, unrelated `src/d_a_eykn_eyekun.c` — both are one-factory files with near-identical `decl_*` include blocks — and presented a content conflict *inside* the latter, offering `// @symbol daPgMthr_c_classInit` as a replacement for `// @symbol daEykn_c_classInit_EYEKUN`. Accepting it would have put the wrong symbol marker in a file the commit never touched, and **no gate downstream reads `@symbol` markers against the deleted file's identity**, so it would have gone through silently. `git show --stat --no-renames <commit>` settles it in one call: the commit's own file list did not contain that file. Similarity detection is a guess about intent; the commit is the record of it.

Also resolved rather than carried: `include/daTBasket_c.h`, where the branch's prose would have **reintroduced** the stale `BooCage_Spawn` / `BooCage_SpawnInfo` spellings that main has since renamed. Both headers are now byte-identical to main.

**Verified:** `rombuild -j 16 --no-rom` **106/106 exact, 100.000000%**; `romdata_check` 648/204/5/414; `layout_check` clean; `tiers_ratchet` PASS 2597/2597; `check_references` 42 unresolved vs baseline 45; `pytest test_tubuild` 68 passed; `test_srcpath`+`test_bytegate` 69/69; `PR validation` passed.

PR #2268, opened 2026-09-05
Add the class-reconstruction agent pipeline

andrewboudreau:
## TL;DR for a reader who has not been in this session

We want to write SM64DS's C++ classes back. That job is too long for one agent to hold in its head, so this PR turns the process into **files in the repo**: five role descriptions, an ordered worklist, and a lock so several agents can work at once without colliding. A Codex instance and a Claude instance read the same files and follow the same protocol — nothing here depends on a model-specific skill or memory.

It also corrects what we thought the remaining job was. See "What the census actually says."

## The five stages

One class moves through four producing stages and one reviewing stage. Each is a separate agent run with a fresh context that reads the previous stage's **written output**, rather than inheriting a conversation.

| stage | role | consumes | produces |
|---|---|---|---|
| 1 | `scout` | the ROM | `notes/data/class-facts/<class>.json` |
| 2 | `writer` | the facts file | the class header and source |
| 3 | `humanizer` | the source | a revision that reads like 2004 EAD C++ |
| 4 | `builder` | the revision | green byte gates, then a PR |
| 5 | `reviewer` | the PR | merge, or a rejection naming the rule broken |

Stages 1–3 never run a byte gate; stage 4 is the only one that decides whether the bytes are right. Stage 3 never touches semantics. Stage 5 is a different instance from stage 4, because a producer's own green output is evidence, not review.

## Claiming, without a claims file

`tools/classqueue.py` makes a claim an **atomic ref create** under `refs/claims/`. The remote accepts a given ref name exactly once, so two agents racing for one class cannot both win:

```
python tools/classqueue.py next    --role writer
python tools/classqueue.py claim   dActor_c --role writer --worktree C:/tmp/sm64ds-dactor
python tools/classqueue.py release dActor_c --role writer
```

Verified: the second push to an existing claim ref is rejected as a non-fast-forward. That non-zero exit *is* the lock.

This is deliberately **not** a claims file. A shared file that every agent edits is what produced the conflict cascade across the wave PRs this week, and `symbols/actor_renames.tsv` is the standing demonstration of the worse failure — its row *order* is load-bearing, so a hand merge silently inverts rename chains while every count still looks right.

## What the census actually says

`notes/data/class-build-worklist.tsv` is derived from `build/rtti.json` (429 classes, 413 base edges, 24 roots), with `tools/rom_class_names.py` resolving the 198 classes whose header still carries a pre-rename coined name.

| bucket | count |
|---|---|
| header already real | 387 |
| shadow / flat header | 14 |
| no header | 27 (about half are nested inner classes whose outer header exists) |

**So writing class declarations is nearly finished, and that is not where the work is.** `src/` holds 10,518 files, of which **9,439 are still auto-named one-function shards** — 6,380 matching `func_ovNN_ADDR` and 3,059 matching `_ZN<len><Name>...`. The classes are declared; their bodies are scattered one function per file.

The real remaining job is gathering those shards into per-class translation units, which is the shape of `72c6dcfb6` — one 240-line `d_a_obj_mc_metalnet.cpp` added, eight one-function shards and one shadow TU deleted. A follow-up commit adds `notes/data/tu-promotion-queue.tsv` ranking those opportunities best-first.

Worth stating plainly: the `Reconstruct N actor profiles (wave NN)` PRs do not touch that pile. They rename registry rows — `<Coined>_SpawnInfo` to `g_profile_<ID>` — which is real work, but it is not class reconstruction, and the role files say so explicitly so no writer agent uses a wave commit as its template.

## What the role files carry

The byte-match rules that have each cost a day when guessed at, stated as constraints rather than advice: the destructor written inline and declared **first** (out-of-line emits D2/D0/D1 in the wrong order plus a homeless D2, where the ROM carries D1-then-D0 and no D2); the vptr store as `&_ZTV...[2]` because the addend loses 8 on rebind; C++ scalarizing a struct copy where C block-moves; bool widening making a bare `.c` to `.cpp` rename cost bytes.

And the gate traps that make a green run a lie — chiefly `check_src_tu_compiles.py` printing `NOT CHECKED` and **exiting 0** when a worktree has no compiler junctioned, which is the default state of a fresh worktree.

## Not covered

- No new CI. Nothing here is enforced by a gate; it is a protocol agents follow.
- `tools/classqueue.py` reads the worklist and the remote only. It cannot start work, and it has no timeout — a crashed agent's claim must be released by hand with `classqueue.py release`.
- The stage-to-stage handoff is a branch name convention, not a state machine. If a class stalls between stages, `classqueue.py list` shows who holds it but not why.


andrewboudreau:
`notes/data/tu-promotion-queue.tsv` has landed on the branch (`db17d2c40`). This is the file the writer agents actually read.

Built by regenerating `build/rtti.json` → `build/rtti_vtables.json` → `build/tu_map.json` (694 TU candidates across 74 modules), then cross-referencing each candidate's live `tubuild.py` disqualifier data against `config/tu_manifest.d/**`, the class worklist, and `c-cpp-classification.tsv`. That last join had to be done **by class name, not by path** — the tsv's `path` column predates many renames.

| | |
|---|---|
| open classed candidates | 226 |
| no hard blocker | **163** |
| shards those 163 absorb | **2,231** |
| blocked | 63 |

Blocked rows are mostly `unmatched:N` (functions with no `complete` delinks marker) combined with `no-legacy-source:N` or heavy `pragma:N` — file-global `#pragma opt_*` TUs cannot be merged as-is. The largest are `Player` (ov002, 596 functions, 8 unmatched) and the `dScMg*` scoreboard classes in ov006.

Two trailing `UNATTRIBUTED` rows record what is deliberately **excluded rather than guessed**: 1,457 shards in modules `tu_map.py` cannot attribute to a class, and 3,103 inside the under-segmented arm9 `main`+`itcm` blob.

One data-quality flag worth knowing before anyone reads a shard count as a win: several candidates — `dScMgJump2_c+dScMgJump_c`, the `dMg*CardObj_c` family — already share one real `.cpp` across all their functions. They are not raw shards at all, just missing a manifest entry, and are tagged `already-one-legacy-file:<path>` rather than ranked.

**First target: `dBgActor_c` (ov002).** Header exists, 11 functions all `complete` with zero pragmas and zero unmatched, D1/D0 pair present, and 22 promoted siblings in the same overlay as style oracle. It owns its own vtable/RTTI, which needs `production_mode: intact-object` — an already-landed pattern in this overlay, not a novel risk.

A pilot agent is running that class through all five stages now, with the protocol defects it hits as the primary deliverable. Findings will be folded back into the role files before the crew is launched.

PR #2283, opened 2026-09-05
ov027: promote daPgDfdr_c into a single C++ TU

andrewboudreau:
## TL;DR for a reader who does not know this codebase

SM64DS was written in C++ in 2004. We only have the shipped cartridge, so the
decompilation starts as ~9,400 tiny files, one per machine-code function, with
names like `func_ov027_02111994.c`. This PR takes fifteen of those files — every
function of the Penguin Defender actor, `daPgDfdr_c` — and puts them back into
the one `.cpp` file Nintendo actually compiled: `src/actors/daPgDfdr_c.cpp`.

Nothing is rewritten. The proof that we got it right is that the 2004 compiler,
fed the merged file, emits **byte-for-byte the same machine code as the
cartridge** for all fifteen functions, and the whole 16 MB ROM still rebuilds
exactly. Fifteen files became one, and the game is bit-identical.

The one interesting subtlety: merging files merges their `#include`s, and this
tree has two different definitions of the same `Matrix4x3` type. The merged file
was reaching the wrong one, which made the compiler emit six extra instructions
in a function nobody had touched. Putting `common.h` first fixed it. That is the
difference between 14/15 and 15/15.

**This PR is a builder verification of work written by another agent. Two
writer-stage defects were found and corrected — see "Defects found" below.**

---

## Proof

```
tubuild verify  15/15 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tubuild linkcheck (pre-promotion)  SCRATCH-LINK-VERIFIED, ROM sha256 d1506e90... identical to baseline
rombuild -j6    106/106 exact, 100.000000% of compared bytes, mismatching 0
romdata_check   7 VERIFIED, 4 PARTIAL, 0 DIFFERS
```

Re-run by the builder in its own worktree (`C:\tmp\sm64ds-dapgdfdr-build`, branch
`cpp/daPgDfdr_c-tu-build`) on the tree merged with `origin/main` @ `379bd67f9`,
which already carries the `dBgActor_c` promotion (#2269) this class inherits from.

`tubuild verify ov027/daPgDfdr_c`:

```
compiler-only     : exact deadstrip ['_ZN7Vector3D1Ev']

MATCH    _ZN10daPgDfdr_cD1Ev                        0x021118c8  size 0x05c
MATCH    _ZN10daPgDfdr_cD0Ev                        0x02111924  size 0x070
MATCH    func_ov027_02111994                        0x02111994  size 0x094
[...]

PR #2281, opened 2026-09-05
ov064: promote daOts_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

The retail cartridge was built from `.cpp` files. This project had recovered
`daOts_c` — the shared base class of the three Bully enemies — as **22 separate
one-function source files**, each compiled on its own and each byte-matching its
slice of the ROM. That is correct but it is not what Nintendo compiled.

This PR puts those 22 functions back into **one translation unit**,
`src/actors/daOts_c.cpp`, compiles it once, and shows the resulting object is
byte-identical to the cartridge across all 22 functions at once. Nothing about
the game's behaviour changes; what changes is that the tree now asserts a real
C++ class with a real compilation boundary instead of 22 fragments.

Two things about this class are worth knowing:

1. **`daOts_c` is abstract.** Vtable slots 0 (`InitResources`) and 6 (`Behavior`)
   are bare zero words. It has no factory and nothing ever allocates one — the
   three Bully subclasses fill those slots in.
2. **Its RTTI is in a different overlay from its vtable.** The vtable is in
   ov064; the type_info record and type name string are in **ov027**. That is a
   real property of the shipped binary, not a bookkeeping error, and it is
   verified below.

I did not write this class. I re-ran every gate from scratch in my own worktree
against the ROM.

---

## Proof block (builder's own runs, worktree `C:\tmp\sm64ds-daots-build`, ref `cpp/daOts_c-tu-build`)

    tubuild verify ov064/daOts_c   22/22 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
    tubuild linkcheck (writer's pre-promotion run)  SCRATCH-LINK-VERIFIED, ROM sha256
      d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8 IDENTICAL to stock control
    rombuild -j6 --no-rom          106/106 exact, 100.000000% of compared bytes, mismatching 0
                                   ROM-build analysis: PASS
    romdata_check                  11 symbols: 7 VERIFIED, 4 PARTIAL, 0 DIFFERS, 0 UNNAMED

The `d1506e90…` sha is corroborated independently here: `rombuild`'s own
`BASELINE CONTROL` stage in this worktree linked the stock scratch ROM to
exactly that sha256, so the writer's pre-promotion linkcheck control and my
build agree bit for bit.

Full gate sweep, in `builder.md` order, all exit 0 unless noted:

| gate | result |
|---|---|
[...]

PR #2286, opened 2026-09-05
ov006: promote dScMgD3DBase_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

The retail SM64DS cartridge was built from C++ source that no longer exists. This repo
rebuilds that source one file at a time and proves each rebuild by compiling it with the
*original* 2004 compiler and checking the output is byte-for-byte the machine code on the
cartridge.

`dScMgD3DBase_c` is the shared base class of the game's four 3D minigame scenes (Jump,
Jump2, Trampoline, Trampoline2). Its code had been sitting in the repo as **27 separate
one-function files** — an artifact of how the code was originally recovered, not how it
was written. This PR merges all 27 back into the single `.cpp` file they came from, and
proves the merge is correct: the compiler, fed the merged file, emits the same 2,616 bytes
of machine code in the same order, and a full ROM rebuilt around it is bit-identical to a
stock cartridge.

Nothing about the game changes — this is a source-organisation change that the byte gates
prove is free.

> **A correction to how this task was framed.** It was handed to me as "the largest TU
> attempted". It is not. Measured across the 95 promoted manifests, 27 functions ties for
> 7th: `ov006/dScMgMemory_c` has 47, `ov006/dScMgTrampoline_c+MgTrampolineTime` 44,
> `ov006/dScMgTrampoline2_c` 42, `ov006/dScMgCard_c` 37, and `ov014/daWanwan_c` and
> `ov006/dScMgJump2_c` 29 each. It is a large TU and the four `dScMgD3DBase_c` children
> above it in that list are its own descendants, but the superlative is wrong and is not
> load-bearing for anything below.

## Proof block

```
tubuild verify   27/27 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tubuild linkcheck (pre-promotion)  SCRATCH-LINK-VERIFIED; whole 0x020e6c28..0x020e7660
                 range reproduced from one object; 106/106 modules exact;
                 == pristine-tree baseline control
tu_order_check   ALL MATCH, ROM order, nothing unlicensed
rombuild -j6     106/106 exact, 100.000000% of compared bytes, mismatching 0
                 ROM-build analysis: PASS
romdata_check    11 symbols: 6 VERIFIED (200 bytes equal), 5 PARTIAL (52 bytes equal), 0 DIFFERS
```

All eighteen gates in `notes/agents/roles/builder.md`'s order were re-run in a fresh
worktree at a named ref, and again after the correction below. All exit 0.

`tubuild linkcheck` was **not** re-run here: once the source is enrolled in `delinks.txt`,
`cmd_linkcheck` routes down the intact path unconditionally and dies at `[4/8]`. It fails
identically for untouched classes on `main`. The writer's pre-promotion run is the evidence.

### The 27 functions

The 27 declared function sizes sum to `0xa38`, exactly the licensed range
`0x020e6c28..0x020e7660` — no gaps, nothing unaccounted. They split as the scout's fact
file predicted: **17 class methods (`0x6e0`)** and **10 free helpers (`0x358`)**.
`config/arm9/overlays/ov006/delinks.txt` collapses 27 entries into one contiguous range,
with the neighbours meeting it exactly (`d_s_mg_curling2.c` ends at `0x020e6c28`,
`_ZN12dScMg3DEsp_cD1Ev.cpp` starts at `0x020e7660`).

## Class identity

| | |
|---|---|
| `_ZTS14dScMgD3DBase_c` | `ov006:0x0213c5d4` → `"14dScMgD3DBase_c"` |
| `_ZTI14dScMgD3DBase_c` | `ov006:0x0213c5c8`, `abi::__si_class_type_info` (`_ZTVN3abi20__si_class_type_infoE` @ `arm9:0x0209a764`), base pointer → `ov004:0x020bbf6c` = `_ZTI11dScMgBase_c` (`"11dScMgBase_c"`) |
| `_ZTV14dScMgD3DBase_c` | `ov006:0x0213c62c` (address point) |

Chain: `dScMgD3DBase_c` → `dScMgBase_c` → `dScene_c` → `dBase_c` → `fBase_c`, five levels,
single inheritance at every edge.

**No factory and no `new`-size literal.** The class is not abstract — all 36 vtable slots
hold real code pointers, no pure-virtual zero words — but nothing in the image allocates
one directly. The only stores of `_ZTV14dScMgD3DBase_c` are the mid-construction base vptr
stores in the four children (`d_s_mg_jump2.cpp`, `d_s_mg_trampoline.cpp`,
`d_s_mg_trampoline2.cpp`). The class boundary `0x5004` therefore comes from the field-span
reasoning in `include/dScMgD3DBase_c.h`, not from an allocation size.
`check_header_offsets` independently confirms it: *22 commented fields, 0 mismatched,
struct spans 0x5004*.

### Vtable extent — proven from the ROM, not inferred

```
0x0213c624  0x00000000            offset-to-top
0x0213c628  0x0213c5c8            -> _ZTI14dScMgD3DBase_c
0x0213c62c  <address point, _ZTV symbol>
            36 consecutive code pointers
0x0213c6b8  last slot
0x0213c6bc  0x6f6d6168            "hamon" -- carries its own symbol data_ov006_0213c6bc
```

`0x0213c6bc - 0x0213c624 = 0x98`. The emitted `_ZTV` section in the compiled object is
**152 bytes = 0x98** — two independent sources agreeing. There is no 37th slot and no
trailing null slot in which an undeclared override could hide, so the emitted table is the
**whole** table, not a prefix.

### Overrides — slot-by-slot against the base

Diffing `_ZTV14dScMgD3DBase_c` against `_ZTV11dScMgBase_c` (`ov004:0x020bc0c0`) word for
word gives **17 overridden slots, which are exactly this TU's 17 class methods** — every
override has a body here, and every class method here is an override. Nothing implicit,
nothing inherited silently.

| slot | method | | slot | method |
|---|---|---|---|---|
| 1 | `BeforeInitResources` | | 25 | `OnPushed` |
| 2 | `AfterInitResources` | | 26 | `OnHitByCannonBlastedChar` |
| 5 | `AfterCleanupResources` | | 27 | `OnHitByMegaChar` |
| 7 | `BeforeBehavior` | | 28 | `OnHitFromUnderneath` |
| 10 | `BeforeRender` | | 29 | `OnAimedAtWithEgg` |
| 11 | `AfterRender` | | 30 | `OnAimedAtWithEggReturnVec` |
| 16 | `~dScMgD3DBase_c` (D1) | | 31 | `Virtual7C` |
| 17 | D0 (deleting) | | 33 | `Virtual84` |
| 24 | `OnKicked` | | | |

Eight of these (1, 2, 5, 7, 10, 11, 16, 17) re-override slots inherited from
`dScene_c`/`fBase_c`; the other nine (24–31, 33) are among `dScMgBase_c`'s own 18 new
slots (18–35). The class adds no new slots of its own — base and derived tables are both
36 words.

### Destructor placement

`~dScMgD3DBase_c()` is **defined inline in the class body** and must stay there. All four
children write this class's vptr store, `mSysTracker` destruction and the chain to
`dScMgBase_c`'s D2 out inline in their own D1s, and `_ZN14dScMgD3DBase_cD2Ev` exists
nowhere in the ROM — an out-of-line definition would leave every child with an undefined
external.

The out-of-line D1/D0 pair at `0x020e6c28`/`0x020e6c60` is emitted anyway, by the **key
function**: `BeforeInitResources` is the first *declared* non-inline virtual (the
destructor is declared first but, being inline, cannot be the key function), and this TU
defines it. That makes mwcc emit `_ZTV14dScMgD3DBase_c` here; slots 16 and 17 odr-use D1
and D0, so the pair is emitted out of line, D1 first — exactly the order retail has. The
`_EmitDestructor`/`_EmitDeletingDestructor` scaffolds the two legacy shards carried are
deleted rather than ported: they were props for files with no key function to lean on, and
keeping them would add two unlicensed `.text` symbols and change the vtable's data.

## Promotion route

**Text-only + `compiler_only_output`**, not intact-object. The TU licenses `.text` only;
the 11 data symbols the compiler emits as vague linkage are externalized and discarded by
production isolation. The manifest carries all 13 keys universal to the 95 promoted
manifests; the keys it lacks (`production_mode`, `relocations`, `partial_isolation`, …) are
intact-object-route keys that do not apply.

**The 11 rows were verified independently, not taken on trust.** The count follows from
inheritance depth — `2 × (5 levels) + 1 vtable = 11`. It is 11 and **not** 12 because this
class has no `Vector3` member, so `writer.md`'s conditional trailing term does not apply;
the header confirms no such member exists. Every canonical module and address was checked
against the *owning* module's own `symbols.txt`, and all 11 matched exactly. Eight live
outside `ov006`:

| module | symbols |
|---|---|
| `arm9` (6) | `_ZTI7fBase_c` `0x02086d70`, `_ZTS7fBase_c` `0x02086e60`, `_ZTI7dBase_c` `0x02086e78`, `_ZTS7dBase_c` `0x02086e54`, `_ZTI8dScene_c` `0x020914d4`, `_ZTS8dScene_c` `0x020914b0` |
| `ov004` (2) | `_ZTI11dScMgBase_c` `0x020bbf6c`, `_ZTS11dScMgBase_c` `0x020bbf84` |
| `ov006` (3) | `_ZTI14dScMgD3DBase_c` `0x0213c5c8`, `_ZTS14dScMgD3DBase_c` `0x0213c5d4`, `_ZTV14dScMgD3DBase_c` `0x0213c62c` |

The compiled object emits exactly these 11 and no others; the manifest's 17
`droppedSections` are these 11 `.data` sections plus their 6 `.rela.data` sections —
nothing licensed is dropped.

The **sibling oracle** `dScMgSingle3DBase_c` also needs 11, but only because it sits at the
same depth under the same base. The count here was derived, not copied.

## What I corrected

The writer's 27/27 reproduced exactly, and every figure in the proof block held. One real
defect, in the class's own evidence trail:

**`include/dScMgD3DBase_c.h` claimed its vtable was "a byte-exact 34-slot prefix".** That
was stale — written while slots 34 and 35 were still unnamed — and wrong in both the count
and the kind. `dScMgBase_c.h` now declares all eighteen of its own slots through 35, so the
emitted table is the whole 36-slot table. Left standing, prefix language is precisely the
sentence that would later excuse a short table, and it flatly contradicted the manifest's
own central claim in this same PR. Replaced with the ROM read above, plus a note not to
reintroduce it without re-reading those words.

The header was unmodified on `main`, so `check_header_offsets` had nothing to check and the
contradiction was invisible to CI. Fixing it brings the header into the gate's scope, where
it now passes and confirms the `0x5004` span.

**Two claims the writer honestly flagged as inference are now measured.** `romdata_check`'s
CLI prints only bucket totals, so I recovered the per-symbol verdicts by importing the
module and calling `check_object()` directly:

- the **5 PARTIAL are exactly the five `_ZTS` strings** (emitted 9+9+10+14+17 = 59, equal
  8+8+8+12+16 = 52) — an alignment-padding artifact, benign;
- the **vtable is VERIFIED**, all 144 emitted body bytes equal to the cartridge.

Also recorded: the tool's `romExtent` for the vtable reads `0x50`, not `0x90`, because
`ov006`'s `symbols.txt` carries an interior phantom `data_ov006_0213c67c` at slot 20 that
truncates the extent lookup. The comparison still covered all 144 emitted bytes, but that
`0x50` must not be read as the table's size.

The header edit is comment-only; the bytes were re-proven after it (`tubuild verify` 27/27,
`rombuild` 106/106 exact, all eighteen gates re-run).

## Gates I had to `--update`

**None.** `tiers_ratchet --check` already passed (`baseline 2674, current 2674`) because
the writer banked correctly and this branch sits directly on an unmoved `main`
(`379bd67f9`), so no rebase disturbed the two ratchet files. Running `--update` anyway
would have appended duplicate rows for the same paths — `builder.md`'s documented
double-banking defect. I audited instead: **301 rows, 301 distinct, zero byte-identical
duplicates**, with 53 paths legitimately recurring under different reasons (dedupe by path
would destroy real history). Four rows out of the CONVERTED set, four back in as
`src/actors/dScMgD3DBase_c.cpp#<member>` — a clean many-to-one fold.

`config/converted-baseline.json` and `config/converted-backslide-exceptions.jsonl` are the
writer's, untouched. `notes/data/c-cpp-classification.tsv` and
`notes/data/tu-merge-candidates.json` are left at `main`'s version, as `builder.md`
requires.

## Two validator lines that are expected, not losses

- **`Contributor credit: 0 added, N changed, 0 lost`** — the pre-push hook restamping
  authorship across the fold. Nothing is lost.
- **`N address range(s) left the byte-verified set`** — the counter reading a many-to-one
  fold: 27 shard ranges became one range. The same run reports
  `Relocation check: N checked; N VERIFIED`.

## Base

`main` @ `379bd67f9`, re-fetched immediately before pushing and unmoved. No stacking: the
queued siblings (#2281 `daOts_c`/ov064, #2283 `daPgDfdr_c`/ov027, #2284
`daObjCtMecha04_c`/ov065) are all in other overlays and none has landed.

Do not merge — builder does not merge its own PR.


andrewboudreau:
Superseded by #2294, which landed this promotion unchanged as part of a six-class integration (main `e98de4be8`).

This PR was validator-green on its own, but `main` moved four times today and all six remaining promotions then conflicted on `config/converted-baseline.json` and `attribution.json` — files every promotion regenerates wholesale, so the conflicts were bookkeeping rather than content. Landing them one at a time would have meant six rebase-and-revalidate cycles, each one re-breaking the other five. Composing them once and gating the merge tree cost a single validator run instead.

`dScMgD3DBase_c` is on `main` at `src/actors/dScMgD3DBase_c.cpp`, 27 functions, manifest `status: promoted`.

Closing as merged-elsewhere. The branch is left in place; nothing here is lost.

PR #2287, opened 2026-09-05
ov091: promote daDsnBase_c into a single C++ TU (9 of 11)

andrewboudreau:
## TL;DR for a reader who does not live in this repo

`daDsnBase_c` is the abstract base of SM64DS's crushers — the Thwomp that drops on you
and the Grindel that slides. Its code was scattered across eleven single-function files,
each a stub the tree had recovered separately. This PR gathers them back into the one
C++ file they were originally compiled from, `src/actors/daDsnBase_c.cpp`, and proves
byte-for-byte that the compiler turns that file back into the exact machine code on the
cartridge. **The class already had a real header; this is gathering, not authoring.**

**Nine of the eleven functions move into the new file. Two do not, and that is a fact
about the cartridge rather than unfinished work.**

The two hold-outs are the class's destructors. C++ compilers emit a destructor two
different ways — `D1` (destroy this object) and `D0` (destroy it *and* free its memory) —
and the compiler decides what order they land in. The cartridge stores them **D0 first,
then D1**. Every source form this compiler accepts emits them **D1 first, then D0**; the
one form that reverses the order also emits a third variant (`D2`) that the cartridge
does not contain at all, so it cannot be right either. That lever set was measured to
exhaustion in earlier work and is closed.

So this is not "two functions we could not decompile." **Both destructors byte-match
perfectly** — the recovered source is correct. Only the *order* the compiler files them
in is unreproducible, and the linker gate rightly refuses to license a block of code
whose internal order disagrees with the ROM. The two therefore stay in their own existing
files, which already reproduce them correctly, and this file's duplicate copies are
declared compiler-only output and discarded.

Nothing is lost by that split, because **nothing in the game ever calls those two
addresses.** They are reached only through the class's virtual dispatch table, at slots
16 and 17. I verified this against the raw cartridge rather than the config: see below.

---

## Why the split is safe — verified against the cartridge, not the config

The load-bearing claim is that the destructor pair is unreachable except through the
vtable, so its position cannot matter to anything.

I re-proved it directly from the ROM dump. A **byte-aligned** scan (not merely
word-aligned) of **all 109 module binaries** — `arm9`, `arm9_dec`, `itcm`, `dtcm` and
every overlay, 4,074,812 bytes — for the 32-bit values `0x02132d04` (D0) and
`0x02132d6c` (D1) finds **exactly two occurrences in the entire image**:

| file offset | address | word | what it is |
|---|---|---|---|
| `ov091.bin +0x433c` | `0x0213523c` | `0x02132d6c` | `_ZTV11daDsnBase_c` **slot 16** → D1 |
| `ov091.bin +0x4340` | `0x02135240` | `0x02132d04` | `_ZTV11daDsnBase_c` **slot 17** → D0 |

A separate decode of every ARM `B`/`BL` instruction in `ov091` finds **zero** branches
resolving to either address. Cross-module branches would require a relocation entry, and
`config/` records none.

So the pair is referenced by two vtable words and nothing else. Their relative order is
invisible to the machine.

**Slot numbering checked, not assumed.** The `_ZTV11daDsnBase_c` symbol sits at
`0x021351fc`, and the two preamble words are *below* it — `0x021351f4` (offset-to-top,
zero) and `0x021351f8` (the RTTI pointer, relocating to `0x021137f8`). The symbol is
therefore the address point, so slot *n* is at `0x021351fc + 4n`, making `0x0213523c`
slot 16 and `0x02135240` slot 17. Storage runs `0x021351f4..0x0213527c` = `0x88`
(2 preamble words + 32 slots), and `0x0213527c` is exactly where the next symbol
(`_ZTI11daObjPile_c`) begins.

---

## Class identity

    _ZTS11daDsnBase_c   ov025  0x02113804   "11daDsnBase_c"
    _ZTI11daDsnBase_c   ov025  0x021137f8   __si_class_type_info, one base, offset 0
      base  dBgActor_c  ov002  0x021089ec
    _ZTV11daDsnBase_c   ov091  0x021351fc   32 slots, storage 0x88

**The RTTI record and the vtable live in different overlays**, which is unusual enough to
be worth stating plainly: the record is in **ov025**, the vtable in **ov091**. The
cartridge settles this on its own — the vtable's preamble RTTI word reads `0x021137f8`,
an ov025 address — so it is read off the image, not inferred from an ambiguous
relocation tag. `romdata_check`'s per-symbol verdicts confirm it from the other
direction: the compiled object's `_ZTI11daDsnBase_c` is **VERIFIED against ov025**
(12 bytes) while its `_ZTV11daDsnBase_c` is **VERIFIED against ov091** (128 bytes).

**No factory and no `new`-size literal, because the class is abstract.** Slots 0
(`InitResources`) and 6 (`Behavior`) are bare `0x00000000` with no relocation reaching
either — the signature of a pure virtual once the linker is done. Nothing in the image
allocates a `daDsnBase_c`; its two descendants do, and they disagree
(`daDkk_c_classInit` passes 928 = `0x3a0`, `Thwomp_Spawn` 932 = `0x3a4`), so no literal
fixes this class's own end. The header records `0x360` as the minimal reading and says so.

### Vtable diff against the base — this class overrides exactly 4 of 32 slots

| slot | target | |
|---|---|---|
| 0 | `0x00000000` | **pure virtual** (`InitResources`) |
| 3 | `_ZN11daDsnBase_c16CleanupResourcesEv` `ov091 0x021331b8` | **override — key function** |
| 6 | `0x00000000` | **pure virtual** (`Behavior`) |
| 9 | `_ZN11daDsnBase_c6RenderEv` `ov091 0x02133210` | **override** |
| 16 | `_ZN11daDsnBase_cD1Ev` `ov091 0x02132d6c` | **override** (stays enrolled) |
| 17 | `_ZN11daDsnBase_cD0Ev` `ov091 0x02132d04` | **override** (stays enrolled) |
| 31 | `_ZN10dBgActor_c4KillEv` | inherited — confirms the base |

The other 25 slots relocate into `arm9` and are inherited verbatim from `dActor_c` /
`fBase_c`. Slot 31 resolving to `dBgActor_c::Kill`, plus the 32-slot length, independently
corroborate the `__si_class_type_info` base clause.

### Destructor placement, and why

The destructor is defined **inline in the class body**. That is required independently of
anything above: both descendants (`daDsn_c` in ov091, `daDkk_c` in ov025) inline the body,
and an out-of-line declaration would make each of them emit a `bl` the ROM does not have —
and would move the ABI key function off `CleanupResources`.

`CleanupResources` is the key function (the first declared virtual that is neither inline
nor pure), so this TU emits the vtable and, with it, both destructor variants. The forcing
scaffolds the two destructor shards used to carry (`p->~daDsnBase_c();` and `delete p;`)
are therefore **dropped, not carried** — owning the key function alone emits both variants
byte-identically.

The cartridge contains **no `_ZN11daDsnBase_cD2Ev`** anywhere in `config/`, which is what
rules out the out-of-line form: it would emit a homeless `D2`.

### Promotion route

**Text-only + `compiler_only_output`.** Not intact-object. `sections` licenses one range,
`.text 0x02132dc0..0x02133254`; `externalized_output` is empty.

**Sibling oracle: there is none.** ov091 has no previously promoted TU to copy a style or
a policy from, which is why the scout commit measured the class facts directly out of
`extracted/dsd/arm9_overlays/ov091.bin` rather than adapting a neighbour's manifest.

### `compiler_only_output` — 15 rows, audited individually

Unusually large, so each row was checked against the **owning module's own**
`symbols.txt`. All 14 homed addresses match exactly:

- **11 × `deadstrip-data`** — 10 inherited/own RTTI records plus `_ZTV11daDsnBase_c`.
  Emitted as vague linkage because this TU owns the key function; each has a real ROM
  home outside the licensed `.text` range, so `dsd` supplies it independently.
  (`fBase_c`, `dBase_c`, `dActor_c` → `arm9`; `dBgActor_c` → `ov002`; `daDsnBase_c` →
  `ov025`; the vtable → `ov091`.)
- **3 × `deadstrip-duplicate`** — `D1`, `D0` (both byte-matching, left out of the licensed
  run for the ordering reason above; their enrolled shards stay canonical) and
  `_ZN7Vector3D1Ev` (`arm9 0x020072c0`, pulled in because `func_ov091_02132f04` holds two
  `Vector3` locals).
- **1 × `deadstrip`** — `_ZN25@class$424daDsnBase_c_cppD1Ev`, the implicit destructor of
  the file-local anonymous struct mwccarm synthesises to hold those two `Vector3` locals.
  mwccarm names such a type after the translation unit, so this four-byte body exists only
  in this object.

**The plain `deadstrip` row is the only genuinely homeless address, and I confirmed the
homelessness rather than assuming it:** that symbol appears **nowhere in `config/` except
this manifest**, and it is the **only `@class$` symbol in the entire config tree** — no
`symbols.txt` in any module names it. It is also the first plain `deadstrip` any class in
this campaign has needed.

---

## Proof block

    tubuild verify    9/9 MATCH, objisolate clean, reloc-destinations clean,
                      emitted .text ROM-ascending                  -> TEXT-VERIFIED
    tubuild linkcheck (pre-promotion, writer's run) SCRATCH-LINK-VERIFIED
                      [4b/8] 9 LICENSED / 0 COLLIDES-GAP / 0 COLLIDES-SRC / 0 HOMELESS,
                      whole range 0x02132dc0..0x02133254 IDENTICAL from one object,
                      ROM sha256 d1506e90... identical to the stock control
    rombuild -j6      106/106 exact, 100.000000% of compared bytes, mismatching 0,
                      ROM-build analysis: PASS
    romdata_check     6 VERIFIED, 5 PARTIAL, 0 DIFFERS  (11 data symbols, 0 UNNAMED)
    tu_order_check    all match in ROM order; 4 extras all carry compiler-only policy

`linkcheck` is deliberately **not** re-run here. Once a source is enrolled in
`delinks.txt`, `cmd_linkcheck` sets `enrolled_intact_candidate = True` unconditionally and
routes down the intact path regardless of `production_mode`, dying at `[4/8]`. That is
pre-existing — it fails identically for `ov002/daBar_c` on untouched `main` — so the
writer's pre-promotion run is the evidence. My `rombuild` run reproduces the same baseline
control ROM (`sha256 d1506e90…`) independently.

The baseline control's `dsd check symbols --fail` failure (three ITCM symbols) is
pre-existing and present in the stock control with no TU substitution: 0 new, 0 resolved.

## Gates

All eighteen gates in the role's order pass. **No gate needed `--update`.**

`tiers_ratchet --check` is **PASS (baseline 2674, current 2674)** as it stands. The writer
had already banked the rows and this branch sits directly on current `origin/main`, so
re-running `--update` would only append duplicate records — the exceptions file opens `"a"`
and never replaces. Audited: **0 byte-identical duplicate records** across all 298 rows
(53 paths legitimately recur under *different* reasons, which is correct and must not be
deduped by path).

`check_dead_references` is clean — no prose file named any of the nine deleted shards.
`port_refcheck` green (402 checked, 0 stale). `check_src_tu_compiles` genuinely compiled
127/127 rather than taking its silent `NOT CHECKED` path.

Base is `main`: `origin/main` (`379bd67f9`) is a direct ancestor of this branch, re-fetched
and re-verified immediately before pushing. No stacking, no rebase, no sibling dependency.

## Two validator lines that are expected and are not losses

- `Contributor credit: 0 added, N changed, 0 lost` — the hook restamping authorship across
  the fold.
- `N address range(s) left the byte-verified set` — the counter reading a many-to-one fold
  (9 delink entries becoming 1). The same run reports `Relocation check: N checked;
  N VERIFIED`.


andrewboudreau:
Superseded by #2294, which landed this promotion unchanged as part of a six-class integration (main `e98de4be8`).

This PR was validator-green on its own, but `main` moved four times today and all six remaining promotions then conflicted on `config/converted-baseline.json` and `attribution.json` — files every promotion regenerates wholesale, so the conflicts were bookkeeping rather than content. Landing them one at a time would have meant six rebase-and-revalidate cycles, each one re-breaking the other five. Composing them once and gating the merge tree cost a single validator run instead.

`daDsnBase_c` is on `main` at `src/actors/daDsnBase_c.cpp`, 9 of 11 functions, manifest `status: promoted` — the deliberate partial, with the destructor pair held out for ROM emission order and proved unreachable.

Closing as merged-elsewhere. The branch is left in place; nothing here is lost.

PR #2292, opened 2026-09-05
Document eight mwccarm colouring levers from the ov006 third wave

andrewboudreau:
Moves the 6bu compiler-codegen findings out of matching PR #2291 so that source matching and notes remain independently reviewable, as required by AGENTS.md. Content is unchanged from the original notes-only commit.


andrewboudreau:
Review complete: no blocking findings. This is notes-only, the eight described levers agree with the committed ov006 matched-source spellings and the explicitly labelled func_ov006_020e1854 near-miss evidence, and all public plus private checks are terminal green. The PR is now over four hours old. Formal APPROVE is unavailable because the authenticated account is the PR author; this comment records the review evidence instead.

PR #2295, opened 2026-09-05
Refresh the TU-promotion queue: it disagreed with both the manifests and the filesystem

andrewboudreau:
## TL;DR

`notes/data/tu-promotion-queue.tsv` is how an agent picks its next class. Two of its columns had drifted out of date, and both errors pushed in the direction of *doing less work* — one by hiding classes that are ready, the other by offering classes already finished. Nothing here changes code; it changes what the next wave sees.

## `already_promoted` — 11 rows

Every promoted class was still listed as open, including the eight that landed today. A stale open row is how a wave duplicates work already on `main`, and the claim lock cannot catch it: a released claim leaves no ref behind.

Re-derived from ground truth rather than hand-edited — a row is `already_promoted` when every class it names has a `config/tu_manifest.d` entry reading `status: promoted`. 102 manifests qualify.

## `has_header` — 161 rows

This was the bigger one. The column read `unknown` for 113 open rows, which reads as "no header, not a candidate". Spot-checking the first dozen found a header on disk for **every one** — `Ukiki`, `OneUpMushroom`, `Scuttlebug`, `MrBlizzard`, `Pokey`, `BobOmb`, `BabyPenguin`, `Rabbit`, `SnowmanBody`, `MovingBar`, `StarSwitch`, `Klepto`. The generator's detector had simply failed to resolve them.

Re-derived by indexing every basename under `include/`. It was wrong in both directions:

| change | rows |
|---|---|
| `unknown` → `yes` | 143 |
| `yes` → `partial` | 9 |
| `unknown` → `partial` | 3 |
| `no` → `partial` | 4 |
| `unknown` → `no` | 1 |

The 9 `yes → partial` matter too — each would have sent a writer at a multi-class row missing one of its headers.

## Why it matters

Clean single-class rows with a header — no unmatched functions, no missing legacy source, not multi-class:

| | rows | shards |
|---|---|---|
| before | 16 | 95 |
| after | **129** | **1,674** |

Read the old column literally and you would conclude the easy tier was exhausted after today's eight promotions, and the next wave had to attack blocked or multi-class rows. It is not exhausted — it had barely been touched.


andrewboudreau:
Review complete on exact head `39395d827`. I fixed the current-main semantic drift by replacing the stale `StarDoor` row with ROM-backed `daStarGate_c` and marking it promoted. The queue now has 228 rows / 12 `already_promoted=yes` rows, with no represented promoted manifest still marked `no`; `dScGameOver_c` correctly remains unpromoted while #2296 is open. Merge against current main is clean, public checks are green, and terminal private PR validation passed. No remaining findings. Leaving open because it is under the four-hour merge threshold.

PR #2296, opened 2026-09-05
ov003: promote dScGameOver_c into a single C++ TU

andrewboudreau:
## TL;DR for a non-specialist

`dScGameOver_c` is the "GAME OVER" screen: Mario is out of lives, eight glyphs
drop in from off screen, and a yes/no continue prompt appears on the touch
screen.

In the original 2004 source it was **one C++ file**. This project had it as
**nine separate one-function files**, because the recovery tool that carves the
cartridge up works a function at a time and cannot see file boundaries. This PR
puts the nine back into one file, `src/actors/dScGameOver_c.cpp`.

The only thing that makes that safe is that the rebuilt file has to produce
**byte-for-byte the same machine code as the retail cartridge** -- not similar,
identical. It does: all nine functions, and the whole 2,968-byte run of memory
they occupy, and the finished 16 MB ROM image has the same SHA-256 as one built
without the change at all.

One detail is worth calling out because it drove the design. C++ compilers emit
a destructor two or three times over, and mwccarm 2004/b56 emits those copies in
a different order depending on *how the destructor is written in the header*.
The cartridge has them in the order that only the `~dScGameOver_c() {}` form
(defined inline in the class) produces. Writing it the other way produces the
wrong order plus a third copy that appears nowhere in the cartridge. So the
header uses the inline form, and this promotes **whole -- all 9 of 9** -- rather
than holding the destructor pair out.

---

## Proof block

Re-run independently by the builder in its own worktree at
`cpp/dScGameOver_c-tu-build`, on a tree identical to `origin/main` plus these
two commits.

```
tubuild verify                    9/9 MATCH, objisolate clean, reloc-destinations
                                  clean, emission order ROM-ascending -> TEXT-VERIFIED
tubuild linkcheck (pre-promotion) SCRATCH-LINK-VERIFIED: [4b/8] 9 LICENSED / 0 COLLIDES-GAP /
                                  0 COLLIDES-SRC / 0 HOMELESS; [6/8] TU .text range 2968 bytes
                                  IDENTICAL, module fidelity 106/106; [8/8] ROM sha256
                                  d1506e90...c478e8, IDENTICAL to the stock control
rombuild -j6 --no-rom             106/106 exact, 100.000000% of compared bytes,
                                  mismatching 0; ROM-build analysis: PASS
romdata_check                     9 symbols: 6 VERIFIED (132 bytes), 3 PARTIAL, 0 DIFFERS
tu_order_check                    ALL MATCH, ROM order, nothing unlicensed
```

The `linkcheck` line is the **writer's pre-promotion** run, and that is the
correct evidence for it. `tubuild.py linkcheck` cannot pass on an
already-promoted TU: once the source is enrolled in `delinks.txt`,
`cmd_linkcheck` sets `enrolled_intact_candidate = True` unconditionally and dies
at `[4/8]` whatever `production_mode` says. That is pre-existing and reproduces
on untouched `main` for other classes, so per the builder role it was not re-run
here.

### Every other gate, all exit 0

```
tiers_ratchet --check          CONVERTED ratchet PASS   baseline 2671  current 2671
eligible.py -j6                11095 / 11234
check_references               42 unresolved (baseline 45); no new unresolvable references
cpp_tu_compat --require-ready  8/8 READY
check_src_tu_compiles          135/135 translation units compile
check_rename_ledger            1818 mangled/vtable rows agree with symbols.txt
check_profile_campaign         OK
langmode_audit --check         langmode ratchet PASS
check_header_offsets           dScGameOver_c header: 10 commented fields, 0 mismatched, spans 0x98
port_refcheck                  402 checked, 0 stale
check_duplicate_sources        10323 stems, none doubled
check_dead_references          no new dead references, no broken markdown links
check_tubuild_conflicts        135 manifest entries; every recorded conflict matches its source
layout_check                   clean
```

**No gate was `--update`d.** `tiers_ratchet --check` passed, so per the role file
`--update` was deliberately *not* run -- it appends rather than replaces and
would have double-banked. The writer's banking was audited instead, two ways:

- `config/converted-backslide-exceptions.jsonl`: 323 to 327 rows. The 4 added
  rows are this class's four shards, and there are **0 new byte-identical
  whole-record duplicates** (and 0 pre-existing ones on the base).
- `config/converted-baseline.json["converted"]`: 2671 to 2671. Exactly four
  paths left and four `path#member` rows arrived, one for one. Nothing else
  moved, so no sibling's rows were clobbered.

---

## Class identity

**`dScGameOver_c : dScene_c`**, ov003, `.text 0x020b0580 .. 0x020b1118`
(2,968 bytes, 9 functions, contiguous and gapless).

| record | module | address | verdict |
|---|---|---|---|
| `_ZTS13dScGameOver_c` (the string `13dScGameOver_c`) | ov003 | `0x020b1784` | VERIFIED, 16 bytes |
| `_ZTI13dScGameOver_c` (`__si_class_type_info`, base pointer to `_ZTI8dScene_c`) | ov003 | `0x020b1758` | VERIFIED, 12 bytes |
| `_ZTV13dScGameOver_c` (address point; header word at `0x020b1794`) | ov003 | `0x020b179c` | VERIFIED, 72 bytes |

The base is proven twice over. The typeinfo record's `+0x08` word is
`0x020914d4`, which arm9's own symbol table names `_ZTI8dScene_c`; and the
vtable is 18 slots, exactly `dScene_c`'s own 18, so the class adds no new
virtual.

**`new`-size literal: `0x98` (152 bytes).** `tools/opnew_sizes.py` finds a single
unambiguous site at `0x020b1120`, inside the factory
`ov003:dScGameOver_c_classInit` (`0x020b1118`, size `0x64`), confidence
`inline_store`. That agrees with `check_header_offsets`, which independently
reports the header's struct spanning `0x98`. The factory begins at exactly the
byte where this TU's licensed range ends, which is its own corroboration of the
boundary.

**Vtable extent: 18 slots, proven rather than inferred.** The known `_ZTV`
extent-overrun defect makes `rtti_vtables.py`'s slot count non-authoritative, so
it was corroborated from a second direction: `romdata_check` compares the
emitted vtable against the cartridge with relocations resolved and returns
`_ZTV13dScGameOver_c ov003 VERIFIED 72 bytes`, and 72 = 18 x 4. The scout also
walked it word by word off the linked overlay and stopped at `0x020b17e4`, where
the next word (`0x81f000f0`) is not a code address in any loaded module and
ov003's symbol table starts a new data symbol. Three sources, one answer.

**Slot by slot against the base -- 7 overrides, 0 new virtuals:**

```
slot 0   0x020b0b3c  InitResources
slot 3   0x020b0b34  CleanupResources
slot 6   0x020b0894  Behavior
slot 9   0x020b0814  Render
slot 12  0x020b0810  OnPendingDestroy
slot 16  0x020b0580  ~dScGameOver_c  [D1]
slot 17  0x020b05bc  ~dScGameOver_c  [D0]
```

Those seven are exactly the seven `dScene_c` gives every direct child. The
remaining two functions in the run are free helpers that carry no mangled name
in the ROM: the glyph drop-in animation step (`0x020b060c`) and the sub-screen
yes/no highlight repaint (`0x020b0730`).

### Destructor placement, and why

**The cartridge puts D1 at `0x020b0580` below D0 at `0x020b05bc`.** That is the
reproducible direction, and inline-in-class is the only admissible source form
that produces it -- out of line, mwcc emits D2, D0, D1, plus a D2 that exists
nowhere in the ROM. The header therefore moves `virtual ~dScGameOver_c();` to
`virtual ~dScGameOver_c() {}`. Because the destructor is inline it cannot be the
key function, so the key function is `InitResources`, the first *declared*
non-inline virtual, and this TU defines it. That emits `_ZTV13dScGameOver_c`,
whose slots 16 and 17 odr-use the destructor pair and drag both out of line, D1
first. **No forcing scaffold is needed.** The same inline form is already used by
`dScene_c` and `dBase_c` one and two levels up; `fBase_c` at the root is the
out-of-line one, which is why D1's tail is a `bl` to it.

This is why the promotion is **whole, 9 of 9**, with no partial hold-out.

### Promotion route: text-only plus `compiler_only_output`

Not intact-object. `externalized_output` is empty (absent, which is the majority
form -- 84 of the 103 promoted manifests carry no such key), and
`compiler_only_output` carries **9 rows, all `deadstrip-data`**.

Nine is what the chain predicts and what the object actually emits. The chain is
`dScGameOver_c -> dScene_c -> dBase_c -> fBase_c`, four levels, so
**2 x 4 + 1 vtable = 9**, with **no `Vector3` term**. That term counts what the
TU *touches*, not only what it stores as a member, and this TU touches none --
neither as a member nor as a function local. That is measured, not argued: the
compiled object emits exactly nine data symbols and `romdata_check` reports
`UNNAMED 0`, so there is no tenth symbol hiding outside the config.

Every canonical address was checked against **the owning module's own** symbol
table, and corroborated from the other direction with
`romdata_check.check_object(..., names=romdata_check.name_index())`:

```
_ZTI13dScGameOver_c  ov003  0x020b1758  VERIFIED 12
_ZTS13dScGameOver_c  ov003  0x020b1784  VERIFIED 16
_ZTV13dScGameOver_c  ov003  0x020b179c  VERIFIED 72
_ZTI8dScene_c        arm9   0x020914d4  VERIFIED 12
_ZTS8dScene_c        arm9   0x020914b0  PARTIAL   8
_ZTI7dBase_c         arm9   0x02086e78  VERIFIED 12
_ZTS7dBase_c         arm9   0x02086e54  PARTIAL   8
_ZTI7fBase_c         arm9   0x02086d70  VERIFIED  8
_ZTS7fBase_c         arm9   0x02086e60  PARTIAL   8
```

All nine resolve to exactly one definition in the module named. The three
`PARTIAL`s are the three ancestor typeinfo *name strings*, whose ROM extent runs
past what this object emits; `DIFFERS` is 0. Cross-module homes are normal here:
the six arm9 records are the ancestors' canonical copies, outside this TU's only
licensed range, so `dsd` delinks them from the cartridge independently and
discarding this object's vague-linkage duplicates cannot cost the image a byte.

**Sibling oracle:** `dScene_c`'s other direct children -- `src/d_s_game_over.cpp`
sits immediately above this run in the same overlay -- use the identical inline
destructor form and the identical 7-slot override set. That is what
`boundary_evidence` leans on to say the run cannot be missing an eighth own
method.

`boundary_confidence` stays **medium**, deliberately. The facts file's
`unproven` list is binding and says why: the two helpers take this object's
address and are called only from inside the class region, which is *consistent
with* static file-scope helpers but does not prove static versus external
linkage, nor the original translation-unit cut.

### The one pragma, and its blast radius

`src/func_ov003_020b060c.c` carried `#pragma opt_strength_reduction off`, which
`tubuild create` dropped with a `[NOT carried -- review]` note. It is carried
here at **file scope**, because in mwccarm 2004/b56 the pragma is file-global
last-wins -- bracketing it around one member does nothing, since the trailing
`on` wins for the whole file. There is no narrower place to put it.

The builder did not take that on trust. Removing the pragma and re-running
`verify` gives **8/9 MATCH**: exactly one function breaks --
`func_ov003_020b060c`, the one the writer named -- and the other eight still
MATCH. So the pragma is genuinely load-bearing, and its blast radius on the rest
of the TU is measurably zero. (The tree was restored and re-verified 9/9
afterwards.)

### Merge conflicts reconciled by hand

Three type conflicts that `tubuild`'s detector does not report, because it
compares declaration spellings and not types:

- `data_0209d45c` and `data_0209d454` arrived as `unsigned char x[]` from one
  shard and `u8 x` from another -- an array against a scalar, a hard compile
  error once merged. The scalar wins; it is what the rest of the tree uses for
  these two BG-enable registers, and the `[0]` subscripts became scalar reads.
- `func_ov003_020b060c` was defined taking `char *` against the common header's
  `void *` declaration, giving `illegal function overloading` reported against
  the *definition* line and never naming the header. The header wins; the
  `char *` view survives as a local alias so the raw-offset accesses keep their
  legacy spelling.
- The legacy shadow types `VT` and `Obj` collided with the common header's real
  ROM symbol `extern int VT[];`. A variable hides a struct tag in C++, so this
  surfaces as `undefined identifier 'VT'` on the struct that *uses* it, never on
  the header. Renamed `GameOverGateVt` and `GameOverGate`; both are file-local
  shadow types with no ROM identity.

A fourth, which `create` *did* report: the two shards declared
`_ZN3G2S12GetBG1ScrPtrEv` as `unsigned short *` and as `void *`. The
`unsigned short *` spelling is kept -- it is the only one that types
`func_ov003_020b0730`'s pointer arithmetic and 16-bit stores, and
`InitResources` only feeds the result to `DecompressLZ16`'s `void *` parameter,
which it converts to implicitly.

### Prose re-read against what shipped

Checked, since no gate looks at it. The manifest's `boundary_evidence`, all four
`notes`, and the comment blocks in both the header and the `.cpp` all describe
the inline-destructor, whole-9-of-9, text-only state that actually shipped. No
superseded rationale survives: none of `SHADOW translation unit`,
`NOT ENROLLED`, `NOT CANONICAL`, `contributes nothing`, `PROMOTION IS REFUSED`
or `IT COSTS THE TU PROMOTION` appears anywhere in the three files. Each
individual prose claim was re-checked against the shipped source -- the
`unsigned short *` declaration, the two scalar `u8` externs, the renamed shadow
tags and the `void *` helper signature are all present as described.

---

## Validator output, explained

`PR validation` is **green**. Four of its lines look like losses and are not.
All four are the same many-to-one fold artifact: nine one-function delinks
ranges become one range, so the counter credits only that range's *first* symbol
and reclassifies the other eight as "claimed".

```
Contributor credit: 0 added, 5 changed, 0 lost
9 address range(s) left the byte-verified set (enrolled totals held steady)
8 more function(s) now claim a match that nothing compiles
Byte-verified functions -8, code bytes -2,908
```

- **-8 functions / -2,908 bytes** reconciles exactly: the run is 2,968 bytes and
  the credited first symbol is D1 at `0x3c` = 60 bytes, so 2,968 - 60 = 2,908.
- **9 ranges left the byte-verified set** because all nine one-function ranges
  were replaced by the single folded range; the validator's own line notes
  *enrolled totals held steady*, i.e. `Enrolled ranges ... +0`.
- **8 now "claim a match that nothing compiles"** is the same eight, seen from
  the other side of the fold.
- **5 credit reassignments** is the attribution fold, explicitly flagged
  `not a blocker`.

> **One reviewer note on that last line.** A third commit,
> `Preserve GameOver TU attribution and state`, was added to this branch by
> another pipeline agent after the promotion had already validated green. It
> touches only `attribution.json` and `notes/cpp-tu-current-state.md` — no
> byte-gate input — and every artifact this PR's proof rests on is byte-identical
> with and without it. But its effect on the ledger is worth a look before merge:
> it moves the five folded functions' credit to `github-actions[bot]`, including
> `func_ov003_020b060c` from **lunavyqo** and `func_ov003_020b0730` from
> **ruspecial**, both real contributors. Without that commit the fold credited
> them to the promoting author instead. Attribution is a documented non-goal and
> the validator calls it not a blocker, so this is flagged rather than reverted —
> but crediting a bot for two humans' matches looks like the wrong end state.

Nothing is actually unverified. The same run reports
**`Relocation check: 9 checked; 9 VERIFIED`**, **`Module fidelity: 106/106
exact; 100.000000% compared bytes`**, and the per-file detail
**`All 1 changed file(s) compile to the ROM byte-for-byte with correct
relocation targets`** — 9 slots checked, verified.

---

## Housekeeping

- `tools/classqueue.py status dScGameOver_c` reports **UNCLAIMED**. There is no
  `refs/claims/*` ref to release.
- Base: `main` moved during validation (`4f6866be4` to `5831b5833`, bringing the
  `daStarGate_c` promotion), which turned this PR `CONFLICTING` on the two
  ratchet ledgers. `config/converted-backslide-exceptions.jsonl` carries
  `merge=union`, and as `.gitattributes` itself warns, GitHub's merge does not
  honour that driver — so a file that merges cleanly here conflicts there. It was
  reconciled by the documented recipe (`reset --hard` onto the new base, then
  cherry-pick only this class's two commits — not `rebase --onto`), the merged
  ledgers were audited for lost rows and duplicate whole records (none of
  either), and **every gate was re-run from scratch on the new base**: 9/9 MATCH,
  106/106 exact, ratchet PASS 2680/2680, and all twelve remaining gates exit 0.
  The five load-bearing artifacts are byte-identical to the versions that
  validated the first time.
- Nine contiguous `delinks.txt` ranges fold into one
  `.text start:0x020b0580 end:0x020b1118`, gapless.

Do not merge on my account -- opened for validation and review.


andrewboudreau:
Review complete on exact head `668626b92`. I repaired the TU promotion's lost per-member attribution and refreshed the generated TU-state note; the local attribution gate now reports 5 tracked consolidations with credit intact, 0 changed, 0 lost. Local gates also pass TU compatibility, header offsets, ratchets, source coverage (0 bytes handed back), all 136 TU compiles, and port references. The full stock-ROM build has 0 mod/ROM-gap fallbacks, 11,107 source-built functions with 0 mismatches, 20/20 source-owned data claims reproducing, and 106/106 modules exact. All public checks and terminal private PR validation now pass on this SHA. No remaining findings. Leaving open because it is under the four-hour merge threshold.

PR #2297, opened 2026-09-05
ov025: promote daDkk_c into a single C++ TU

andrewboudreau:
## TL;DR (plain English)

The original game was built from `.cpp` files. Our tree had this class scattered
across **eight separate one-function source files**, each compiled on its own.
This PR puts those eight functions back into the single file they came from,
`src/actors/daDkk_c.cpp`, and tells the build that one file now supplies the
whole address range `0x021118c8..0x02111cf4` instead of eight files supplying
one function each.

Nothing about the shipped game changes. Every one of the eight functions still
compiles to **byte-identical** machine code, and the full 16 MB ROM built from
this tree has the same SHA-256 as the stock cartridge. What changes is that we
now assert something stronger than "each function matches": we assert that *one
translation unit, compiled once, in this order, with these includes* produces
that whole stretch of the ROM — which is what the original build actually did.

The one source-level change outside the merge is in `include/daDkk_c.h`: the
destructor moves from an out-of-line declaration to an in-class definition.
That is a byte requirement, not a style preference — see below.

**Expect the validator to report four "loss"-looking lines. They are not losses.**
They are the arithmetic of folding eight credited ranges into one; the section
"What the validator will say" reconciles the byte figure exactly.

---

## Proof block

```
tubuild verify                     8/8 MATCH, objisolate clean,
                                   reloc-destinations clean, emission order
                                   ROM-ascending, 0 unlicensed  -> TEXT-VERIFIED
tubuild linkcheck (pre-promotion)  SCRATCH-LINK-VERIFIED
                                   .text 0x021118c8..0x02111cf4 differingBytes 0
                                   106/106 modules exact
                                   ROM sha256 d1506e90...c478e8, matchesStockRom true
                                   moduleSetSha256 f9852ffa...dd802b
dsd check symbols --fail           still FAILS on the same 9 pre-existing errors
                                   the baseline control has: 0 new, 0 resolved.
                                   NOT green, and not reported as green.
relocation destinations            PASS (reloc_audit check_destinations);
                                   45 relocation records originate inside this
                                   range in config/arm9/overlays/ov025/relocs.txt
```

`tubuild linkcheck` is deliberately **not** re-run post-promotion: once the
source is enrolled in `delinks.txt`, `cmd_linkcheck` routes down the intact path
and dies at `[4/8]` regardless of `production_mode` (pre-existing; it fails the
same way for `ov002/daBar_c` on untouched `main`). The pre-promotion run above is
the evidence.

## Class identity

| fact | value | where it is proven |
|---|---|---|
| `_ZTS7daDkk_c` | ov025 `0x021137e0` | RTTI name string |
| `_ZTI7daDkk_c` | ov025 `0x021137ec`, `__si_class_type_info`, 0x0c bytes | `relocs.txt` lines 311-313 relocate exactly its three words |
| base `_ZTI11daDsnBase_c` | ov025 `0x021137f8` | third word of the typeinfo; its own base word points at `_ZTI10dBgActor_c` (ov002 `0x021089ec`) |
| ancestor chain | `daDsnBase_c -> dBgActor_c -> dActor_c -> dBase_c -> fBase_c` | read off the cartridge's own `__si_class_type_info` chain, not inferred |
| `_ZTV7daDkk_c` | address point ov025 `0x02113850`, storage `0x02113848..0x021138d0` = `0x88` = 2 preamble words + **32 slots** | see extent note |
| object size | **0x3a0 (928)** | `daDkk_c_classInit` (ov025 `0x02111cf4`) passes the literal 928 to `fBase_c::operator new` |

**Vtable extent is proven by relocations, not by symbol spacing.** `relocs.txt`
relocates every word from `0x0211384c` (the typeinfo pointer) through
`0x021138cc`, one per entry, and the next relocated word `0x021138d0` is already
`_ZTI14daObjDpBrock_c`'s kind-vtable pointer. Three `ambiguous data_ov025_*` rows
(`021138a8`, `021138bc`, `021138c4`) sit *inside* the table and are cross-overlay
phantom interior symbols — reading the extent from the next `symbols.txt` row
would give a false `0x58` / 22 slots. `verify`'s own reported `_ZTV` section size
of `0x88` agrees independently. (`rtti_vtables.py` is not authoritative here.)

**Own slots, diffed against the base's table** — this leaf overrides only:

| slot | target | symbol |
|---|---|---|
| 0 | `0x02111c24` | `daDkk_c::InitResources` |
| 6 | `0x02111b64` | `daDkk_c::Behavior` |
| 16 | `0x021118c8` | `_ZN7daDkk_cD1Ev` |
| 17 | `0x02111928` | `_ZN7daDkk_cD0Ev` |
| 29 | `0x0211199c` | `daDkk_c::OnAimedAtWithEgg` |

Slot 3 resolves to ov091 `0x021331b8` and slot 9 to ov091 `0x02133210` —
`daDsnBase_c`'s own concrete `CleanupResources` / `Render`, inherited. Slot 31 is
ov002 `0x020ee55c`, `dBgActor_c::Kill`; its presence is what makes the table 32
words rather than `dActor_c`'s 31 (the `dBgActor_c` length rule). The class adds
no new virtual, so its table is the same length as `daDsnBase_c`'s.

## Destructor placement, and why

The cartridge orders **D1 (`0x021118c8`) below D0 (`0x02111928`)** and carries no
D2. mwccarm 2004/b56's destructor-variant lever set is closed and measured: a
destructor **defined in the class body** emits D1 then D0 and nothing else; an
out-of-line definition emits D2, D0, D1 *plus a homeless D2 the ROM does not
carry*. So `include/daDkk_c.h` moves `~daDkk_c()` into the class body. That is a
byte requirement.

The body is empty but not inert — 0x60 bytes: it stores this class's vptr, then
`daDsnBase_c`'s (inlined, destroying `ShadowModel@0x338` then
`TextureSequence@0x324`), then `dBgActor_c`'s (also inlined, `dBgW_KcMbg@0x124`
and `Model@0xd4`), before chaining to `dActor_c::~dActor_c`.

Inlining it also moves the ABI **key function** down to `InitResources` — the
first declared virtual that is now neither inline nor pure — which this TU
defines, so this TU emits `_ZTV7daDkk_c`.

This is the direction the class's own base could *not* reproduce: ov091 orders
`daDsnBase_c`'s pair D0-below-D1, which is why that one promoted 9-of-11 while
this leaf promotes **whole, 8 of 8**.

## Promotion route

**Text-only + `compiler_only_output`.** 14 rows, **MEASURED**, not estimated —
this class had no sibling oracle in the queue, so the route defaulted to
text-only with `compiler-only:unknown` and the count had to be taken from
`verify`. Thirteen are RTTI/vtable records: the `__si_class_type_info` chain
gives five ancestors, so `2 x 6 + 1 = 13`. The fourteenth is `_ZN7Vector3D1Ev`,
emitted because the TU takes `Vector3` **locals** — `daDkk_c` has no `Vector3`
member. Every one of the 14 has a configured ROM home (arm9, ov002 or ov025), so
all license as `deadstrip-data` / `deadstrip-duplicate` and
**`externalized_output` is empty**. Nothing takes the non-promotable
canonical-import route. Unlike `daDsnBase_c` there is no `@class$`
anonymous-struct destructor — that appears only when the source declares a
file-local anonymous struct to hold the `Vector3` locals.

Sibling oracle: none. `daDsnBase_c` (the base, ov091) is the nearest precedent
and it is the *opposite* destructor case, which is exactly why its 9-of-11
subset licensing is not needed here.

## Reconciliations made during the merge

All recorded in the manifest notes:

- **`decl_common.h` wins.** It already declares every address-named symbol this
  TU touches — `func_ov025_021119a4` (line 1516), `func_ov025_021119f4` /
  `func_ov025_02111a84` (2644-2645), `func_ov091_02132dc0` (1586),
  `func_ov091_02132e64/02132e98/02132f04/02132ff4/02133020/02133098` (2853-2858)
  and `data_ov025_02113814` (1098) — all taking `char*`, where several shards had
  spelled the parameter `void*`. Merged, both spellings become visible and
  mwccarm rejects it as illegal function overloading. Shadow declarations
  deleted; definitions now match the header. Bytes unchanged.
- **`func_0201267c` had two contradictory return types** across the merged shards
  (`int` in `func_ov025_021119f4.c`, `void` in `func_ov025_02111a84.cpp`). This
  is *not* reported by `tubuild`'s conflict detector, which does not compare
  return types; both spellings survived into the generated TU, one in the shared
  block and one in a RAW per-member block — an unconditional redeclaration error.
  No header declares the symbol; eight other shards in `src/` spell it `void` and
  one `int`; **no call site here reads the result**, so the return type cannot
  move a byte. Settled on `void`.
- **The two RAW-flagged shards** (`func_ov025_021119a4`, `func_ov025_02111a84`)
  were carried verbatim inside nested `extern "C"` blocks; both are now single
  file-scope `extern "C"` definitions with the block-local declarations hoisted.
  Bytes unchanged.
- **Both destructor forcing scaffolds were dropped, not carried.** Owning the key
  function is enough to emit the vtable and both variants byte-identically.
- Include order is load-bearing: `daDkk_c.h -> daDsnBase_c.h -> dBgActor_c.h`
  pulls `common.h` before `Model.h`, fixing `Matrix4x3` to the flat `s32 m[12]`
  spelling every shard compiled against.

`func_ov025_021119a4/021119f4/02111a84` keep their address-derived names and stay
C-linkage free functions: nothing in `config/` or `include/` spells them any
other way, and this promotion does not invent member names.

## Gates

`tiers_ratchet --check` **passed**, so `--update` was deliberately **not** run
(running it anyway appends duplicate rows). Three CONVERTED-tier rows move from
shard paths to member-qualified paths in `converted-baseline.json`
(`src/_ZN7daDkk_c16OnAimedAtWithEggEv.cpp`, `...D0Ev.cpp`, `...D1Ev.cpp` ->
`src/actors/daDkk_c.cpp#<symbol>`), a net-zero set change, with matching
`converted-backslide-exceptions.jsonl` rows. **No gate was `--update`d.**

MEASURED on the three-way merge result against current `origin/main`
(`git merge-tree --write-tree`): merges **clean**; `converted-baseline.json`
keeps `main`'s `count` and `main`'s eight `daStarGate_c` rows — no silent
clobber of the sibling promotion that landed in the meantime.

`check_dead_references`: MEASURED zero Markdown references to any of the eight
deleted `src/` paths in the merged tree, so no prose scrub was needed here.

## What the validator will say, and why none of it is a loss

Four lines are expected on any promotion. All four are the *same* many-to-one
fold artifact: `delinks.txt` now carries **one** range where it carried eight,
and the counter credits only that range's **first** symbol, reclassifying the
other seven as "claiming a match that nothing compiles" — because no *single*
file compiles them any more; one file compiles all eight together.

```
Contributor credit: 0 added, 8 changed, 0 lost
7 address range(s) left the byte-verified set
7 more function(s) now claim a match that nothing compiles
Byte-verified functions -7, code bytes -972
```

- **0 added / 8 changed / 0 lost** — the eight attribution keys are re-keyed from
  `src/<shard>.cpp#<symbol>` to `src/actors/daDkk_c.cpp#<symbol>`. Same eight
  functions, same author, new path. Nothing is lost.
- **7 address ranges left the byte-verified set** — eight `delinks.txt` entries
  became one. The surviving entry `src/actors/daDkk_c.cpp .text
  start:0x021118c8 end:0x02111cf4` is credited to its first symbol,
  `_ZN7daDkk_cD1Ev` at `0x021118c8`; the other seven ranges no longer exist as
  separately-credited ranges.
- **7 functions now claim a match nothing compiles** — the same seven, seen from
  the other side.
- **Byte-verified functions -7, code bytes -972** — reconciles exactly:

```
  0x021118c8  0x60 =   96   _ZN7daDkk_cD1Ev              <- RETAINED (first symbol)
  0x02111928  0x74 =  116   _ZN7daDkk_cD0Ev
  0x0211199c  0x08 =    8   _ZN7daDkk_c16OnAimedAtWithEggEv
  0x021119a4  0x50 =   80   func_ov025_021119a4
  0x021119f4  0x90 =  144   func_ov025_021119f4
  0x02111a84  0xe0 =  224   func_ov025_02111a84
  0x02111b64  0xc0 =  192   _ZN7daDkk_c8BehaviorEv
  0x02111c24  0xd0 =  208   _ZN7daDkk_c13InitResourcesEv
                     -----
  whole range 0x02111cf4 - 0x021118c8 = 0x42c = 1,068 bytes  (sum checks: 1,068)
  minus D1's retained 0x60 = 96
                     -----
  net                       -972 bytes, over 7 functions
```

The same run should report `Relocation check: 45 checked; 45 VERIFIED` — 45 is
the MEASURED number of relocation records originating inside
`0x021118c8..0x02111cf4` in `config/arm9/overlays/ov025/relocs.txt`.

Every byte of all eight functions is still in the ROM and still byte-identical;
the counter is measuring *how many separately-credited ranges exist*, not how
much matched code exists.

https://claude.ai/code/session_013qeuLvkrjFxZriAZ93kFnF


andrewboudreau:
Superseded by #2302, which carries the current-main build branch, builder corrections, and contributor-attribution/state repair. Closing this stale duplicate so the queue has one authoritative head and validator result.

PR #2298, opened 2026-09-05
ov025: promote daDgr_c into a single C++ TU, absorbing all 9 shards

andrewboudreau:
## TL;DR (plain English)

`daDgr_c` is a swinging platform. In our tree its code lived in **nine separate
one-function source files**, each compiled on its own. The original game was
built from one `.cpp`, so this PR puts the nine back together into
`src/actors/daDgr_c.cpp` and tells the build that this single file now supplies
the whole address range `0x021111a0..0x02111898` — one `delinks.txt` entry where
there were nine.

Nothing about the shipped game changes. All nine functions still compile to
**byte-identical** machine code, and the full 16 MB ROM built from this tree has
the same SHA-256 as the stock cartridge. The claim just got stronger: not "each
function matches in isolation" but "*one* translation unit, compiled once, with
this include order, reproduces that whole stretch of ROM" — which is what the
original build did.

One source-level change outside the merge: `include/daDgr_c.h` moves the
destructor from an out-of-line declaration to an in-class definition. That is a
byte requirement, and it corrects a wrong analogy the header used to carry.

**Expect the validator to report four "loss"-looking lines. They are not losses.**
They are the arithmetic of folding nine credited ranges into one; the section
"What the validator will say" reconciles the byte figure exactly.

---

## Proof block

```
tubuild verify                     9/9 MATCH, objisolate clean,
                                   reloc-destinations clean, emission order
                                   ROM-ascending, 0 unlicensed  -> TEXT-VERIFIED
                                   (re-run from the promoted path after the move:
                                   still 9/9)
tubuild linkcheck (pre-promotion)  SCRATCH-LINK-VERIFIED
                                   .text 0x021111a0..0x02111898 differingBytes 0
                                   106/106 modules exact, 100.000000%
                                   ROM sha256 d1506e90...c478e8, matchesStockRom true
                                   moduleSetSha256 f9852ffa...dd802b
dsd check symbols --fail           still FAILS on the same 9 pre-existing errors
                                   the baseline control has: 0 new, 0 resolved.
                                   NOT green, and not reported as green.
romdata_check                      DIFFERS 0 over this file's 11 data records
relocation destinations            PASS (reloc_audit check_destinations);
                                   51 relocation records originate inside this
                                   range in config/arm9/overlays/ov025/relocs.txt
other gates                        layout_check clean; check_tubuild_conflicts
                                   clean; check_dead_references clean after
                                   rewriting the daDgr_c section of
                                   notes/bgobject-provenance.md to name the
                                   merged path; eligible.py lists all 9 members;
                                   cpp_tu_compat --require-ready READY on all 9
                                   contracts; port_refcheck 402/402;
                                   check_header_offsets clean at span 0x334
```

`tubuild linkcheck` is deliberately **not** re-run post-promotion: once the
source is enrolled in `delinks.txt`, `cmd_linkcheck` routes down the intact path
and dies at `[4/8]` regardless of `production_mode` (pre-existing; it fails
identically for `ov002/daBar_c` on untouched `main`). The pre-promotion run is
the evidence.

Full-ROM `rombuild.py` was not run separately — `linkcheck` already built the
complete ROM and compared its SHA-256 against a source-independent stock control
(`matchesStockRom: true`, 106/106 modules exact). The off-box validator repeats
that build against the exact PR SHA.

## Class identity

| fact | value | where it is proven |
|---|---|---|
| `_ZTS7daDgr_c` | ov025 `0x02113724` | RTTI name string |
| `_ZTI7daDgr_c` | ov025 `0x02113730`, `__si_class_type_info`, 0x0c bytes | its three words: `0x0209a764` (kind vtable), `0x02113724` (name), `0x021089ec` (base) |
| base `_ZTI10dBgActor_c` | ov002 `0x021089ec` | third word of the typeinfo |
| ancestor chain | `dBgActor_c -> dActor_c -> dBase_c -> fBase_c` | the cartridge's own `__si_class_type_info` chain |
| `_ZTV7daDgr_c` | address point ov025 `0x02113760`, storage `0x02113758..0x021137e0` = `0x88` = 2 preamble words + **32 slots** | see extent note |
| object size | **0x334 (820)** | the class factory (`daDgr_c_classInit`, historical alias `func_ov025_02111898`) passes the literal 820 to `fBase_c::operator new`, read off the retail instruction. Closes exactly on `dBgActor_c`'s `0x320` plus this class's own `0x14` tail. |

**Vtable extent is proven by the following word.** Read from
`extracted/overlays/overlay_0025.bin` at base `0x021111a0`: 32 consecutive
function pointers run from the address point `0x02113760` through `0x021137dc`.
The next word at `0x021137e0` is `0x44616437` — little-endian bytes `37 64 61
44`, the ASCII `"7daD"` that begins the following `_ZTS` string, **not** a code
pointer. So `0x021137e0` is the exclusive end and emitted storage is exactly
`0x88`, which is the `_ZTV` section size `verify` reports independently.
`rtti_vtables.py` would say 34 for this class — the known extent overrun,
counting the two preamble words as slots.

**Own slots, MEASURED by word-diffing against `_ZTV10dBgActor_c`** (ov002
`0x0210ae38`, base `0x020ad660`): **exactly six words differ**, and every other
word is byte-identical to the base's.

| slot | target | symbol |
|---|---|---|
| 0 | `0x021117dc` | `daDgr_c::InitResources` |
| 3 | `0x02111384` | `daDgr_c::CleanupResources` |
| 6 | `0x021113f0` | `daDgr_c::Behavior` |
| 9 | `0x021113c8` | `daDgr_c::Render` |
| 16 | `0x021111a0` | `_ZN7daDgr_cD1Ev` |
| 17 | `0x021111e4` | `_ZN7daDgr_cD0Ev` |

So the class overrides nothing else and adds no new virtual of its own. Slot 31
is `0x020ee55c`, `dBgActor_c`'s own `Kill`, inherited — and slot 31 existing at
all is what makes the table 32 words rather than `dActor_c`'s 31 (the
`dBgActor_c` length rule).

## Destructor placement, and why

The cartridge puts **D1 at `0x021111a0` below D0 at `0x021111e4`** and carries no
D2 anywhere in ov025. That is the reproducible direction: it is exactly what
mwccarm 2004/b56 emits from an **inline in-class** virtual destructor. The
out-of-line form would emit D2, D0, D1 plus a homeless D2 and would fail
`linkcheck`'s ROM-ascending `.text` audit. So `include/daDgr_c.h` now declares
`virtual ~daDgr_c() {}` first, and nothing in the `.cpp` defines it.

The header previously declared it out of line, reasoning by analogy with
`include/BigBrickBlock.h`'s leaf-class convention. **The ROM's own D1/D0
addresses refute that analogy.** Each of the two destructor shards had also
carried its own out-of-line definition — harmless in isolation, a duplicate
definition once merged; both are deleted.

Because the pair is reproducible, this TU promotes **whole, 9 of 9**, with no
subset licensing.

## Promotion route

**Text-only + `compiler_only_output`**, and this class is the *measurement pass*
for the row count. `notes/data/tu-promotion-queue.tsv` lists `daDgr_c` with no
`sibling_oracle` and `compiler-only:unknown(... measure directly ...)`,
defaulting the route to text-only. The route **is** text-only and the count **is
12**, derived and then MEASURED by `verify`:

```
  2 x (4 ancestors + self) = 10 RTTI rows
  + 1  _ZTV7daDgr_c
  + 1  _ZN7Vector3D1Ev
  = 12
```

The `Vector3` term is a **function local, not a member** — `daDgr_c` has no
`Vector3` field at all; `func_ov025_0211123c` declares `volatile struct Vector3
v;` on the stack, and that odr-use alone re-emits `Vector3`'s vague-linkage D1.
No file-local anonymous struct holds it, so there is no 13th genuinely-homeless
row. All 12 have a configured ROM home — arm9 for the six `fBase_c` /
`dBase_c` / `dActor_c` records and `Vector3`'s D1, ov002 for the two
`dBgActor_c` records, ov025 for this class's own three — so all license as
`deadstrip-data` (`deadstrip-duplicate` for the one function) and
**`externalized_output` is EMPTY**. Nothing takes the non-promotable
canonical-import route.

Sibling oracle: **none** — which is why the count was measured rather than
predicted, and why it is worth recording here.

## Merge conflicts the detector does not look for

Both would have failed to compile, and both were reported under `no cross-file
conflicts detected`:

1. **Contradictory `SharedFilePtr` types.** `extern int data_ov025_02113a68[]`
   from the `CleanupResources` shard against `extern SharedFilePtr
   data_ov025_02113a68` from the `InitResources` shard, same for `_02113a60`.
   Unified on the object spelling, which lets both call sites be ordinary C++
   (`.Release()` and a plain reference argument) with no cast.
2. **A local 30-slot `struct dActor_c`** (`f00..f29`) that
   `func_ov025_0211123c` carried purely to spell one virtual call — a hard
   redefinition once `daDgr_c.h -> dBgActor_c.h -> dActor_c.h` brings in the real
   one. Deleted; the call now goes through `daDgr_c::OnAimedAtWithEgg()`, the
   real slot 29.

The detector compares neither types nor real headers, so it saw neither. Both
`create` RAW warnings in the manifest are **RESOLVED**; neither is a `CONFLICT:`
note and no `/* TUBUILD` marker was ever emitted, which is why
`check_tubuild_conflicts` is clean.

### The one byte-level regression the merge caused

`func_ov025_021112e0` was the **sole DIFF** in the first `verify` of the merged
TU — a function nobody had touched. `Matrix4x3` has two guarded 0x30-byte
spellings, `common.h`'s flat `s32 m[12]` and `math/Matrix.h`'s `Matrix3x3 r;
Vector3 t`, and whichever a TU reaches first stands. As a one-function shard the
file included `common.h` with nothing ahead of it and got the flat one; merged,
`daDgr_c.h -> dBgW_KcMbg.h -> math/Matrix.h` arrives first and `common.h`'s guard
stands down, changing the codegen of the whole-matrix assignment. Spelling that
one copy through a local flat type restored **9/9**. Include order in the merged
file is therefore load-bearing and is documented at the include block — do not
reorder without re-running `verify`.

`func_ov025_0211123c/021112e0/02111344` keep their address-derived names and stay
`extern "C"` free functions: `ov025/symbols.txt` and `include/decl_common.h`
spell all three by address, and this promotion does not invent member names.

## Gates, and the CONVERTED-tier dilution

The manifest `status` was flipped to `promoted` **before** running the ratchet,
which is what makes `tiers_ratchet` resolve the shards as `MOVED -- absorbed
into` rather than `GONE`. `tiers_ratchet --check` then **passes**, so
`--update --reason` was not re-run. **No gate was `--update`d beyond that.**

Four backslide rows were logged and they are real, and they are the known
CONVERTED-tier dilution of a TU promotion, not a readability loss: `D1`, `D0`,
`Render` and `CleanupResources` each scored CONVERTED on their own because each
was a two-line file with nothing to fail on. Merged, the file is scored as a
whole and inherits the three auto-named helpers it now contains — `func_<addr>`
names, raw offset arithmetic and mangled call spellings, all imposed by
`ov025/symbols.txt` and `decl_common.h` and none of which this promotion may
invent. Every one of the nine bodies is byte-identical to the cartridge.
Recovering the tier is stage-3 humanizer work on the merged file.

`converted-baseline.json` is a net-zero set change: four shard paths out, four
`src/actors/daDgr_c.cpp#<symbol>` paths in.

MEASURED on the three-way merge result against current `origin/main`
(`git merge-tree --write-tree`): merges **clean**; `converted-baseline.json`
keeps `main`'s `count` and `main`'s eight `daStarGate_c` rows — no silent clobber
of the sibling promotion that landed in the meantime.

`delinks.txt` is CRLF and was spliced at byte level, so the diff is the 36 lines
that actually changed rather than the whole file.

## What the validator will say, and why none of it is a loss

Four lines are expected on any promotion. All four are the *same* many-to-one
fold artifact: `delinks.txt` now carries **one** range where it carried nine, and
the counter credits only that range's **first** symbol, reclassifying the other
eight as "claiming a match that nothing compiles" — because no *single* file
compiles them any more; one file compiles all nine together.

```
Contributor credit: 0 added, 9 changed, 0 lost
8 address range(s) left the byte-verified set
8 more function(s) now claim a match that nothing compiles
Byte-verified functions -8, code bytes -1,716
```

- **0 added / 9 changed / 0 lost** — the nine attribution keys are re-keyed from
  `src/<shard>.cpp#<symbol>` to `src/actors/daDgr_c.cpp#<symbol>`. Same nine
  functions, same author, new path. Nothing is lost.
- **8 address ranges left the byte-verified set** — nine `delinks.txt` entries
  became one. The surviving entry `src/actors/daDgr_c.cpp .text
  start:0x021111a0 end:0x02111898` is credited to its first symbol,
  `_ZN7daDgr_cD1Ev` at `0x021111a0`; the other eight ranges no longer exist as
  separately-credited ranges.
- **8 functions now claim a match nothing compiles** — the same eight, from the
  other side.
- **Byte-verified functions -8, code bytes -1,716** — reconciles exactly:

```
  0x021111a0  0x44  =    68   _ZN7daDgr_cD1Ev            <- RETAINED (first symbol)
  0x021111e4  0x58  =    88   _ZN7daDgr_cD0Ev
  0x0211123c  0xa4  =   164   func_ov025_0211123c
  0x021112e0  0x64  =   100   func_ov025_021112e0
  0x02111344  0x40  =    64   func_ov025_02111344
  0x02111384  0x44  =    68   _ZN7daDgr_c16CleanupResourcesEv
  0x021113c8  0x28  =    40   _ZN7daDgr_c6RenderEv
  0x021113f0  0x3ec =  1004   _ZN7daDgr_c8BehaviorEv
  0x021117dc  0xbc  =   188   _ZN7daDgr_c13InitResourcesEv
                        -----
  whole range 0x02111898 - 0x021111a0 = 0x6f8 = 1,784 bytes  (sum checks: 1,784)
  minus D1's retained 0x44 = 68
                        -----
  net                          -1,716 bytes, over 8 functions
```

The same run should report `Relocation check: 51 checked; 51 VERIFIED` — 51 is
the MEASURED number of relocation records originating inside
`0x021111a0..0x02111898` in `config/arm9/overlays/ov025/relocs.txt`.

Every byte of all nine functions is still in the ROM and still byte-identical;
the counter is measuring *how many separately-credited ranges exist*, not how
much matched code exists.

https://claude.ai/code/session_013qeuLvkrjFxZriAZ93kFnF


andrewboudreau:
Superseded by #2300, which carries the current-main build branch, corrected manifest evidence, and contributor-attribution/state repair. Closing this stale duplicate so the queue has one authoritative head and validator result.

PR #2300, opened 2026-09-05
ov025: promote daDgr_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

`daDgr_c` is the swinging platform in overlay 25. Its code was scattered across
nine separate source files, one per function, because that is how the
decompiler carved the ROM up. The original developers wrote it as **one** C++
file, and this PR puts it back that way: nine files in, one
`src/actors/daDgr_c.cpp` out, with every one of the nine function bodies
compiling to bytes **identical** to the retail cartridge.

The interesting part is the destructor. C++ compilers emit up to three flavours
of a destructor (`D0`, `D1`, `D2`), and which flavours you get *and in what
order* depends on whether you write the destructor's body inside the class or
outside it. The cartridge has exactly two — `D1` at `0x021111a0`, `D0` just
above it at `0x021111e4`, and no `D2` at all. Only the "inside the class"
spelling reproduces that. The header had guessed the other way, copying the
convention from a similar class; the ROM's own addresses say that guess was
wrong, so this PR corrects the header. Getting this right is what lets the
whole class ship as one unit instead of leaving the destructor pair behind.

Everything else is bookkeeping to prove the claim: the file links at the right
address, the whole 16 MB ROM still builds byte-for-byte, and the class's
type-information and virtual-function table match the cartridge too.

---

## Proof block

```
tubuild verify      9/9 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tubuild linkcheck (pre-promotion)
                    SCRATCH-LINK-VERIFIED; whole range 0x021111a0..0x02111898 (1,784 bytes)
                    reproduced from ONE object, 0 differing bytes; 106/106 modules exact;
                    full ROM sha256 d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8,
                    identical to the source-independent stock control
rombuild -j6        106/106 exact, 100.000000% of compared bytes, mismatching 0
                    ROM-build analysis: PASS
romdata_check       6 VERIFIED, 5 PARTIAL, 0 DIFFERS, 0 UNNAMED  (11 data symbols)
tu_order_check      all match in ROM order; 1 EXTRA (_ZN7Vector3D1Ev) carries a compiler-only policy
check_header_offsets  include/daDgr_c.h: 0 mismatched, 0 unparsed, struct spans 0x334
tiers_ratchet --check PASS  baseline 2680  current 2680
```

`linkcheck` was **not** re-run here: once the source is enrolled in
`delinks.txt`, `cmd_linkcheck` routes down the intact path unconditionally and
dies at `[4/8]`. It fails identically for an untouched class on `main`. The
writer's pre-promotion run is the evidence.

### The nine members, all byte-identical

| ROM addr | size | symbol |
| --- | --- | --- |
| `0x021111a0` | `0x044` | `_ZN7daDgr_cD1Ev` |
| `0x021111e4` | `0x058` | `_ZN7daDgr_cD0Ev` |
| `0x0211123c` | `0x0a4` | `func_ov025_0211123c` |
| `0x021112e0` | `0x064` | `func_ov025_021112e0` |
| `0x02111344` | `0x040` | `func_ov025_02111344` |
| `0x02111384` | `0x044` | `_ZN7daDgr_c16CleanupResourcesEv` |
| `0x021113c8` | `0x028` | `_ZN7daDgr_c6RenderEv` |
| `0x021113f0` | `0x3ec` | `_ZN7daDgr_c8BehaviorEv` |
| `0x021117dc` | `0x0bc` | `_ZN7daDgr_c13InitResourcesEv` |

## Class identity

**RTTI, read out of `extracted/overlays/overlay_0025.bin` at base `0x021111a0`:**

- `_ZTS7daDgr_c` — ov025 `0x02113724`, the 9 bytes `"7daDgr_c\0"`. The RTTI
  spelling is the cartridge's own; no name is coined.
- `_ZTI7daDgr_c` — ov025 `0x02113730`, three words
  `[0x0209a764, 0x02113724, 0x021089ec]` =
  `_ZTVN3abi20__si_class_type_infoE`, its own type name, and
  **`_ZTI10dBgActor_c`**. Single inheritance from `dBgActor_c`, stated by the
  cartridge rather than inferred.
- `_ZTV7daDgr_c` — ov025 `0x02113760` (the address point), 32 slots.

**Vtable extent is proven by the following word.** The 32 pointers run
`0x02113760`..`0x021137dc`; the next word at `0x021137e0` is `0x44616437` —
little-endian `37 64 61 44`, the ASCII that begins `"7daDkk_c"`, the *next*
type-name string, not a code pointer. Three independent witnesses agree:
`verify` reports the emitted section at `0x88` (2 preamble words + 32 slots),
`romdata_check` reports `_ZTV7daDgr_c ov025 VERIFIED, 128 bytes` (= 32 x 4),
and `rtti_vtables.py --own daDgr_c` reports 32 slots (base 32).

**Overrides, measured by word-diffing against `_ZTV10dBgActor_c` (ov002
`0x0210ae38`):** exactly six words differ and every other word is byte-identical
to the base's, so this class overrides nothing else and adds no new virtual.

| slot | target | member |
| --- | --- | --- |
| 0 | `0x021117dc` | `InitResources` |
| 3 | `0x02111384` | `CleanupResources` |
| 6 | `0x021113f0` | `Behavior` |
| 9 | `0x021113c8` | `Render` |
| 16 | `0x021111a0` | `D1` |
| 17 | `0x021111e4` | `D0` |

Slot 31 (`Kill`) is `0x020ee55c` in both tables — inherited, not overridden.
Slot 31 existing at all is what makes this table 32 words rather than
`dActor_c`'s 31.

**Factory and size.** `daDgr_c_classInit` at ov025 `0x02111898` passes
`0x334` (820) to `operator new` — `opnew_sizes.py`: 1 site, `size_confidence:
exact`, unambiguous. `check_header_offsets` computes the header's span as
`0x334` independently. The factory's address is also the exclusive end of this
TU's `.text` range, so the right-hand boundary sits exactly between
`InitResources`'s end and the factory's first instruction.

**Destructor placement.** `virtual ~daDgr_c() {}`, declared inline and first.
The cartridge has `D1` (`0x021111a0`) **below** `D0` (`0x021111e4`) and no `D2`
anywhere in ov025. That is the reproducible direction: mwccarm 2004/b56 emits
exactly that pair in exactly that order only for a destructor defined in the
class body. The out-of-line form emits `D2, D0, D1` plus a `D2` with no ROM
home, and fails the ROM-ascending `.text` audit. Had the cartridge ordered `D0`
below `D1`, no admissible source form would reproduce it and the pair would have
had to stay outside the licensed range — this class promotes **whole, 9 of 9**
because the ROM happens to sit on the reproducible side.

**Promotion route: text-only + `compiler_only_output`.** No sibling oracle
existed for this class; `notes/data/tu-promotion-queue.tsv` listed it as
`compiler-only:unknown(... measure directly ...)`, so this PR is the measurement
pass. The count is **12**, derived as `2 x (4 ancestors + self) = 10` RTTI rows,
`+ 1 _ZTV7daDgr_c`, `+ 1 _ZN7Vector3D1Ev`. All twelve have a configured ROM home
in some module's own `symbols.txt` — arm9, ov002 and ov025 — so all twelve
license as `deadstrip-data` (`deadstrip-duplicate` for the one function) and
**`externalized_output` is empty**. Nothing takes the non-promotable
canonical-import route.

The `Vector3` term is a **function local, not a member**: `daDgr_c` has no
`Vector3` field at all, but `func_ov025_0211123c` puts a `volatile Vector3` on
the stack and that odr-use alone re-emits `_ZN7Vector3D1Ev`. The object's symbol
table shows no file-local anonymous struct holding it, so there is no 13th
genuinely-homeless row.

Cross-module homes were confirmed from both directions — against each owning
module's own `symbols.txt`, and by `romdata_check.check_object()` with
`names=name_index()` so `module` resolves:

```
_ZTI7fBase_c      arm9  0x02086d70  VERIFIED 12    _ZTS7fBase_c      arm9  0x02086e60  PARTIAL
_ZTI7dBase_c      arm9  0x02086e78  VERIFIED 12    _ZTS7dBase_c      arm9  0x02086e54  PARTIAL
_ZTI8dActor_c     arm9  0x0208e390  VERIFIED 12    _ZTS8dActor_c     arm9  0x0208e384  PARTIAL
_ZTI10dBgActor_c  ov002 0x021089ec  VERIFIED 12    _ZTS10dBgActor_c  ov002 0x02108a04  PARTIAL
_ZTI7daDgr_c      ov025 0x02113730  VERIFIED 12    _ZTS7daDgr_c      ov025 0x02113724  PARTIAL
_ZTV7daDgr_c      ov025 0x02113760  VERIFIED 128
_ZN7Vector3D1Ev   arm9  0x020072c0  (function; deadstrip-duplicate)
```

The five `PARTIAL`s are all `_ZTS` rows: the emitted string is shorter than the
ROM's padded extent, so coverage falls short while every compared byte is equal.
Each of the twelve is defined exactly once across all modules' `symbols.txt`.

## Two merge conflicts `tubuild`'s detector did not see

Both were reported under `no cross-file conflicts detected`, and **both would
have failed to compile.** The detector does not compare types and does not
compare against real headers.

1. **Contradictory types for the same object.** The `CleanupResources` shard
   declared `extern int data_ov025_02113a68[]` and cast the decayed pointer; the
   `InitResources` shard declared `extern SharedFilePtr data_ov025_02113a68` and
   passed a reference. Same for `_02113a60`. Unified on the object spelling,
   which makes both call sites ordinary C++ (`.Release()` and a plain reference
   argument) with no cast. Confirmed no residue: the merged file carries exactly
   one spelling.
2. **A local `struct dActor_c` redefining the real one.**
   `func_ov025_0211123c` declared a 30-slot `f00()..f29()` stand-in purely to
   spell one virtual call. Once `daDgr_c.h -> dBgActor_c.h -> dActor_c.h` brings
   in the real `dActor_c`, that is a hard redefinition. Deleted; the call now
   goes through `daDgr_c::OnAimedAtWithEgg()`, the real slot 29. Confirmed no
   `f29`/stand-in residue survives.

## The one byte-level regression the merge caused, and its fix

`func_ov025_021112e0` — a function nobody edited — was the sole `DIFF` in the
first verify of the merged TU. `Matrix4x3` has two guarded `0x30`-byte
spellings, `common.h`'s flat `s32 m[12]` and `math/Matrix.h`'s
`Matrix3x3 r; Vector3 t`, and whichever a TU reaches first stands. As a
one-function shard this file included `common.h` with nothing ahead of it and
got the flat one; merged, `daDgr_c.h -> dBgW_KcMbg.h -> math/Matrix.h` arrives
first, `common.h`'s guard stands down, and the whole-matrix assignment's codegen
changed.

Fix: spell that one copy against a local flat `struct Mat12 { s32 m[12]; }`,
which is flat by construction and cannot be reordered by a future include
change. Verified the name does not shadow anything reachable: the only other
`Mat12` in the tree is a `typedef` inside `src/MulMat4x3Mat4x3.c`, a separate
translation unit, not a header. `Transform`'s declaration is `extern "C"`, so
the parameter type is not part of the mangled name and neither the symbol nor
the call moves.

## Gates

All green; **none needed `--update`.**

`tiers_ratchet --check` **passed** (2680/2680) both before and after the rebase,
so `--update` was deliberately skipped — it appends rather than replaces, and
running it on an already-banked branch is the double-banking defect. Audited
instead, per both the pre-rebase base and the post-rebase one: **0 byte-identical
whole records** in `converted-backslide-exceptions.jsonl`, and a set-diff of
`converted-baseline.json["converted"]` showing the four "removals" are exactly
four renames (`src/_ZN7daDgr_c*.cpp` -> `src/actors/daDgr_c.cpp#<symbol>`), with
the count unchanged. `main` moved during this build; the rebase used
`reset --hard` + cherry-pick (not `rebase --onto`), and the ledger auto-merge was
audited afterwards: **0 of main's rows lost, 0 duplicates**, main's own 9 new
rows intact.

`check_references` (no new unresolvable references), `cpp_tu_compat
--require-ready` (8/8 READY), `check_src_tu_compiles` (135/135),
`check_rename_ledger`, `check_profile_campaign`, `langmode_audit`,
`check_header_offsets`, `port_refcheck` (402 checked, 0 stale),
`check_duplicate_sources`, `check_dead_references`, `check_tubuild_conflicts
--list`, `layout_check` — all exit 0.

Two gates report a *shrinkable baseline*, which each labels explicitly as not a
failure and neither is banked here: `check_references` notes 3 references fixed
since its baseline, and `check_dead_references` notes 1 baselined reference that
now resolves.

`notes/bgobject-provenance.md` named seven of the deleted shards; the literal
`src/` tokens were removed, not just the sentences around them, since the gate
reads bare tokens.

## Expected validator lines

A promotion folds many delinks ranges into one, and the counter credits only the
range's first symbol, reclassifying the rest as "claimed". Expect all four, and
none is a loss:

```
```

The byte figure should reconcile against the nine members' `0x6f8` (1,784)
bytes of `.text` — which is also exactly the range
`0x02111898 - 0x021111a0`, i.e. the nine bodies tile the licensed run with no
gap — minus whatever the single surviving credited range keeps.

## Corrections made in this PR beyond the writer's branch

Three, all prose; the bytes were already right.

1. **`_ZTV7daDgr_c` was described as "emitted as vague linkage".** The object
   says otherwise: it is the only `GLOBAL`/`OBJECT` row in
   `compiler_only_output` (`0x88` in `.data`), while all ten RTTI rows and
   `_ZN7Vector3D1Ev` carry mwcc's vague-linkage binding `13`. Strong emission
   *is* the key-function rule, so the `deadstrip-data` disposition was right and
   only the word was wrong.
2. **Nothing recorded which function the key function now is.** Making the
   destructor inline skips it for key-function selection, so the key function
   moved from `~daDgr_c` to `InitResources`, the first non-inline virtual
   *declared*. Both are defined in this TU, so the same eleven data symbols come
   from the same object either way and the row count stays at 12 — **the header
   correction did not move the eligibility bracket.** Now stated in both the
   manifest and the header, with the warning that declaring a new virtual above
   `InitResources` would move it again.
3. **A stale claim that `rtti_vtables.py` over-reports 34 slots here.** On this
   tree it reports 32 (base 32) — its following-table tail trim removes exactly
   those two words. It is now a third witness for the extent rather than a caveat
   against it.

Do not merge; opening for validation.


andrewboudreau:
Superseded by #2307, which landed `daDgr_c` and `daDkk_c` together as one integration branch after both PRs went CONFLICTING on the shared ratchet files. The class content here landed unchanged — the integrator cherry-picked the content commits onto a fresh `origin/main` and re-proved them (verify 9/9 MATCH, rombuild 106/106 exact, romdata_check 6 VERIFIED / 5 PARTIAL / 0 DIFFERS, row count 12 re-derived, `externalized_output` empty). Nothing from this branch was dropped except the attribution hook commit.

PR #2302, opened 2026-09-05
ov025: promote daDkk_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

`daDkk_c` is the Grindel — the stone block that rises, falls and turns in Whomp's
Fortress. Its machine code lived in eight separate one-function files, each of
which compiled on its own. That is fine for checking one function at a time but
it is not what the original developers wrote: the cartridge shows those eight
functions laid out in one contiguous run, which means one `.cpp` file produced
them all. This PR reconstructs that file and proves it is right by compiling it
and comparing every byte against the cartridge.

The interesting part is the destructor. C++ compilers emit *several* destructor
variants per class, and the 2004 Metrowerks compiler this game was built with
picks their **order** based on where you write the destructor body:

| where the destructor body is written | what mwccarm emits |
|---|---|
| inside the class body (`~daDkk_c() {}`) | D1, then D0 — nothing else |
| outside, in the `.cpp` | D2, D0, D1 — plus a D2 the cartridge does not have |

The cartridge has D1 at `0x021118c8` and D0 at `0x02111928`, i.e. D1 first, and
no D2 anywhere. Only the first form reproduces that. The header this branch
started from had the destructor written out of line — inherited from the
one-function shards, none of which could ever have noticed, because a file with
one function in it has no ordering to get wrong. Moving it into the class body
is what lets all **8 of 8** functions ship in one translation unit instead of
carving the destructor pair out. The class's direct base, `daDsnBase_c`, is
ordered the *other* way in the cartridge and could only be promoted 9-of-11 for
exactly this reason; this leaf goes whole.

Moving the destructor also moves the class's *key function* — the C++ rule that
decides which object file owns the vtable — from the destructor down to
`InitResources`. That function is defined in this same file, so this file still
emits `_ZTV7daDkk_c`, and the promotion is unaffected.

Nothing about the game changes. Every byte of the ROM built from this branch is
identical to a ROM built from stock.

---

## Proof

    tubuild verify      8/8 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
    tubuild linkcheck   (pre-promotion, writer's run) SCRATCH-LINK-VERIFIED
                        range 0x021118c8..0x02111cf4 IDENTICAL, 1068 bytes, 0 differing
                        [4b/8] 8 LICENSED / 0 COLLIDES-GAP / 0 COLLIDES-SRC / 0 HOMELESS
                        emission order ROM-ascending, zero unlicensed sections
                        module fidelity 106/106, dsd check symbols 0 new vs the baseline control
                        == stock
    rombuild -j6        106/106 exact, 100.000000% of compared bytes, mismatching 0
                        ROM-build analysis: PASS
    romdata_check       8 VERIFIED, 5 PARTIAL, 0 DIFFERS, 0 UNNAMED
    tu_order_check      all 8 in ROM order; 1 EXTRA (_ZN7Vector3D1Ev) carries a policy row
    tiers_ratchet       PASS 2680 / 2680

Every other gate in the builder list exits 0: `eligible` (11094/11234),
`check_references` (42 unresolved vs baseline 45, no new), `cpp_tu_compat
--require-ready`, `check_src_tu_compiles` 136/136, `check_rename_ledger` 1843
rows, `check_profile_campaign`, `langmode_audit`, `check_header_offsets` (1
changed header, 0 mismatched, struct spans 0x3a0), `port_refcheck` 402/0 stale,
`check_duplicate_sources`, `check_dead_references`, `check_tubuild_conflicts`
136/136, `layout_check`.

**`--update` was run on nothing.** `tiers_ratchet --check` passed on the writer's
banking, and after the rebase onto current `main` the auto-merged ratchet files
were verified byte-identical to what a fresh `--update` regenerates from the
restored base — so the merge was correct and no second banking was committed. The
exceptions file carries 334 rows, 334 distinct: no double-banking.

`tubuild linkcheck` is **not** re-run here. Once the source is enrolled,
`cmd_linkcheck` routes down the intact path unconditionally and dies at `[4/8]`;
it fails identically on untouched `main`. The writer's pre-promotion run is the
evidence.

## Class identity

| | |
|---|---|
| `_ZTS7daDkk_c` | `ov025:0x021137e0` |
| `_ZTI7daDkk_c` | `ov025:0x021137ec`, base `_ZTI11daDsnBase_c` @ `ov025:0x021137f8` |
| `_ZTV7daDkk_c` | `ov025:0x02113850`, **32 slots**, section `0x88` |
| factory | `daDkk_c_classInit` @ `ov025:0x02111cf4` |
| `new`-size literal | `0x02111cfc` → `928` = `0x3a0`, exact, single site (`opnew_sizes.py`) |
| base chain | `daDsnBase_c → dBgActor_c → dActor_c → dBase_c → fBase_c` |

**Overrides**, diffed slot-by-slot against `daDsnBase_c` (32 slots, base 32 — this
leaf adds no new virtual):

| slot | symbol | address |
|---|---|---|
| 0 | `InitResources` | `0x02111c24` — override of a slot the base leaves pure |
| 6 | `Behavior` | `0x02111b64` — the base's other pure slot |
| 16 | `~daDkk_c` (D1) | `0x021118c8` |
| 17 | `~daDkk_c` (D0) | `0x02111928` |
| 29 | `OnAimedAtWithEgg` | `0x0211199c` — returns `0xce000` (206.0) against `dActor_c`'s default `0x14000` (20.0) |

### The `_ZTV7daDkk_c` extent is a trap in both directions — settled here

Reading the extent off the next `symbols.txt` row (`data_ov025_021138a8`) gives a
false `0x58` / 22 slots, and three `ambiguous` `data_ov025_*` rows
(`021138a8`, `021138bc`, `021138c4`) are phantom interior symbols *inside* the
table. `rtti_vtables.py` errs the other way and over-reports unless its
following-table trim fires. Three independent readings agree on **32**:

1. `relocs.txt` — every word from `0x0211384c` (the typeinfo pointer, to
   `_ZTI7daDkk_c`) through `0x021138cc` (slot 31, to `ov002:0x020ee55c`
   `dBgActor_c::Kill`) is a relocated entry, and the next relocated word
   `0x021138d0` already belongs to `_ZTI14daObjDpBrock_c`.
   `(0x021138cc - 0x02113850) / 4 + 1 = 32`.
2. The compiler emits exactly **128 bytes** = 32 words for this symbol.
3. `romdata_check.check_object()` compares all 128 emitted bytes against the
   retail `ov025` image with relocations applied, `blindWords: 0` — all equal,
   verdict `VERIFIED`. Note that the same record's own `romExtent` field reads
   `88` decimal, which *is* the false `0x58`; the compared-and-equal 128 is the
   number that is proven.

## Promotion route

**Text-only**, whole promotion, 8 of 8, `.text 0x021118c8..0x02111cf4`. The single
`.text` range is the only licensed range; everything the object emits outside it
is `compiler_only_output`.

**14 rows, and the count is derived, not guessed** — this class had **no sibling
oracle**: `2 x (5 ancestors + self) + 1 vtable + 1 Vector3 = 14`.

- **13 `deadstrip-data`** — the `_ZTI`/`_ZTS` pair for each of the six classes in
  the chain, plus `_ZTV7daDkk_c`. Every canonical address was re-checked against
  the owning module's own `symbols.txt` and each symbol is defined exactly once
  tree-wide. Three modules are involved — `arm9` owns six, `ov002` two, `ov025`
  five — so **two of the three homes are cross-module**, which is normal, and
  corroborated from the other direction: the vtable's typeinfo word at
  `0x0211384c` relocates to `_ZTI7daDkk_c`, and `check_object()` returns
  `_ZTI10dBgActor_c ov002 VERIFIED 12 bytes`.
- **1 `deadstrip-duplicate`**, `_ZN7Vector3D1Ev` @ `arm9:0x020072c0` (4 bytes).
  The trigger is **locals-only**: `func_ov025_02111a84` takes a `Vector3 v[2]`
  and `InitResources` four more. **No member of `daDkk_c` has that type.**
- **No `externalized_output`, and deliberately no `@class$` anonymous-struct
  row.** The base class `daDsnBase_c` gets one because *its* source declares
  `struct { Vector3 part; Vector3 quake; } sp;`; a plain `Vector3 v[2]` array
  does not produce that symbol. Both points re-checked here — confirmed.

**Sibling oracle: none.** The nearest comparison is the direct base
`daDsnBase_c` (ov091), and it does **not** transfer: `ov091` orders that class
D0-below-D1, which no admissible source form emits, so it promoted 9-of-11 with
its destructor pair left in shards. `daDkk_c`'s cartridge order is the
reproducible one, so it promotes whole. The `_ZTV` extent trap above is also
untested by any oracle — this is the only check it gets.

## What the builder changed

Byte gates were green throughout; the following were **documentation** defects
found by reading the manifest and header prose against what actually shipped
(commit `6479d94d7`):

1. The manifest claimed the destructor shards' "forcing scaffolds were DROPPED,
   not carried." **There were no scaffolds.** Both shards were an out-of-line
   `daDkk_c::~daDkk_c() {}` with an empty body. What moved is the out-of-line
   definition itself.
2. The `func_0201267c` return-type conflict — real, and correctly resolved to
   `void` — was justified by "eight shards spell it void and one spells it int."
   **Measured: 83 files spell it `void`, 21 spell it `int`**, and the far
   stronger citation was missed entirely: the symbol's own *enrolled* definition,
   `src/func_0201267c.cpp` (in `config/arm9/delinks.txt`), defines
   `void func_0201267c(unsigned int id, const Vector3 *v)`. Conclusion unchanged,
   evidence replaced. Neither call site reads the result, so no byte can move.
   `tubuild`'s conflict detector still does not compare return types.
3. `boundary_evidence` said "three of the modules involved are not ov025."
   Three modules are involved *in total*; two of them are not `ov025`.
4. The header asserted the class's own `0x360..0x3a0` storage held **two**
   fields. The TU touches **six** words there. `0x394` in particular is read by
   the state-5 fall as the ground height and is *not* `mProbeHeight@0x390` under
   another name — both are separate words and both accesses byte-match. `0x394`,
   `0x39c`, `0x39e` and `0x39f` are now documented as live-but-unnamed rather
   than silently sitting inside `pad_`. No struct field moved; every access was
   already a raw offset.
5. Recorded the `_ZTV` corroboration above in the manifest, including the warning
   that `romdata_check`'s own `romExtent` for this symbol carries the false
   `0x58`.

## Expected validator lines — all four are the many-to-one fold, not losses

The counter credits only the delinks range's **first** symbol and reclassifies
the rest as "claimed":

The byte figure reconciles against this TU: the eight functions total
`0x60 + 0x74 + 0x8 + 0x50 + 0x90 + 0xe0 + 0xc0 + 0xd0 = 0x42c` = **1,068 bytes**,
which is the whole licensed range, and the surviving credited entry is the range's
first symbol `_ZN7daDkk_cD1Ev` at `0x60` — so the expected drop is `1,068 - 96 =
972` code bytes across 7 functions. The same run reports `Relocation check: N
checked; N VERIFIED`.

Do not merge on my say-so; `PR validation` is the gate.

---


andrewboudreau:
Superseded by #2307, which landed `daDgr_c` and `daDkk_c` together as one integration branch after both PRs went CONFLICTING on the shared ratchet files. The class content here landed unchanged — the integrator cherry-picked the content commits onto a fresh `origin/main` and re-proved them (verify 8/8 MATCH, rombuild 106/106 exact, romdata_check 8 VERIFIED / 5 PARTIAL / 0 DIFFERS, row count 14 re-derived, `externalized_output` empty). Nothing from this branch was dropped except the attribution hook commit.

PR #2304, opened 2026-09-05
ov006: promote dScMgHanachan_c into a single C++ TU (22 of 61, deliberate)

andrewboudreau:
## TL;DR for a reader who is not deep in this repo  The SM64DS cartridge was built from C++ source files that no longer exist. This project rebuilds them one at a time: take the N tiny one-function files that the disassembler produced, glue them back into a single `.cpp`, compile it with the original 2004 compiler, and check that every byte comes out identical to the cartridge. When that works, the N files collapse into one real translation unit.  Here the disassembler says ov006 contains one 61-function run that plausibly was one original file. **This PR takes 22 of those 61, not all 61, and that is a finished result, not an abandoned one.** Two independent things cap it:  1. **There is a hole in the middle of the run that nobody has source for.**    `func_ov006_020ea914` (0x020ea914, 0x324 bytes) has no `src/` file and no    entry in `delinks.txt` — the build just copies those bytes out of the    cartridge. A translation unit has to claim one unbroken address range, so the    61-function run can only be taken as one of its two contiguous sides. This one    takes the larger side.  2. **Two compiler switches on the larger side want opposite settings, and both    are whole-file.** Four functions only reproduce with    `#pragma opt_strength_reduction off`; three only reproduce with it on. One    `.cpp` cannot have it both ways, so whichever way you set it, some functions    in the middle of the range stop matching and break the range into pieces. The    longest unbroken all-matching stretch under any setting is 22 functions, and    that is what this file claims.  Nothing is lost by taking 22: **38 of the other 39 functions keep their own source shard and their own `delinks.txt` entry** and still build into the ROM exactly as before (the 39th is the sourceless hole, which has neither). The ROM this branch produces is byte-identical to the stock cartridge.  **This is *not* a destructor-order refusal.** Several recent partials were capped because the cartridge orders `D1` *below* `D0`, which this compiler cannot be made to emit. Here the cartridge orders D1 (0x020ea280) below D0 (0x020ea2c8) — the *reproducible* direction — so the destructor pair would have promoted fine. It is out only because it sits on the far side of the sourceless hole.  ---  ## Proof  ``` tubuild verify    22/22 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED                   emission order: all 22 in expected ROM-ascending section order tubuild linkcheck (pre-promotion, by the writer)                   LINK-VERIFIED. [4b/8] 22 LICENSED, 0 COLLIDES-GAP, 0 COLLIDES-SRC,                   0 HOMELESS; module fidelity 106/106 exact; ROM sha256 IDENTICAL to                   the stock control. `dsd check symbols --fail` FAILs on the same 9                   pre-existing errors the baseline control has -- 0 new. rombuild -j6      106/106 exact, 100.000000% of compared bytes, mismatching 0                   ROM-build analysis: PASS                   source-built functions 11,111 reproducing / 0 mismatching romdata_check     0 data symbols emitted by this source (0 VERIFIED, 0 PARTIAL, 0 DIFFERS)                   -- correct: this TU owns no vtable or RTTI, see below tu_order_check    ALL MATCH, ROM order, nothing unlicensed rombuild -j6 (full, ROM built)                   build/sm64ds.nds sha256                   d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8                   -- identical to the no-TU baseline control's ROM tiers_ratchet     --check PASS, baseline 2680 / current 2680 -- NOT updated ```  `tubuild linkcheck` was independently re-run during review against a temporary current-main pre-promotion configuration after refreshing the content-bound stock control. It passed with 22 licensed symbols, a 5,184-byte identical TU range, 106/106 exact modules, zero new symbol errors, and a stock-identical ROM. The exact promoted tree was then restored and the normal production `rombuild.py` gate passed independently.  All other gates green at exit 0 on the current base: `eligible.py` 11094/11234, `check_references --against origin/main` (no new unresolvable references), `cpp_tu_compat --require-ready` (8/8 READY), `check_src_tu_compiles` 136/136, `check_rename_ledger`, `check_profile_campaign`, `langmode_audit --check` (PASS), `check_header_offsets --changed` (0 files under `include/`), `port_refcheck` (402 checked, 0 stale), `check_duplicate_sources`, `check_dead_references` (no new dead references), `check_tubuild_conflicts --list`, `layout_check` (clean).  **No gate was `--update`ed.** `tiers_ratchet --check` passed both before and after the restack onto current `main`, so `--update` was deliberately skipped — running it on an already-banked tree appends duplicate rows. `config/converted-baseline.json` and `config/converted-backslide-exceptions.jsonl` are byte-identical to `main`'s in this branch's diff. `check_references` and `check_dead_references` each report one banked item that now resolves and could be shrunk; neither is a failure and neither is banked here.  ## Where the two edges come from  **Edge 1 — the sourceless hole, verified independently.** Parsing `config/arm9/overlays/ov006/delinks.txt` over the whole run `0x020ea280..0x020ede18` yields **39 `.text` blocks and exactly one coverage gap: `0x020ea914..0x020eac38`** — the hole, and nothing else. There is no defining `src/` file anywhere in the tree; the only mentions of the symbol are a call site in `_ZN15dScMgHanachan_c6RenderEv.cpp`, a declaration in `include/decl_common.h`, and its `symbols.txt` row (`kind:function(arm,size=0x324) addr:0x020ea914`). A scan of all 136 TU manifests finds **zero** with more than one `.text` section, so a claim with a hole in it cannot be expressed here. The run is therefore takeable only as ordinals 0..10 (11 functions, the smaller side, holding the destructor pair) or ordinals 12..60 (49 functions). This takes the larger side.  **Edge 2 — contradictory file-global pragmas, corroborated by where the edges land.** The four functions the writer names as needing `opt_strength_reduction off` each carry that pragma in their own surviving legacy shard — checked directly:  | function | ordinal | position relative to this TU | |---|---|---| | `func_ov006_020eac38` | 12 | immediately **below** the claimed range | | `func_ov006_020ec458` | 35 | immediately **above** the claimed range | | `func_ov006_020ec93c` | — | above | | `func_ov006_020ec9c0` | — | above |  The claimed range's two edges are *exactly* the two nearest strength-reduction-off functions, and the range contains two of the three functions that need it on (`func_ov006_020ebf20` ord 32, `func_ov006_020ec2bc` ord 34). That is an independent check on the whole story: 0x020eb018..0x020ec458 is the maximal gap between strength-reduction-off members, and the shipped `.cpp` sets no file-global `opt_strength_reduction` (default on) accordingly.  Measured whole-side scores were SR off / OCS on = 44/49, SR on / OCS on = 42/49, SR off / OCS off = 40/49. **The 44 is not the number that decides this** — promotion needs a *contiguous* run in which every function matches, and `tu_promote` refuses a manifest with more than one `.text` span. The longest contiguous all-matching run is 20 under the 44/49 setting and **22 under the 42/49 setting**, so the lower-scoring setting is the correct choice. Taking the higher total would have been the natural mistake.  ## Class identity  ``` _ZTS15dScMgHanachan_c   0x0213ca88   "15dScMgHanachan_c" _ZTI15dScMgHanachan_c   0x0213ca7c   abi::__si_class_type_info                                      { 0x0209a764, 0x0213ca88, 0x020bbf6c }                                      base 0x020bbf6c -> ov004 "11dScMgBase_c" _ZTV15dScMgHanachan_c   0x0213cab8   36 slots, ends 0x0213cb48 (next word 0xffffffff) factory                 dScMgHanachan_c_classInit @ 0x020ede18 (0x68 bytes) new-size literal        0x4f68 (20,328) stored inline at 0x020ede24, unambiguous,                         1 site -- matches the header's sizeof == 0x4f68 assert ```  Vtable diffed slot-by-slot against `dScMgBase_c` (`rtti_vtables.py --own`, which agrees with the direct byte read at 36 slots — worth stating given that tool's known `_ZTV` extent overrun):  | slot | target | override | |---|---|---| | 0 | 0x020edb04 | `InitResources` | | 3 | 0x020ecec8 | `CleanupResources` | | 6 | 0x020ed18c | `Behavior` | | 9 | 0x020ecee4 | `Render` | | 16 | 0x020ea280 | `D1` | | 17 | 0x020ea2c8 | `D0` | | 18 | 0x020eda48 | `OnYoshiTryEat` (own new slot) |  Sibling oracle: `ov006/dScMgD3DBase_c` shares the same base and the same `0x020bbf6c` type-info record, and also carries 36 slots.  ## Promotion route, and why there is no `compiler_only_output`  Route is **text-only**, with **zero** `compiler_only_output` rows and no `externalized_output`. The class's first *declared* virtual is the out-of-line `~dScMgHanachan_c()`, so the destructor is the key function — and the destructor pair lies on the far side of the sourceless hole, outside this TU. A TU that does not own the key function emits no `_ZTV`/`_ZTI`/`_ZTS`, which is why `include/dScMgHanachan_c.h` needed **no edit at all** (it keeps the out-of-line destructor declaration that `_ZN15dScMgHanachan_cD1Ev.cpp` and `..D0Ev.cpp` require). Corroborated from three directions: the manifest's `compilerOnlyOutput` is empty on all four lists, `romdata_check --files` reports **0 data symbols** emitted by the source, and linkcheck's object audit reports 0 non-licensed symbols and 0 unlicensed sections.  ## Tool change: `tools/tubuild.py` gains `_legacy_rel()`  `cmd_create` was already written to tolerate a member with no source file (it becomes a banner and `verify` reports it MISSING), but `build_manifest_entry` was not: `SP.path_for()` returned `None` and `.relative_to(REPO)` raised `AttributeError`, so `create` aborted before any manifest existed and the one shape `cmd_create` was taught to survive was unreachable. `_legacy_rel()` is a null guard around exactly that call.  Reviewed for whether it can strand a `null` downstream:  - `tu_promote.py:126-128` **refuses** a falsy `legacy_source` with   `PromoteError(... has no legacy_source)`. A null can therefore never be   promoted, which is why the shipped manifest is clean. - `tubuild.is_complete()` already begins `if legacy_rel is None: return False`. - `tiers_ratchet.promoted_moves()` guards with `if legacy and legacy != dest`. - `BP.version_for(None, sym)` does not crash: it stems `str(None)`, misses the   pin table, and falls through to `DEFAULT_VERSION`. In `derive_version` that   can only add a spurious entry to `seen`, which yields a **CONFLICT** rather   than a false green — fail-closed. - `tu_production.py` (the intact-object route, lines 407/410/455) would raise   `TypeError`/`AttributeError` on a null. Loud, not silent, and unreachable   here: this TU is text-only and the sourceless member is held out.  Verified tree-wide: **0 null `legacy_source` entries across all 136 manifests.**  ## Review corrections

- Removed the draft-only tools/tubuild.py change so this remains a source-promotion PR.
- Restored all 22 per-member attribution identities and refreshed the generated TU-state note.
- Replaced comments that cited deleted shard paths with the live promoted member list.
- Merged current main and independently reran both the pre-promotion link proof and the exact normal production ROM build.

## Expected validator lines (not losses)  A promotion always produces these, all the same many-to-one fold artifact — the counter credits only the delinks range's first symbol and reclassifies the rest as "claimed":  ``` Contributor credit: 0 added, N changed, 0 lost N address range(s) left the byte-verified set N more function(s) now claim a match that nothing compiles Byte-verified functions -N, code bytes -N,NNN ```  The 22 absorbed functions total 0x1440 = 5,184 code bytes, which is the figure the byte line should reconcile to. The same run reports `Relocation check: N checked; N VERIFIED`.  ## Corrections in this branch  - Removed three shadow types left behind by the 49-function draft and referenced   by none of the 22 licensed members: `struct E`, `struct W2` (also named in the   array-spelling comment as if live) and `struct Thing`, whose comment asserted a   0x98-stride requirement belonging to `func_ov006_020ed494` — a function above   this TU's upper edge. The `Thing` observation is kept as prose so a later pass   does not rediscover it. `tubuild verify` re-run after the edit: still 22/22. - Fixed a count in the manifest that contradicted the entry directly below it.   `boundary_evidence[0]` read "The other 39 keep their own shards and their own   delinks.txt entries"; `boundary_evidence[2]` says the sourceless hole has   neither. Counted from the post-promotion `delinks.txt`: 38 shards, 1 hole. The   source header's "the remaining 27" was correct but scoped to the larger side   only, and now says so.  ## Known, pre-existing, and deliberately not fixed here  `include/dScMgHanachan_c.h` and `src/_ZN15dScMgHanachan_c13InitResourcesEv.cpp` both still say slot 18 "stays an unmigrated raw `extern "C"` helper (`src/func_ov006_020eda48.c`)". That path no longer exists — the method is migrated as `src/_ZN15dScMgHanachan_c13OnYoshiTryEatEi.cpp` and declared in the header two lines below its own stale comment. This predates the branch, is present on `main`, and touches files this promotion does not, so it is reported rather than folded in.  🤖 Generated with [Claude Code](https://claude.com/claude-code)  https://claude.ai/code/session_01G1Gdj5fTjMsW2VjRLpXxYA


andrewboudreau:
Review complete on exact head 8c04b522e. Independent pre-promotion link proof reproduced the whole 5,184-byte promoted range from one object; partitioned verification is 22/22 MATCH with objisolate and relocation destinations clean. The production stock build is 106/106 exact, 11,111 of 11,111 source-built functions reproduce, all 20 source-owned data claims reproduce, and the retail ROM SHA-256 is d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8. Source coverage is net zero with no byte handed back, all 22 member credits remain intact, generated TU state and ratchets are current, and terminal private PR validation passed. No findings. This is deliberately 22 of 61; the held-out area is not represented as source-owned. The PR remains open because it is under the four-hour merge threshold.

andrewboudreau:
Superseded by #2309, which landed `dScMgHanachan_c` at **49 of 61** rather than 22.

The 22-of-61 cut here was correct given what was known: the role file stated that `opt_strength_reduction` and `opt_common_subs` are file-global last-wins with no positional escape, so a pragma contradiction forced a subset, and the longest contiguous all-matching run was 22.

That rule turned out to be true only by default. `#pragma defer_codegen off` makes positional `push`/`pop` brackets bind, and all 27 pragma-driven exclusions came back. The decisive probe: one `opt_strength_reduction off` member under a bracket gave 23/23 while the two `opt_strength_reduction`-**on** members still matched. The control is equally clean — deleting that single pragma line from the finished file drops it to 44/49 *and* 48 ordinal pairs out of ROM order.

Still held back, correctly: the sourceless function `func_ov006_020ea914` and the 11 functions on its smaller side. A licensed claim cannot have a hole, and no pragma affects that.

Gates on the landed version: verify 49/49 MATCH, `linkcheck [4b/8]` 49 LICENSED with `emittedTextOrderIsRomAscending = true`, rombuild 106/106 exact, ROM sha256 identical to the baseline control, 0 B handed back to the cartridge.

Nothing from this branch was lost — #2309 contains its commits and extends them.

PR #2303, opened 2026-09-05
Match four ov006 minigame functions from the m100 OV63 lane: Flower round setup, two Luigi board routines, and the Sound note tick

tangosdev:
Four ov006 minigame functions recovered by the m100 OV63 lane, byte-exact at the pinned mwccarm 2004/b56 with strict relocation checking, plus the two headers they needed and the delinks entries that put them in the ROM link.

## The four functions

| Function | Address | Size | What it is |
| --- | --- | --- | --- |
| `func_ov006_0212a764` | `0x0212a764` | `0x310` (784 B) | `dScMgFlower_c` round setup: clears the cursor pair and the 0x16 petal records, rolls the petal count, lays the petals out on a circle (angle = base + i * turn / count, a (0, 3.0) offset rotated through `func_0203d388`), shuffles angle and position between random pairs, and resets the round state. Called from `InitResources` and again from `OnYoshiTryEat` on a round restart. |
| `func_ov006_020f15ac` | `0x020f15ac` | `0x250` (592 B) | `dScMgLuigi_c` per-slot picture mover: arms an unstarted slot from the speed-level tables, otherwise advances position by the two speeds, runs the x phase machine (phase 0 decelerates and flips to 1 or 2 at the crossing, phases 1 and 2 accelerate to the level top speed and drop back to 0), and wraps through `func_ov006_020f1dbc`. |
| `func_ov006_0211bc8c` | `0x0211bc8c` | `0x2b8` (696 B) | `dScMgSound_c` per-note animation tick: advances the slot frame counter against the per-pattern length, steps the pattern index at the wrap, and past index 6 pins the index, sets the slot state, appends the note id to the playback queue with the slot number packed into the high byte, arms the queue timer, and routes the note to a player chosen by `dScMgBase_c::Virtual8C` and the instrument select. |
| `func_ov006_020f2224` | `0x020f2224` | `0x56c` (1388 B) | `dScMgLuigi_c` board layout: for boards whose `data_ov006_0213ce98` entry is non-zero, lays the whole board out at once in one of four fixed layouts (2x2 at (0x70,0x50), 4x4 at (0x50,0x30), 6x8 at (0x10,0x10), or a single row of eight at y = -5), each with a random count and a per-slot random speed level nudged off `unk_545a` on collision. Called from `func_ov006_020f2790`. |

## Byte gate, strict relocations, at the pinned compiler

[...]

PR #2306, opened 2026-09-05
Restore provenance for four OV63 matches

andrewboudreau:
Follow-up to #2303, which merged while its review repair was still running. Adds the four final matched attempt/provenance rows and removes the corresponding solved near-miss tips. The source tree was independently verified at 17/17 relocation-aware consumers, 106/106 exact modules, 11,111 reproducing source functions, zero source mismatches, and zero stock replacement/fallback leakage.


andrewboudreau:
Review complete on exact head 278f77db. This follow-up is bookkeeping-only: it restores four matched attempt nodes, four exact-match provenance rows, and removes only the four solved near-miss tips for the already-merged OV63 functions. It adds no src/ file and changes no delink or source-coverage range. Public checks and terminal private PR validation passed. No findings. The PR remains open because it is under the four-hour merge threshold.

andrewboudreau:
Superseded by #2399. Three of the four rows this PR wanted to add (020f15ac, 0211bc8c, 020f2224) were already handled when dScMgLuigi_c and dScMgSound_c were promoted into single TUs (their ledger rows were updated and near-miss tips cleared as part of those promotions). The remaining gap — func_ov006_0212a764 never getting a matched ledger row — is carried forward in #2399.

PR #2312, opened 2026-09-06
ov006: promote dScMgRoulette_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

SM64DS's "mushroom roulette" minigame scene was one C++ file when Nintendo built it in 2004.
This repo had it as **40 separate one-function files**, each holding a single machine-verified
function body. This PR puts them back into one `src/actors/dScMgRoulette_c.cpp` and proves the
merged file still compiles, under the original 2004 CodeWarrior compiler, to **exactly** the
bytes on the cartridge — all 40 functions, byte for byte, relocations included.

The interesting constraint is the destructor. C++ compilers emit up to three destructor bodies
per class (D0 "deleting", D1 "complete", D2 "base"). The cartridge has D1 at `0x0210788c`,
*below* D0 at `0x02107920`, and no D2 anywhere. mwccarm 2004/b56 produces that exact order from
exactly one source shape: a destructor **defined inline in the class, declared first**. Written
out of line it emits D2/D0/D1 — the wrong order plus a D2 the ROM does not have, which would
split the run and force a partial promotion. So the inline destructor is what lets this TU
promote whole at 40/40 rather than in pieces.

This is a **builder** re-verification: everything below was measured in a fresh worktree at
`cpp/dScMgRoulette_c-tu-build`, not copied from the writer's output.

---

## Proof block

```
tubuild verify    40/40 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
tu_order_check    ALL MATCH, ROM order, nothing unlicensed
tubuild linkcheck (pre-promotion, writer's record)  SCRATCH-LINK-VERIFIED
                  tuRange 0x0210788c..0x0210a400, differingBytes 0, ROM sha256
                  d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
rombuild -j6      profile stock: 0 mod source replacements, 0 ROM-gap fallbacks
                  source-built functions 11,118 reproducing, mismatching 0
                  ROM-build analysis: PASS
                  build/sm64ds.nds  16,777,216 B  sha256 d1506e90...c478e8
                  == the run's own source-independent strict stock BASELINE CONTROL
                     (no TU substitution), and == the recorded pre-promotion linkcheck ROM
romdata_check     13 data symbols: 7 VERIFIED (212 B), 6 PARTIAL (72 B), 0 DIFFERS, 0 UNNAMED
source_coverage   clean - 0 B handed back to the cartridge
prepush_attrib.   10548 tracked, 32 consolidated with credit intact, 0 changed, 0 lost
tiers_ratchet     PASS  baseline 2680  current 2681 (+1 gained, not banked - see below)
check_header_offs 22 commented fields, 0 mismatched, 0 unparsed, struct spans 0x5400
```

Every other gate exits 0: `check_references` (42 unresolved vs baseline 45, no new),
`cpp_tu_compat --require-ready` (8/8 READY), `check_src_tu_compiles` (139/139),
`check_rename_ledger`, `check_profile_campaign`, `langmode_audit`, `port_refcheck`
(402 checked, 0 stale), `check_duplicate_sources`, `check_dead_references`,
`cpp_tu_state --check-note`, `check_tubuild_conflicts`, `layout_check`.

**No gate needed `--update`.** `tiers_ratchet --check` PASSES with the ledger files untouched:
the 40 shard identities resolve through `tu_manifest` as `absorbed_clean` into a source that is
itself CONVERTED, so no exception row and no re-bank is required. Audited rather than assumed —
`converted-baseline.json["converted"]` is set-identical to the base (0 in, 0 out), and
`converted-backslide-exceptions.jsonl` carries **0 byte-identical duplicate records** across its
342 rows (53 paths legitimately recur under different reasons). The `+1 gained, not yet banked`
and the `3 fixed since the baseline -- run --update to bank it` line from `check_references` are
shrinkable-baseline nudges, not failures, and belong to the automated refresh rather than a class PR.

---

## Class identity

**Names, from the cartridge's own RTTI.** ov006 `symbols.txt`:

| symbol | module | address |
|---|---|---|
| `_ZTS15dScMgRoulette_c` | ov006 | `0x0213e30c` |
| `_ZTI15dScMgRoulette_c` | ov006 | `0x0213e300` |
| `_ZTV15dScMgRoulette_c` | ov006 | `0x0213e39c` (address point) |

The base is proved from the image, not assumed. `_ZTI15dScMgRoulette_c` is a three-word
`__si_class_type_info`: `0x0213e300 -> 0x0209a764` (the `__si_class_type_info` vtable),
`0x0213e304 -> 0x0213e30c` (`_ZTS`), `0x0213e308 -> 0x0213bc64` = `_ZTI19dScMgSingle3DBase_c`.
Walking the same third word up gives the whole chain out of `relocs.txt`:

```
dScMgRoulette_c      0x0213e300 -> 0x0213bc64  dScMgSingle3DBase_c   (ov006)
dScMgSingle3DBase_c  0x0213bc64 -> 0x020bbf6c  dScMgBase_c           (ov004)
dScMgBase_c          0x020bbf6c -> 0x020914d4  dScene_c              (arm9)
dScene_c             0x020914d4 -> 0x02086e78  dBase_c               (arm9)
dBase_c              0x02086e78 -> 0x02086d70  fBase_c               (arm9)
fBase_c              0x02086d70 -> 0x0209a774  __class_type_info -- root, 8 bytes, no base
```

**Size.** `tools/opnew_sizes.py` (`build/opnew_sizes.json`): factory
`ov006:dScMgRoulette_c_classInit` at `0x0210a400` (size `0xac`), one allocation site at
`0x0210a40c`, size `21504 = 0x5400`, `size_confidence: exact`, `attribution: inline_store`,
installing vtable `0x0213e39c`. That is an independent corroboration of
`include/dScMgRoulette_c.h`'s own `sizeof(dScMgRoulette_c) == 0x5400` assert, and
`check_header_offsets` now walks all 22 commented fields against it with 0 mismatches.

**Vtable, slot by slot against the base.** `tools/rtti_vtables.py --own dScMgRoulette_c`:

```
dScMgRoulette_c : dScMgSingle3DBase_c   36 slots (base 36)
  slot 0   0x0210a194  InitResources
  slot 3   0x0210980c  CleanupResources
  slot 6   0x02109aac  Behavior
  slot 9   0x02109834  Render
  slot 16  0x0210788c  D1
  slot 17  0x02107920  D0
  slot 18  0x021095cc  OnYoshiTryEat
  slot 19  0x021096c8  OnTurnIntoEgg
```

Same width as the base: all eight are **re-overrides of inherited slots, not new virtuals**, so
no field moves and the table does not grow.

**`_ZTV` extent — the number is 0x98 storage / 0x90 from the address point, and both are right
for different questions.** Derived from `relocs.txt` and `symbols.txt` rather than from any tool's
extent field:

- The contiguous relocated run is `0x0213e398 .. 0x0213e428` inclusive — 37 relocated words,
  beginning at the typeinfo word (`0x0213e398 -> 0x0213e300`).
- `0x0213e42c` is **not** relocated, and the next relocated word is `0x0213e444`
  (`dScMgSingle3DBase_c`'s typeinfo). So there is a real `0x18` gap and the run does not bleed
  into the next object.
- Storage = `0x0213e394` (offset-to-top, an unrelocated zero) through `0x0213e42c` = **`0x98`**.
- The `_ZTV` symbol **is** the address point, so the symbol's own extent is
  `0x0213e39c .. 0x0213e42c` = **`0x90` = 36 slots**, which is the naive next-symbol delta and
  also what `romdata_check` measures.

`romdata_check`'s per-symbol record for `_ZTV15dScMgRoulette_c` (via `check_object()` with
`names=name_index()`) is `bytes 144, emitted 144, romExtent 144, blindWords 0, VERIFIED` —
all 36 slots equal to the cartridge with no blind words.

**Promotion route: text-only + `compiler_only_output`.** `sections` is a single `.text` claim
`0x0210788c..0x0210a400`; `data` and `bss` are empty; `externalized_output` is empty. The 40
functions tile that span exactly — no gaps, no overlaps, ending precisely at the claim — and no
other ov006 delinks entry overlaps it. One `.text` block, which is the invariant a licensed claim
depends on.

**`compiler_only_output` is 13 rows, and 13 is derivable.** `2 x (5 ancestors + self) + 1 vtable`.
The `Vector3` term other classes carry is **zero** here: the two `Model`-shaped subobjects at
`0x531c`/`0x536c` are declared as raw `u8[0x50]` and destroyed by explicit calls in the ROM's own
order, so no typed member contributes vague-linkage output. The queue's `compiler-only:~11` is an
estimate copied straight from the sibling oracle `dScMgSingle3DBase_c` (which really is 11); it
undercounts by exactly the `_ZTI`/`_ZTS` pair for `dScMgSingle3DBase_c`, which is that class's own
self and this class's direct base.

All 13 canonical addresses were checked against the owning module's `symbols.txt` and confirmed
**defined exactly once tree-wide**, and corroborated from the other direction by
`romdata_check.check_object()`: `_ZTI19dScMgSingle3DBase_c ov006 VERIFIED 12 bytes`,
`_ZTI11dScMgBase_c ov004 VERIFIED 12 bytes`, `_ZTI8dScene_c`/`_ZTI7dBase_c`/`_ZTI7fBase_c`
`arm9 VERIFIED`. The six PARTIALs are all `_ZTS` name strings where `emitted` falls exactly 2
bytes short of `romExtent` — the ROM's 4-byte padding of the string, not a disagreement.

**The sibling oracle** is `dScMgSingle3DBase_c` (ov006, `status: promoted`, same one-`.text`-claim
text-only shape, `.text 0x0210a4b0..0x0210a8c0`). This TU copies its route exactly.

---

## Things a reviewer should not be surprised by

**`tubuild linkcheck` cannot be re-run on this TU now, and that is pre-existing.** Once the
source is in `delinks.txt`, `cmd_linkcheck` sets `enrolled_intact_candidate = True`
unconditionally and routes to the intact-object path whatever `production_mode` says; the
text-only manifest then trips `tu_production.prepare_intact_object`, which raises when there is
one `.text` claim and no non-text claim. **Control run, on a class this PR does not touch and
whose `delinks.txt` block and manifest are byte-identical to `origin/main`:**

```
$ python tools/tubuild.py linkcheck ov006/dScMgSingle3DBase_c
[4/8] ... 1 failed
FAIL src/minigames/d_s_mg_single3_d_base.cpp: isolate: intact TU preparation refused:
     ov006/dScMgSingle3DBase_c: intact production requires one .text claim and at
     least one non-text claim
exit 1
```

`ov006/dScMgRoulette_c` produces the identical message at the identical phase. The link evidence
for this TU is therefore the **pre-promotion** `linkcheck`, recorded in the manifest.

**The recorded link evidence and the shipped source are from the same tree — verified, not
argued.** The writer removed the two shard-level pragmas after the recorded `linkcheck` run.
Re-measured here with an isolated compile at the ROM build's own flags:

| source form | object |
|---|---|
| as shipped, no pragmas | 29,744 B  `9569de726c9193c1b70f0414aee978a432b7124bcf187dfc4995ab533b565be3` |
| `optimize_for_size on/off` bracketed around `func_ov006_021094ac` | **identical** |
| `opt_propagation off/on` bracketed around `func_ov006_02108b90` | **identical** |
| both bracketed | **identical** |
| both transplanted **verbatim** from the shards (one-way, no closing pragma) | 29,584 B `aa30cbc1...` — **different** |

So the claim holds in its scope-preserving form, with one caveat worth recording: these pragmas
are file-global last-wins, so a shard's one-way file-scope pragma is *not* a no-op when copied
verbatim into a merged TU — it must be bracketed. The shipped TU carries no pragma at all, which
sidesteps the question. And independently of all of it, the full `rombuild` of the shipped source
produces the same ROM sha256 as the recorded pre-promotion `linkcheck`, which settles it directly.

**`dsd check symbols --fail` reports 9 errors.** They are identical to the run's own
`_baseline` control's `symbolCheckBaselineErrors` (ITCM `func_01ff*`, `_deq`, `data_020ad5*`,
`overlay_100/102`) — pre-existing and unrelated to this change.

**The many-to-one credit fold is banked, not tolerated — and a shallow clone nearly poisoned
it.** A promotion normally reports `0 added, N changed, 0 lost` because the counter credits only
the delinks range's first symbol and reclassifies the rest as "claimed". Here
`prepush_attribution` reported `0 changed, 37 lost`: git records the 40 deletes plus one add
under a different directory as delete+add, not a rename, so the lineage ends and credit
re-points at whoever pushes. `c5c0551` therefore adds 40
`src/actors/dScMgRoulette_c.cpp#<symbol>` rows to `attribution.json`'s `overrides` (append-only,
in manifest order, after the ov025 block; no existing row touched or reordered), including the
three members the gate did not flag (`D1`, `D0`, `func_ov006_02108b90`) so the whole TU carries
an explicit owner rather than three of forty depending on a similarity heuristic.

**Its first version had every owner wrong, and `39dc24b` fixes it.** This checkout is a
**shallow clone** — 1,325 of 4,743 commits. `chaos_db_ci.first_matchers()` and
`match_finishers()` both replay `src/` history, so truncated history makes every shard's
lineage appear to begin at the newest commit that touched it, which for all forty was the bot's
`[skip ci]` progress refresh. `prepush_attribution` consumes those same maps, so its
`[github-actions[bot]]` handles read as fact and are an artifact. After `git fetch --unshallow`
the answer is completely different:

```
tangosdev 24, andrewboudreau 7, ruspecial 7, lunavyqo 1, aitddlabs 1
(resolved via: 32 first-adder, 6 finisher, 2 pre-existing shard override)
```

This PR's first validation run printed 25 of the 40 pre-fold owners; the recomputed set agrees
with **all 25, 0 disagreements**, which is what the 15 it truncated are derived from. The
superseded rows would have moved 16 functions off four named contributors and mislabelled the
other 24 — permanently, since an `attribution.json` override is highest priority and is never
pruned. With the corrected rows the gate reports **`32 consolidated with credit intact,
0 changed, 0 lost`**: the fold artifact is gone rather than explained away, and every
contributor keeps exactly the credit the validator itself assigns them.

**One commit is not part of the promotion.** `80a861c` moves
`include/dScMgRoulette_c.h`'s inline destructor opening brace onto the declaration line.
`check_header_offsets.py` only skips an inline method body when the opening brace is on the
declaration line — the `n == 0` branch computes `depth` from the declaration line alone — so the
Allman form left `skip_body` at 0 and the bare `{` fell through to declaration parsing. The cost
was not only `exit 1`: it reported `0 commented fields, 0 mismatched, 1 unparsed, struct spans
0x4f38`, i.e. it checked **none** of the 22 field offsets and read the span as the first field's
offset. With the brace moved: `22 commented fields, 0 mismatched, 0 unparsed, struct spans
0x5400`. Whitespace only; `tubuild verify` still reports 40/40 MATCH. The parser defect itself is
unfixed and worth a separate change.

---

## Manifest audit

Re-read against what actually shipped; no superseded prose found. Specifically:

- All **40** `legacy_source` rows are real `delinks.txt` entries on `origin/main` with matching
  extensions (the `dScMgMemory2_c` `.c`-vs-`.cpp` defect does not recur here), and the
  `legacy_source` set is exactly the set of 40 deleted files — symmetric difference empty.
- `boundary_evidence[3]` says "36-word table"; the ROM says 36 slots. Agrees.
- `boundary_evidence[5]`'s destructor-order claim reproduces: D1 `0x0210788c` below D0
  `0x02107920`, and `_ZN15dScMgRoulette_cD2Ev` does not exist anywhere in `config/` or
  `extracted/`.
- `boundary_evidence[6]`'s pragma controls reproduce, including its own explicit retraction of an
  earlier draft's positionality claim.
- No dead declarations: of 126 declared externs and 15 struct/typedef tags, every one is
  referenced. The 17 identifiers a naive scan flags (`v4`, `d0`..`d17`) are deliberate
  vtable-slot placeholders inside the `Sub` and `VtObj` shadow structs, not residue.

## Files

50 files: `src/actors/dScMgRoulette_c.cpp` added, 40 shards deleted, plus the manifest, the class
facts, `delinks.txt`, `include/dScMgRoulette_c.h`, `include/decl_common.h`, `attribution.json`,
the generated `notes/cpp-tu-current-state.md`, and three prose notes that named a deleted shard.


andrewboudreau:
The validation above passed, but read its **Contributor credit** table before the verdict — it
caught a real defect I had pushed.

`c5c0551` credited all 40 folded members to `github-actions[bot]`. The validator's own "Before"
column shows what that would have cost: `andrewboudreau`, `ruspecial` and `lunavyqo` losing
functions they matched, and 24 more mislabelled off `tangosdev`.

Root cause, worth recording because it will hit the next promotion the same way: **this checkout
is a shallow clone** — 1,325 of 4,743 commits. `chaos_db_ci.first_matchers()` and
`match_finishers()` replay `src/` history, so truncated history makes every shard's lineage look
like it begins at the newest commit that touched it — for all forty, the bot's `[skip ci]`
progress refresh. `tools/prepush_attribution.py` consumes those same maps, so its
`CREDIT LOST ... [github-actions[bot]]` lines read as a statement of fact and are an artifact of
the clone depth. Nothing warns you.

After `git fetch --unshallow`, recomputing with the tools' own precedence
(`override > finisher > first-adder`, then `identity_aliases`) gives:

```
tangosdev 24, andrewboudreau 7, ruspecial 7, lunavyqo 1, aitddlabs 1
(32 first-adder, 6 finisher, 2 pre-existing shard override)
```

That set agrees with **all 25** pre-fold owners this validation run printed — 0 disagreements —
which is the basis for the 15 it truncated. `39dc24b` replaces the rows with it, and
`prepush_attribution` now reports `32 consolidated with credit intact, 0 changed, 0 lost`.

Nothing else moved: `attribution.json` is the only file the correction touches, so the byte
gates are unchanged (`tubuild verify` 40/40 MATCH, `rombuild` 106/106 exact 100.000000%, built
NDS `d1506e90…c478e8` identical to the run's own source-independent stock baseline control).
Please validate `39dc24b1c` rather than `c5c05519d`.

PR #2332, opened 2026-09-06
Regenerate the queue #2319 left stale, and stop two docs asserting a retracted rule

andrewboudreau:
Found by the `dScMgTeresa_c` builder (PR #2329).

**`queue_audit --check` has been red on `main` since I merged #2319** without regenerating the queue — `dScMgBomroom_c`''s row still read `promoted no` with its pre-promotion shard count. Regenerated; `--check` agrees with the tree again.

The gate is **not wired into any workflow**, so nothing would have caught it, and `builder.md`''s gate list did not mention it either. Now listed, with the two ordering traps: it needs `build/rtti.json` and a `build/tu_map.json` **newer than `config/`**, and it goes red after every promotion until `--write` is run.

Worth deciding separately: whether `queue_audit --check` should join CI. It would keep the queue honest without anyone remembering, but it makes every promotion PR carry a regenerated TSV, so I have not added it here.

**`tools/tu_order_check.py`''s docstring still asserted the rule #2321 retracted** — "a TU that defines its destructor OUT OF LINE makes mwcc emit `D2, D0, D1`", stated unconditionally. `dScMgTeresa_c` refutes it: out-of-line under `#pragma defer_codegen off` emits `D1, D0, D2`, and an inline destructor emits its variants in the reverse order of their first odr-use. The docstring now says the order is a per-TU measurement — which is what the tool exists to make.

**And "a failing `verify` rewrites the prose fields" understates the damage.** On a measured negative control it also rewrote the entire `droppedSections` list — the compiler-only policy''s section index list, not prose — alongside `functions_matched: 48 -> 45`. A five-second control can leave a wrong licensing list in a manifest you then commit.


andrewboudreau:
Superseded by #2339, which regenerates the queue against the post-train head (all six promotions landed since this branch was cut) and carries this PR's two content changes forward: the `tu_order_check` docstring rewrite and the `droppedSections` correction to builder.md. Both were cherry-picked, not retyped.

Closing rather than rebasing because the queue file this PR regenerates was stale again within an hour of being written — regenerating once, after the train drained, is the right shape.

PR #2347, opened 2026-09-06
ov006: promote dScMgPanel_c into a single C++ TU (71 of 71)

andrewboudreau:
## ov006: promote dScMgPanel_c into a single C++ TU (71 of 71)

The whole linker run. `tools/tu_map.py` calls `0x0210428c..0x02107858` one contiguous run of 71 functions and this TU licenses all of it — no sourceless hole, no held-out shard. 71 one-function shards are absorbed and deleted.

### Proof block, re-derived at `4ccb54de3`

```
tubuild verify    71/71 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
object audit [4b] order_ok True, {'LICENSED': 71}, 0 extra sections, 0 refusals
object symtab     0 mangled undefined symbols (41 before the repair below)
rombuild -j6      module fidelity: 106/106 exact, 100.000000% of compared bytes
                  source-built functions 11,174 reproducing, mismatching 0
                  source-owned data claims 20 reproducing, 0 mismatching
                  ROM-build analysis: PASS
                  profile stock: 0 mod source replacements, 0 ROM-gap fallbacks
ROM sha256        d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
                  identical to the run's own source-independent BASELINE CONTROL
romdata_check     6 VERIFIED, 5 PARTIAL, 0 DIFFERS, 0 UNNAMED (11 data symbols)
tu_order_check    all match in ROM order; EXTRA _ZN12dScMgPanel_cD2Ev has its policy
source_coverage   clean, 0 B handed back to the cartridge, net +0 B
prepush_attrib.   59 consolidated with credit intact, 0 changed, 0 lost
tiers_ratchet     CONVERTED ratchet PASS, baseline 2680, current 2680
```

`rombuild`'s BASELINE CONTROL prints `dsd check symbols --fail FAIL`: **9 errors, the same 9 that are on untouched `main`, 0 new.**

`git rev-parse --is-shallow-repository` returned `false`, checked before any attribution row was banked.

### Stage 4 found a real blocker: the TU did not link

As it arrived from stages 2 and 3 this promotion compiled, byte-matched 71/71, passed `objisolate`, passed reloc-destinations and passed the `[4b]` object audit — and `mwldarm` aborted with 22 `Undefined`, every one of them `Referenced from "dScMgPanel_c::InitResources()"`.

The mechanism. Each of the 67 `func_ov006_*` members sits inside its own file-scope `extern "C" { }` region, so a block-scope `extern` declaration in its body has **C linkage** and names the ROM symbol — that is the whole reason 71 independently recovered spellings can coexist. A **class member function cannot sit inside such a region**, so the identical block-scope declarations in the four C++-named members (ordinals 67-70) had **C++ linkage**. The object carried 41 mangled undefined symbols that exist nowhere in the image: `_Z8LoadFilei`, `_Z10DeallocatePv`, `_Z19func_ov006_021063a0Pv`, and so on — including both `_Z19func_ov006_02106168Pc` *and* `..._02106168Pv`, two mangled ghosts of one ROM symbol that this TU itself defines.

Nothing before the linker looks at this. Every relocation is against a name, and every byte gate is satisfied by a relocation against a name nobody defines.

The repair takes the shape `dScMgTeresa_c` and `dScMgRoulette_c` already use:

- **one file-scope `extern "C"` region** carrying the 24 genuinely external ROM symbols those four members call, placed **after ordinal 66** so that none of the 67 members above it can see it and their recovered spellings stand exactly as they were. `mwccarm` 2004/b56 rejects a block-scope linkage specification outright — measured: `extern "C" { extern void f(int); }` inside a function body is `Error: declarator expected`. There is no in-body fix.
- the **13 declarations naming members of this TU** are deleted rather than moved — those symbols are defined earlier in the same file, so the call binds to the definition. Three call sites take the definition's own pointer type (`func_ov006_02106168(this)` in ordinals 67 and 70, `(PanelObj_ab4 *)` in 68, `(PanelC_ac4 *)` in 69).
- `func_02073244` in ordinal 69 was a **dead declaration** — nothing in the body referenced it — and is dropped.
- **data declarations are untouched and did not need to move**: `mwccarm` leaves a file-scope variable's name unmangled in C++, so `data_0209d454`, `data_0209d45c` and `data_ov006_02142888` stay in their own bodies.

Bytes are unchanged by all of it: 71/71 MATCH before and after, emission order still ROM-ascending.

**The manifest's recorded `linkcheck: scratch-link-verified` was therefore not reproducible on the source as shipped.** That snapshot is now marked `superseded-by-production-link` rather than left standing, and `verification.production` records what stage 4 measured. The live link proof is `rombuild`'s own BASELINE CONTROL, which does a full scratch delink, lcf, whole-tree `mwccarm`, `mwldarm` link and ROM build of the **working tree** — a text-only promoted TU is compiled and linked inside it like any other enrolled source. That control is what caught this, in a fresh worktree that had to bootstrap it.

### Class identity

| | |
|---|---|
| RTTI name | `dScMgPanel_c` — a cartridge string, not coined |
| base | `dScMgBase_c`; confirmed leaf, no RTTI record names it as a base |
| `_ZTS12dScMgPanel_c` | ov006 `0x0213dd84` |
| `_ZTI12dScMgPanel_c` | ov006 `0x0213dd40` (base `_ZTI11dScMgBase_c` ov004 `0x020bbf6c`) |
| `_ZTV12dScMgPanel_c` | ov006 `0x0213e24c`, 36 slots, storage size `0x98` |
| factory `new` size | `0x4fec` (`tools/opnew_sizes.py`, one inline_store site, `ov006:dScMgPanel_c_classInit`) — equal to the header's `sizeof(dScMgPanel_c) == 0x4fec` assert |

Six of the 36 slots are this class's own, and every one lands inside the claimed range; the other 30 relocate out to ov004 or arm9:

```
slot 0   0x021073b0  InitResources()
slot 6   0x02107358  Behavior()
slot 9   0x0210730c  Render()
slot 16  0x0210428c  D1     <- key function, lowest address of the run
slot 17  0x021042b0  D0
slot 18  0x021071fc  OnYoshiTryEat(int)   override of dScMgBase_c's
```

**Destructor**: out-of-line and declared FIRST in the header, so it is the key function and this TU emits the whole chain's vtable and typeinfo as vague-linkage passengers. Under `#pragma defer_codegen off` the variants come out D1, D0, D2; `D2` is homeless — no address in `config/` names one — and is deadstripped. The inline form was rejected for a second reason: it would move the key function to `InitResources`.

**Route**: text-only + `compiler_only_output`, matching the sibling oracle `dScMgTeresa_c`. It cannot be intact-object — ov006's entire `.data` segment is one section owned by no file, so `_ZTV`/`_ZTI`/`_ZTS` are licensed as compiler-only output and every static stays `extern`. All 11 homed `compiler_only_output` rows were checked against their owning module's own `symbols.txt`: **11/11 agree on the address and each is defined exactly once**. All 71 `legacy_source` rows were audited against the pre-promotion tree: **71/71 present, 0 with the wrong extension**, and all 71 are deleted here.

**Pragmas**: eight bracketed optimisation pragmas over three families, under a file-level `#pragma defer_codegen off`. Negative control re-run here rather than taken from the writer — deleting only `defer_codegen off` gives `order_ok False` and `licensed .text functions are not emitted in ROM address order`. It is load-bearing, and it is why the file is written ROM-ascending.

### Two pre-existing tool findings, both controlled

**1. `tubuild.py linkcheck` cannot pass on any already-promoted text-only TU.** It refuses at `[4/8]`:

```
FAIL src/actors/dScMgPanel_c.cpp: isolate: intact TU preparation refused:
     ov006/dScMgPanel_c: intact production requires one .text claim and at
     least one non-text claim
```

Control, on `ov006/dScMgTeresa_c` — already promoted, text-only, landed on `main`, untouched by this branch, and with its manifest and its `delinks.txt` block byte-identical to `origin/main`:

```
[1/8] scratch configuration
      delinks.txt already contains the exact final intact-object entry; no scratch splice is needed:
        src/actors/dScMgTeresa_c.cpp: complete
          .text            0x0211cbd0..0x0211e72c
[2/8] dsd delink        ok (3.3s)
[3/8] dsd lcf           ok (0.8s)
[4/8] mwccarm: 9329 enrolled source file(s), -j20
      9327 reused from cache, 1 compiled, 1 failed (50.1s)
      FAIL src/actors/dScMgTeresa_c.cpp: isolate: intact TU preparation refused:
           ov006/dScMgTeresa_c: intact production requires one .text claim and
           at least one non-text claim
```

Identical refusal, different class, nothing to do with this change. The chain is `tubuild.py:4613` — `enrolled_intact_candidate = True` for any entry already in `delinks.txt` that is neither `partial` nor `partitioned`, whatever `production_mode` says — into `tubuild.py:4777` (`intact_override[entry["source"]] = entry`) into `tu_production.py:41`, which refuses a manifest with no non-text claim. A text-only promotion has `data: []`, `bss: []` and no `production_mode` key at all, so it can never satisfy it. `rombuild.intact_tu_policies` (`rombuild.py:773`) skips entries without `production_mode: "intact-object"`, which is why the normal build never takes this path. **No `production_mode` and no fake non-text claim was added to silence it** — this TU cannot own its vtable, and `rombuild` would have had to eat the lie later.

**2. `d_a_obj_rc_carpet.cpp` and `d_a_propeller_heyho.cpp` failing `[4/8]` with `baseline content fingerprint differs for ['trackedConfigArm9Sha256']` is not tree-wide scratch drift.** It is a worktree with no `build/tu/_baseline` yet: both are `production_mode: intact-object` and their vtable-partition proof needs that control to exist. Before `rombuild` generated it, the control run above reported `3 failed`; after, `1 failed`, and the only survivor is the Teresa refusal. Nothing to fix, and nothing specific to this class.

### Deliberately not in this PR

- **`notes/data/tu-merge-candidates.json`.** Generated, ungated, and with no generator anywhere in `tools/` or `.github/workflows/`. Measured: it names **5,859 dead `src/` paths**, of which **118 are this promotion's** and **5,741 predate it**. It is a tree-wide stale artifact rather than this class's debt, and rewriting it from a class PR would conflict with every other open promotion.
- **`queue_audit --check`'s `unmatched: 14 row(s) disagree with the tree`.** All 14 are foreign rows moved by main's own #2337 and #2338. `dScMgPanel_c`'s own row **is** refreshed here — `shard_count 72 -> 2`, `promoted no -> yes`, one line.
- the `--update` nudges from `check_references` (3 fixed since the baseline) and `check_dead_references` (1 baselined reference now resolves). Both are shrinkable baselines belonging to the tree.

### Gates run

`tubuild verify`, `rombuild -j6`, `romdata_check`, `tu_order_check`, the `[4b]` object audit and its negative control, `tiers_ratchet --check` (PASS, so **no `--update`** — the ledger audit found 0 byte-identical repeats in `converted-backslide-exceptions.jsonl` and a balanced 3-out/3-in swap in `converted-baseline.json` with no foreign row touched), `eligible.py`, `check_references`, `cpp_tu_compat --require-ready`, `check_src_tu_compiles` (148/148), `check_rename_ledger`, `check_profile_campaign`, `langmode_audit`, `check_header_offsets`, `port_refcheck` (402 checked, 0 stale), `check_duplicate_sources`, `check_dead_references`, `queue_audit`, `cpp_tu_state --check-note`, `check_tubuild_conflicts`, `layout_check`, `source_coverage`, `prepush_attribution`.

Nothing was rebased — `origin/main` moved 7 commits during the build and was **merged** in twice. `attribution.json` resolved as an exact union of both parents (2,764 = 2,705 + 59, 0 value conflicts, 0 invented); `converted-baseline.json` held at 2,680 with the 3 shard paths correctly replaced by 3 member rows.

### Role files corrected

`writer.md` on `main` calls this TU "71/71, link-verified" and presents the block-scope declarations as the resolution. Both are corrected from measurement, and `builder.md` gains the real link proof, the full `[4/8]` refusal chain with its control, and a cheap undefined-symbol scan of the compiled object.


andrewboudreau:
## Stage-5 review — independently re-derived at `b7dc904d4`

I am not the agent that built this. Everything below I ran myself, in two fresh
`C:\tmp\sm64ds-rev2347*` worktrees wired per `decomp-worktree` (both since torn down with
`wt-remove.ps1`; ROM dump and `tools/mwccarm/` verified intact after).

### Re-derived, not taken on trust

| what | my result |
|---|---|
| `tubuild verify ov006/dScMgPanel_c` | **71/71 MATCH**, objisolate clean, reloc-destinations clean, ROM-ascending order — at `4ccb54de3` and again at `b7dc904d4`. `git status` clean after every run (no manifest rewrite). |
| `rombuild.py -j 6` | **106/106 exact, 100.000000%**; 11,174 reproducing / 0 mismatching; 20 data claims reproducing; ROM-build analysis **PASS**; stock profile, 0 mod replacements, 0 ROM-gap fallbacks. |
| built ROM | `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8` == the run's own BASELINE CONTROL ROM. The control's `dsd check symbols --fail` FAIL is the same **9** pre-existing errors, **0 new**. No `LINK FAILED`, no `Undefined`. |
| object `.symtab` scan | **104 undefined, 0 mangled-ghost `_Z<len><name>` symbols.** All 42 `_Z*` undefined are legitimate `_ZN…` names (inherited `dScMgBase_c` / `dScene_c` / `fBase_c` slots, `G2` / `GX` / `G2S` / `GXS`, `Sound::PlayBank2_2D`, `Memory::Deallocate`) plus the two `abi` type_info vtables. The linkage repair holds. |
| `mwccarm` block-scope `extern "C"` | reproduced: `extern "C" { extern void f(int); }` inside a function body gives **`declarator expected`** and the tool aborts. There is genuinely no in-body fix. |

### The linkage repair, checked line by line

- **44 removed declarations → 24 kept + 14 dropped.** All 13 `func_ov006_*` dropped as
  "this TU's own member" carry a `// @symbol` marker and a definition earlier in the file.
  `func_02073244` has no definition here **and no remaining reference** — genuinely dead.
  None of the 24 kept in the file-scope region is defined in this TU.
- **No spelling was silently canonicalised.** Only three names had two different signatures
  across the four C++-named members (`func_ov006_02106168`, `_021063a0`, `_021067a4`, each
  `char *p` vs `void *p`) — and all three are among the 13 *deleted*, so both spellings
  collapse onto the real definition instead of one guess being picked over the other. Every
  one of the 24 externals kept has a single, unchanged spelling.
- **The three retyped call sites do not change codegen** — proven by the 71/71 MATCH after
  the change. `func_ov006_02106168` is defined `(dScMgPanel_c *)`, so `this` is required;
  `PanelObj_ab4` and `PanelC_ac4` are plain structs with no base, so those casts are
  value-preserving.
- **Placement after ordinal 66 does not disturb emission order.** Declarations emit nothing;
  `verify` re-reports `all 71 function(s) in the expected ROM-ascending section order` and the
  `[4b]` audit is `order_ok True`. The region sits at line 2351, after the last member region
  at 2323, so none of the 67 wrapped members can see it.

### The five `romdata_check` PARTIALs — settled

They are **all five `_ZTS` typeinfo name strings**, every one PARTIAL for the same benign
reason: the emitted NUL-terminated string is shorter than the ROM's 4-byte-aligned extent,
and `verify_data_symbol` compares whole words only, so the verdict falls through
`len(linked) >= extent`. **0 differing bytes and 0 blind words in all five.**

```
PARTIAL  _ZTS12dScMgPanel_c   ov006 0x0213dd84  emitted=15  romExtent=16  compared=12  blind=0
PARTIAL  _ZTS11dScMgBase_c    ov004 0x020bbf84  emitted=14  romExtent=16  compared=12  blind=0
PARTIAL  _ZTS8dScene_c        arm9  0x020914b0  emitted=10  romExtent=12  compared=8   blind=0
PARTIAL  _ZTS7dBase_c         arm9  0x02086e54  emitted=9   romExtent=12  compared=8   blind=0
PARTIAL  _ZTS7fBase_c         arm9  0x02086e60  emitted=9   romExtent=12  compared=8   blind=0
VERIFIED _ZTV12dScMgPanel_c   ov006 0x0213e24c  emitted=144 romExtent=144 compared=144 blind=0
```

**None is a vtable extent defect in either direction.** `_ZTV12dScMgPanel_c` is VERIFIED with
`emitted == romExtent == compared == 144` (36 slots plus the 8-byte preamble = the manifest's
`0x98` storage size) — the table matches the cartridge end to end. And the class does **not**
own it: `data: []`, `bss: []`, a single `.text` claim, and `_ZTV` licensed as
`compiler_only_output` / `deadstrip-data` homed at ov006 `0x0213e24c`.

### The `linkcheck [4/8]` refusal — control re-run, and strengthened

Reproduced on this branch, then **on pure untouched `origin/main` `9126e60de`** with no part
of this promotion present:

```
[1/8] scratch configuration
      delinks.txt already contains the exact final intact-object entry; no scratch splice is needed:
        src/actors/dScMgTeresa_c.cpp: complete
          .text            0x0211cbd0..0x0211e72c
[2/8] dsd delink        ok (3.1s)
[3/8] dsd lcf           ok (0.8s)
[4/8] mwccarm: 9400 enrolled source file(s), -j20
      9397 reused from cache, 0 compiled, 3 failed (42.1s)
      FAIL src/actors/dScMgTeresa_c.cpp: isolate: intact TU preparation refused:
           ov006/dScMgTeresa_c: intact production requires one .text claim and at
           least one non-text claim
```

Identical refusal, different class, on a tree this PR never touched. It is a general
pre-existing tool defect affecting every promoted text-only TU. `rombuild`'s BASELINE CONTROL
is the link proof here, and the PR says so plainly. No `production_mode` key and no fake
non-text claim were added — confirmed by reading the manifest.

(The other two failures in that same control run were `daObjRcCarpet_c` and
`daPropeller_Heyho_c` with `baseline content fingerprint differs for
['trackedConfigArm9Sha256']` — a stale `build/tu/_baseline` for that config, exactly the
artifact the PR describes, demonstrated here in the reverse direction.)

### Merge tree, not the branch

Before the re-merge I composed the merge myself (`origin/main` plus this branch, resolving
`attribution.json` as a union and regenerating `cpp-tu-current-state.md`) and gated it:
`verify` 71/71 for Panel **and** 33/33 for `dScMgCoin_c`, `rombuild` 106/106 100.000000%,
the same ROM hash, `source_coverage` net **+0 B**, `tiers_ratchet` PASS 2680/2680,
`langmode_audit` PASS, `cpp_tu_state --check-note` current, plus `check_dead_references`,
`check_duplicate_sources`, `check_tubuild_conflicts`, `layout_check`, `check_rename_ledger`
and `check_profile_campaign` all green.

`git diff` between my composition and the pushed `b7dc904d4` is **one file** —
`notes/data/tu-promotion-queue.tsv` — so the content I gated and the content that shipped
are the same bytes.

### Ledger audits, re-derived against the true merge base `091a71601`

- `ov006/delinks.txt`: exact union. 1270 − 71 (Panel) − 33 (Coin) + 2 = **1168**; no block's
  content matches neither parent; both splices present; the two ranges
  (`0x020dbf7c..0x020dd0e0` and `0x0210428c..0x02107858`) are disjoint.
- `attribution.json`: **2796 = exact union** of 2764 and 2737 over base 2705; 0 value
  conflicts; 0 invented keys; `merged[:len(base)] == base`, so **key order is preserved
  byte-for-byte and nothing was re-serialised**; `github-actions[bot]` credits unchanged at
  968, and **none of the 59 new rows credits the bot**.
- `converted-baseline.json`: **2680 on all four sides**, the `count` field agrees with the
  list length, 0 duplicates, removals are exactly the 3 Panel shard paths, additions exactly
  the 3 `src/actors/dScMgPanel_c.cpp#…` member rows, 0 foreign rows touched.
- `config/match_attempts.jsonl` and `match_provenance.jsonl`: 4 in-place `srcPath` rewrites,
  line counts unchanged (2072 / 720), author `lunavyqo` and every other field byte-identical.
  No row deleted, none added.
- `symbols/actor_renames.tsv` and `notes/data/c-cpp-classification.tsv`: **untouched**.
- Queue regeneration: `queue_audit --check` returns `queue agrees with the tree`, exit 0
  (after `rtti_extract` and `tu_map`). The 14 foreign rows that moved differ **only** in the
  `blockers` column and **only** by dropping stale `unmatched:` tokens; row count 267 is
  unchanged; no `shard_count`, `already_promoted`, `sibling_oracle` or `promotion_route`
  was touched on any foreign row.

### Class identity

`compiler_only_output`: **11 of 11 homed rows agree with their own module's `symbols.txt`**
on the address, and each is defined exactly once; the 12th, `_ZN12dScMgPanel_cD2Ev`, is
homeless and deadstripped. D1 (`0x0210428c`) then D0 (`0x021042b0`), no D2 in ROM order.
All **71 of 71 `legacy_source` rows equal the 71 deleted shards exactly** — none missing,
none extra. The header's layout re-derived from its field list closes at **exactly `0x4fec`**,
matching the `opnew_sizes.py` factory size and the static assert, and every field's comment
offset matches its computed offset. `OnYoshiTryEat` is a cartridge symbol
(`_ZN12dScMgPanel_c13OnYoshiTryEatEi` in `ov006/symbols.txt`), not an invented name, and the
English field readings stay in comments over `unk_*` identifiers behind an explicit
"these are readings, not recovered identifiers" disclaimer — which is what the scout's
`unproven` list requires.

### Verdict: MERGE

Two things worth recording rather than fixing:

1. The validator reports **10 changed, 0 lost**, not the PR body's local `0 changed`. Those
   10 are exactly the ten members with **no `attribution.json` override row** —
   `func_ov006_02104a10`, `_02104c60`, `_02105d20`, `_02106080`, `_02106664`, `_021067a4`,
   `_02106aa8`, `_02106fdc`, `_0210713c` and `_ZN12dScMgPanel_c13OnYoshiTryEatEi` — the
   fallback-to-commit-author mechanism `builder.md` documents. `lost` is 0, which is the
   merge bar. Adding those 10 rows from the validator's own *Before* column would close it.
2. The `-70` byte-verified figure in the validator table is the known many-to-one fold
   artifact (enrolled ranges `+0`, module fidelity 106/106, `source_coverage` `+0 B`), not a
   regression.

PR #2336, opened 2026-09-06
Repair 20 counted files that had drifted from the ROM

tangosdev:
A sweep of the delinks configuration found 116 single-function source files whose entry is a plain address range with
no `complete` line. dsd fills those from the cartridge, so the ROM build never compiles the file and never byte-compares
it. Anything that drifted out of shape since it landed stayed invisible.

Of the 116, 56 still reproduce byte-for-byte under the pinned compiler and 60 did not. This branch repairs 22 of the 60
and marks 36 entries `complete`, so from now on the ROM build compiles those files and compares the result to the
cartridge.

## The 22 repairs

Every one is the same family of problem: an offset that the cartridge folds into a load or store had drifted into a
literal-pool constant, or a read-modify-write had let the compiler common up an address the cartridge recomputes. All 22
verify byte-for-byte at their symbol address and size under 2004/b56.

| Symbol | Module | Address | Size | What had drifted |
| --- | --- | --- | --- | --- |
| `func_02005098` | arm9 | 0x02005098 | 0x44 | int-typed base materialized the 0x644 field offset through a literal pool; char* base folds it into the load |
| `_ZN11dCapEnemy_c12UpdateCapPosERK7Vector3RK10Vector3_16` | arm9 | 0x020062b8 | 0xf8 | a local CapFlags* at +0x17f materialized the offset through the literal pool; inline casts fold it into ldrb |
| `func_0201fb4c` | arm9 | 0x0201fb4c | 0xc4 | the ==2 flag needs unsigned int width or mwcc folds the moveq/movne/cmp materialization into a single beq |
| `_ZN15TextureSequence17UpdateFileOffsetsER8BTP_File` | arm9 | 0x020468a0 | 0x10c | record loops reuse the loaded offset word for the add; the += form CSEd p+4 and reloaded through it |
| `func_0204a17c` | arm9 | 0x0204a17c | 0x34c | Flags2* local CSEd rec+4 into r6 and read the raw word through it; inline casts re-issue ldr [r1,#4] |
| `func_020584d0` | arm9 | 0x020584d0 | 0x68 | int-typed list nodes let mwcc CSE prev+0x68 for the read and write; char* nodes keep the folded offsets |
| `func_02065de4` | arm9 | 0x02065de4 | 0xd4 | second +0x1718 access shared the literal-pool constant with the first; stepping +0x1000 then +0x718 restores add/ldr |
[...]

PR #2348, opened 2026-09-06
ov006: promote dScMgCup_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

`dScMgCup_c` is the cup-shuffle minigame scene. Its code was scattered across
32 one-function source files with machine names like `src/func_ov006_020def80.c`
— an artifact of how the decompilation started, not how Nintendo wrote it. The
ROM proves those 32 functions were one `.cpp` file: they occupy a single
uninterrupted address run, `0x020de988..0x020e0638`, with `dScMgCoin_c` ending
immediately below and `dScMgCurling_c` beginning immediately above.

This PR folds them back into one file, `src/actors/dScMgCup_c.cpp`, and the
compiler still produces the cartridge's bytes exactly — all 32 functions, and
all 103 overlays plus both main modules still link byte-identical to the retail
ROM.

Two things here are worth a reviewer's attention beyond the byte match:

1. **A symbol rename in this PR is load-bearing, not cosmetic.** Eight functions
   are renamed from `func_ov006_*` to class-anchored member names. Eight 8-byte
   pointer-to-member records in ov006's `.data` point at those eight addresses,
   and `dsd` resolves those relocations *by symbol name*. Before the rename all
   eight code words linked as `0x00000000` — 32 differing bytes, the promotion's
   only module-fidelity failure. They now link non-zero and correct; the check
   is reproduced below.
2. **One recovered function reads an uninitialised local, on purpose.** That is
   undefined behaviour in C++. It is also what the cartridge does, and removing
   it breaks the match. Full disclosure below — please decide deliberately
   rather than discover it.

---

## Proof block

Every figure below was re-derived in this stage's own worktree
(`C:\tmp\sm64ds-cupmatch`, branch `cpp/dScMgCup_c-match` at `7d7e336da`), on the
tree as merged with `origin/main` at `9d45c0082`, not inherited from the
producing stages.

```
tubuild verify ov006/dScMgCup_c
  32/32 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
  emission order: all 32 function(s) in the expected ROM-ascending section order

rombuild -j 16 --no-rom
  [4/6] mwldarm
  source-built functions: 11,175  (2,121,644 / 2,211,124 code bytes, 95.95%)
    reproducing: 11,175
    mismatching: 0
  source-owned data claims: 20  (reproducing 20, mismatching 0)
  ROM-build analysis: PASS

rombuild BASELINE CONTROL (regenerated this run; no TU substitution)
  [5/8] mwldarm (scratch module link)  ok (29.1s)
  BASELINE CONTROL: modules PASS, dsd check symbols --fail FAIL (the 9 pre-existing
  error lines: overlay_100, overlay_102, data_020ad524, data_020ad560 and five ITCM
  symbols), ROM built.

romdata_check --files src/actors/dScMgCup_c.cpp
  13 data symbols: VERIFIED 7 (212 bytes), PARTIAL 6 (68 bytes), DIFFERS 0, UNNAMED 0

prepush_attribution --base origin/main
  9852 tracked, 28 consolidated with credit intact, 0 changed, 0 lost
```

Every other gate in the builder list exits 0 on this tree:
`cpp_tu_compat --require-ready`, `check_src_tu_compiles --quiet` (149/149),
`eligible -j 16` (11161/11256), `check_references --against origin/main`
(*"no new unresolvable references"*, 42 unresolved vs a baseline of 45),
`check_rename_ledger`, `check_profile_campaign`, `langmode_audit --check`,
`check_header_offsets --changed origin/main`, `port_refcheck`,
`check_duplicate_sources`, `check_dead_references`,
`check_tubuild_conflicts --list`, `tiers_ratchet --check`, `queue_audit --check`,
`cpp_tu_state --check-note`.

**No gate needed `--update` in this stage.** `tiers_ratchet` was already banked
once by stage 2 (2680 -> 2682) and reads `PASS baseline 2682 current 2682` here;
it was not re-banked. `port_refcheck` never went red — the `port/` slice
manifests name none of the deleted shards.

---

## The link proof, which is the part the byte gates cannot give you

`ov006/dScMgPanel_c` (#2347) passed byte compare, `objisolate`, `reloc_audit`
and the `[4b]` object audit and then died in `mwldarm` with 22 `Undefined`.
Byte match is not a link proof, so this PR carries two independent link
arguments.

**1. The object's symbol table has no mangled ghosts.** The failure mode in
#2347 was a `func_*` declared inside a class member body picking up C++ linkage
and referencing a mangled symbol nothing defines. I read the ELF `.symtab` of
both the isolated production object and the `tubuild` object directly:

```
build/src/actors/dScMgCup_c.o   125 UNDEF, 59 mangled, 0 ghosts of locally-defined symbols
build/tu/ov006-dScMgCup_c/dScMgCup_c.o  111 UNDEF, 45 mangled, 0 ghosts
```

Every mangled undefined is a genuine external: `dScMgBase_c`,
`dScMgSingle3DBase_c`, `dScene_c`, `fBase_c`, `G2S`, `GXS`, `OAM`, `Sound`,
`Memory`, `Particle::SysTracker`, `abi::__class_type_info`. Not one names a
`func_ov006_*` in this TU's own range, and not one is a mangled ghost of a
symbol this file defines. The four `dScMgCup_c`-named undefineds in the
production object are the intended externalisation —
`_ZN10dScMgCup_cD2Ev`, `_ZTI10dScMgCup_c`, `_ZTS10dScMgCup_c`,
`_ZTV10dScMgCup_c` — i.e. **this TU does not own its vtable**, which is
required here.

The file's shape is one file-scope `extern "C" { ... }` region holding the
genuine externals, with `extern "C"` on each of the fourteen `func_*`
definitions this TU owns. There are no block-scope declarations left inside any
member body.

**2. `mwldarm` actually ran, and its output is the cartridge.** `rombuild`'s
`[4/6] mwldarm` links the working tree; that is the real link, not a scratch
substitution. Its output compares byte-for-byte against the extracted ROM:

```
linked overlays byte-identical to cartridge: 103/103
build/build/arm9_ov006.bin  sha256 9432d4ba3e7f4afe3b023a4a9eb407a17c529ff5a2b52aaec93cabf6e366eeec
extracted/overlays/overlay_0006.bin  sha256 9432d4ba3e7f4afe3b023a4a9eb407a17c529ff5a2b52aaec93cabf6e366eeec
```

**`tubuild.py linkcheck` is unavailable for this TU, and that is a pre-existing
tool defect rather than anything about this change.** It refuses at `[4/8]`:

```
FAIL src/actors/dScMgCup_c.cpp: isolate: intact TU preparation refused:
     ov006/dScMgCup_c: intact production requires one .text claim and at least one non-text claim
```

I reproduced the identical refusal on **`ov006/dScMgTeresa_c`, an
already-promoted TU this PR does not touch**:

```
FAIL src/actors/dScMgTeresa_c.cpp: isolate: intact TU preparation refused:
     ov006/dScMgTeresa_c: intact production requires one .text claim and at least one non-text claim
```

This TU is text-only by design (`data: []`, `bss: []`) so it can never satisfy
that predicate. No `production_mode` key and no fake non-text claim was added to
silence it — doing so would make this TU claim the vtable it must not own.
`rombuild.intact_tu_policies` is why the normal build never takes that path.

---

## The eight `State*` renames, verified

`config/arm9/overlays/ov006/symbols.txt` renames eight addresses. Read back out
of the **linked** `build/build/arm9_ov006.bin` (base `0x020bfec0`), the eight
pointer-to-member code words in `0x0213bfe8..0x0213c048`:

```
  0x0213bfe8: 020df1bc 00000000  _ZN10dScMgCup_c9StateIdleEv
  0x0213bff8: 020df540 00000000  _ZN10dScMgCup_c17StateWaitForInputEv
  0x0213c000: 020dfcd8 00000000  _ZN10dScMgCup_c19StatePrepareShuffleEv
  0x0213c008: 020df1c0 00000000  _ZN10dScMgCup_c11StateFinishEv
  0x0213c010: 020df3bc 00000000  _ZN10dScMgCup_c11StateSelectEv
  0x0213c028: 020df5b8 00000000  _ZN10dScMgCup_c12StateShuffleEv
  0x0213c030: 020dfd48 00000000  _ZN10dScMgCup_c10StateSetupEv
  0x0213c040: 020df28c 00000000  _ZN10dScMgCup_c11StateResultEv
```

8/8 non-zero and each equal to the retail overlay's own word.
`check_rename_ledger --repo .` is green: *"every mangled/vtable row agrees with
its module's symbols.txt"*, 1843 rows checked.

Two small corrections to the manifest's prose, for the record: the eight records
are **not contiguous** — four unrelated 8-byte rows are interleaved at
`0x0213bff0`, `0x0213c018`, `0x0213c020` and `0x0213c038` — and the span ends at
`0x0213c048`, not `0x0213c040` (`StateResult` sits *at* `0x0213c040`).

The names themselves remain class-anchored **inferences**: `__sinit_ov006_021303d0`
fixes the table's order and each body's behaviour fixes its role, but the
original spellings are not in the cartridge. The tree already carried a coined
member symbol for this class (`_ZN10dScMgCup_c9Virtual50Ev`) before this change.

---

## Class identity

| record | ROM address | note |
|---|---|---|
| `_ZTS10dScMgCup_c` | ov006 `0x0213c054` | bytes `10dScMgCup_c\0`, read from the image |
| `_ZTI10dScMgCup_c` | ov006 `0x0213c048` | `abi::__si_class_type_info`, 0x0c bytes; `+0x00` = `0x0209a764` (the `__si_class_type_info` vtable), `+0x08` = `0x0213bc64` = `_ZTI19dScMgSingle3DBase_c` |
| `_ZTV10dScMgCup_c` | ov006 `0x0213c154` | 36 slots; `-8` is `0x00000000` (offset-to-top), `-4` is `0x0213c048`, so the symbol *is* the address point |

**Base: `dScMgSingle3DBase_c`, single edge, offset 0** — read directly out of the
`__si_class_type_info` record, not inferred. All three records are local to
ov006 and adjacent. The name is real RTTI, so the coined-name wall does not
apply.

**Object size `0x5470` (21,616)** from the factory's own literal:
`dScMgCup_c_classInit` at `0x020e0574` loads the constant at `0x020e0608`
(`0x00005470`) and passes it straight to `_ZN7fBase_cnwEj`.

**Vtable diff against `dScMgSingle3DBase_c`** — 36 slots in both; seven differ,
and every one of the seven points inside this TU's text run:

| slot | symbol | address | size |
|---:|---|---|---:|
| 0 | `_ZN10dScMgCup_c13InitResourcesEv` | `0x020e0308` | `0x26c` |
| 6 | `_ZN10dScMgCup_c8BehaviorEv` | `0x020e0204` | `0x104` |
| 9 | `_ZN10dScMgCup_c6RenderEv` | `0x020e0068` | `0x19c` |
| 16 | `_ZN10dScMgCup_cD1Ev` (complete-object) | `0x020de988` | `0x094` |
| 17 | `_ZN10dScMgCup_cD0Ev` (deleting) | `0x020dea1c` | `0x0a8` |
| 18 | `_ZN10dScMgCup_c13OnYoshiTryEatEi` | `0x020dfeec` | `0x17c` |
| 20 | `_ZN10dScMgCup_c9Virtual50Ev` | `0x020dfed4` | `0x018` |

**Destructor placement.** One source-level destructor definition, written
**first** in the file, because the ROM puts D1 at `0x020de988` — the lowest
address in the run — and D0 immediately after it at `0x020dea1c`. That ordering
is held by `#pragma defer_codegen off`: generating at parse time emits `.text` in
source order where deferred codegen reverses it, and it also keeps the five
member-local optimiser brackets scoped. D2 is emitted as a compiler-only
passenger and exact-deadstripped (`compiler-only: exact deadstrip
['_ZN10dScMgCup_cD2Ev']`). No `operator delete` is written here; `dScMgBase_c`,
two levels up, already provides one.

**Promotion route: text-only + `compiler_only_output`.** `data: []`, `bss: []`.
The three self records (`_ZTI`/`_ZTS`/`_ZTV`) are *homeless* — their canonical
addresses sit inside ov006's still-unowned `.data` span — so they are
externalised to the canonical cartridge copies rather than claimed. The ten
ancestor RTTI rows are homed elsewhere already. Sibling oracle
`dScMgSingle3DBase_c`'s other promoted children in this overlay
(`dScMgTeresa_c`, `dScMgRoulette_c`, and now `dScMgCoin_c` #2344 and
`dScMgPanel_c` #2347) take the same route with the same one-file-scope
`extern "C"` shape.

---

## Review corrections (a claim withdrawn)

Review spot-checked the three "measured" `StateShuffle` spellings by removal and
found one of them false. I re-ran all three myself on this tree:

| spelling removed | result | verdict |
|---|---|---|
| cached base reused at the STORE site | `31/32`, `1 reloc destination(s) WRONG (data_ov006_0213c085 != 0x0213c084:ov006)`, objisolate FAIL | load-bearing |
| `*(u16*)(c + 0x545c) += 0x8000` instead of the plain RMW | `31/32 MATCH` | load-bearing |
| `*(s16*)(c + 0x545e) = -speed;` fused into the store | **`32/32 MATCH`** | **NOT load-bearing** |

The `speed = -speed;` statement split buys nothing. Both the source comment and
the manifest note claimed it "reverses the pair"; **both are corrected**, and the
split is kept for readability with an explicit label saying it is byte-neutral.

This matters more than one line of prose. `dScMgCup_c` sat shadowed for months
partly because its own manifest described the residual as "scratch-register
encoding differences" with seven lever families exhausted -- a hypothesis written
as a measurement, and wrong. Shipping a second one in the same file would be the
same mistake in the same place, so the manifest now carries a standing note: no
load-bearing claim goes in this file that was not produced by deleting the
spelling and re-running the gate.

---

## Disclosure: `func_ov006_020def80` reads an uninitialised local

```c
extern "C" void func_ov006_020def80(char *c, int i)
{
    ...
    int cup;
    ...
        goto epilogue;      /* twice, on two of the three arms */
    ...
    cup = i;                /* only the third arm assigns it */
epilogue:
    char *out = c + (cup << 2);
    ...
}
```

Two of the three arms reach `epilogue` with `cup` never assigned, so the
epilogue reads whatever register mwcc left `i` in. **This is undefined behaviour
in C++.** It is also exactly what the cartridge does, and it is the only
spelling that reproduces it — I measured the obvious repair myself rather than
taking the comment's word for it. Hoisting to `int cup = i;` and deleting the
late assignment:

```
DIFF     func_ov006_020def80   0x020def80  size 0x0a4
byte comparison   : 31/32 MATCH -> NOT verified
```

The working tree was restored and re-verified at 32/32 afterwards.

**The hazard is closed, and the port's repair is known.** Review pointed out this
is more provable than I first wrote it, and the ROM agrees -- disassembled from
`extracted/overlays/overlay_0006.bin`:

```
0x020df00c  add   r0, r0, r1, lsl #2      <- the epilogue's index
0x020df010  add   r0, r0, #0x5000
0x020df014  mov   r1, #0                  <- r1's FIRST write, after that use
0x020df018  str   r1, [r0, #0x440]
0x020df01c  str   r1, [r0, #0x44c]
```

`r1` is the incoming `i`, and it is never written on any path before the epilogue
reads it. So the cartridge's behaviour is `cup == i` **unconditionally** --
`int cup = i;` is not a guess at an undefined value, it is a provably
behaviour-preserving rewrite, and it is exactly the edit a host port should make.
Only the byte-matching build needs the split. The source comment now says all of
this.

Four other spellings in this file are likewise measured and commented — do not
tidy them: `StateShuffle`'s three read-modify-write spellings on `+0x545c` /
`+0x545e`, `Render`'s two pointer-arithmetic towers, `Behavior`'s three `(int)`
launders, and `func_ov006_020ded00`'s two spellings of `+0x12`.

---

## What this stage changed on top of stages 2 and 3

- **Merged `origin/main` twice.** Main moved under this branch mid-run: #2344
  (`dScMgCoin_c`) and then #2347 (`dScMgPanel_c`) both landed, both in **ov006**,
  both editing `config/arm9/overlays/ov006/delinks.txt`. Merged rather than
  rebased (the branch is on the remote), and the base was verified after the
  fact: `origin/main` `bedcaf5f0` is an ancestor of `9260a8bba`. The two
  conflicted files — `notes/cpp-tu-current-state.md` and
  `notes/data/tu-promotion-queue.tsv` — are generated, so they were regenerated
  (`cpp_tu_state.py --write-note`, `queue_audit.py --write`) rather than hand
  resolved, and both `--check` modes are green. I audited the merged ov006
  `delinks.txt` independently: **1107 `.text` entries across 1138 file blocks**
  (plus 31 `.init` and 2 `.data`), **0 overlaps**, and this TU's
  `0x020de988..0x020e0638` sits flush between `src/d_s_mg_coin.cpp` (ends
  `0x020de988`) and `src/_ZN14dScMgCurling_cD1Ev.cpp` (starts `0x020e0638`).
  An earlier revision of this body said 1176 entries. That figure was correct
  when I measured it and stale by the time I wrote it down -- #2347 folded 71
  shards into one file in between -- which is my error, not a tool's: a count
  carried across a merge is a fresh measurement, not a copy.
- **`attribution.json`: 41 override rows.** The fold is a rewrite *and* a move in
  one step, so git records no rename and every function's credit re-pointed at
  the pusher — `prepush_attribution` read `0 changed, 28 lost`. Added the same
  `<path>#<symbol>` overrides #2344 and #2347 use, resolved from `origin/main`'s
  own composite. Nine members need two key spellings because this promotion also
  renames them: `validate_merge.attribution_snapshot` keys `path#<current
  symbol>`, `prepush_attribution` keys `path#<legacy basename>`. Now `0 changed,
  0 lost, 28 consolidated with credit intact`.
- **Regenerated both derived notes again after merging `9d45c0082`**, and note
  that `queue_audit --check` is **already red on `origin/main`**: #2346 re-landed
  `src/func_ov006_020d1ba0` as `.cpp` without re-running it. The single data row
  that moves is `dScMgAmida_c` 24/2006 -> 25/2001 -- that PR's drift, carried here
  only because the file is generated and leaving it stale would ship a red gate.
- **`include/dScMgCup_c.h`: comment-only.** The header still cited four source
  files this PR deletes, and its closing paragraph (*"Until production promotion,
  the enrolled D1/D0 bodies remain in their legacy one-function files"*) is now
  false. Repointed the six path references, restated the destructor paragraph,
  and added an explicit note that the struct's non-`unk_`/`pad_` field names are
  **inferences** (see below). `check_header_offsets` on the changed header: *25
  commented fields, 0 mismatched, 0 unparsed, struct spans 0x5470*, and
  `tubuild verify` is unchanged at 32/32.

## Flagged, not fixed

- **`include/dScMgCup_c.h` field names are invented.**
  `mShuffleAngle`, `mShuffleSpeed`, `mOnes`, `mIds`, `mFlags`, `mShuffleSound`
  and `mState` are inferences from an earlier stage;
  `notes/data/class-facts/dScMgCup_c.json` proves eleven offsets in
  `0x5400..0x5470` and names **none** of them. This PR does not propagate them —
  `src/actors/dScMgCup_c.cpp` addresses that tail as `*(T *)(c + 0x54xx)` on
  purpose, and only `mState` is used from the header at all. The header now says
  so in its own banner.
- **`notes/data/c-cpp-classification.tsv` goes stale by 22 rows**, not one:
  lines 4442-4462 and 4465 all name shards this promotion deletes. That file is
  an **append log** whose row order is load-bearing, so it is flagged and not
  hand-edited. For scale, the same file already carries 1,579 dead `src/` rows
  out of 7,067 on `origin/main` — this is a systemic backlog, not a regression
  introduced here.
- **`// @symbol` markers cover 15 of 32 members, and here that costs nothing.**
  `tiers.py:_marked_member_fragment` slices a member from its marker to the
  *next* marker, so the seventeen unmarked bodies are scored as part of the
  marked member above them. I measured the repair rather than assume it: adding
  all seventeen markers moves CONVERTED **2 -> 2**. Every remaining member is
  blocked by `real_name` (the fourteen `func_ov006_*` helpers the ROM gives no
  other name) or by `no_raw_offset` (this TU addresses the object tail as
  `*(T *)(c + 0x54xx)` on purpose, because no field in it has a proven name), and
  one member — `Virtual50` — gets *worse*. Both banked records
  (`#_ZN10dScMgCup_c8BehaviorEv`, `#_ZN10dScMgCup_c9StateIdleEv`) are marked, so
  there is no backslide exposure either; `tiers_ratchet --check` is PASS. Left
  alone deliberately, and recorded here so the next reader does not re-derive it.

- `docs/class-reference.html` (generated) and `notes/data/tu-merge-candidates.json`
  (generated) also name deleted shards and will settle on their next regeneration.
  `src/actors/dScGameOver_c.cpp` carries one comment reference to
  `src/_ZN10dScMgCup_c13InitResourcesEv.cpp`; that is another file's comment, so
  it is left alone. `check_dead_references` walks `.md` only, so none of these
  are gated — it exits 0 with *"no new dead references"*.


andrewboudreau:
## Reviewer sign-off (stage 5) — independent re-derivation

Reviewed in my own worktree (`C:\tmp\sm64ds-rev2348`, branch `review/pr2348-dScMgCup_c`
at `9260a8bba`), and again on the tree **composed with current `origin/main` `9d45c0082`**
— main moved again after this branch's last merge (#2345, #2346, #2336, plus a progress
refresh). The merge is conflict-free.

**Verdict: MERGE.**

### Re-derived, not taken on trust

- `tubuild verify ov006/dScMgCup_c` — **32/32 MATCH**, objisolate clean, reloc-destinations
  clean, all 32 in ROM-ascending emission order. At the PR head *and* on the merge tree.
- `rombuild -j 16 --no-rom` at the head: 11,174 reproducing / 0 mismatching, 20 source-owned
  data claims / 0 mismatching, **module fidelity 106/106 exact, 100.000000%**, ROM-build
  analysis PASS. On the merge tree: 11,175 / 0, same fidelity, PASS. The baseline control's
  9 `dsd check symbols` errors are the documented pre-existing set.
- **The link proof, hashed myself.** All 103 built overlays vs `extracted/`: **103/103
  byte-identical**, on both trees. `build/build/arm9_ov006.bin` =
  `9432d4ba3e7f4afe3b023a4a9eb407a17c529ff5a2b52aaec93cabf6e366eeec` =
  `extracted/overlays/overlay_0006.bin`. Confirms the body's figure exactly.
- **The eight `State*` renames.** Read the PMF code words straight out of the linked
  `arm9_ov006.bin` at `0x0213bfe8..0x0213c048`: all eight non-zero, each equal to the retail
  overlay's own word and to its `symbols.txt` address. Your two corrections are right — the
  records are not contiguous (four unrelated rows interleaved) and the span ends at
  `0x0213c048`. Every rename keeps address and size; `_ZN10dScMgCup_c9Virtual50Ev` was
  already coined in that file, so the convention is consistent. `check_rename_ledger` green,
  1843 rows.
- **Vtable extent, by eye.** `_ZTV10dScMgCup_c` at `0x0213c154`: 144 bytes emitted = 144 ROM
  extent, 0 blind words, and the next `symbols.txt` record is `data_ov006_0213c1e4` at
  exactly `+0x90` — the table ends where the next record begins. No overrun, no truncation.
  Preamble `-8` = 0, `-4` = `0x0213c048` = `_ZTI`. Dumped all 36 slots: exactly seven point
  inside this TU's run (0, 6, 9, 16, 17, 18, 20), **D1 at slot 16 then D0 at slot 17, no D2**;
  D2 is the exact deadstrip.
- **`romdata_check` per symbol** (via the module API — the CLI hides the verdicts). All six
  PARTIALs are `_ZTS` name strings whose NUL-terminated content is shorter than its 4-aligned
  extent — 12/16, 12/16, 20/24, 8/12, 8/12, 8/12, zero differing bytes. **No vtable is
  PARTIAL, no record DIFFERS.** Benign, exactly the `dScMgPanel_c` shape.
- **Object symbol tables.** Production object: the four `dScMgCup_c`-named undefineds are
  `D2`/`_ZTI`/`_ZTS`/`_ZTV`, 0 ghosts, no own-range `func_ov006_*` undefined. Non-`.text`
  defined symbols are only RTTI + vtable — no guards, statics, thunks, or bss.
- **Ledgers.** `symbols/actor_renames.tsv` and `notes/data/c-cpp-classification.tsv` are
  untouched. `config/converted-baseline.json` 2680 → 2682: two rows appended in sorted
  position, list is 2682 unique and sorted, no double-banking, no row naming a deleted file.
- **Attribution shape.** 41 rows appended at the end of the block, one line touched (the
  comma), no re-serialisation, no new `github-actions[bot]` credit. `prepush_attribution`
  reproduces `28 consolidated, 0 changed, 0 lost`.
- **Merged `delinks.txt`, audited myself:** 0 overlaps, no duplicate file blocks, and
  `0x020de988..0x020e0638` flush between `src/d_s_mg_coin.cpp` (ends `0x020de988`) and
  `src/_ZN14dScMgCurling_cD1Ev.cpp` (starts `0x020e0638`).
- **Gate sweep on the merge tree**, all exit 0: `check_rename_ledger`, `tiers_ratchet --check`
  (2682/2682), `check_profile_campaign`, `check_dead_references`, `port_refcheck`,
  `check_header_offsets --changed`, `check_duplicate_sources`, `check_src_tu_compiles`
  (149/149), `cpp_tu_compat --require-ready`, `check_references` (42 unresolved vs a baseline
  of 45), `eligible` (11181/11257), langmode ratchet.
- **The header's invented names do not propagate.** `mShuffleAngle`, `mShuffleSpeed`, `mOnes`,
  `mIds`, `mFlags` appear in `src/actors/dScMgCup_c.cpp` only inside the banner that labels
  them inferences; the one member used is `mState`.

### Spot-checks of the load-bearing spellings — one of the three does not hold

I removed each and re-verified, then restored and re-verified 32/32 with a clean `git status`
(the failing runs did rewrite the manifest, as expected; reverted).

| spelling | result |
|---|---|
| cached base dropped at the **store** site (`c + 0x545c/0x545e`, loads keep `state`) | **load-bearing** — restoring `state + 0x5c/0x5e` gives DIFF + 1 wrong reloc destination |
| plain read-modify-write rather than `+= 0x8000` | **load-bearing** — `+=` gives DIFF |
| `speed = -speed;` as its own statement | **not load-bearing** — `*(s16*)(c + 0x545e) = -speed;` still gives **32/32 MATCH** |

The third claim is overstated in the source comment and in the manifest note ("storing `-vv`
directly reverses the pair"). It was presumably measured in an intermediate state before the
other two edits landed. Worth correcting in a follow-up so the next reader does not treat a
free spelling as frozen — not a merge blocker, since the tree as committed matches.

### The undefined behaviour — I agree with keeping it, and it is more provable than stated

Disclosed prominently in the body and commented at the site. Keeping it is the right call:
faithfulness to the cartridge is the goal, and the repair was measured to break the match.

One addition worth putting in the comment. I disassembled `0x020def80` out of the ROM: the
epilogue at `0x020df00c` is `add r0, r0, r1, lsl #2`, and **`r1` (`i`) is never written on any
path**. So the cartridge's behaviour is `cup == i` unconditionally — the UB is a
source-spelling artifact, not real indeterminacy. That means the `port/` repair is exactly
`int cup = i;` and is *provably* behaviour-preserving, rather than "a different compiler is
free to put anything in that slot". Saying so turns an open hazard into a closed one.

### Smaller notes

- The body's "1176 `.text` entries" is wrong; the merged ov006 `delinks.txt` has **1106**
  (1137 file blocks). The claims that matter — 0 overlaps, flush neighbours — reproduce.
- At the PR head alone, `launder_or_forced`/`launder_sites` read 192/417 against
  `chaos-data`'s 190/413; the ratchet is green only because the root `langmode-baseline.json`
  override (209/438, banked by #2270) is present. On the merge tree the counts fall back to
  190/413 and pass even the stricter banked baseline. Pre-existing debt, not this PR's.
- On the composed merge tree two local-only generated-file gates go red — `queue_audit --check`
  (row `dScMgAmida_c`, shard_count 24→25, total_lines 2006→2001) and `cpp_tu_state
  --check-note` (MATCHED 11294→11295, `.c` 5425→5424). **Every delta is main's own #2346**, so
  they are already red on `origin/main`; neither is in CI. Somebody should regenerate both on
  main — nothing for this PR to do.

PR #2360, opened 2026-09-06
Split func_020610fc so the wireless manager registration routine has its own symbol

tangosdev:
## What the cartridge says

`func_020610fc` was declared 0x3c bytes, covering 0x020610fc to 0x02061138.
Disassembled from `extracted/arm9_dec.bin` (base 0x02004000):

```
020610FC  70402de9  push    {r4, r5, r6, lr}
02061100  0d60a0e3  mov     r6, #0xd
02061104  0250a0e3  mov     r5, #2
02061108  0040a0e3  mov     r4, #0
0206110C  0600a0e1  mov     r0, r6
02061110  0510a0e1  mov     r1, r5
02061114  0420a0e1  mov     r2, r4
02061118  1aeaffeb  bl      #0x205b988
0206111C  000050e3  cmp     r0, #0
02061120  f9ffff1a  bne     #0x206110c
02061124  feffffea  b       #0x2061124
02061128  04109fe5  ldr     r1, [pc, #4]
0206112C  000081e5  str     r0, [r1]
02061130  1eff2fe1  bx      lr
02061134  a4890a02  .word   0x020a89a4
02061138  10402de9  push    {r4, lr}          <- func_02061138 starts here
```

That is two routines, not one.

**0x02061124 terminates the first one.** `feffffea` decodes as an unconditional
branch with offset -2, so the target is 0x02061124 itself. It is a deliberate
hang. Nothing falls through it into 0x02061128, and there is no other path there
from inside func_020610fc: the only other branch in the routine, the `bne` at
0x02061120, goes backwards to 0x0206110c. The stack agrees as well. The first
routine pushes {r4, r5, r6, lr} and never pops; the second ends on a bare `bx lr`
with no matching pop, so it cannot be a continuation of the first without
returning with the stack 16 bytes out.

**0x02061128 is address-taken.** A scan of every module binary for branches and
literal words pointing into the range turned up the word 0x02061128 stored at
0x02019ec8, inside `func_02019ebc`:

```
02019EBC  04c09fe5  ldr     ip, [pc, #4]      ; -> 0x02019ec8 = 0x02061128
02019EC0  04009fe5  ldr     r0, [pc, #4]      ; -> 0x02019ecc = 0x02019f10
02019EC4  1cff2fe1  bx      ip
02019EC8  28110602  .word   0x02061128
02019ECC  109f0102  .word   0x02019f10
```

So `func_02019ebc` loads r0 with `func_02019f10` and tail-jumps to 0x02061128,
which stores r0 into the global at 0x020a89a4. `IRQ::GameCardIREQMCHandler` at
0x020610ac reads that same global back as a function pointer and calls it, so
0x02061128 is the registration half of a live callback pair.

The tree already knew the name. `src/func_02019ebc.c` is written as
`func_02061128(func_02019f10);`, and `config/unresolved-baseline.json` carries
`arm9:0x02019ebc` with `missing: ["func_02061128"]` precisely because no symbol
defined it. After this change that entry resolves and `func_02019ebc` becomes
eligible.

The routine's own literal pool word at 0x02061134 is what `ldr r1, [pc, #4]`
reads (pc is 0x02061130 there, plus 4), so the routine is 0x10 bytes including
the pool, ending exactly where `func_02061138` begins.

## Lines changed

`config/arm9/symbols.txt`:

```
-func_020610fc kind:function(arm,size=0x3c) addr:0x020610fc
+func_020610fc kind:function(arm,size=0x2c) addr:0x020610fc
+func_02061128 kind:function(arm,size=0x10) addr:0x02061128
```

`src/func_020610fc.c` is a hand-written asm draft that covered the whole old 0x3c
span. Its three trailing instructions and the `data_020a89a4` extern now belong
to the new symbol, so they were removed; the file compiles to exactly
0x020610fc..0x02061128 under the pinned compiler. Its banner note said the span
"also contains an unreachable epilogue-less tail" as the reason it was kept as
asm, which is no longer true, so that sentence was replaced with the real reason
(the routine never returns).

## Everything else that touches the range

- `config/arm9/delinks.txt` has no entry covering 0x020610fc..0x02061138 at all.
  The neighbours run 0x020610ac..0x020610fc and 0x02061138..0x0206116c, so the
  split point sits in a gap and no delink range had to move.
- `config/arm9/relocs.txt` has two entries touching the old range:
  `from:0x020608f0 kind:arm_call to:0x02061138` and
  `from:0x020610e4 kind:arm_call to:0x020610fc`. Both point at the two ends,
  neither points inside, so neither changes. The config also already carries
  `from:0x02061134 kind:load to:0x020a89a4 module:main`, which is the new
  routine's own literal pool word, so the relocation data already treats
  0x02061134 as belonging to code rather than to the hang above it.
- No reloc records the pointer word at 0x02019ec8. Its neighbour does
  (`from:0x02019ecc kind:load to:0x02019f10`), and the reason 0x02019ec8 does not
  is that its destination had no symbol to name. Adding that entry is left out of
  this pull request on purpose so the symbol change stands alone; it can follow
  afterwards.
- Source coverage: `src/func_020610fc.c` (trimmed here) and `src/func_02019ebc.c`
  (already referencing the new name, unchanged). Nothing in `src_tu/` covers the
  range.

## Why the port needs it

The port dispatches through the config symbol table, and it cannot dispatch a
routine that has no symbol. Today the registration routine is invisible: its
bytes live inside a function whose declared behaviour is an infinite hang, so a
host that honours the symbol table either never reaches it or reaches it by
guessing at an interior offset. With its own symbol the port can call it the way
the cartridge does, through the pointer word at 0x02019ec8.

## Gates

```
$ python tools/eligible.py -j 8
classifying 11263 enrolled functions with -j8 ...
  11188  ELIGIBLE
     41  unresolvable
     23  isolate
      3  extra sections
      1  .text section 0xb6c != declared 0xb0
      1  .text section 0xb6c != declared 0x39c
      1  .text section 0xb6c != declared 0xec
      1  .text section 0xb6c != declared 0x518
      1  .text section 0x4d0 != declared 0x4c8
      1  .text section 0x390 != declared 0x38c
      1  .text section 0x300 != declared 0x2fc
      1  .text section 0x1364 != declared 0x1360
eligible: 11188 / 11263

$ python tools/check_references.py
unresolved symbols: 41  (baseline 45)
eligible:           11188  (baseline 11044)
  4 fixed since the baseline -- run --update to bank it
OK: no new unresolvable references

$ python tools/check_src_tu.py
check_src_tu: 30 translation unit(s), 316 include(s), 783 mangled reference(s), against 31764 ROM symbols
check_src_tu: every reference resolves.

$ python tools/prepush_linkcheck.py --range origin/main..HEAD
  [OK  ] func_02057198                                VERIFIED
prepush-linkcheck: skipped 1 non-matching draft(s)
prepush-linkcheck: 1 checked - 1 verified, 0 warning(s), 0 blocking

$ python tools/langmode_audit.py --check langmode-baseline.json
langmode ratchet PASS

$ python tools/check_duplicate_sources.py
duplicate-sources: 9476 stems, none doubled

$ python -m unittest tools.test_srcpath tools.test_bytegate tools.test_enroll
Ran 80 tests in 26.361s
OK
```

The baseline was left alone rather than banked, because three of those four fixes
came from other work already on main and are not this change's to record.

Byte gate spot check on the neighbours, so that a symbols.txt edit cannot break
them silently (`build_pin.verify`, module arm9):

```
func_020610fc                      0x020610fc size 0x2c  -> (True, '2004/b56')
func_02061138                      0x02061138 size 0x34  -> (True, '2004/b56')
func_0206116c                      0x0206116c size 0x1c  -> (True, '2004/b56')
_ZN3IRQ21GameCardIREQMCHandlerEv   0x020610ac size 0x50  -> (True, '2004/b56')
func_02019ebc                      0x02019ebc size 0x14  -> (True, '2004/b56')
func_02019f10                      0x02019f10 size 0x54  -> (True, '2004/b56')
func_02060890                      0x02060890 size 0x7c  -> (True, '2004/b56')
```

This adds one function symbol, so the validator's function-count denominator
moves by one. That is the whole reason this is on its own branch with nothing
else in it.


andrewboudreau:
Ruling from the metric side, since this has been sitting on us: **this PR is correct on its merits and should land.** The only thing stopping it is our gate, not anything about the change.

Verified against `main` rather than taken on trust:

- `src/func_02019ebc.c:5` calls `func_02061128`.
- `func_02061128` is absent from `config/arm9/symbols.txt`.
- `config/unresolved-baseline.json` records it as missing under `arm9:0x02019ebc`.

So `main` currently ships a source file calling a symbol its own symbol table does not have, and the cause is one symbol claiming a span the cartridge splits between two routines. The deleted tail in `src/func_020610fc.c` was unreachable code sitting after a `b self`.

The block is `tools/validate_merge.py:587`, which fails on any change to `totalFunctions` or `totalBytes` in either direction. Here `0x3c == 0x2c + 0x10`, so `totalBytes` does not move; only the partition over the same bytes does. `matchedFunctions` does not move either — `func_020610fc` carries a NONMATCHING banner and the new symbol is unimplemented. Denominator +1, numerator +0: 99.271% to 99.262%.

That is the correct price for not shipping a false symbol table, and it is the shape the file's own WITHDRAWN carve-out was written for — "treating both as loss made the gate reward the lie."

Worth recording: the merge precedent this resembles, `5030b9b89` ("symbols: merge func_020729e8 into func_02072168 (severed epilogue, one function)", #964), landed 2026-08-01 — and the denominator check landed the same day, *after* it, in `e3a5dd048` (#979). The route was closed by the gate rather than by a decision.

Fix is on our side. A `SPLIT` carve-out for `validate_merge.py` is being written and independently reviewed now; this PR needs no changes and should not be closed.

PR #2375, opened 2026-09-06
ov006: promote dMgJump3DMario_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

SM64DS's jump minigames give each of three players an object of one class. Its
code was sitting in the repo as **28 separate one-function C files** with names
like `func_ov006_020c7c68` — one file per machine-code function, which is how
this project first recovers code it has decompiled but not yet organised.

This PR puts them back into **one C++ file**, `src/actors/dMgJump3DMario_c.cpp`,
which is what the original 2004 source almost certainly was: the linker laid all
28 down in one unbroken run, `0x020c762c..0x020c8a30`, with no gaps.

Twenty of the 28 are now real `dMgJump3DMario_c::` methods instead of loose
functions. The other **eight cannot move yet** — they are still called by name
from seven neighbouring C files that have not been promoted, and a C file cannot
call a C++ method. They convert the day those seven do.

The whole thing is checked by rebuilding the game ROM from source and comparing
it byte-for-byte against the cartridge: **106 of 106 modules identical,
100.000000% of compared bytes**. Nothing about the game changed. Only the source
got more like what Nintendo wrote.

---

## Proof block

```
tubuild verify        28/28 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
object audit [4b/8]   emittedTextOrderIsRomAscending True, {'LICENSED': 28}, 0 refusals, 0 unlicensed sections
rombuild -j6          106/106 exact, 100.000000% of compared bytes, reproducing 11,190 / mismatching 0
rombuild -j6 --no-cache  same figures with all 9,257 objects recompiled (shared objcache bypassed)
  BASELINE CONTROL    modules PASS, ROM sha256 d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
  profile             stock, 0 mod source replacement(s), 0 ROM-gap fallback(s)
romdata_check         4 VERIFIED, 2 PARTIAL, 0 DIFFERS, 0 UNNAMED
tu_order_check        all match in ROM order; 4 EXTRA covered by the manifest's compiler-only policy
source_coverage       clean, 0 B handed back to the cartridge
prepush_attribution   19 consolidated with credit intact, 0 changed, 0 lost
tiers_ratchet --check PASS   baseline 2686   current 2686
```

Every gate was re-run in this worktree at this branch's ref; none of the figures
above is quoted from an earlier stage.

---

## Scope of the run

One contiguous linker run, `0x020c762c..0x020c8a30`, 28 functions, `0x1404`
bytes. The 28 `symbols.txt` sizes sum to exactly `0x1404`, every function starts
where the previous one ends, and every one had both a `src/` file and a
`complete` `delinks.txt` entry — so there is no sourceless hole and no choice of
segment. Promotion route is **text-only + `compiler_only_output`**: the manifest
claims `.text` and nothing else, `data: []`, `bss: []`, `externalized_output`
empty, and there is deliberately **no `production_mode` key**. Sibling oracle
`dScMgSingle3DBase_c`, used as a shape reference only.

`compiler_only_output` carries **10 rows** where the queue estimated a floor of
5 — six `deadstrip-data` (the class's and its abstract base's `_ZTV`/`_ZTI`/
`_ZTS`) and four homeless `deadstrip` (`C2`, `D0`, `D2`, and the base's `D2`).
All six canonical addresses were checked against `ov006/symbols.txt` and each of
the six symbols is defined exactly once tree-wide.

---

## Class identity, read out of the cartridge

Re-read at this ref straight from `extracted/dsd/arm9_overlays/ov006.bin`
(module base `0x020bfec0`), not taken from the facts file:

| record | ov006 address | content |
|---|---|---|
| `_ZTS16dMgJump3DMario_c` | `0x0213b0b0` | `16dMgJump3DMario_c\0`, 19 bytes, file offset `0x7b1f0` |
| `_ZTI16dMgJump3DMario_c` | `0x0213b0a4` | `{0x0209a764, 0x0213b0b0, 0x0213b000}` — `abi::__si_class_type_info`, base `_ZTI22dMg3DHeyhoObjAdapter_c` |
| `_ZTV16dMgJump3DMario_c` | `0x0213b0cc` | slots `0x020c76d8`, `0x020c76d0`, `0x020c762c`; preamble at `0x0213b0c4` is `{0, 0x0213b0a4}` |
| `_ZTS22dMg3DHeyhoObjAdapter_c` | `0x0213afe4` | `22dMg3DHeyhoObjAdapter_c\0`, 25 bytes |
| `_ZTI22dMg3DHeyhoObjAdapter_c` | `0x0213b000` | `{0x0209a774, 0x0213afe4}` — root `abi::__class_type_info`, **no base word**, so the chain is exactly two deep |
| `_ZTV22dMg3DHeyhoObjAdapter_c` | `0x0213afd8` | `{0, 0, 0}` — three pure-virtual words, so the base is ABSTRACT |

**The class name is the cartridge's own RTTI spelling, not coined.** So is the
base's, and so is the inheritance edge.

**Vtable diff against the base: all three slots are overrides.** The base's table
is three zero words (pure virtual), the derived table fills all three, and there
is nothing else to enumerate — the base declares exactly three virtuals and the
derived class adds none.

Corroborated from a second direction, since one tool's slot count is not
evidence: `rtti_vtables.py --own dMgJump3DMario_c` prints
`3 slots (base 0)` with slots 0/1/2 at `0x020c76d8`, `0x020c76d0`, `0x020c762c`
— agreeing word-for-word with the direct ROM read above. Its `base 0` is the
known trailing-null trim, not a contradiction: the cartridge really does hold
three zero words at `0x0213afd8`, which is what "abstract" looks like, and the
manifest's `_ZTV22dMg3DHeyhoObjAdapter_c` row predicts this exact tool behaviour
in advance. The RTTI spelling and the coined name are the same string here, so
the `--own <coined name>` trap that prints "no vtable for X" does not apply.

**No factory and no `new`-size literal.** Nothing in the image heap-allocates one
of these; the two jump minigames hold them as `dMgJump3DMario_c mPlayers[3]`.
That is stated rather than invented, and confirmed positively: `opnew_sizes.py`
was run and `build/opnew_sizes.json` carries **no row for either class**.

The size is nailed down from the other direction instead, and more strongly than
a single factory literal would be: three independent caller shards
(`func_ov006_020c70d0`, `_020c712c`, `_020c719c`) index the array with a
**`* 0xb8`** stride, and `include/dMgJump3DMario_c.h` carries
`sizeof(dMgJump3DMario_c) == 0xb8` and `sizeof(dMg3DHeyhoObjAdapter_c) == 0x4c`
as static assertions. `check_header_offsets --changed origin/main` reports
`struct spans 0xb8`, 0 mismatched.

**Destructor placement, and why.** The destructor is declared **first** in the
class and is **not virtual** — none of the three vtable slots is `0x020c893c`
and there is no deleting stub in the ROM. The key function is therefore the
first non-inline virtual *declared* in the header, `Unk_020c76d8` at
`0x020c76d8`, which is inside the licensed run — which is why this TU emits the
class's `_ZTV`/`_ZTI`/`_ZTS` at all, and the abstract base's too (`D1` restores
that base vptr at `0x020c89c0`, odr-using it).

**`#pragma defer_codegen off` is load-bearing, measured here, not asserted.** The
`[4b/8]` object audit is unreachable through `tubuild linkcheck` on an enrolled
text-only TU, so it was driven directly (`_compile_tu` -> `apply_compiler_only_policy`
-> `audit_tu_object` -> `object_audit_refusals`). With the pragma:
`emittedTextOrderIsRomAscending True, {'LICENSED': 28}, 0 refusals`. With **only
that line deleted** as a negative control: `emittedTextOrderIsRomAscending
False`, 1 refusal, `licensed .text functions are not emitted in ROM address
order`.

---

## 20 of 28 members are methods; the other 8, and how that ceiling was derived

Five members already carried mangled names (three virtuals, `D1`, `C1`); the
conversion pass added fifteen. It cost nothing: 28/28 MATCH before and after,
objisolate clean, emission order still ROM-ascending.

The remaining eight keep `func_ov006_*` names because each is called by a direct
`bl` from a shard **below** this run that is still a plain `.c` file, and a C
shard cannot name a C++ member. Renaming the definition without editing the
caller is an undefined symbol at link.

| member | sites | callers |
|---|---|---|
| `func_ov006_020c76e0` | 4 | `func_ov006_020c7388` / `_020c7418` / `_020c7490` / `_020c7574` |
| `func_ov006_020c7734` | 1 | `func_ov006_020c70d0` |
| `func_ov006_020c7860` | 1 | `func_ov006_020c712c` |
| `func_ov006_020c8084` | 1 | `func_ov006_020c7388` |
| `func_ov006_020c81e0` | 1 | `func_ov006_020c7418` |
| `func_ov006_020c862c` | 1 | `func_ov006_020c7490` |
| `func_ov006_020c8658` | 2 | `func_ov006_020c719c` / `_020c7574` |
| `func_ov006_020c87d0` | 1 | `func_ov006_020c7574` |

Twelve call sites, seven shards, eight targets.

**Disclosed honestly: this is not the linker's `Undefined` list, and the brief
asked for one.** The producers derived it from a source grep plus two binary
scans (`relocs.txt` filtered on `module:overlay(6)`, and a raw ARM `BL` decode of
the overlay image). This stage reproduced it a third way and independently — a
tree-wide grep of `src/` and `include/` for each of the eight names, plus a
config-wide scan confirming none of the eight is spelled by any `.data` pointer
record or any other module — and got **exactly the same 8 targets, 7 shards and
12 call sites**. All seven caller shards are confirmed still enrolled as their
own `delinks.txt` entries.

**The verdict: acceptable, and here is the reasoning rather than a shrug.** A
linker `Undefined` list is a *post-hoc* artifact — you only get one by performing
the rename and letting `mwldarm` fail — so it could never have been the input to
the decision. What bounds the same set from the linker's side is already in this
PR: `rombuild` links all seven caller shards against the promoted TU and reports
`106/106 exact`, which is the linker agreeing that the boundary as shipped (20
converted, 8 not) is exactly right. The grep bounds it from above, the green link
bounds it from below, and they meet. What is *not* proven either way is that
converting a ninth would fail — that claim rests on the reference scan alone, and
nobody ran the experiment. **The eight were not converted here.**

---

## The rename is one edit with the source, and the names are coined

Fifteen 8-byte pointer-to-member records sit in ov006's **unowned** `.data` at
`0x0213b020..0x0213b098`, each `{code pointer, 0}`. `dsd` delinks them straight
from the cartridge and resolves each code word **by the name `symbols.txt` gives
at the target address** — so a converted member whose `symbols.txt` row still
said `func_*` would have its `.data` word link as `0x00000000`, with a perfectly
green compile. Fifteen `symbols.txt` rows therefore move in the same commit as
the source.

Audited at this ref: the `symbols.txt` diff is **15 renames and nothing else** —
every address and every `size=` is byte-identical, only the name changes.
`symbols/actor_renames.tsv` goes **3393 -> 3408 lines**, and the blob prefix is
**byte-identical**, so the append log was appended to and not rewritten. The two
sets agree address-for-address and old-name-for-new-name, with zero mismatches.
`check_rename_ledger` passes (1875 mangled/vtable rows agree with their module's
`symbols.txt`); `check_profile_campaign` passes (786 claims, 0 diverge).

**The fifteen member names are COINED.** What the cartridge proves for each: the
address, the size, that the first argument is the object, and — for the nine
reached through a pointer-to-member — which record installs it and which member
writes that record. What it proves for none of them: the spelling. Every `why`
sentence in the ledger separates the two, and
`include/dMgJump3DMario_c.h` carries the full derivation. The three virtuals stay
address-derived (`Unk_020c762c` and kin) because the ROM carries no spelling for
them, and no field below `+0x4c` was named for the same reason. The **class**
name, by contrast, is real ROM RTTI — see the table above.

---

### The CONVERTED ratchet was red on the branch as pushed

`tiers_ratchet --check` failed at the pushed ref — **before any edit of mine**,
confirmed by re-running it against the committed source. Stage 3 did not re-run
it after humanizing.

The mechanism: stage 2 banked `EnterMove` and `EnterBounce` as CONVERTED while
their bodies still read `*(struct S *)(p + 0x3c)` and `*(struct S60 *)(p + 0x3c)`.
`tiers.RAW_OFFSET` matches only a **single-word** type name before the star, so an
elaborated-type-specifier evades it — the evasion `tiers.py:102-104` documents and
deliberately leaves in place. Stage 3 then folded eight one-off shadow structs
(`G`, `S`, `V2`, `W2`, `S60`, `S50`, `S48`, `G38`) plus a lone `double` spelling
of the same eight bytes into one named `PmfRecord`; both bodies became
`*(PmfRecord *)(c + 0x3c)`, and the regex fires. **The code got more readable and
the metric got worse.**

Banked once with that reason. Audited after: `converted-baseline.json` 2688 ->
2686 with exactly those two member rows removed and nothing else touched;
`converted-backslide-exceptions.jsonl` at 387 rows with **0 byte-identical
duplicate records**; the earlier writers' `reason` strings intact.

Against `origin/main` the promotion is still a **net gain**: 2685 -> 2686, with
two whole-file shard identities out (both plain shard paths) and three TU member
identities in (`StateIdle`, `C1`, `D1`, all `src/actors/dMgJump3DMario_c.cpp#…`
rows), and every other row on `main` untouched. **Note for the reviewer:** stage
2's own exception row says "2 paths out, 5 member rows in". The real figure is
**3 member rows in**. That row is append-only history and was left as written;
the two rows appended here record the correction.

### Three lever comments claimed measurement they did not have

Stage 3 kept five spellings that cost bytes when removed, and flagged that its
own explanations were inference. Checked here: two of the five (`flag` reuse in
`StateHold`, the `int t` in `func_ov006_020c87d0`) state a measured word count
and nothing else. Three asserted a *mechanism* as fact beside the count —
register spill, stack demotion, operand evaluation order. The counts are real
(11, 5, 4 words, and two whole-function rewrites with 4 and 13 wrong relocation
destinations); the mechanisms are readings of the diff. Reworded so each says
which half is which, plus one paragraph in the file header saying it for all
five. Comment-only — `verify` reports 28/28 MATCH after.

### Attribution: 19 lost -> 0 lost

`prepush_attribution` measured `0 changed, 19 lost` at the pushed ref. Those 19
are the many-to-one fold artifact, and leaving them would have stripped credit
from **four named humans** — `ruspecial` (×3), `lunavyqo`, `alexsobolew7`,
`andrewboudreau`, plus `tangosdev` (×14). Clone checked with
`git rev-parse --is-shallow-repository` -> **false** before anything was banked.

31 `path#symbol` override rows added to `attribution.json`. The gate resolves a
consolidated member by the symbol name it had **before** the fold
(`member_overrides_at` keys on the text after the `#`, and the caller looks it up
with the projected old name), so the 12 renamed members needed their
`func_ov006_*` spelling as well as their new mangled one; both are kept, never
pruned. Result: **19 consolidated with credit intact, 0 changed, 0 lost**, exit 0.

`attribution.json` **looks** sorted and is not. Inserting the rows in sorted
position produced a **2,057-line reorder** that would have conflicted with every
other open PR in the train; that was reverted and the rows appended instead, for
a 20-insertion/1-deletion diff, with an audit confirming every pre-existing row
keeps both its value and its position.

**Please apply the `attribution-override` label.** It is needed because the local
tool is *not* the same computation as the validator's credit line on a
many-to-one fold — the two disagree about which side is "before" — so a local
`0 changed` does not predict the gate. Judge this on `lost`, which is 0.
Reconciling `changed` further is a stated non-goal in this repo.

---

## Disclosures the reviewer should not have to find

**1. The queue row's `classif:no-header` blocker is stale — verified, not
chased.** `include/dMgJump3DMario_c.h` exists on `origin/main` and six committed
sources include it. The row **contradicts itself**: its own `has_header` column
reads `yes` beside a `no-header` blocker. No tool under `tools/` emits the string
`no-header` — `queue_audit.py` only ever *preserves* a `classif:` blocker, so
`--write` cannot clear one. Measured at this ref: **15 rows carry
`classif:no-header`, and exactly 3 of them are stale in this way**
(`dMgJump3DMario_c`, `dThIcon_c`, `dMgPsOpt_c` — all three read `has_header=yes`
with the header present). Neither the queue TSV nor
`notes/data/c-cpp-classification.tsv` was hand-edited; the latter is untouched by
this PR. This is a generator gap and wants its own change.

**2. `queue_audit --check` is red on the base, for reasons that are not this
class's.** Regenerating with `--write` into a scratch copy and diffing (then
restoring — the TSV in this PR is untouched beyond stage 2's own row) gives, at
the pushed ref, **five** disagreeing rows, and **none of them is this class's**:

| row | drift |
|---|---|
| `dEnemyBase_c` (ov002) | `shard_count` 30→31, `total_lines` 945→1093, `no-legacy-source:1` cleared |
| `dScStarSel_c` (ov003) | `shard_count` 12→14, `total_lines` 663→1042 |
| `dScMgSnowball_c` (ov006) | `shard_count` 20→22, `total_lines` 1271→1892, `no-legacy-source` 3→1 |
| `dScMgPachinko2_c+dScMgPachinko_c` (ov006) | `unmatched:1` cleared |
| `Dorrie+DorriePlatform+daDossyCap_c` (ov065) | `shard_count` 28→29, `total_lines` 810→932, `no-legacy-source:1` cleared |

**This was watched happening, which settles the ownership question rather than
arguing it.** Measured before the second `origin/main` merge the set was **four**
rows; after the merge it was **five**, and `dScMgSnowball_c` moved further in the
same direction (21→22 shards, 1550→1892 lines). The only thing that changed in
between was merging main's #2374, which landed `src/func_ov006_02126ee4.cpp` —
the snowball course generator. The drift is main's own unregenerated queue,
observed drifting further in real time, and none of it is downstream of this
promotion. Corroborating that from the geometry: three rows are in ov002 / ov003
/ ov065, modules this PR does not touch at all, and the two ov006 rows sit at
`0x0212568c..0x021295ac` and beyond — disjoint from and nowhere near this
promotion's `0x020c762c..0x020c8a30`.

Stage 2's decision to update only its own row rather than sweep unrelated drift
into a promotion commit is confirmed correct and left standing; a sweep here
would also have been stale again within the hour. For the record, `queue_audit`
and `cpp_tu_state` are wired into **no** workflow, so none of this gates CI.

**3. `tubuild linkcheck` refuses at `[4/8]` and that is expected.** `intact
production requires one .text claim and at least one non-text claim` — a
text-only promotion has none, and this fails identically on untouched TUs on pure
`origin/main`. No `production_mode` key and no fake non-text claim was added to
silence it. The link evidence is `rombuild`'s own full scratch delink + lcf +
whole-tree `mwccarm` + `mwldarm` + ROM build, above.

**4. `dsd check symbols --fail` FAILs inside the baseline control** with 9
pre-existing errors — 4 in ARM9 main, 5 in ITCM, **none naming ov006**. Present
on untouched `main`; **0 new**.

**5. `romdata_check`'s two PARTIAL rows are the two typeinfo-name strings, and
they are not mismatches.** Per-symbol verdicts (reachable only by calling
`check_object()` directly): `_ZTV16`, `_ZTV22`, `_ZTI16`, `_ZTI22` all VERIFIED
(12/12/12/8 bytes, `blindWords: 0`); `_ZTS16` and `_ZTS22` PARTIAL because the
comparison covers whole words only — 16 of 19 and 24 of 25 emitted bytes — and
each `romExtent` (20, 28) is the padded distance to the next symbol, not the
string. 44 + 40 reconciles exactly against the summary line. **0 DIFFERS.**

**6. Cheap link pre-check, since byte match + objisolate + reloc_audit + `[4b]`
is famously not a link proof.** The compiled object's `.symtab` was walked: 82
`SHN_UNDEF` entries, none of them a mangled name whose demangling is a plain C
ROM symbol. Every `func_*`/`data_*` reference is correctly undecorated, and all
seven distinct `_Z…` names resolve to a definition in `config/`.

**7. `check_dead_references` prints `1 baselined reference now resolves — run
--update to shrink the baseline`.** That is a nudge on someone else's baseline,
not a failure, and was deliberately not banked from this PR. `check_references`
prints the same shape of nudge (`3 fixed since the baseline`) and was likewise
left alone.

**8. No dead declarations, checked rather than assumed.** A twice-narrowed TU
leaves behind `extern`s for functions it no longer contains, and no gate looks.
Parsed at this ref: the file has **nine** `extern "C"` regions, eight of which
are the per-member wrappers around the eight surviving free functions (bodies
included, which is where they get C linkage). Exactly **one** is the
declaration-only file-scope region the header comment describes; it declares 15
symbols and **every one of the 15 is referenced elsewhere in the file — zero
dead**. The prose that says a class member may not sit inside a linkage
specification, so the file-scope region must follow the last `func_*` member,
matches the shipped layout.

**9. The manifest prose was re-read against what shipped.** The two hits for
`REFUS`/`SHADOW` are both accurate documentation, not residue from a superseded
plan: `verification.round` describes the compile mode, and
`verification.linkcheck.intact_object_route` records the expected `[4/8]`
refusal. All **28** `legacy_source` rows were audited against the tree — every
one names a file that really existed on `origin/main` with the extension the row
gives, so the `.c`-for-`.cpp` defect that made another class's recorded
`link-verified` unreproducible is not present here. 28 rows checked, 0 bad.

**10. Two small prose imprecisions in inherited text, left as they are.** The
manifest calls ov006's `.ctor` table "31-entry" where the span
`0x0213356c..0x021335ec` is 32 words. The load-bearing half of that claim was
re-derived here and holds: **0 entries point inside the run**, so there is no
static-initialiser order to simulate. Separately, several
`compiler_only_output` rows say an emitted size "equals the cartridge's own
storage extent"; for the two `_ZTS` rows the emitted size is the string length
(`0x13`, `0x19`) and the cartridge's extent to the next symbol is one and three
bytes larger — alignment padding, not a disagreement.

---

## Gates run at this ref

All re-run after merging `origin/main`, which moved **four times** during this
run. Both merges were clean — no conflict in either ratchet file, in
`attribution.json`, or in the generated note. The second brought
`src/func_ov006_02126ee4.cpp` (#2374) into **this same overlay** and spliced
ov006's `delinks.txt`, so `verify`, `rombuild` and every base-relative gate were
re-run afterwards rather than carried over. That landing is also, incidentally,
the `dScMgSnowball_c` queue drift disclosed above.

| gate | verdict |
|---|---|
| `tubuild.py verify ov006/dMgJump3DMario_c` | `28/28 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED` (3 runs; tree clean after each — no manifest write) |
| `[4b/8]` object audit, driven directly | `order_ok True, {'LICENSED': 28}, 0 refusals` |
| `[4b/8]` negative control, pragma removed | `order_ok False`, 1 refusal — the pragma is load-bearing |
| `rombuild.py -j 6` | `106/106 exact, 100.000000%`, `ROM-build analysis: PASS`, `reproducing 11,190 / mismatching 0` (3 runs, the last at the pushed SHA) |
| `rombuild.py -j 6 --no-cache` | identical, all 9,257 objects recompiled — `build/objcache` is a symlink to the shared clone-wide cache, so this is the run that makes the compile independent of the producers' |
| `romdata_check --files src/actors/dMgJump3DMario_c.cpp` | `4 VERIFIED, 2 PARTIAL, 0 DIFFERS, 0 UNNAMED` |
| `tu_order_check ov006/dMgJump3DMario_c` | `all match in ROM order`; 4 EXTRA carry a compiler-only policy |
| `tiers_ratchet --check` | `PASS  baseline 2686  current 2686` (red before this PR's correction) |
| `eligible.py -j 6` | see below |
| `check_references.py --against origin/main` | see below |
| `cpp_tu_compat.py --require-ready` | all `[READY]` |
| `check_src_tu_compiles.py --quiet` | `151/151 translation unit(s) compiled` |
| `check_rename_ledger.py --repo .` | `every mangled/vtable row agrees with its module's symbols.txt` (1875 checked) |
| `check_profile_campaign.py --repo .` | `786 claims resolved, 0 diverge` |
| `langmode_audit.py --check langmode-baseline.json` | `langmode ratchet PASS` |
| `check_header_offsets.py --changed origin/main` | `2 commented fields, 0 mismatched, 0 unparsed, struct spans 0xb8` |
| `port_refcheck.py` | `402 checked - all references resolve` |
| `check_duplicate_sources.py` | `9424 stems, none doubled` |
| `check_dead_references.py` | `no new dead references`, `no broken markdown links` |
| `queue_audit.py --check` | FAIL — 5 unrelated rows, disclosed above (4 drift on `shard_count`/`total_lines`; the 5th, `dScMgPachinko2_c+dScMgPachinko_c`, disagrees only on `unmatched`) |
| `cpp_tu_state.py --check-note` | `cpp-tu state note is current` (regenerated twice: once for the ratchet correction, once for main's #2364 denominator change) |
| `check_tubuild_conflicts.py --list` | `every recorded conflict matches the source` |
| `layout_check.py` | `layout-check: clean` |
| `source_coverage.py --check --base origin/main` | `clean (no bytes handed back to the cartridge)` |
| `prepush_attribution.py --base origin/main --head HEAD` | `19 consolidated with credit intact, 0 changed, 0 lost` |

Gates this PR had to `--update`: **`tiers_ratchet --update --reason`, once**, for
the two-row backslide explained above — the tool's own documented remedy, and
`--check` was genuinely red rather than red by construction. Nothing else was
banked; in particular the `check_dead_references` baseline-shrink nudge was left
alone.


andrewboudreau:
### Follow-up note: the validator's 9 credit reassignments

`PR validation` is green and reports **0 added, 9 changed, 0 lost**, flagged as *"not a blocker"*. The `attribution-override` label is applied. Recording the exact set here so the divergence from the local tool is on the record rather than left as a surprise.

**The local tool and the validator disagreed, and the disagreement is fully explained.** `prepush_attribution` finished at *"19 consolidated with credit intact, 0 changed, 0 lost"*. That 0 is not a contradiction of the validator's 9 — the two tools partition this TU's 28 members exactly complementarily:

| | count |
|---|---|
| members in the manifest | 28 |
| members carrying an `attribution.json` override row | 19 |
| members carrying **no** override row | **9** |

The 9 uncovered members are *precisely* the 9 the validator reports as changed — verified as a set equality, not a count match.

**Mechanism.** `member_overrides_at()` (`tools/prepush_attribution.py:64-105`) keys its map on the text after the `#` and the caller looks it up under the **projected (old)** symbol name, so the 19 rows silence the local check. A member with no row at all has nothing to project and falls back to the promotion commit's author — which the local tool does not model, and the validator does.

**The 9, with the credit the validator recorded as "Before":**

| address | current symbol | before |
|---|---|---|
| `0x020c762c` | `_ZN16dMgJump3DMario_c12Unk_020c762cEv` | lunavyqo |
| `0x020c76d0` | `_ZN16dMgJump3DMario_c12Unk_020c76d0Ev` | tangosdev |
| `0x020c76d8` | `_ZN16dMgJump3DMario_c12Unk_020c76d8Ev` | tangosdev |
| `0x020c7734` | `func_ov006_020c7734` | lunavyqo |
| `0x020c78ec` | `_ZN16dMgJump3DMario_c9StateDampEv` | tangosdev |
| `0x020c8270` | `_ZN16dMgJump3DMario_c8EnterHitEv` | tangosdev |
| `0x020c85bc` | `_ZN16dMgJump3DMario_c11StateWindUpEv` | tangosdev |
| `0x020c893c` | `_ZN16dMgJump3DMario_cD1Ev` | tangosdev |
| `0x020c8a04` | `_ZN16dMgJump3DMario_cC1Ev` | tangosdev |

**Deliberately not fixed on this branch.** Attribution reconciliation is a stated non-goal for this run, `lost` is 0, and the validator validates the *final* SHA — pushing 9 more override rows would discard a green `PR validation` to chase a non-blocker. If someone wants the credit repaired, it is nine `src/actors/dMgJump3DMario_c.cpp#<symbol>` rows taken from the "Before" column above, appended (never sorted) to the `overrides` block, in a separate follow-up PR.

PR #2381, opened 2026-09-07
Phase 5: Add overlay and data links in provenance documentation, biggest one yet

Alberto12345678999:
Enhance the documentation by adding overlay and data links in the `ITCM` section and continuing work on the two biggest docs alongside `plan-tu-merge-queue` :

`minigame-provenance`;
`ghidra-dsd-landscape`


Alberto12345678999:
Hang on, @andrewboudreau, i saw some errors that ChatGPT pointed out, it was right, i will push a further commit to resolve these tiny mistakes.

Alberto12345678999:
Yes! Managed to fix the errors, and push safely with the real number of commits! I do get a dead reference, that isn't in the scope of this PR.

Alberto12345678999:
Should be all set now.

PR #2386, opened 2026-09-07
notes: add an index, a lifecycle policy, and a retrieval benchmark

andrewboudreau:
Adds a generated `notes/` index, a lifecycle policy, a registry, and a ten-question retrieval benchmark. The index routes readers to existing evidence; the benchmark measures answer reachability and the bytes along those lookup paths.

This remains a draft. Its current-main conflicts still require a separate integration review, including preservation of newer research, retired-source decisions, and the current agent policy. The September 20 tooling repair does not resolve those conflicts or establish source acceptance for the entire PR.

## Tooling repairs

The generated index and benchmark now count LF-normalized UTF-8 bytes. A Windows checkout therefore produces the same budget as an LF checkout, including the index's self-counting total. Equal-hop benchmark routes use a deterministic order.

The expanded reference scanner distinguishes historical Git provenance in `config/port_linkage.json` from live file references. Only its top-level string `branch` and `basis` fields receive that treatment; comments, other fields, nested values, malformed input, and other files retain their checks. Paths containing `+`, such as combined class filenames, remain complete tokens.

The original index, registry, lifecycle policy, codegen digest, and broader JSON/JSONL/JavaScript citation scanning remain part of this PR. The repair adds one registry row for its handoff and does not edit either reference baseline. It preserves the original branch ancestry.

## Current verification

At `3991b404f2c9b3d7186d987c44ba70cced9cc5df`, independently checked in a separate fresh Windows worktree:

- 66 tests pass across the notes index, benchmark, and reference checker. Negative controls still reject stale indexes, missing or unregistered notes, changed facts, malformed reference metadata, and actual missing paths.
- `notes_index.py --check` passes: 156 indexed files; 364,788 normalized doctrine bytes across 19 files. The doctrine cap remains advisory.
- `notes_bench.py --strict` passes: 10/10 answers resolve, all are reachable in one hop, and summed normalized path bytes are 393,760. Producer runs under two hash seeds agree.
- `check_dead_references.py` passes: no new dead references, 981 Markdown links, zero broken links. Existing baselined debt remains.
- The exact current-main port-linkage JSON was exercised as a build-only fixture. It retains real comment paths without treating the recorded branch names as files.

These figures describe a ten-question path-byte proxy. They do not measure reader time, model tokens, or a general 67% reduction in question-answering effort; the earlier headline overstated the result. Historical raw-byte totals are not directly comparable to these normalized measurements.

## Remaining integration

The published draft still has overlapping changes with current main, including `AGENTS.md`, reference bookkeeping, newer provenance notes, and paths main has retired. Reconcile those deliberately and obtain fresh current-base Source review and private validation before considering readiness. No source, header, ROM input, or shared attribution/declaration ledger changed in the nine-file tooling repair; no new full-ROM acceptance is claimed for this draft.

The committed [repair handoff](https://github.com/tangosdev/sm64ds-decomp/blob/3991b404f2c9b3d7186d987c44ba70cced9cc5df/notes/agents/handoffs/pr2386-tool-repair-0920.md) records the measured scope and limitations.


andrewboudreau (commented):
Bounded tooling/integration review at draft head `ece7c4303cd9502dae7572d7a5ef01699cdbdd5b` found a reproducible generated-index failure:

**P2: the index does not survive a fresh Windows checkout.** With `core.autocrlf=true`, `python tools/notes_index.py --check` exits 1 on the untouched candidate. The committed budget is 370,228 bytes; the fresh render expects 370,425. `notes/README.md` gains exactly 197 bytes from its 197 CRLF line endings. `doctrine_bytes()` counts raw filesystem bytes, while `--write` emits the self-counted README with LF, so regenerating locally does not establish a stable check after Git checks it out again. Make the indexed size calculation independent of checkout line endings, and verify a fresh checkout after regeneration.

The strict retrieval benchmark passes 10/10, with one-hop answers and 398,038 total path bytes in this checkout. These are benchmark proxy measurements, not measured agent time or a general 67% reduction in question-answering cost; qualify the PR wording and reconcile its differing percentages.

Composition against main `c1c8846df622f3a49b6b107e0845e9929473d059` has four conflicts: AGENTS.md, enemy-leaf-provenance.md (modify/delete), minigame-provenance.md and mwccarm-codegen.md. Preserve the newly landed protocol and evidence when resolving them, then rerun the index/reference checks on that actual composition. In particular, the conventions' current-status command should use the active v2 protocol rather than `classqueue.py list`.

The draft remains unmerged. This is not a complete audit of every condensed research paragraph or a source-acceptance pass.

andrewboudreau:
Recording what actually blocks this PR, since the branch is otherwise in good shape.

**The conflict is resolved.** `notes/audit-0905-restack-0908` @ `0bb529421` merges cleanly into current `main` — `git merge-tree --write-tree origin/main` returns no conflict lines.

**What blocks it is a false positive introduced by this branch's own checker change.**

```
python tools/premerge_check.py 0bb529421d2bd31435accaf395e55bd6fae0fe02 --base origin/main

  dead-references     pass     fail     REGRESSION

  FAIL: 1 prose/comment reference(s) name a path that does not exist:

    config/port_linkage.json
        names `port/link100`, which is not in the tree
```

`config/port_linkage.json:34` reads `"branch": "port/link100"`. That is a **git branch**, not a path — it exists at `refs/heads/port/link100`, currently `597e584fcdeee54f1d77c18e344c4e4e2799b3c4`, which is the exact sha the same JSON object cites in its `basis` field. It is correct as written and there is nothing in the tree for it to name.

The regression appears because this branch widens the scan to config JSON, and `port` is in `TOPDIRS`, so a slash-separated branch name is shaped like a repo-rooted path. `main`'s current checker does not scan that file, which is why the base side is green.

Both baselines are the wrong remedy here — banking it would put a waiver in the file whose own docstring says a baseline nobody needs "is a place for real breakage to hide." The fix belongs in the scanner: a value under a `"branch"` key (or a string that resolves as a ref) is not a path reference.

Two other signals from the same run, neither a failure:

- 2 baselined prose references now resolve — the prose baseline can shrink with `--update`.
- 15 baselined C/C++ comment references now resolve — refresh with `--update-code` from a clean, pinned `main`.

Adjacent but distinct: #2534 covers a separate defect in the same file, where `PATH_RE` truncates a path at a `+` and reports the non-existent prefix as dead. Different root cause; both would be natural to fix together.

I am not able to edit `tools/check_dead_references.py` in this session, so I have left the branch as it stands rather than work around the finding.

andrewboudreau (commented):
## The dead-reference blocker is real, but it is a composition failure, not a head failure

Run on this PR's head alone, `tools/check_dead_references.py` is **green**:

```
check_dead_references: 10507 scan target(s), 9930 C/C++ file(s), 11937 repo-rooted
path reference(s), 1717 that do not resolve; 980 markdown link(s), 0 that do not
resolve relative to their own file
  no new dead references
  no broken markdown links
EXIT 0
```

`pytest tools/test_check_dead_references.py` is 48 passed, and `tools/notes_bench.py`
is 10/10. So the PR's own verification block is accurate as written.

It goes red the moment it meets today's `main`. This PR extends the prose scan
surface to `config/**/*.json`. On the merge base, `config/port_linkage.json` held
`"branch": "port-mount-noseat-cluster"`, which the path regex does not match. On
`main` today that same field reads `"branch": "port/link100"`, which the regex does
match, and `port/link100` is a git branch, not a tracked path.

Isolated — this PR's head with only `config/port_linkage.json` replaced by main's
version, nothing else changed:

```
FAIL: 1 prose/comment reference(s) name a path that does not exist:

  config/port_linkage.json
      names `port/link100`, which is not in the tree

EXIT 1
```

That is the whole delta. Every other entry that appears when this PR's scanner is
run against main's content (`notes/pret-idioms.md` in `tangos.json`,
`tools/refine_run.js`, `tools/sched_run.js`, `tools/archive/crack_pr104.js` and
`notes/levers.jsonl`; `notes/stale-tu-map-overcut-ov006.md` and
`notes/minigame-family-decisions.md` in three `config/tu_manifest.d` manifests;
`src/actors/daPropeller_Heyho_c.cpp` in `config/profile-campaign-baseline.json`)
is already fixed by this branch — all of those files are in its diff.

### Minimal fix

`config/port_linkage.json` is rewritten by the progress bot on every restamp — the
last twenty-odd commits touching it are all "Restamp LINKED off port/link100 …" —
so the branch name in it will keep changing. Banking `port/link100` in the prose
baseline therefore re-breaks on the next rename, and leaves a waiver for a path
nobody wrote, which is the failure mode `tools/check_dead_references.py`'s own
docstring warns about.

The durable minimal fix is to keep `config/port_linkage.json` out of the prose
scan surface the same way this PR already excludes the other bot-written ledgers:
it is generated state whose `branch` field is a ref name by construction, not
prose anyone can fix. One line alongside the existing `ROOT_JSON_LEDGERS` /
`JSONL_LEDGERS` sets does it. Excluding the `branch` key specifically would work
too and is narrower, but costs a special case in the scanner.

I have not edited the tool.

### Note: this is not issue #2534

#2534 is the `+`-truncation defect in `PATH_RE` (paths like
`src_tu/actors/Koopa+KoopaSmall.cpp` matching as `src_tu/actors/Koopa`). I could
not get that one to fire from this branch, on the head or in the composition
above. Landing this PR would still close it if the regex fix ships with it, but
the thing actually blocking the gate today is the `port/link100` value, which is a
different false positive with a different fix.

### Two blockers ahead of that one

* This PR is still marked **draft**.
* It **conflicts with `main`** in 8 files, four of them modify/delete where main
  edited a note this branch retires: `AGENTS.md`,
  `config/dead-reference-baseline.json`, `notes/minigame-provenance.md`,
  `notes/mwccarm-codegen.md`, and the deletions of
  `notes/enemy-leaf-provenance.md`, `notes/overlay-ambiguous-references.md`,
  `notes/sinit-ownership-pilot-ov002.md`, `notes/sinit-wave-a-ownership.md`.
  The four modify/delete pairs are decisions, not merges — main added content to
  files this branch retires, and the retire rule in the new `notes/CONVENTIONS.md`
  §3 requires the citations to move in the same commit.

PR #2395, opened 2026-09-07
ov081: promote daGmch_c into a single C++ TU (36 of 37 functions become real methods)

andrewboudreau:
## TL;DR for a non-specialist

Overlay 81 held a hopping coin-purse enemy. The cartridge's own type information
calls the class **`daGmch_c`**; our tree had been calling it `Moneybag`, a name
that appears nowhere in the ROM. This PR retires that coined name and folds the
37 one-function files that made up the class into a single C++ translation unit
that compiles to **byte-identical machine code**.

The interesting part is not the fold, it is the conversion. Of the 37 functions,
**36 are now real `daGmch_c::` methods** — including all 26 that previously had
machine-generated `func_ov081_<address>` names. Nothing was left as a free
function because it refused to match, and nothing hit a compiler or scoping wall.
The 37th, `daGmch_c_classInit`, stays a free function **by design**: the
cartridge's own symbol for it is a free symbol, and it *constructs* the object
rather than running on one. That is the same shape the landed `ov006/dScMgLuigi_c`
promotion shipped.

So the honest headline is **36/37 converted, with the 37th excluded on purpose** —
not 37/37.

---

## Proof block

```
tubuild verify ov081/daGmch_c   37/37 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
                                emission order: all 37 in expected ROM-ascending section order
rombuild.py -j 16               106/106 exact, 100.000000% of compared bytes
                                11,191 source-built functions reproducing, 0 mismatching
                                profile stock: 0 mod source replacements, 0 mod-entry gap fallbacks
                                ROM-build analysis: PASS
romdata_check --files ...       5 VERIFIED, 4 PARTIAL, 0 DIFFERS, 0 UNNAMED
tu_order_check ov081/daGmch_c   all match in ROM order; 2 EXTRA licensed by the manifest
tiers_ratchet --check           PASS, baseline 2698 / current 2698
```

`rombuild` is also the **link** proof here. This is a text-only promotion with no
`production_mode`, so the merged source is compiled and linked inside rombuild's
own baseline control like any other enrolled file; reaching `106/106 exact` means
it linked. Its ROM sha256 is identical to the pre-promotion control.

### Full gate sweep

| Gate | Result |
|---|---|
| `tubuild.py verify ov081/daGmch_c` | 37/37 MATCH, TEXT-VERIFIED; manifest unchanged after the run |
| `rombuild.py -j 16` | 106/106 exact, 100.000000%, PASS |
| `romdata_check.py` | 5 VERIFIED, 4 PARTIAL, 0 DIFFERS |
| `tu_order_check.py` | ROM order OK, extras licensed |
| `tiers_ratchet.py --check` | PASS 2698/2698 — **no `--update` was run** |
| `eligible.py -j 6` | 11,197 / 11,273 |
| `check_references.py --against origin/main` | OK, 42 unresolved (baseline 45), no new |
| `check_data_definitions.py` | 20,107 objects, none define a ROM data symbol |
| `cpp_tu_compat.py --require-ready` | all five concerns READY |
| `check_src_tu_compiles.py --quiet` | 154/154 TUs compile |
| `check_rename_ledger.py --repo .` | 1,949 mangled/vtable rows agree with `symbols.txt` |
| `check_profile_campaign.py --repo .` | 391 complete, 0 diverged |
| `check_header_offsets.py --changed origin/main` | 2 headers, 0 mismatched; `daGmch_c.h` spans 0x3f4 |
| `port_refcheck.py` | 402 checked, 0 stale |
| `check_duplicate_sources.py` | 9,299 stems, none doubled |
| `check_dead_references.py` | no new dead references, no broken markdown links |
| `queue_audit.py --check` | queue agrees with the tree |
| `cpp_tu_state.py --check-note` | note is current |
| `check_tubuild_conflicts.py --list` | 154 entries; every conflict recorded matches its source |
| `layout_check.py` | clean |
| `source_coverage.py --check --base origin/main` | net +0 B, **0 B handed back to the cartridge** |
| `langmode_audit.py --check` | PASS at 165 / 168 — see the withdrawn claim below |
| `prepush_attribution.py` | 0 changed, **27 lost** — see below |
| `dsd check symbols` | **not green: 9 errors, 0 new, none in ov081** |

No gate required an `--update`. `tiers_ratchet --check` already passed, so it was
audited rather than re-banked: the exceptions file has **0 byte-identical
whole-record duplicates** (404 rows, 7 added), and `converted-baseline.json` went
**2696 → 2698** as 7 plain shard paths left and 9 `daGmch_c.cpp#symbol` rows
entered. The unbalanced swap is expected — two members qualify under the
readability criteria in merged form that did not qualify as standalone shards.

## Not green, and one claim withdrawn

**`dsd check symbols` does not pass.** It reports the same **9 pre-existing
errors** present on untouched `main`, and **0 new** ones. None of the 9 is in
ov081 or touches this class — they are 4 ITCM functions, `_deq`, two
`data_020ad5xx` symbols in ARM9 main, and `overlay_100` / `overlay_102`.
`rombuild`'s own intact-TU gate independently reports `zero new symbol errors`.

### A claim I made, tested, and withdrew

An earlier revision of this PR said `langmode_audit`'s extern-vtable ratchet rose
by one and that the rise was **unavoidable** — that the factory *must* declare
`extern int _ZTV8daGmch_c[];` locally, and that putting it in a header would move
the counter without changing a byte and was therefore gaming the metric. A
reviewer asked me to test that "must". **It does not hold, and I withdraw it.**
Leaving the retracted reasoning visible on purpose, because the correction is the
useful part.

**What was actually wrong.** Before this promotion the shard `src/d_a_gmch.c`
reached the same symbol through `decl_common.h` and carried **no `extern` of its
own** — it only *used* the name (`p[0] = (int)_ZTV8Moneybag;`). So a header is the
declaration site this class already had. Moving the declaration back into
`include/daGmch_c.h`, which the TU already includes, is conformance with what
`main` did, not evasion of a counter. That is the fact that reverses the
judgement.

**Measured free.** Compiling `src/actors/daGmch_c.cpp` with the pinned mwccarm
2004/b56 before and after the move gives byte-identical objects — 22,176 bytes,
sha256 `0c6eee50a63564d9ba625d3645e2c33e89f7514a6134fb579c9579e1b6f4a5d6` both
times. An extern array declaration used only for `&_ZTV8daGmch_c[2]` emits no code
of its own; the address is a relocation either way. (Checked with a direct
compiler invocation, deliberately **not** `tubuild verify` — a failing verify
rewrites manifest prose and `droppedSections` in place, which is not a risk worth
taking for a hypothesis test.)

**Result: there is no rise, and no baseline was re-banked.**
`langmode_audit.py --check` now passes at exactly **165 / 168**, its previous
values.

**Where the rise had come from.** Replaying the metric's own regex over both trees:
`extern_vtable` 165 → 166 was **exactly one file added** (`src/actors/daGmch_c.cpp`)
with none removed, and `extern_vtable_classes` 168 → 169 was **exactly one class
added** (`_ZTV8daGmch_c`) with none removed. **Zero** of it came from the 26
`func_ov081_*` → method renames — the metric matches only `extern … _ZTV…;`
declarations, and a function rename cannot move it. So there was never a component
bookable under the ratchet's "renames giving previously-unnamed functions their
evidenced C++ symbols" rule; the whole thing was one declaration in the wrong place.

**One further correction.** The manifest and the TU banner justify not including
`decl_common.h` by citing three `illegal function overloading` errors from 7
auto-named members. That reasoning describes the **pre-promotion** header: this
PR's own edit removes those 7 rows. Against the post-edit header the surviving
conflict is a single line — `func_02038414`, which `decl_common.h:2059` declares
`void func_02038414(void*)` against the `int func_02038414(void *clsn)` this TU
requires — and it has nothing to do with the vtable. Only 1 of the TU's 61
file-scope declarations overlaps `decl_common.h` at all.

**`prepush_attribution` reports `0 changed, 27 lost`** (24 `tangosdev`, 2
`lunavyqo`, 1 `ruspecial`). This is the standard many-to-one fold artifact and the
project's standing instruction is that attribution is a non-goal and the
`attribution-override` label is the way it lands; I have applied that label rather
than banking 27 override rows. Note also that a local run is not the validator's
computation on a fold — overrides must be built from the validator's own "Before"
column, never from the local tool's.

## The one thing blocking this PR: the vtable rename, not the bytes

`PR validation` fails with exactly one reason:

```
ROM data verification lost 1 exact symbol(s): ov081:_ZTV8Moneybag
```

**No bytes were lost.** `_ZTV8daGmch_c` is present in the head run's verified set
(686 exact symbols), same module `ov081`, same address `0x02128c04`, same 0x84
bytes. Module fidelity is `106/106 exact, 100.000000%`, byte-verified functions
and code bytes both move `+0`, and the validator did **not** raise its companion
`ROM data verified bytes fell from …` check — so the verified byte total did not
drop either. What moved is the *name key*, and only because this PR retires the
coined `Moneybag` for the cartridge's own RTTI spelling.

The check (`tools/validate_merge.py`, `rom_data_regressions`) is a raw set
difference of `(module, symbol)` pairs between base and head `verifiedSymbols`. It
has no rename awareness: `symbols/actor_renames.tsv` is not consulted, and the only
`aliases` map in the tool is for *contributor* names in `attribution.json`. So any
rename of an exactly-verified data symbol reports as a loss, by construction.

That behaviour looks deliberate — the surrounding comment says "known-good symbol
identities and bytes may not be lost" — so **this needs a maintainer decision, not
a workaround.** I have not touched it. For the record, the two workarounds I can
see are both wrong and I did not take either: adding an alias row for the retired
`_ZTV8Moneybag` to `ov081/symbols.txt` would fabricate a symbol identity to silence
a gate, and editing the validator is tooling work that does not belong in a source
reconstruction PR.

Attribution, by contrast, is explicitly **not** blocking here — the validator
records `0 added, 36 changed, 0 lost` and labels it "contributor attribution
changed, not a blocker".

## Class identity, from the cartridge

The class name is **the ROM's own RTTI, not a coined name**:

- `_ZTS8daGmch_c` at **0x02128bc8** — `overlay_0081.bin` file offset 0x5488 holds
  `38 64 61 47 6d 63 68 5f 63 00`, the Itanium length-prefixed `"8daGmch_c"`.
- `_ZTI8daGmch_c` at **0x02128bd4** — an `__si_class_type_info` record pointing at
  that string, with base **`dActor_c`**.
- `_ZTV8daGmch_c` address point at **0x02128c04**; the object runs
  0x02128bfc..0x02128c80 = **0x84 bytes** (mwcc's two-word preamble plus 31 slots).

`sizeof` is corroborated three ways at **0x3f4 (1012)**: the factory
`daGmch_c_classInit` at 0x02127adc passes the literal 1012 to
`fBase_c::operator new` (`opnew_sizes` calls it `exact`, one site, resolved in its
own module); `include/daGmch_c.h` asserts 0x3f4; and `check_header_offsets`
independently measures the struct span as 0x3f4.

**Vtable: 31 slots against `dActor_c`'s 31** — the length itself is what proves the
base. Ten slots are overrides:

| Slot | Symbol |
|---|---|
| 0 | `daGmch_c::InitResources` |
| 3 | `daGmch_c::CleanupResources` |
| 6 | `daGmch_c::Behavior` |
| 9 | `daGmch_c::Render` |
| 12 | `daGmch_c::OnPendingDestroy` |
| 16 | `daGmch_c::~daGmch_c` (D1) |
| 17 | `daGmch_c::~daGmch_c` (D0) |
| 18 | `daGmch_c::OnYoshiTryEat` |
| 19 | `daGmch_c::OnTurnIntoEgg` |
| 29 | `daGmch_c::OnAimedAtWithEgg` |

**Destructor placement.** The cartridge puts D1 (0x02126504) *below* D0
(0x02126554) and carries no `daGmch_c` D2 at all. Out of line under
`#pragma defer_codegen off`, mwccarm 2004/b56 emits D1, D0, D2 in that order, so
the two real symbols land in the ROM's own order and the D2 is a homeless extra.
The same pragma also makes `.text` come out in source order rather than reversed,
which is why members appear lowest-ROM-address first. The two are one decision:
flipping either alone makes the ROM-ascending emission-order audit refuse.

I verified the D2 really is homeless: no `symbols.txt` or `delinks.txt` anywhere in
`config/` names a `daGmch_c` or `Moneybag` D2, so a plain deadstrip is right.

**Promotion route: text-only plus `compiler_only_output`** (no `production_mode`,
`data: []`, `bss: []`). The manifest licenses 11 compiler-only symbols. I checked
every one of the 10 that carries a canonical address against its owning module's
`symbols.txt` — all 10 match exactly, and each is defined in exactly one module:

```
_ZN7Vector3D1Ev   arm9  0x020072c0   deadstrip-duplicate
_ZTV8daGmch_c    ov081  0x02128c04   deadstrip-data
_ZTI8daGmch_c    ov081  0x02128bd4   deadstrip-data
_ZTS8daGmch_c    ov081  0x02128bc8   deadstrip-data
_ZTI8dActor_c     arm9  0x0208e390   deadstrip-data
_ZTS8dActor_c     arm9  0x0208e384   deadstrip-data
_ZTI7dBase_c      arm9  0x02086e78   deadstrip-data
_ZTS7dBase_c      arm9  0x02086e54   deadstrip-data
_ZTI7fBase_c      arm9  0x02086d70   deadstrip-data
_ZTS7fBase_c      arm9  0x02086e60   deadstrip-data
```

I also audited all **37 of 37** `legacy_source` rows in the manifest against the
pre-promotion tree and its `delinks.txt`: every spelling resolves, with **0**
extension mismatches of the kind that silently made one landed manifest's
`link-verified` status non-reproducible.

## What the naming does and does not claim

The 18 state handlers (`EnterStateN` / `UpdateStateN`) are ROM-proven on two
independent witnesses, but they assert only the state **index** and the **half** —
enter versus update. **They claim nothing about what any state means.** Please read
them that way; the header records the same caveat per name. The other coined member
names say what the byte-matched body computes, and the ledger rows carry that
qualification.

## One measured result worth keeping

Making every member a method forced all **54** external declarations to file scope,
because a member function may not sit in a linkage specification. Nine of the 54
had multiple, disagreeing type spellings across the shards; **eight collapsed for
free**. The ninth is load-bearing: declaring `RandomIntInternal` as `int` rather
than `unsigned int` drops the TU to **36/37**, breaking `ChooseNextState` alone —
the only caller that takes `% 10` of the result. (This is the producer's
measurement, quoted rather than re-run: reproducing it means a deliberately failing
`verify`, and a failing `verify` rewrites manifest prose and the compiler-only
section list in place.)

Relatedly, `decl_common.h` is deliberately **not** included. It declared 7 of this
TU's auto-named members, and 3 of those 7 contradict the byte-matched definitions
(0x02126950, 0x02127708, 0x0212777c) — including it would raise three
`illegal function overloading` errors aimed at the definitions. Those 7 rows named
nothing else in the tree, so this PR removes them.

## A pre-existing `main` failure this PR also clears

`queue_audit --check` is currently **red on `main`**, and not because of this
class. PR #2387 landed `src/func_ov006_020d36a4.cpp` inside `dScMgAmida_c`'s shard
set and left that queue row stale. Re-deriving the file with `--write` corrects it:

```
dScMgAmida_c   shard_count 25 -> 26, total_lines 2001 -> 2148, no-legacy-source 2 -> 1
```

That row is kept deliberately — it is a genuine regeneration of state `main` is
already wrong about, not scope creep. `queue_audit --check` passes on this branch.
The inputs were regenerated in the correct order (`rtti_extract` → `rtti_vtables` →
`tu_map`), giving the non-degraded **691 TUs / 333 classed**.

## Expected validator lines

This is a many-to-one fold: 37 shards become 1 file and `ov081/delinks.txt` goes
from 113 source blocks to 77. The credit counter attributes only the delinks
range's *first* symbol and reclassifies the rest as "claimed", so these are
artifacts of the fold and not losses:

```
```

Those four N's are **not one number**. The ranges-left count runs exactly one
higher than the byte-verified function delta, because the TU's own new range
enters the set as the 37 shard ranges leave.

## Ledger hygiene

`symbols/actor_renames.tsv` is an append log: **26 new member rows are appended at
the end**, and 8 pre-existing ov081 rows are rewritten in place for the
`Moneybag` → `daGmch_c` rename, each carrying an explicit `(was ...)` note. The
vtable row deliberately **keeps its `alloc=0x3f4`**: `check_rename_ledger --fix`
withdraws that field by default, but here the corrected name is the *same class*
under its cartridge RTTI spelling, not a different owner, so the figure still
stands. Please do not let a re-run of `--fix` strip it.

---

Producer: reconstruction and byte verification. This PR is the independent gate
pass, re-run from scratch in its own worktree at `b2bd6a323`.


andrewboudreau:
Advisory humanizer review for #2412 at **`2f34c64aa90c712c0da8b936842592a39ad853cf`**. Source-quality recommendation: **corrections needed**.

1. **Preserve original contributor credit (P1).** This folds 37 functions without adding `daGmch_c.cpp#symbol` attribution overrides. The private report records 36 contributor reassignments. A fresh `prepush_attribution.py` comparison using tools/base `f327f7b6460e157153eb7fc0749dbbe60dd854f1` and this exact head exits 1 with **27 lost source-lineage identities, zero changed**. These are different metrics. Restore each original symbol owner's credit through the promoted overrides, including renamed identities, and verify both readers. The [folded definitions](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/src/actors/daGmch_c.cpp#L286) must not all inherit their latest editor's credit.
2. **Correct name provenance (P2).** The [source preamble](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/src/actors/daGmch_c.cpp#L53) and [header](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/include/daGmch_c.h#L92) claim the cartridge supplies ten mangled members and the free factory spelling. RTTI establishes the class name; these function spellings are reconstructed. The facts already disclose this for the factory. Correct those claims and the two ledger rationales saying “cartridge-mangled”; retain the useful class, address and behavior evidence.
3. **Disclose the specific signature choices (P2).** [`SetState(int)` and `ApplySlopeToVertSpeed(void*)`](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/include/daGmch_c.h#L127) encode signed-int and void-pointer choices that the cited operations do not establish as original types. Promotion convention section 1 requires those caveats beside the declarations, in the rename rationales and in manifest notes. This requests accurate disclosure, not a speculative ABI change.
4. **Make the handoff describe the final implementation (P2).** [Manifest conflict resolutions](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/config/tu_manifest.d/ov081/daGmch_c.json#L296) repeatedly say function declarations were moved into member bodies. The [final source](https://github.com/tangosdev/sm64ds-decomp/blob/2f34c64aa90c712c0da8b936842592a39ad853cf/src/actors/daGmch_c.cpp#L185) reconciles them at file scope. Mark the earlier resolution superseded while retaining measured probe results.

The genuine destructor/subobject work and bounded text promotion have substantial evidence. Raw offsets and shadow views remain reconstruction opportunities: short state setters are suitable focused probes using already modeled fields. Preserve measured compiler barriers, and require fresh relocated-byte proof for any source improvement.

The [private gate remains FAILURE](https://github.com/tangosdev/sm64ds-decomp/pull/2395#issuecomment-5565531893), completed 2026-09-07 06:41:15 UTC, for lost exact data identity `ov081:_ZTV8Moneybag`. Coordinate that cause with #2409. The report's 37 verified TU functions do not override its failed overall result or unresolved affected-consumer totals (4 BLIND, 1 DRAFT).

This review inspected source, header, facts, manifest and relevant changes, and freshly ran attribution. It did not rerun compilation/full-ROM checks or accept a v2 verifier stage. `fleet-0907` should assign narrow follow-ups through #2412 and obtain fresh independent verification of the resulting candidate.

andrewboudreau:
Superseded by #2431, which carries this PR's content in full rather than replacing it.

**Why this is a supersession and not a competing PR.** This PR's head
`2f34c64aa90c712c0da8b936842592a39ad853cf` is a strict *ancestor* of #2431's branch:

```
2f34c64aa  Declare _ZTV8daGmch_c in the header, not the factory body   <- this PR's head
4644212ac  Name daGmch_c's fields in every member and record the v2 handoff
22602102f  Merge origin/main into integ/daGmch_c-2412
7b04f3676  Reconcile daGmch_c's integration lane and correct its handoff counts
```

Nothing here was rebased, squashed or rewritten, so every commit on
`cpp/Moneybag-tu` and its authorship travels into #2431 unchanged. The branch is
left in place; it is not deleted.

**What #2431 adds on top of this PR.** `4644212ac` names the fields in every
member -- raw `(char *)this + 0xNNN` accesses drop from about 240 to the 6 in the
free factory and shadow type definitions from 9 to 4 -- and that commit passed
independent verification (37/37 `tubuild verify` MATCH, 106/106 modules exact,
`linkcheck` driven by symbol for all 37 functions: 37 VERIFIED / 0 BLIND /
0 WRONG). #2431 then composes it on current `main` and re-proves the composition.

**The blocker this PR recorded is fixed on `main`.** This PR failed private
validation with exactly one reason,
`ROM data verification lost 1 exact symbol(s): ov081:_ZTV8Moneybag`, and correctly
declined both workarounds. #2425 (`1c93d2663`) landed the real fix: the ROM-data
diff is now anchored on `(module, address, bytes)`, so retiring a coined
`_ZTV<Coined>` for the cartridge's own `_ZTV<RomName>` at the same address reads as
a rename. #2431 is its first end-to-end test on this class.

**Contributor credit.** This PR could only record the fold's credit movement with
the `attribution-override` label. #2431 banks it as 37 real
`src/actors/daGmch_c.cpp#<symbol>` override rows in `attribution.json`, built with
the same lineage computation `tools/tu_promote.py` uses, so the six members owed
to `lunavyqo` (4), `ruspecial` (1) and `andrewboudreau` (1) keep their credit
through the fold instead of it moving to whoever pushed it.

Closing so only one PR claims `daGmch_c`. Review continues on #2431.

PR #2397, opened 2026-09-07
ov030: promote daMky_c into a single C++ TU

andrewboudreau:
## TL;DR for a CS generalist

SM64DS's monkey — the one that steals Mario's cap — was scattered across 44
separate one-function C files, because that is what an automated splitter
produces when it does not know which functions came from the same original
`.cpp`. This PR puts those 44 functions back into one file,
`src/actors/daMky_c.cpp`, and proves the compiler turns that file into
**byte-identical machine code** to what is on the cartridge. It also gives 11
more of those functions their real C++ method names, so the class now has 20
real `daMky_c::` methods instead of 9.

The class is named `daMky_c` because the cartridge says so: the RTTI string
`7daMky_c` is literally in the ROM at `0x02115b78`. The tree used to call it
`Ukiki`, which appears nowhere in the cartridge.

Two gates are red and neither is hidden below — one is inherited from `main`,
one I was not able to clear. Details in "Not green".

## Proof block

    tubuild verify    44/44 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
    rombuild -j16     106/106 exact, 100.000000% of compared bytes, mismatching 0
                      profile stock: 0 mod source replacements, 0 ROM-gap fallbacks
                      ROM-build analysis: PASS
                      build/sm64ds.nds sha256 d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8
    romdata_check     4 VERIFIED, 5 PARTIAL, 0 DIFFERS, 0 UNNAMED
    tu_order_check    all match in ROM order; 3 EXTRA carry a compiler-only policy (PASS form)
    source_coverage   0 B handed back to the cartridge
    prepush_attrib    35 consolidated with credit intact, 0 changed, 0 lost

`rombuild` is the link evidence. A text-only promoted TU is compiled and linked
inside its BASELINE CONTROL like any other enrolled source, so `106/106 exact`
is the proof that the merged TU actually links — `tubuild linkcheck` cannot pass
on an already-promoted TU and was not used for that claim.

## Class identity — read from the cartridge, not coined

| | |
|---|---|
| `_ZTS7daMky_c` | ov030 `0x02115b78` — bytes `37 64 61 4d 6b 79 5f 63 00` = `7daMky_c` |
| `_ZTI7daMky_c` | ov030 `0x02115b84`, `__si_class_type_info`, base word -> `_ZTI8dActor_c` (arm9 `0x0208e390`) |
| `_ZTV7daMky_c` | ov030 `0x02115bfc`, 31 slots, against a 31-slot `dActor_c` base |
| instance size | `0x3cc` (972) — both factories do `mov r0, #0x3cc` then `bl _ZN7fBase_cnwEj` |

The class name is **not** coined. `Ukiki` was, and it is withdrawn: a scan of
2,286 images / 21 MB (every overlay, arm9, arm7 and the whole ROM filesystem)
found 0 hits for the length-prefixed `5Ukiki` and 0 for bare `Ukiki`
case-sensitive.

**Vtable overrides** (diffed slot-by-slot against `_ZTV8dActor_c`): slot 0
`InitResources`, 3 `CleanupResources`, 6 `Behavior`, 9 `Render`, 12
`OnPendingDestroy`, 16/17 `~daMky_c` D1/D0, 18 `OnYoshiTryEat`, 19
`OnTurnIntoEgg`. Every other slot holds the base's word and is inherited.

**Destructor placement.** `~daMky_c` is the first virtual declared in the header
and is defined out of line, so this TU owns the key function and emits
`_ZTV`/`_ZTI`/`_ZTS`. Written under `#pragma defer_codegen off`, which makes an
out-of-line destructor emit D1, D0, D2 in source order — matching the cartridge,
which puts D1 at `0x02111688` and D0 at `0x021116d0`, the two lowest addresses
in the run. D2 has no ROM home and is deadstripped.

**Route:** text-only + `compiler_only_output` (12 rows: 9 RTTI data symbols,
`_ZN7daMky_cD2Ev`, `_ZN7Vector3D1Ev`, and the position-dependent
`_ZN21Bundle$724daMky_c_cppD1Ev`). Single `.text` claim
`0x02111688..0x021145e0`, `data: []`, `bss: []`, no `production_mode`. No
sibling oracle exists for ov030; the shape follows ov006/`dScMgPanel_c` in
keeping `decl_common.h` out.

## Methods: 20 of 44 — the other 24 are stopped by NAMING

Up from 9. Nine were already C++-named (the whole virtual interface); this
branch converts eleven more to `daMky_c::EnterState0..EnterState10`.

Their member-ness is ROM-read: all eleven are among the 22 pointer-to-member
records at `0x02115ac8`. Their *numbers* are ROM-read too: each writes exactly
one immediate to the state word at `+0x3b4`, and across all 44 functions those
immediates are 0..10 with none repeated and none missing, while `0x02115e0c`
holds exactly eleven `0x10`-byte descriptors. Eleven handlers, eleven
descriptors, eleven consecutive ids — a bijection, not a guess. The word
"EnterState" is **coined** and disclosed as such in `symbols/actor_renames.tsv`.

**The wall for the remaining 24 is naming — not codegen and not scope, and both
were measured absent before any conversion was written.** Two probes converted
four members (three declaration-free, one declaration-carrying, forcing its
declarations to file scope above three members that redeclare the same symbol
with a different return type); all four still byte-matched. Eleven of the 24 are
pointer-to-member targets whose member-ness is equally proven but which the ROM
neither names nor numbers; the other thirteen are not proven members at all. A
coined name with no evidence behind it is worth less than the address, so they
stay `func_ov030_*`. That is a result, not a failure.

The renames land **in the same commit** as the promotion because they have to:
the 22 pointer-to-member records are unowned `.data` that dsd resolves by NAME,
so a half-landed rename links them as `0x00000000`. Measured — writing the
renames into `symbols.txt` while the shards were still unpromoted turned the
linkcheck baseline red with `modules NOT byte-exact: ['ov030']`.

## A wrong class attribution in the tree, corrected

`src/func_ov030_02112a14.c` was labelled `daObjHmMaruta_c::AfterClsn` and read
its fields through a `struct daObjHmMaruta_c *`. **It is
`daMky_c::EnterState8`.** Four independent ROM facts:

- `0x02112a14` is inside this TU's `.text` run `0x02111688..0x021145e0`;
  `d_a_obj_hm_maruta.c`'s run is `0x0211164c..0x02111688` and holds one function.
- it is record 0 of daMky_c's 22-entry pointer-to-member array.
- it writes the immediate 8 to daMky_c's state word, and is the only one of the
  44 that writes 8.
- `daObjHmMaruta_c_classInit` allocates **836 (0x344)** bytes. This body stores
  to `+0x3c7` — 131 bytes past the end of that object. daMky_c is 972 (0x3cc)
  and contains it.

This is a correction, not a rename. `include/daObjHmMaruta_c.h` is
auto-generated "from 1 functions" — this one — so its entire field set
(`0x098`, `0x130`, `0x3a0`, `0x3b4`, `0x3c7`) was derived from the
misattribution, and three of those offsets its own class cannot reach. The two
classes share `dActor_c`'s low offsets, which is why the wrong header still
compiled.

The shipped body now uses `char *c = (char *)this` with raw offsets, exactly as
its ten sibling `EnterState` members do; it was the only one of the eleven
reaching for a foreign shadow struct. Byte-neutral — 44/44 MATCH before and
after, `EnterState8` unchanged at size `0x070`. The header is left in place but
no source includes it any more; a follow-up should decide its fate.

## Gates

Green: `tubuild verify`, `rombuild`, `romdata_check`, `tu_order_check`,
`tiers_ratchet --check` (PASS 2697/2697), `eligible` (11197/11273),
`check_references` (42 unresolved vs baseline 45), `cpp_tu_compat --require-ready`,
`check_src_tu_compiles` (154/154), `check_profile_campaign`, `langmode_audit`,
`check_header_offsets`, `port_refcheck` (402 checked, 0 stale),
`check_duplicate_sources`, `check_dead_references`, `check_tubuild_conflicts`,
`layout_check`, `source_coverage`, `cpp_tu_state --check-note`,
`prepush_attribution`.

**Nothing needed `--update`.** `tiers_ratchet --check` already passed, so it was
deliberately **not** re-run with `--update` — it appends rather than replaces.
Audited instead: 0 byte-identical duplicate whole records in the exceptions
file, and `converted-baseline.json` went 2696 -> 2697 with every removal a plain
shard path (5 `Ukiki` shards) and every addition a TU member row (6
`src/actors/daMky_c.cpp#symbol`). The unbalanced 5-out/6-in is `EnterState0`
newly qualifying. The two `--update` nudges from `check_references` and
`check_dead_references` are shrinkable-baseline hints, not failures, and were
not banked from a class PR.

Also audited, because no gate does it: all **44** `legacy_source` rows in the
manifest resolve exactly — path *and* extension — against both the
pre-promotion tree and its `delinks.txt`. 0 defects. All 12 hoisted file-scope
declarations are referenced by at least one member, so no dead declarations
shipped. The compiled object's `SHN_UNDEF` set carries 0 mangled `func_ov*`
names, the signature of the wrong-linkage bug that every byte gate misses.

One reading hazard for the next builder: 119 of the manifest's 130 notes are
auto-generated `tubuild create warning` strings that end "this TU hoists nothing
a function body can hold ... the two spellings never meet". Stage 3b later added
a 12-declaration file-scope region, so that clause is superseded — by note 128,
which describes the hoisting correctly and is the authority. The stale clause was
left as the create-time record rather than rewritten 119 times.

## Not green — please read before merging

**1. `check_rename_ledger` FAILS (8 rows, all ov030) — introduced by this
branch, not fixed here.** Renaming the class leaves 8 historical rows in
`symbols/actor_renames.tsv` whose 4th column still says `_ZN5Ukiki*` /
`_ZTV5Ukiki` where `symbols.txt` now says `_ZN7daMky_c*` / `_ZTV7daMky_c`. The
branch **does** append correct superseding rows for all 8 addresses, but the
checker validates every row independently — it has no last-row-wins rule — so
the historical rows still fail. This gate runs in CI
(`.github/workflows/rename-ledger.yml`, triggered by both paths this PR
touches), so **the PR will show it red**.

The remedy is the tool's own `python tools/check_rename_ledger.py --repo . --fix`,
which rewrites column 4 in place and preserves the retired claim as `(was ...)`
in column 5. I was not permitted to run it in this environment, so I left it for
the reviewer rather than hand-editing an append-log ledger. It is a one-command
fix and it touches only these 8 rows.

**2. `queue_audit --check` FAILS (3 rows) — inherited from `main`, not this
branch.** All three are `dScMgAmida_c` (25->26 shards, 2001->2148 lines,
`no-legacy-source` 2->1), arriving with a merge that landed today. Verified this
branch does not add to it: `queue_audit --write` wanted to change exactly two
rows, `daMky_c` and `dScMgAmida_c`, and only the `daMky_c` row was taken.
This gate is not wired into CI.

**3. `dsd check symbols` is NOT green, and that is not being laundered into a
pass.** It carries **9 pre-existing errors** present on untouched `main`. This
PR resolves **0** of them and adds **0** new ones — `rombuild`'s intact-TU gate
reports `dsd modules PASS, zero new symbol errors`. The 9 are not this change's
defect and are not fixed by it.

## Validator lines expected on this fold

A 44-shard many-to-one fold credits only the delinks range's first symbol and
reclassifies the rest as claimed, so expect `N address range(s) left the
byte-verified set` to be exactly **one higher** than the byte-verified function
delta — the TU's own new range re-enters the set as the 44 shard ranges leave —
plus `N more function(s) now claim a match that nothing compiles` and a negative
`Byte-verified functions` / `code bytes` line. Locally this measures
`0 changed, 0 lost`. A local green does not predict the validator's credit line
on a many-to-one fold; `changed` is explicitly not a merge blocker, `lost` is,
and it is 0.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


andrewboudreau:
Advisory humanizer review for #2410 at **`3e931ef05c60621201384397652bfdc73c2e86cd`**. Recommendation: **corrections needed before approval**.

1. **Repair the eight stale rename-ledger entries.** [These rows](https://github.com/tangosdev/sm64ds-decomp/blob/3e931ef05c60621201384397652bfdc73c2e86cd/symbols/actor_renames.tsv#L1272) still end in Ukiki spellings while the active symbols use daMky_c, producing the [failed public check](https://github.com/tangosdev/sm64ds-decomp/actions/runs/34087300972/job/101633684715). Preserve rename provenance and resolve ledger ownership before using the task's proposed `--fix` command.
2. **Reconcile five contributor reassignments.** The [private report](https://github.com/tangosdev/sm64ds-decomp/pull/2397#issuecomment-5565558444) changes five credits to `andrewboudreau`: ov030 `0211172c` (`OnYoshiTryEat`, tangosdev), `02111b20` (helper, lunavyqo), `021132d4` (`EnterState5`, tangosdev), `02113324` (helper, lunavyqo), and `02113a80` (`EnterState3`, tangosdev). They lack active promoted `path#symbol` overrides. A fresh attribution comparison using tools/base `f327f7b6460e157153eb7fc0749dbbe60dd854f1` **passes: 35 consolidated, zero changed/lost**. That separate result does not reconcile these five address-specific reassignments. Check their lineage, preserve original credit in the authorized attribution lane, and verify both readers.
3. **Update promotion provenance references.** [Actively renamed header citations](https://github.com/tangosdev/sm64ds-decomp/blob/3e931ef05c60621201384397652bfdc73c2e86cd/include/daMky_c.h#L48) point to nonexistent method shards. Reference the promoted TU and symbol, or immutable historical evidence.

There is a small, concrete source-quality experiment: [newly converted `EnterState0`](https://github.com/tangosdev/sm64ds-decomp/blob/3e931ef05c60621201384397652bfdc73c2e86cd/src/actors/daMky_c.cpp#L2040) still casts `this` to `int*` and writes `p[237]`, although the header models that word as `mState`. Test `mState = 0; return 1;` with exact byte and relocation proof. This review has not established equivalence.

Inherited header assertions about unevidenced matrix padding and positions being unread or written only once are contradicted by [matrix use](https://github.com/tangosdev/sm64ds-decomp/blob/3e931ef05c60621201384397652bfdc73c2e86cd/src/actors/daMky_c.cpp#L659) and [position updates](https://github.com/tangosdev/sm64ds-decomp/blob/3e931ef05c60621201384397652bfdc73c2e86cd/src/actors/daMky_c.cpp#L492). Correct the prose without inventing original field names. Those are inherited documentation issues, not new byte defects.

Keep the scope precise: 20 named method symbols across 19 C++ definitions, including D1/D0 from one destructor, plus 24 remaining free functions. Eleven remaining functions are documented PMF targets; thirteen are not proven members. `EnterStateN` is disclosed coinage. The out-of-line destructor/defer-codegen choice has recorded ordering evidence; preserve it unless a replacement passes equivalent proof. Remaining shadow dispatch receivers and raw fields are bounded follow-ups.

The private **SUCCESS**, completed 2026-09-07 06:00:44 UTC, coexists with the failed public ledger and affected-consumer totals of 4 BLIND and 1 DRAFT. This review found no new demonstrated byte or ABI defect in the TU. Source review and fresh attribution are not a new compilation/full-ROM run or v2 verifier acceptance. `fleet-0907` should assign the corrections through #2410, then obtain independent verification tied to the new exact head and tested base.

andrewboudreau:
Superseded by #2430 — closing, with the branch preserved.

`cpp/daMky_c-v2` (the head of #2430) **descends from `3e931ef05c60621201384397652bfdc73c2e86cd`**, this PR's head, so every commit and every byte of this work is contained in #2430; `git merge-base --is-ancestor 3e931ef0 <#2430 head>` returns 0. Nothing here needs rescuing and nothing is being rewritten under a different author.

What #2430 adds on top of this PR:

- The coined spelling `Ukiki` is retired in favour of the name the cartridge holds. ov030 `0x02115b78` carries the `_ZTS` payload `7daMky_c`; the string `Ukiki` appears nowhere in the ROM.
- `symbols/actor_renames.tsv`: the 8 rows that still asserted the retired spelling at those addresses are pointed at the live names, reproducibly (`check_rename_ledger.py --fix` output, byte for byte).
- Stage 3b converted 11 further members into real `daMky_c::EnterState*` methods, 9 → 20 methods of 44.
- An independent verification (queue task `issue-2410`, verify stage) and a validator run on the current head against a base that contains #2425.

**Attribution is carried, not reset.** `attribution.json` gains 54 override rows keyed both by the old `_ZN5Ukiki*` and the new `_ZN7daMky_c*` symbol names, so credit to **@ruspecial**, **@lunavyqo**, **@andrewboudreau** and **@tangosdev** survives the file move and the rename. `tools/prepush_attribution.py` on the composed result reports *35 consolidated with credit intact, 0 changed, 0 lost*.

The branch `cpp/Ukiki-tu` is **not** being deleted — it stays at `3e931ef0` as the recorded input of the v2 task.

PR #2398, opened 2026-09-07
ov002: promote da1up_c into a single C++ TU, under the ROM's class name

andrewboudreau:
## TL;DR for a CS generalist

SM64DS's 1-Up / Mega Mushroom actor was scattered across 36 separate one-function
source files, and the project called its class `OneUpMushroom` — a name somebody
invented years ago. This PR folds those 36 files into one real C++ translation
unit, proves byte for byte that the compiler turns the merged file back into
exactly the machine code on the cartridge, and renames the class to `da1up_c`,
which is what the cartridge itself calls it: at `0x02108370` in overlay ov002 the
bytes literally spell `7da1up_c`, Itanium-ABI length-prefixed. The word
`OneUpMushroom` appears nowhere in any executable image, in any encoding.

Nine of the 36 functions become genuine `da1up_c::` methods. **The other 27 stay
plain C-style functions, and that is a deliberate, reported result, not a
shortfall.** The cartridge proves 14 of them are members and even proves which
dispatch slot each occupies, but it records no identifier for any of them, and
this PR invents none. The ceiling here is **name recovery, not codegen and not
scope**: every member compiles byte-neutrally and the licensed range is whole.

---

## What changed in this update

This is the integration pass. The reconstruction itself was independently
verified three times and is unchanged; what is new is that it is now composed
onto current `main` and its bookkeeping is reconciled.

**Composed, not rebased.** `origin/main` moved **six times** while this sat in
the integration lane — the ov081 `daGmch_c` promotion, the ov102 `daBmb_c`
promotion (#2434), the ov030 `daMky_c` promotion (#2430), the ov002 lava batch
(#2435), and two automated progress refreshes. Each was merged in with `--no-ff`, so every commit on this
branch keeps its identity and its author. The head branch was **fast-forwarded**
at every step, never force-pushed: the accepted candidate `bb5bbfae6` and every
previous PR head are strict ancestors of the current one, so all earlier commits
and their authorship survive untouched.

**Base drift, re-measured rather than inherited.** A drift count ages every time
main moves, and this one aged five times. The handoff's original "21 files, only
2 under `src/`/`config/`/`include/`" was stale by construction; the successive
re-measurements 79/48, 123/88 and now **177 files, 140 under those roots** are
each superseded in turn, and the note now says to re-measure the figure rather
than quote it. The conclusion survives every re-measurement: main touches **0**
ov002 files, and its `include/decl_common.h` hunks land at lines 517, 704, 752,
1565, 2817 and 2909 while this change's single hunk removes the
`_ZTV13OneUpMushroom` declaration at line 510 — the nearest, `daMky_c`'s at 517,
is seven lines clear and git merged the two with no conflict.

**Append logs resolved as a union, then audited rather than trusted.** No
`--ours`/`--theirs` on any whole file. `attribution.json`,
`config/converted-baseline.json`,
`config/converted-backslide-exceptions.jsonl`, the two `config/match_*.jsonl`
logs and `symbols/actor_renames.tsv` each reproduce
`(base + head-added + main-added) − (head-removed + main-removed)` exactly, with
**zero records missing and zero extra**. `attribution.json` reconciles as 3023 base + 43 added here + 60 added by
`daMky_c` − 2 removed here = **3124**; `daMky_c` also placed its rows beside its
own class's rows, so the two blocks did not collide. `notes/cpp-tu-current-state.md`
was **regenerated** with `cpp_tu_state.py --write-note` rather than hand-merged.
The final remerge absorbed **#2435**, the ov002 lava batch, which is the one
that shared ov002 `delinks.txt` and `symbols.txt` with this branch rather than
just the ledgers. It merged **positionally, not semantically**, exactly as
predicted: its ov002 edits begin at `0x020b5734` while this branch's span
`0x020aee40..0x020b0530`, and the two diffs share **zero changed lines**. Of the
54 paths main moved in that batch, **45 are blob-identical** in the composed
tree and the only 9 that differ are the shared bookkeeping files both sides
legitimately append to. Nothing main deleted was resurrected and nothing main
added is missing.

`notes/data/tu-promotion-queue.tsv` was the one genuine *semantic* conflict: the
base carried unfolded `Ukiki` (ov030) and `OneUpMushroom` (ov002) rows, main
folded the first into `daMky_c` and this branch folded the second into `da1up_c`,
so each side's replacement was taken in the slot its original occupied — 267
rows, both superseded rows gone, neither class duplicated. Every one of the 177
paths main moved that this branch does not own is **blob-identical** in the
composed tree.

---

## Proof block

All figures below were measured on this composed tree, not carried over.

```
head ROM build      106/106 modules exact, 100.000000% of compared bytes
                    source-built functions 11,192, mismatching 0
                    source-owned data claims 20 (reproducing 20)
                    ROM-build analysis: PASS
base ROM build      106/106 exact, 11,192 source-built, mismatching 0
validate_merge      Contributor credit 0 added, 0 changed, 0 lost
                    Byte-verified functions 11,164 / 11,347 (+0)
                    Byte-verified code bytes 2,126,760 (+0)
                    Module fidelity 106/106 exact
                    Port reference check 423 checked, 0 stale
premerge_check      8/8 static gates pass on BOTH base and merge tree
                    source-coverage entries 9066 -> 9031 (-35, consolidation;
                    bytes flat)
romdata_check       696 VERIFIED / 220 PARTIAL / 4 DIFFERS   (base: 695/219/4)
                    0 new DIFFERS; net +1 verified symbol, +12 verified bytes
                    0 new DIFFERS; _ZTV7da1up_c and _ZTI7da1up_c both VERIFIED
linkcheck by symbol 38/38 VERIFIED, 0 BLIND, 0 WRONG
prepush_attribution exit 0 -- 28 consolidated, 0 changed, 0 lost
check_references    exit 0, no new unresolvable references
check_rename_ledger exit 0, 2,006 mangled/vtable rows agree with symbols.txt
check_dead_refs     exit 0, no new dead references, no broken links
langmode ratchet    PASS
port_refcheck       423 checked, 0 stale
```

`linkcheck` is driven **by symbol**, not by file stem. `prepush_linkcheck` keys a
file by its stem and so returns `NO-SYM` — a warning, not a verdict — on a
multi-symbol promoted TU. Driving `linkcheck.py` once per symbol is what actually
links and byte-compares every relocation slot in
`[0x020aee40, 0x020b0530)` plus both registry factories.

### The one symbol that leaves the verified set is a rename, not a loss

`romdata_check` shows `_ZTV13OneUpMushroom` leaving and `_ZTV7da1up_c` arriving.
Anchored on `(module, addr, bytes)` — `(ov002, 0x021083c8, 124)` — the record is
present and exact on **both** sides. The net movement is `_ZTI7da1up_c`, a new
12-byte record, for +1 verified symbol and +12 verified bytes.

Two workarounds were available and both were refused: **no** alias row was
re-added to `config/arm9/overlays/ov002/symbols.txt` (the coined `_ZTV`
row is gone and stays gone), and **nothing** was made to consult
`symbols/actor_renames.tsv`. No file under `tools/` differs from main.

---

## Attribution — what the gate proves, and what it does not

Folding 36 shards into one file left eight functions whose credit belonged to
someone other than the person who ran the fold. Ten override rows land here for
those eight shards, placed **beside `da1up_c`'s own retired-shard rows** rather
than appended at the tail of the object, because tail appends collide.

Two of the eight are members this promotion renamed, and they need **both**
mangled spellings, because the two readers key these rows differently:
`validate_merge` builds `path#symbol` from the built report and looks up the
**new** spelling, while `prepush_attribution` consults its map with each deleted
shard's **basename**, which carries the **old** one. Values are written as the
base handle **verbatim** — row 8 reads `alexsobolew7`, not the canonical
`ruspecial` — because `prepush_attribution` string-compares with no alias
resolution while `validate_merge` applies the map afterwards. Writing the
canonical name would pass today by accident and read as CREDIT CHANGED later.
Counting people rather than handles: `ruspecial` ×2, `lunavyqo` ×1, `tangosdev`
the other five.

**The gate passed and no contributor's totals moved.** That is stated as two
separate measured facts. `validate_merge` reports `0 added, 0 changed, 0 lost`,
and its per-contributor table is identical base-vs-head for **all 13
contributors** — same function count and same byte total for every one, 11,279
attributed functions and 2,168,780 bytes on each side.

It is deliberately **not** stated as "credit was preserved in the contributor
chart". `chaos_db_ci.py:611` resolves overrides as `overrides.get(src_path)`, a
whole-path lookup, so every `path#symbol` row is accepted and then never read
when `contributions.json` is regenerated. That is not specific to this PR: the
composed tree carries **2,387** such member rows, 2,344 of which were already on
main. The 43 added here behave exactly like the 2,344 already there. Closing that
gap is a tooling change and does not belong in a class reconstruction PR.

A related defect is filed as **#2433**: `prepush_attribution` keys overrides on
the vanished shard basename while `validate_merge` and `tu_promote` key them on
the surviving symbol.

---

## Class identity — all three records are ROM-read

| record | address | module |
|---|---|---|
| `_ZTS7da1up_c` | `0x02108370` | ov002 |
| `_ZTI7da1up_c` | `0x0210837c` | ov002 |
| `_ZTV7da1up_c` | `0x021083c8` | ov002 |

`0x02108370` holds `37 64 61 31 75 70 5f 63 00` = `7da1up_c\0`, read directly out
of the extracted ov002 image, not from a generated index. `_ZTI7da1up_c` is the
matching `__si_class_type_info`; its +8 base word reaches `_ZTI12dEnemyBase_c` at
`0x021081c0`, so the ROM states the base class directly.

The `_ZTV` symbol IS the address point: storage runs `0x021083c0..0x02108444`
(33 words = a two-word RTTI prologue plus 31 slots), payload `0x7c`. It ends
exactly where `_ZTS7daBar_c` begins, so there is no trailing-null ambiguity.

**`new`-size literal `0x398` (920).** Both factories —
`da1up_c_classInit_ONEUPKINOKO` (`0x020b0580`, profile id 276) and
`da1up_c_classInit_SCALEUP_KINOKO` (`0x020b0530`, profile id 277) — execute
`mov r0,#0x398`, call `fBase_c::operator new` and store this vtable. One class,
two registry profiles.

### Nine overridden slots, read three independent ways that agree

| slot | address | member |
|---:|---|---|
| 0 | `0x020b01c0` | `InitResources` |
| 3 | `0x020affe8` | `CleanupResources` |
| 6 | `0x020b00e8` | `Behavior` |
| 9 | `0x020b0070` | `Render` |
| 12 | `0x020b006c` | `OnPendingDestroy` |
| 16 | `0x020aee40` | `~da1up_c` (D1, also this TU's key function) |
| 17 | `0x020aee88` | `~da1up_c` (D0) |
| 18 | `0x020af3a0` | `OnYoshiTryEat` |
| 19 | `0x020af2b0` | `OnTurnIntoEgg` |

**Five of these nine were declared NON-virtual in the old header**, and the
consequence was not cosmetic: the inherited `dActor_c` addresses were silently
standing in five slots of the vtable this class's key-function TU now emits.

---

## The wall: 9 of 36 converted, and it is a NAMING wall

The 27 unconverted members split cleanly, and the split is the point:

- **14 have ROM-proven member-ness AND a proven dispatch index.** The 14
  `{function pointer, 0}` descriptors at `0x02108300..0x02108370` are
  pointer-to-member-function objects with a zero `this` adjustment, and
  `__sinit_ov002_02100adc` copies them into the 14-element dispatch array at
  `0x0210dc00` that `Behavior` indexes. The ROM proves *that* they are members
  and *which slot* each occupies — and records **no identifier for any of them**.
- **13 have no member-ness evidence at all.**

Converting either group would mean inventing 27 names and pushing them into
`symbols.txt`. This PR invents none.

---

## Scope of the diff

59 files differ from `main`: 36 shard deletions, the new
`src/actors/da1up_c.cpp`, a header swap, the ov002 config and manifest, the two
factory files, `include/decl_common.h`, and the shared ledgers. No ov081, ov102, ov030
or any other class's artefact is touched: of the 177 paths main moved, every one
this branch does not own is blob-identical in the composed tree.

---


andrewboudreau:
Advisory humanizer review for #2411 at **`f38a8f74f584658e02e0ab4ae435a0fa2e949c5f`**. Source-quality recommendation: **corrections needed**. Introduced changes were compared with the actual PR merge-base `b2bd6a323832d244a12e750713b1405b26a94c66`.

1. **Resolve the new non-void fallthrough (P1).** [`func_ov002_020af684`](https://github.com/tangosdev/sm64ds-decomp/blob/f38a8f74f584658e02e0ab4ae435a0fa2e949c5f/src/actors/da1up_c.cpp#L500) changes from `void` to `int`, but never returns a value; its last callee is also declared `void`. Callers consume the supposed result at lines 336/339 and elsewhere. The comment explicitly relies on residual `r0`. Falling off the function does not establish a valid C++ return contract. The manifest records that reverting moves bytes; preserve that measurement, then trace the real return use and probe consistent declarations/control flow. An arbitrary return value or warning suppression would not fix the cause. No replacement has been byte-proved by this review. Other pre-existing fallthroughs are separate debt.
2. **Reconcile eight contributor reassignments (P1).** The [private report](https://github.com/tangosdev/sm64ds-decomp/pull/2398#issuecomment-5565596161) changes eight credits to `andrewboudreau`, at ov002 addresses `020aee88`, `020aefa4`, `020aefb8`, `020af724`, `020af950`, `020afc68`, `020aff10`, and `020b00e8`. These lack promoted per-symbol overrides; previous authors include tangosdev, ruspecial and lunavyqo. A fresh pinned local attribution comparison against main `f327f7b6460e157153eb7fc0749dbbe60dd854f1` **passes: 26 consolidated, zero changed/lost**. Those eight are absent from its consolidated results, so that pass does not resolve the private warning. Preserve original credit using verified lineage and the authorized attribution writer; remove the PR body's instruction to judge only “lost” until the difference is reconciled.
3. **Correct the claimed virtuality fix and conversion scope (P2).** The [new header explanation](https://github.com/tangosdev/sm64ds-decomp/blob/f38a8f74f584658e02e0ab4ae435a0fa2e949c5f/include/da1up_c.h#L68) says five methods were non-virtual because their declarations omitted `virtual`. They already override matching inherited virtual functions. Adding the keyword does not establish the claimed five-slot repair. Correct the header, facts and PR explanation. All nine mangled functions already had native member definitions at the PR base; this change consolidates and renames them. RTTI proves `da1up_c`, not the claimed nine original ROM method names. Missing original names also do not establish a final ceiling on further reconstruction.
4. **Correct the new “unused padding” fact (P2).** The [facts entry for `0x38c`](https://github.com/tangosdev/sm64ds-decomp/blob/f38a8f74f584658e02e0ab4ae435a0fa2e949c5f/notes/data/class-facts/OneUpMushroom.json#L146) says no access was observed, but [Behavior](https://github.com/tangosdev/sm64ds-decomp/blob/f38a8f74f584658e02e0ab4ae435a0fa2e949c5f/src/actors/da1up_c.cpp#L1137) increments/resets its 16-bit value; other functions read and initialize it. The inherited byte array is not evidence of a new layout break. Correct the fact now; a typed field replacement needs separate consumer and byte proof.

The real base/subobjects, compiler-generated destructor and explicit text-only ownership are useful progress. The inherited PMF placeholder, fake model vtable and offset-based helpers remain focused follow-up opportunities, subject to measured compiler constraints.

The private gate remains **FAILURE**, completed 2026-09-07 06:12:00 UTC, for lost exact data identity `ov002:_ZTV13OneUpMushroom`; coordinate the rename-aware cause with #2409. Its 36 verified TU functions and 106/106 exact modules are narrower evidence than a passing overall gate. Affected-consumer totals still include 4 BLIND and 1 DRAFT.

This is source review plus a fresh attribution check, not a new compiler/full-ROM run or v2 verifier acceptance. `fleet-0907` should assign the corrections through #2411 and obtain fresh independent proof of the resulting immutable candidate.

andrewboudreau (commented):
Source-review refresh at current head `19df96cc21d5e2a87a97a22d4de67191319505bd`: the OneUp TU, header and class-facts file are unchanged from the previously reviewed `f38a8f74f584658e02e0ab4ae435a0fa2e949c5f`. The non-void helper fallthrough, false virtuality/conversion claims, live storage described as unused padding, and inherited fake rendering/PMF interfaces therefore still need the recorded corrections.

The latest published private validation now reports Passed, 106/106 exact modules, and zero changed/lost contributor credit. That supersedes the older failing private status quoted in the original review; do not keep reporting that historical failure as current. It does not resolve the source findings. The affected relocation report still distinguishes its BLIND/DRAFT rows from VERIFIED rows.

Continue the existing correction under #2411 and independently verify the final composed candidate. This PR remains held for source acceptance; no class restart or mass rebase is needed.

andrewboudreau (commented):
This one's been overtaken. #2512 landed the same promotion on main on the 9th, and it went further than this branch does.

I went through both versions of `src/actors/da1up_c.cpp` side by side. They agree on all the deletions — the old `include/OneUpMushroom.h`, the `_ZN13OneUpMushroom*` shards, the `func_ov002_*` files are gone on both sides. Where they disagree, main is the one to keep:

`Render()` here casts `&mModel` to a hand-written six-slot `struct Obj` and dispatches through slot 5. Main just calls `mModel.Render(0)`. That's the same call with the real type on it.

`mStateTimer` is spelt `*(unsigned short*)(c + 0x38c)` in three places here, and once as `*(unsigned short*)((char*)this + 0x38c)`. Main uses the named member throughout.

`func_ov002_020af684`, `func_ov002_020afa6c`, `func_ov002_020af924` and `da1up_c::OnTurnIntoEgg(Player&)` return `int` here and `void` on main. That wasn't drift — correcting the `OnTurnIntoEgg` contract to `void` is half of #2512's title, so main's spelling is a deliberate fix of this one.

The only thing on this branch that isn't on main is the comment explaining that ordinal 33 reaches the model's vtable by offset rather than by name, slot 5. It's a real observation, but it exists to justify the `struct Obj` cast, and main doesn't do the cast any more, so there's nothing for it to explain.

So I don't think there's anything here to rescue before it closes. The branch is also conflicting against main now, and rebasing it would just be re-resolving in favour of main every time.

Recommend closing as superseded by #2512.

andrewboudreau (commented):
Reviewed for supersession against main `99c676b8a` and the preserved repair candidate `9c03a19a7480548cac5aad7fa9fa53405cb2fe2c`.

The source and contract repairs landed through #2512 at `3fa6ed95901e0e4080f5d087618372e84ac8cdb1`. Independent comparison accounted for all 122 changed paths: all 109 source/header/port/config paths apart from the shared generated conversion baseline are identical, including the 37 retired files. All candidate attribution overrides remain with the same credit. The remaining differences are shared/generated bookkeeping.

Closing this older PR as superseded preserves the existing merge and attribution. This does not approve the obsolete source at this PR's head. Remaining class reconstruction stays open under #2505. The stale OneUpMushroom promotion-queue row is being repaired separately.

PR #2400, opened 2026-09-07
Record the stopped 2026-09-07 wave and the blocker that stops all five

andrewboudreau:
The class run was stopped deliberately, to upgrade the workflow. Five classes
were promoted under the ROM-name ruling and every branch is pushed -- nothing is
uncommitted and nothing is local-only. This writes down where each one stopped,
so the next session reads it from the repo rather than rediscovering it.

184 shards folded into 5 TUs; 110 members became real C++ methods. `daMip_c`
converted 32 of 32 and `daGmch_c` 36 of 37 (the 37th is the factory, free by
design). The wall on every unconverted member is naming or scope -- never codegen.

**The blocker is one line of the validator, and it stops all five.**
`validate_merge.rom_data_regressions` diffs raw `(module, symbol)` sets with no
rename awareness, so retiring a coined `_ZTV<Coined>` reads as a lost symbol.
Nothing is lost: `_ZTV8daGmch_c` verifies at the same module, address `0x02128c04`
and `0x84` bytes as the `_ZTV8Moneybag` it replaces. Every promotion under the
ROM-name ruling trips it, which makes that ruling unmergeable as things stand.

The note records the fix that does not weaken the gate -- anchor on the ROM
address, since the new name must actually byte-verify there to qualify -- and the
two workarounds that must not be taken, both of which let a PR certify its own
rename. It also records the sequencing constraint: the validator restores all of
`tools/` from the base commit, so the fix cannot ride along in a class PR.

One concrete cleanup is named too. `cpp/BobOmb-tu` carries a commit banking a
langmode `extern_vtable` rise that a later run proved is not byte-necessary --
moving the vptr declaration into the class header restored the old counts with
byte-identical objects, confirmed against the pinned mwccarm.

Notes-only; `check_dead_references` 0. No `src/`, `config/` or `include/` file is
touched, so no byte can move.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01G1Gdj5fTjMsW2VjRLpXxYA

### DIFF
--- a/notes/agents/IN-FLIGHT.md
+++ b/notes/agents/IN-FLIGHT.md
@@ -60,6 +60,59 @@ left where only a branch listing can find it.
 Two of the older ones carry inline-dtor flips that exist nowhere else. Audit before
 deleting anything in this set.
 
+## The 2026-09-07 wave, stopped mid-flight
+
+Five classes were promoted under the ROM-name ruling and the run was stopped
+deliberately, to upgrade the workflow. **Nothing is uncommitted and nothing is
+local-only** -- every branch below is pushed.
+
+| class | overlay | bytes | methods | where it stopped |
+|---|---|---|---|---|
+| `daGmch_c` | ov081 | 37/37 | **36/37** | PR #2395, CI green, validator red |
+| `daMky_c` | ov030 | 44/44 | 20/44 | PR #2397 |
+| `da1up_c` | ov002 | 36/36 | 9/36 | PR #2398 |
+| `daMip_c` | ov085 | 32/32 | **32/32** | gated, **no PR** -- `cpp/Rabbit-tu` |
+| `daBmb_c` | ov102 | 35/35 | 13/35 | gated, **no PR** -- `cpp/BobOmb-tu` |
+
+184 shards folded into 5 TUs; 110 members became real C++ methods. The wall on
+every unconverted member is **naming or scope, never codegen**.
+
+### The blocker that stops all five
+
+`validate_merge.rom_data_regressions` diffs raw `(module, symbol)` sets and has no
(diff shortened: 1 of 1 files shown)


andrewboudreau:
Closing as superseded — its central claim has been overtaken by events, and landing it now would write a false statement into the tracked record.

**The blocker it names is fixed and on main.** `rom_data_regressions` now anchors the ROM-data diff on `(module, address, bytes)` instead of the symbol name, so a rename reads as a rename. That landed as #2425 (`1c93d2663`), which is exactly the address-anchored fix this PR's own body recommends. Neither forbidden workaround was used.

**The headline was also measurably wrong before that.** This says "the blocker is one line of the validator, and it stops all five." It stopped one PR. `daMky_c` validated clean — 686 symbols exact, no lost-symbol line — and its red gate was `check_rename_ledger`, eight historical rows still asserting the retired coined spelling. `daGmch_c`'s failure was a prediction at the time of writing, later confirmed as a result; `da1up_c`'s was the only confirmed one. Independent verification of `daGmch_c` has since re-run the gate with the fix on main and found the defect bucket empty.

**The file it edits is already superseded.** `notes/agents/CUTOVER-0907.md` is on main and supersedes the operational parts of `IN-FLIGHT.md`, and #2428 is open to correct it further with what the lane scouts measured.

Nothing factual is lost. The per-class conversion figures this PR records — 184 shards folded into 5 TUs, 110 members becoming real methods, `daMip_c` at 32/32 and `daGmch_c` at 36/37 with the factory free by design — are being folded into #2428, and each class's detail lives in its committed handoff under `notes/agents/handoffs/` and on its own issue.

PR #2429, opened 2026-09-07
Let an EVIDENCE-BACKED merge through the denominator gate

andrewboudreau:
The split rule that landed in #2392 says merges stay blocked, and gives the reason: a merge lowers `totalFunctions` and so **raises** the headline, which is the direction an attack uses, and no arithmetic over the symbol snapshot can separate a true merge from a false one.

That is still true. This PR does not find a cleverer sum — it stops looking in the symbol table.

## The separating fact is in `delinks.txt`

A range carrying `complete` is compiled, linked into its module and byte-compared against retail. Everything else is filled by a gap object holding the ROM's own bytes, so it is exact by construction and proves nothing — `verification_split` says this in as many words.

So the rule is: **every address the merge removes must be newly covered by a `complete` range, and `sourceBytes` must rise by exactly the extent of those new ranges.**

That is not forgeable in the shape the first draft died on:

- A hundred junk symbols merged into one produces no new `complete` range, because nothing would compile.
- Forging `complete` over a range that does not reproduce fails module fidelity **in the same validator run**.
- Borrowing an unrelated new range does not work either — the evidence has to *cover* the merged addresses, not merely exist in the diff.

**Nothing matched may leave.** A merge is licensed to absorb an ASM stub or a severed epilogue, never a function someone had already matched. Surviving matched records keep their sizes, with one exception: a survivor may grow, and only into the new `complete` range.

## The live case

Measured on the port side, not argued here. `__destroy_arr` at `0x0207328c` is declared `0x5c` in `config/arm9/symbols.txt` while the ROM's own `.exceptix` record gives `0x74`. The trailing `0x18`, carried as `func_020732e8` with its own src file, is not a function at all — it is the catch handler.

Matched as real C++ with try/catch: `.text` 116, `.exception` 20, `.exceptix` 12. `.exception` byte-identical with no relocations; `.text` and `.exceptix` differ only at relocation slots and every slot resolves. `match.py` reports MATCHING; `prepush_linkcheck` reports `__destroy_arr` VERIFIED.

| | totalFunctions | totalBytes | sourceBytes |
|---|---|---|---|
| base | 11,347 | 2,211,124 | 2,126,548 |
| head | 11,346 | 2,211,124 | 2,126,664 |

`sourceBytes` **+116 = 0x74 exactly**. Those are the first test in the suite.

## Two premises for why this could not land, both wrong

Recorded in the commit message so nobody re-derives them:

- The global `-Cpp_exceptions off` in `tools/rombuild.py` does **not** block it — the in-file `#pragma exceptions on` overrides it.
- `notes/rom-build.md`'s rule that a file emitting `.exception`/`.exceptix` cannot enroll does **not** fire — `eligible.classify` issues no rejection.

The real blocker is neither, and is not in this file at all: `objisolate.isolate` drops those sections to size 0 for any single-function source (`tools/objisolate.py:1912` — the legacy rule keeps the single kept function's section and nothing else). The multi-section shape is the `config/tu_manifest.d` path, so the range has to be registered as a TU source first. **That is the port side's next step, not this gate's** — and if `isolate` still drops the sections, this rule simply never fires on that case.

## Tests

Ten new, and the refusals are the ones worth reading: no new coverage; a hundred junk symbols; a hundred junk symbols borrowing an unrelated new range; absorbing a matched function; `sourceBytes` short because an unrelated range left under cover of the merge; and a range that was already `complete` in base.

**69/69 pass.**


andrewboudreau:
## Blocking: `classify_merge` accepts a range that reproduces *vacuously*

First, the good news this PR is built on: **#2392 is confirmed working end to end.** #2360's head now passes all 13 checks with `Symbol-table split | totalFunctions +1 over identical covered bytes`, on a base that includes #2392. Nothing below disputes the stacked base.

### The property that changed

#2392's safety argument is *structural*: a split can only ever lower the headline, so it needs no other gate to be working. This PR replaces that with a *compositional* one — "a new `complete` delinks range means the ROM build compiled and byte-compared it, and forging one fails module fidelity in the same run."

The forging half is correct. Module fidelity is a blocking reason, so a range that fails to reproduce is caught.

But that answers the wrong attack. The problem is not producing a range that **fails** to reproduce — it is producing one that reproduces **vacuously**.

### The path

`asm_policy.classify` returns `None` (clean) as soon as `HAND-ASM PRIMITIVE` appears anywhere in the file. So a `dcd`-word transcription of a span carrying that banner is:

- not in `new_transcribed`, so no blocking reason;
- not stamped `RAW-ASM` by `pr_linkcheck`;
- counted as **matched**;
- byte-identical to retail by construction, so module fidelity passes.

Enroll it `complete`, delete the interior symbols, and `classify_merge` has every piece of evidence it asks for. This is not hypothetical about the banner: `classify` never applies the objective test behind it, and a large share of existing uses in-tree are boilerplate rather than genuine hand-written primitives.

### Measured, against this PR's head, using its own test helpers

```
100 junk symbols WITH a covering new complete range ->
  {'kind': 'merge', 'functionDelta': -99, 'sourceByteDelta': 400, 'removed': 99 removed}
```

That is `test_a_hundred_junk_symbols_cannot_borrow_an_unrelated_new_range` with one change: the range **covers** the merged addresses instead of sitting elsewhere.

The suite refuses the unrelated-range variant and never tries the covering one — which is the only variant an attacker would write, and the one a `dcd` transcription produces for free. Denominator −99, numerator +1, headline up, no decompilation performed.

### The fix looks cheap

`classify_merge` takes no source-classification input at all, and the missing input is already computed a few dozen lines away in `build_report`. Either of these would close it:

- require each new `complete` range's source to be real compiled C/C++ — `AP.classify(text) is None` **and** no `HAND_BANNER`; or
- require its `link` row to be `VERIFIED` and not `RAW-ASM`.

Please also add the covering-range case as a test, since its absence is what let this through.

### Two smaller notes

**The 10 new tests do not run when the file is executed as a script.** `class MergeEvidence` is appended *after* the existing `if __name__ == "__main__": unittest.main()` block. Measured:

- `python tools/test_validate_merge.py` → **Ran 59 tests**
- `python -m unittest tools.test_validate_merge` → **Ran 69 tests**

CI uses the `-m unittest` form (`tool-tests.yml:219`), so this is latent rather than live — but this repo has already been bitten by a test module that silently collected zero tests. Moving the class above the `__main__` guard fixes it.

**The docstring now contradicts itself.** `classify_repartition` still opens a paragraph with "MERGES STAY BLOCKED, AND THAT IS DELIBERATE" (line 383), directly below a `classify_merge` that unblocks them. These docstrings *are* this file's normative argument, so that header needs rewriting, not just the paragraph beneath it.

### Not a blocker for #2360

`classify_merge` is not on #2360's path — `classify_merge` is only consulted when `totalFunctions` decreases, and #2360 is +1. #2392 alone is sufficient there, as the green run proves. This PR should not be treated as gating it.

andrewboudreau:
Pushed `f176934bb` (fix `f9bdc8a58`, plus a `--no-ff` merge of `origin/main` at `cccae7692`). This closes the hole reported in https://github.com/tangosdev/sm64ds-decomp/pull/2429#issuecomment-5572824406 without giving up what the PR is for.

## The hole, reproduced first

`classify_merge` accepted a new `complete` delinks range as proof the fold was really decompiled. The argument was that the ROM build compiles, links and byte-compares such a range, and that forging one fails module fidelity in the same run.

The forging half is right. It just does not cover a range that reproduces **vacuously**. `asm_policy.classify` returns `None` as soon as `HAND-ASM PRIMITIVE` appears anywhere in the file, so a `dcd`-word transcription under that banner:

* is not in `new_transcribed` (no blocking reason),
* is not stamped `RAW-ASM` by `pr_linkcheck` (`pr_linkcheck.py:299` asks `AP.classify(text) == "transcribed"`),
* counts as **matched**,
* and byte-compares against retail **by construction** -- the words typed into the file *are* the ROM's words -- so module fidelity passes too.

Enroll it `complete`, delete the interior symbols, and every condition the gate checked was satisfied with nothing decompiled.

Measured on `660900ee3` with this PR's own test helpers -- `test_a_hundred_junk_symbols_cannot_borrow_an_unrelated_new_range` with the range moved to **cover** the merged addresses instead of sitting elsewhere:

```
100 junk symbols, covering new complete range, sourceBytes +400
-> {'kind': 'merge', 'functionDelta': -99, 'sourceByteDelta': 400, 'removed': 99 removed}
```

The suite only ever tried the unrelated-range variant, which is why this got through.

## Before / after

| case | `660900ee3` | `f176934bb` |
|---|---|---|
| 100 junk symbols, **covering** new range, transcribed source | `merge`, `functionDelta -99`, `sourceByteDelta 400` | **REFUSED** (`None`) |
| the live `__destroy_arr` merge, compiled source | `merge`, `-1`, `+0x74` | `merge`, `-1`, `+0x74` |
| the same 100-symbol fold, if its source really were compiled C | `merge`, `-99` | `merge`, `-99` (unchanged -- a genuinely compiled 400-byte range *is* a real merge) |
| #2360's split shape | `split`, `+1` | `split`, `+1` |

## The fix

`classify_merge` took no source-classification input at all, and the missing input was already computed a few dozen lines away in `build_report`. It now takes it, and refuses any newly-`complete` range whose source is not real compiled code.

Of the two candidate rules, I implemented the **source-classification** one rather than the linkcheck-verdict one:

* it is the direct negation of the attack, which is a *source* fact (a banner plus `dcd` words), not a relocation fact;
* `changed_text` / `changed_cls` already exist in `build_report`, one block above, so the value is threaded rather than recomputed -- `compiled_src` sits with `new_transcribed` / `new_unbannered` / `stranded_markers` and reads like them;
* the `link` route would refuse legitimate merges: `_link_state` gives a file with no relocations the verdict `NONE`, not `VERIFIED`, and `link_rows` is optional (`pr_linkcheck` need not have run). It is also only a proxy -- `RAW-ASM` is the *same* `asm_policy` call, made indirectly, and it is exactly the call the banner defeats.

One deliberate widening of the stated rule: `compiled_src` excludes **both** banners, not just `HAND_BANNER`.

```python
compiled_src = {p for p, t in changed_text.items()
                if changed_cls[p] is None
                and AP.HAND_BANNER not in t and AP.DRAFT_BANNER not in t}
```

`AP.classify` returns `None` for a `NONMATCHING`-bannered file too, so `classify(text) is None and HAND_BANNER not in text` still admits a `dcd` blob under a `NONMATCHING` banner -- byte-identical by construction in exactly the same way. Excluding `DRAFT_BANNER` costs nothing legitimate: a file that declares it does not reproduce cannot be evidence that the ROM build byte-compared the merged range.

A non-compiled new range refuses the merge outright rather than being skipped, because skipping it would leave its bytes in the `sourceBytes` delta -- the arithmetic that stops an unrelated range joining under cover of the merge.

Known conservatism, documented in the docstring: `compiled_src` covers only sources this merge adds or modifies, so a merge that newly enrolls a file it does not edit is refused. A fold that never touches the source absorbing it is not a shape this carve-out is for.

## The other three

1. **Tests.** Added `test_a_transcribed_source_is_not_evidence_even_covering_the_merge` (the attack above) and `test_a_covering_range_from_compiled_source_still_lands` (the positive: a legitimate merge backed by genuinely compiled source must still classify as `merge`).

2. **The 10 tests did not run as a script.** `class MergeEvidence` sat *below* `if __name__ == "__main__": unittest.main()`. Measured on `660900ee3`: `python tools/test_validate_merge.py` -> **Ran 59**, `python -m unittest tools.test_validate_merge` -> **Ran 69**. CI uses the `-m unittest` form (`tool-tests.yml`), so it was latent rather than live. The class and its `_enr` helper now sit above the guard; both forms report **Ran 71** (69 + the 2 new).

3. **The contradictory docstring.** `classify_repartition` still opened a paragraph with "MERGES STAY BLOCKED, AND THAT IS DELIBERATE" directly below a `classify_merge` that unblocks them. Rewrote the header rather than only the text under it -- it now says merges are not *that* rule's to allow and no arithmetic can allow them, that `classify_merge` is the evidence route, and adds a paragraph arguing the banner exclusion instead of leaving it implicit. (It also said "`classify_merge` below" when `classify_merge` is above it; fixed.)

## #2360 is not affected

`classify_repartition` is untouched and takes no new input -- `classify_merge` is only consulted when `totalFunctions` decreases. Verified directly: #2360's shape still returns `{'kind': 'split', 'functionDelta': 1}`.

## Gates

* `python -m unittest tools.test_validate_merge` -> Ran 71, OK
* `python tools/test_validate_merge.py` -> Ran 71, OK (matching counts)
* full CI tool-test list (35 modules) -> Ran 637, OK (skipped=3)
* `premerge_check --base origin/main HEAD` -> all 8 static gates pass on base and on the merge tree; nothing green -> red

Tools-only change: no source, header, symbol or delinks file moved, so no ROM build is implicated and none was run. The static gates are the whole of the proof here, and they do not speak for the ROM.

andrewboudreau:
Pushed `b7b2c80ed`, plus a `--no-ff` restack onto `origin/main` at `788643caf`. All 9 checks green.

**On your two suggested fixes: only the first one works.** I implemented the source-classification rule. The linkcheck-verdict alternative would not have closed this — `pr_linkcheck.py:299-300` stamps `RAW-ASM` only when `AP.classify(text) == "transcribed"`, and `asm_policy.classify:151` returns `None` as soon as `HAND-ASM PRIMITIVE` appears anywhere in the file. The `link` route is the same `asm_policy` call made indirectly, defeated by the same banner that *is* the attack. It is also optional (`link_rows` may be absent) and gives a relocation-free file `NONE` rather than `VERIFIED`, so it would refuse legitimate merges too.

One widening of your stated rule: `compiled_src` excludes **both** banners, not just `HAND_BANNER`. `classify(t) is None and HAND_BANNER not in t` still admits a `dcd` blob under a `NONMATCHING` banner — vacuous in exactly the same way, and a file declaring that it does not reproduce cannot be evidence that the build byte-compared the range.

**The covering-range test is in, and so is one you did not ask for.** Every `MergeEvidence` test calls `classify_merge` directly and invents the `compiled` set it is handed, so none of them touches the wiring — and the wiring is where this rule can silently die. `compiled_src` is keyed on `diff_snapshot` paths; the ranges are keyed on the paths written in `delinks.txt`. If those spellings ever stopped agreeing, the clause would refuse **every** merge, the carve-out would be dead, and all 71 tests would still be green.

`MergeEvidenceThroughBuildReport` drives one committed base through the real `build_report`, folded two ways:

| fold | `660900ee3` | `b7b2c80ed` |
|---|---|---|
| `dcd` words under `HAND-ASM PRIMITIVE` | `merge`, −1, `sourceByteDelta 116`, reasons `[]` | **refused** — `function/byte coverage denominator changed` |
| real C++ | `merge`, −1, `+116`, reasons `[]` | `merge`, −1, `+116`, reasons `[]` |

The first row is your attack reproduced end-to-end on a real tree, not just at the unit boundary. The second is the `__destroy_arr` shape this PR exists for, unchanged.

Each new test was checked to fail deliberately: clause removed → the two transcription tests fail and nothing else; clause inverted → both positives and both refusals fail; `compiled_src = set()` → **only** the new end-to-end positive fails. That last one is the point — the refuse-everything mode is invisible to all 71 pre-existing tests.

**Script/module counts.** Fixed; `MergeEvidence` and `_enr` moved above the `__main__` guard. Both invocations now report **73**. The CI `tools` job reports **639, OK**.

**Docstring.** `classify_repartition`'s header now reads "MERGES ARE NOT THIS RULE'S TO ALLOW, AND NO ARITHMETIC CAN ALLOW THEM", points at `classify_merge` as the evidence route, and argues the banner exclusion rather than leaving it implicit.

**#2360 unaffected.** `classify_repartition` is untouched, and `classify_merge` is only consulted when `totalFunctions` decreases.

Known conservatism, documented in the docstring: `compiled_src` covers only sources this merge adds or modifies, so a merge that newly enrolls a file it does not edit is refused. A fold that never touches the source absorbing it is not a shape this carve-out is for.

PR #2446, opened 2026-09-07
Consolidate daBombking_c text and use real model and state interfaces

andrewboudreau:
Consolidate BombKing's 52 ov078 functions into `src/actors/daBombking_c.cpp` and use the cartridge's `daBombking_c` class identity. Behavior uses the real `BlendModelAnim::UpdateVerts` and `Advance` interfaces; Render retains the required direct `Model::Render` call. The shadow helper and shared declaration agree on `void`.

Two state handlers now call the local ov078 helper `func_ov078_02123804`. Their former name identified an unrelated MontyMoleRock destructor in ov080 at the same address. Independent relocation checks verify the module as well as the address for all 427 references.

The RTTI name is at ov078 `0x02126e18`, typeinfo at `0x02126e0c`, and its vtable reference at `0x02126e48`. The vtable address point is `0x02126e4c`: 0x7c bytes of slots, or 0x84 bytes including the preceding storage. Factory and field names remain identified as reconstruction choices.

Production owns text only. The TU has seven native definitions producing eight method symbols, plus 44 free functions. Remaining raw actor/player fields, matrix views, ABI helpers and manual construction stay tracked in #2479. This is partial reconstruction; current `int` state dispatch does not establish every original handler signature.

Validation for `48465098684a516e5402ec57454227f018164a4d` against main `df70231d3cf547597b2cb36bed8b1494e0753dbd`:

- Independent source review covers the exact candidate and current base, retaining all ten BombKing findings and the six inherited Snowman findings. The complete accepted source diff is unchanged after the current-main restack.
- Independent class proof verifies all 52 functions, 12,064 text bytes and 427 module-correct references, with no blind or differing results. Metadata remains five VERIFIED records and six PARTIAL records; no data ownership is added.
- The fresh composed ROM build reproduces all 106 modules and 11,199 source-built functions, with 26 owned-data and three BSS claims passing. The last restack changes only four progress files; all compiler inputs are unchanged.
- Local committed-merge validation passes with no coverage or contributor-credit loss. All 52 original function credits are preserved, and all 423 port references resolve.

The full consumer report contains 2,080 function results: 2,075 VERIFIED, four inherited BLIND results and one draft. The five limited results exactly match the independent base control; they remain proof gaps. Global metadata retains four inherited differing records. Reused reports are tied to their original commits and checked for unchanged inputs.

The official private validator reached terminal SUCCESS at 23:31:14 UTC for this exact head/base. All public checks and the fresh live Source review gate passed. Merged as `c63b53e0ebbc0ef29adcd87ec18130a11a6c3c47` at 2026-09-10T23:36:10Z. Details and reconstruction limits are recorded in `notes/agents/handoffs/pr-2446-source-review-fixes.md`, `pr2446-callee-repair-0910.md`, and `pr2446-composition-0910.md`.


andrewboudreau (commented):
Source review of `564d701f0c41cec7101786c0e37345cf2e5b0ddc` requests changes:

- `src/actors/daBombking_c.cpp:1702-1727`: `func_ov078_02125c98` changes from the base's void function to int, but still falls through after void DropShadowRadHeight. Both Behavior callers discard the result. Restore a coherent void contract across definition and declarations, with affected-consumer and byte proof; do not invent a return value.
- The PR describes intact-object ownership although the manifest licenses text only. Distinguish the emitted metadata comparison from production ownership.
- Correct the stale manual-D0 rationale and source-order explanation; they do not describe the final compiler-owned lifecycle form. Distinguish the 0x7c vtable slot payload from the 0x84 full object storage.
- The fake slot-3 dispatch has an existing declared target, BlendModelAnim::UpdateVerts. Probe the real virtual interface. Preserve the direct Model::Render call shape when probing that separate bridge; an unqualified virtual call is not interchangeable.

The published exact ROM/byte results remain valid historical evidence. The return-contract defect and current provenance claims need correction before integration, followed by independent verification of the revised candidate.

andrewboudreau (commented):
Final independent source acceptance and current-main composition are recorded in task `pr2446-final-restack-0910` for `48465098684a516e5402ec57454227f018164a4d` against `df70231d3cf547597b2cb36bed8b1494e0753dbd`. BombKing findings BKG-01 through BKG-10 and inherited SNM-01 through SNM-06 retain their explicit dispositions. Partial reconstruction remains tracked in #2479 and #2476.

The local committed test merge passes. Independent proof verifies 52 functions / 12,064 text bytes / 427 module-correct references, and all original credits are preserved. Full-ROM and consumer evidence is pinned to its tested commits; the final progress update changes no compiler inputs. Four inherited BLIND consumer results, one draft and four global metadata differences remain explicit limitations.

The final private validation result is still required before merge.

PR #2447, opened 2026-09-07
Promote daSanbo_c to a verified class translation unit

andrewboudreau:
Promote the 36 Pokey functions to `src/actors/daSanbo_c.cpp` using the cartridge's `daSanbo_c` identity. The TU uses the real actor, Model, matrix and segment-link interfaces where verified. It corrects the void `OnTurnIntoEgg` contract and removes the render and virtual-call shadows.

The RTTI name is at ov096 `0x02137980`, typeinfo at `0x0213798c`, and vtable storage at `0x021379d0` (address point `0x021379d8`). Typeinfo identifies `dActor_c` as the base. The 36-function run includes both factories and spans `0x02135700..0x02136db0`.

This remains a text-only, partial reconstruction: ten native method symbols, 24 address-named helpers and two manual factories. Remaining raw interfaces and field/state reconstruction are tracked in #2558. Compiler-emitted class metadata is checked separately and adds no production data ownership.

Review also corrected the state-table explanation. The initializer copies twelve PMF constants into six entry/update pairs in BSS, in a different order from the constants. The class facts record the verified mapping and preserve the earlier scout notes as history. The promotion queue now records one TU / 1,146 lines; the 36-shard census describes the inputs before consolidation.

Validation for `ccac1f85144fe1eed327efcf3f15d5bb80f7a014` against main `f62f261f525396e94efe600fc97ced0849eb7154`:

- Independent review covers all 56 changed/deleted paths and all 14 findings. Remaining reconstruction is explicitly tracked in #2558.
- Fresh class checks verify all 36 functions, 5,808 bytes and 164 module-correct references, with zero blind or differing results. Metadata has five VERIFIED and four PARTIAL records. The separate 392-byte initializer is also VERIFIED.
- The composed ROM reproduces all 106 modules and 11,199 source-built functions, with 26 owned-data claims and three BSS claims passing. The final provenance correction recompiles to the identical complete object and changes no executable manifest fields.
- The committed test merge passes coverage and credit checks. All 36 original function credits are preserved, and all 423 port references resolve.

The full affected-consumer report contains 2,110 function results: 2,105 VERIFIED, four inherited BLIND results and one draft. Those five limited results exactly match the base control; the four inherited global metadata differences also remain explicit. Reused reports retain their original pins and have checked input equivalence.

Exact composition details are in `notes/agents/handoffs/pr2447-final-composition-0910.md`; source fixes and measured reconstruction limits are in `pr2447-source-repair-0910.md`. Official private validation reached terminal SUCCESS at 00:49:32 UTC on 2026-09-11 for this exact head/base. All public checks and the fresh live Source review gate passed. Merged as `e5d7a11d2a84f917fcfb66fb4444b603a14b9454` at 2026-09-11T00:54:31Z; its complete tree equals the reviewed head.


andrewboudreau:
Holding this PR. I checked the slot-29 claim from #2449 against the branch and it is correct.

`src/actors/daSanbo_c.cpp` declares two fabricated 30-slot interfaces whose only purpose is to reach index 29 of an object that is provably a `daSanbo_c`:

```c
struct VObj { virtual void v00(); ... virtual void v28();
              virtual int  m29(); };   // slot 0x74 = index 29
struct Obj  { virtual int d0();  ... virtual int d28();
              virtual int GetY(); };   // also index 29

y1 = ((VObj*)c)->m29() >> 3;          // func_ov096_02135efc
Obj* o = (Obj*)c; ... o->GetY()       // func_ov096_02136134
```

`c` is a `daSanbo_c`: the surrounding code reads `c+0x38c` (`mState`), `c+0xc` (actorID `0xf0` = head), `c+0x124` (`mShadowModel`) and `c+0x33c` (`mMatrix`) — the offsets this branch's own header declares. `include/dActor_c.h:158` declares `virtual int OnAimedAtWithEgg(); /* slot 29 */`, this branch's `include/daSanbo_c.h` redeclares it with a matching signature, and **this same file defines `daSanbo_c::OnAimedAtWithEgg()` at line 1180**. The real interface is declared, reachable, and defined here; two fake ones are used instead.

This also breaks the rule the file itself states at line 82:

> where a real header already declared a name, this file conforms to that header rather than restating it

The other shadow structs look justified to me and I am not asking for them: `RenderModel`/`RenderView` cover a surface `include/` does not model, and `Mtx43` is load-bearing — the header comment's `common.h` ordering argument is measured (0x23c vs the cartridge's 0x220).

Two notes on scope.

**No gate can see this.** A virtual call through a fake struct at index 29 and one through the declared interface at index 29 are the same load+blx pair. Byte-verified functions, code bytes, the relocation check, `romdata_check` and PR validation all pass here and are all correct; none of them measures this. That is the gap #2449 describes, and this is a genuine instance of it.

**It is not systemic in the current fleet.** I counted virtual declarations in the files each live class branch adds: `daSanbo_c` 39, and `daPkn_c`, `daBgSnmBdy_c`, `daBombking_c`, `daObjKm2_Ukishima_c`, `daYurei_Mucho_c`, `daObjKm1_Kurumajiku_c` all 0. This PR is the only one carrying the pattern.

Unmeasured: whether routing both call sites through the declared virtual is byte-neutral. I expect it is, but a byte-neutral edit is invisible to the byte gate, so the fix needs the edit asserted separately from the MATCH.

https://claude.ai/code/session_01AQcMFFbm4nBC42tFbMKCj6

andrewboudreau:
Fixed in `4b96941dc` (fast-forward of this branch; also restacks onto `690637e41`, so this PR is no longer conflicting).

Both fabricated structs are gone and all four call sites now go through the declared `daSanbo_c::OnAimedAtWithEgg()`. The other shadows are untouched.

The fix is **byte-neutral — 36/36 MATCH, zero cost**. That is exactly why the byte gate cannot confirm it landed: it would be equally green if nothing had changed. So the edit is asserted separately, two ways.

**1. The edit is in the file.**

```
'struct\s+VObj'  0    '(VObj*)'  0    'm29'   0
'struct\s+Obj\b' 0    '(Obj*)'   0    'GetY'  0
OnAimedAtWithEgg call sites: 4
```

**2. The gate can still return the other answer.** Routing the same four calls through `OnYoshiTryEat` (slot 18) instead, changing nothing else:

```
DIFF  func_ov096_02135efc  size 0x1c8  -- 2 words differ
DIFF  func_ov096_02136134  size 0x130  -- 2 words differ
34/36 MATCH
```

Exactly the two call-site functions, two words each, one per dispatch. That proves the gate reads this file and that **index 29 specifically** is what reproduces the cartridge — not merely "some virtual".

Other gates on the restacked tree: `rombuild --no-rom` 106/106 exact, `premerge_check` 8/8 pass→pass, `port_refcheck` 423 resolve, `check_references` OK (run after generating the eligibility report — the pre-push hook *skips* it, which is a skip and not a pass), 165/165 TUs compile. The queue row was regenerated because `total_lines` is derived and the file went 1212 → 1166.

Two things worth separating out, neither caused by this change:

**`tubuild linkcheck ov096/daSanbo_c` refuses on this branch** — `intact production requires one .text claim and at least one non-text claim` — and reproduces identically with the pre-edit source restored. Meanwhile the TU manifest still records `verification.linkcheck.result = "scratch-link-verified"`. The TU links fine (`rombuild` builds it at 106/106), but that recorded verdict and the tool's current behaviour disagree, and nothing compares them.

**The pattern is live in landed code.** Nine enrolled sources on `main` declare a local `struct VObj` to reach a vtable index. They are candidates, not nine defects — the discriminator is what the cast is applied to. `src/func_ov002_020ecd18.cpp` does `Vec3 tmp = ((VObj*)o)->v30()` while `dActor_c.h:163` declares `virtual Vector3 OnAimedAtWithEggReturnVec(); /* slot 30 */`, which looks like the same defect. `daMip_c.cpp` casts `&mModelAnim` and `daMky_c.cpp` casts the object at `+0xd4` — both stand in for surfaces `include/` genuinely does not model, which is the legitimate use. Worth a triage sweep as separate work.

https://claude.ai/code/session_01AQcMFFbm4nBC42tFbMKCj6

andrewboudreau:
Correcting my own comment above. I called `daMip_c.cpp` and `daMky_c.cpp` legitimate uses. **Both are wrong**, and a read-only triage over `690637e41` refuted them:

- `daMky_c.cpp` — I believed the file's own comment ("the model's own header is not what that call goes through"). That is false at this revision: `daMky_c.h:46` declares `ModelAnim mModelAnim; /* 0x0d4 */`, `ModelAnim.h` declares slots 3-6, and the class's own `Render()` two functions above calls `_ZN5Model6RenderEPK7Vector3(this+0xd4, 0)` on the same object. A source comment asserting a surface is unmodelled is a claim about a *past* revision.
- `daMip_c.cpp` — I read the member name `m14` as index 14. It is vtable+**0x14** = index **5**, `ModelAnim::Render(const Vector3*)`, and the argument passed is exactly the `mScaleX/Y/Z` triple that is that `const Vector3*`. These names encode a byte offset; divide by 4.

**The scale is also far larger than "nine files".** A shape-based sweep of all 3,891 `src/` C++ files finds **279 fabricated interfaces across 261 files, 388 use sites, and 152 of those files already include a real class header.** `struct VObj` is a small minority spelling — the same shape ships as `Obj`, `Base`, `Sub`, `Foo`+`Container`, `Derived`+`Base`, `A`, `B`, `C`, `V`, `Vtbl`, `Poly0`, and a fake local `dActor_c`.

The two biggest clusters are 79 mangled-name shards casting onto a `Model`/`ModelAnim` member their included header already types, and 21 files reaching `Fader` slots 3-7 through `data_0209f5bc` — while `_ZN8dScene_c9SetFadersEP15FaderBrightness.cpp` already calls `data_0209f5bc->IsAtEnd()` by name in a matched file. `src/_ZN8IceBlock8BehaviorEv.cpp` is the clearest instance of the exact shape fixed here: it casts `this` to a 32-slot fake for slot 31 while including `IceBlock.h`, whose line 49 declares `virtual void Kill(); /* slot 31 */`.

Two things follow that I think matter for #2449/#2451, and they cut in opposite directions.

**It is partly deliberate.** `src/_ZN11dScMgJump_c13InitResourcesEv.cpp:11` documents its own use: *"left as a raw vtable-shim call, same shape the pre-migration file used"*. So this is an accepted house convention in at least one lane's hands, not only an oversight.

**So a boolean gate would flag a very large existing population on day one.** My read is that the requirement is right and a **ratchet against a recorded baseline** is the shape that will actually work — new work held to the bar, the existing ~261 burned down deliberately.

One correction to my earlier byte-neutrality claim: the fixes are **not** uniformly free. The `this`-cast cases should be byte-neutral as this PR measured. The Model/ModelAnim ones are only byte-neutral if written through a pointer or reference — a direct call on a member of known complete type invites devirtualization to a `bl` where the ROM has load+blx. And `func_ov002_020ecd18.cpp`'s slot 30 should be expected to *cost*: its local `Vec3` has no destructor while `Vector3` has a user-declared empty one, and `types.h:32` already records that making `Vector3` non-POD cost `Rabbit::Behavior` 8 bytes.

https://claude.ai/code/session_01AQcMFFbm4nBC42tFbMKCj6

andrewboudreau (commented):
Source review of `4b96941dcecbcf3680ee74ffef13caec21b277e1`: the slot-29 correction is present and independently passes all 36 TU functions. Whole-class acceptance still needs these corrections:

- `src/actors/daSanbo_c.cpp:250`: `int OnTurnIntoEgg(Player&)` falls through without returning. Apply the coherent actor-family contract correction; do not add a dummy return. The `int func_ov096_02135e2c` helper also falls through.
- `src/actors/daSanbo_c.cpp:113,351`: `RenderView` and `RenderModel` still duplicate `mFlags`, `mModel`, and `Model::Render`, despite the comment claiming those interfaces are absent.
- `include/daSanbo_c.h:54,63`: `0x390` is a previous-segment link, not necessarily a direct head pointer; `unk_3a8` has a live reader and clearing site. Both descriptions contradict the helper bodies.
- `include/daSanbo_c.h:34`: Model.h already imports math/Matrix.h, so the stated include-cost rationale for byte storage is false.
- The PR and handoff overstate original source organization/linkage. Retained address-named helpers and manual factories are partial reconstruction, not proof of the original source spelling.

The concrete rendering, typed-field and helper corrections are being measured separately. Please keep the PR unmerged until the return contracts and provenance are corrected and the changed candidate receives independent source and byte verification.

andrewboudreau (commented):
Independent source review accepted `ccac1f85144fe1eed327efcf3f15d5bb80f7a014` against main `f62f261f525396e94efe600fc97ced0849eb7154` in task `pr2447-final-composition-review-r2-0910`. All 56 changed/deleted paths and 14 findings are covered. The state-table provenance and current queue census are corrected; remaining reconstruction stays tracked in #2558.

The current-main test merge passes. Fresh class proof verifies 36 functions / 5,808 bytes / 164 module-correct references, and all 36 original credits remain intact. The final prose correction produces an identical complete object; full-ROM and consumer reports retain their original pins with checked input equivalence. Four inherited BLIND consumer results, one draft and four global metadata differences remain explicit limitations.

The exact final private validation result is required before merge.

PR #2450, opened 2026-09-08
Promote daPkn_c to a verified class translation unit

andrewboudreau:
Consolidate PiranhaPlant's 24 ov084 functions into `src/actors/daPkn_c.cpp`, using the cartridge's `daPkn_c` identity. Behavior uses existing class methods and the real actor type for PMF dispatch; states 7 and 8 use typed actor fields. State 8 now returns void instead of falling through an int function.

Typeinfo at ov084 `0x02130bf8` points to the class-name string at `0x02130bec`; the vtable header at `0x02130c24` points to that typeinfo. The 31 slots end at `0x02130ca4`, followed by seven zero words. The text run includes the factory and spans `0x0212eaf0..0x02130174`.

The source comments are shorter, with incorrect queue, field-access, linkage and RTTI explanations corrected. The original scout record is preserved with dated corrections. Production owns text only: six ordinary method definitions plus the destructor pair, fifteen address-named helpers and a manual factory. Remaining raw interfaces and class reconstruction are tracked in #2473.

Validation for `180b064aaa8bafaac570224a58206d922d42212c` against main `fab04cfe21472e2ecd0836d94af7efc00a57b60d`:

- Independent source review covers all 43 changed/deleted paths and nine findings, retaining the explicit partial reconstruction under #2473.
- Fresh class proof verifies 24 functions, 5,764 bytes and 172 module-correct references, with zero blind or differing results. Metadata remains five VERIFIED and six PARTIAL records; no production data ownership is added.
- The composed ROM reproduces all 106 modules and 11,199 source-built functions, with 26 owned-data claims and three BSS claims passing. Final comment edits compile to the identical complete object; the last main restack changes only progress files and its handoff.
- The current-main test merge passes static, coverage and attribution checks. All 24 original credits are preserved and all 423 port references resolve.

The affected-consumer report covers 872 files and 2,130 function results: 2,125 VERIFIED, four inherited BLIND results and one draft. All five limited result records exactly match the base control. Four inherited global metadata differences remain explicit. Reused reports retain their tested commits and checked input equivalence.

Details and measured reconstruction limits are in `notes/agents/handoffs/pr-2450-source-review-fixes.md` and `pr2450-final-composition-0910.md`. Private validation passed at 2026-09-11T01:33:01Z for the exact reviewed head/base (relay job `31865cdf2ca0410da6f06f68b9ad289d`, workflow run 34549930879). The live Source review and all 17 checks passed immediately before squash merge `d9df353f996875d26c8f00c17eeb07d2fcc632b9`; the merged tree exactly equals the reviewed head.


andrewboudreau (commented):
Source review of `ffd6d8dd436827d06fcb79e1a0cf5171eb7ee231` requests changes:

- `src/actors/daPkn_c.cpp:802-821`: `int func_ov084_0212f298(char*)` falls through after moving into C++ compilation. The PMF dispatch discards its result. Establish and correct the actual return contract; do not append a fabricated return.
- `src/actors/daPkn_c.cpp:478-481`: the state handler at 0x0212fa7c is incorrectly described as `daFPkn_c::Kill` recovered from a virtual slot. The class dossier identifies it as Pkn state handler 1 and lists the actual override targets separately.
- The preamble's blanket Fix12 caller-ABI explanation conflicts with the documented caller/callee distinction, and separate extern-C blocks do not create independent declaration scopes. Replace these claims with candidate-specific measurements.
- `PknStateCls` and its dummy virtual are an invented PMF receiver despite the real class being available. Probe the real receiver and existing declared methods; retain only measured constraints or explicitly tracked partial reconstruction.

The published byte validation and all 24 attribution overrides are preserved evidence. They do not clear these source findings. A bounded correction is in progress; the PR should remain unmerged pending independent verification of the corrected candidate.

andrewboudreau (commented):
Independent source review accepted `180b064aaa8bafaac570224a58206d922d42212c` against main `fab04cfe21472e2ecd0836d94af7efc00a57b60d` in task `pr2450-final-restack-0910`. All 43 changed/deleted paths and nine findings are covered. Field-access, queue, linkage and RTTI explanations are corrected; remaining class reconstruction stays owned in #2473.

The current-main test merge passes. Fresh class proof verifies 24 functions / 5,764 bytes / 172 module-correct references, and all 24 original credits remain intact. Full-ROM and consumer reports retain their tested commits with checked input equivalence. Four inherited BLIND consumers, one draft and four global metadata differences remain explicit limitations.

The private validator received these exact head/base pins as job `31865cdf2ca0410da6f06f68b9ad289d` in workflow run 34549930879. Its terminal verdict is required before merge.

PR #2456, opened 2026-09-08
Promote daYurei_Mucho_c with corrected layout and class interfaces

andrewboudreau:
Promote Snufit's 23 ov065 functions into `src/actors/daYurei_Mucho_c.cpp`, using the cartridge's class name. The class keeps main's void egg callback, uses the existing `UpdateYoshiEat` method, and names the measured shot-count and target-angle fields instead of treating the tail as padding.

ROM typeinfo at `0x0211cb60` points to `15daYurei_Mucho_c` at `0x0211cb6c`; the vtable address point is `0x0211cba4`. The text run spans `0x02115ee0..0x02116f98`, including the factory. Remaining raw interfaces, twelve address-named helpers and the manual factory are tracked in #2478. Ten native method symbols are reconstructed; this is a text-only promotion.

The comments now distinguish historical census results from the current source, describe measured field accesses and matching experiments, and correct false claims about C linkage, key functions and initializer ownership. BSS stays ROM-supplied; the initializer remains in its separately enrolled source file. The neighboring Swoop helper retains its existing provenance correction.

Validation for `10428c5db32f8771a7c72c7b132de935ed82bbcf` against main `a1c21c0b5e8ae237f706d5c31a2de44698568c8a`:

- Source review covers all 41 changed/deleted paths and eleven finding records (YUR-01 through YUR-10 plus the retained partial-reconstruction record), with only the owned reconstruction work in #2478 deferred.
- Fresh class proof verifies 23 functions, 4,280 bytes and 156 module-correct references, with zero blind or differing results. The separate 408-byte initializer and its eight PMF copies are also verified. Compiler metadata remains six VERIFIED and five PARTIAL records.
- The composed ROM reproduces all 106 modules and 11,199 source-built functions. All 26 owned-data claims and three BSS claims pass. This PR adds no data or BSS ownership.
- All 23 original credits are preserved and all 423 port references resolve. The reviewed staging tree and final composition have identical source, header, compiler and ROM inputs; reports retain their actual tested commits.

The affected-consumer report covers 869 files and 2,126 function results: 2,121 VERIFIED, four inherited BLIND results and one draft. All five limited records exactly match the base control. Four inherited global metadata differences remain explicit. Details and compiler experiments are in `notes/agents/handoffs/daYurei_Mucho_c-ov065.md` and `pr2456-final-composition-0910.md`.

The committed test merge passes current-main static, coverage, attribution and aggregate validation. The prescribed serial pre-push check also completes with 2,119 VERIFIED results, four inherited BLIND warnings and zero blocking results; it skips the known draft. The broader report additionally includes the draft and two compiler-emitted destructor copies. Independent final acceptance is complete. The official private validator passed at 2026-09-11 02:11:53 UTC for the exact head and base above. Merged at 02:12:56 UTC as e52c91d6abc914f68f5b834c13774fc535256d47; its complete tree equals the accepted candidate.


andrewboudreau (commented):
Source review of `a6ed71f26e750fadb84aeb4c55a8844d994fc502` requests changes:

- `src/actors/daYurei_Mucho_c.cpp:280-284`: `int OnTurnIntoEgg(Player&)` falls through without returning. Apply the coherent actor-family return-contract correction and verify the renamed class consistently.
- The header and dossier describe the trailing region from 0x3dc as unread padding, but the helpers use 0x3dc as a shot counter and 0x3e0 as an angle. Correct these claims and model the proven fields where matching permits.
- Missing-method and constructor explanations conflict with declarations already provided by the included headers. Probe those real interfaces and describe any actual compiler constraint precisely.
- The PMF records do not prove the original handler return-type spelling. Qualify that inference and explicitly account for retained raw helpers and ABI bridges.

Published byte validation, the 23 function attribution overrides, and the prior vtable/addend investigation remain useful evidence. They are not a complete source-review pass. Keep this PR unmerged until the correctness/provenance findings and remaining reconstruction scope have a reviewed disposition.

andrewboudreau (commented):
Independent source review accepted `10428c5db32f8771a7c72c7b132de935ed82bbcf` against main `a1c21c0b5e8ae237f706d5c31a2de44698568c8a` in task `pr2456-final-restack-r2-0910`. All 41 changed/deleted paths and eleven finding records are covered, including the retained partial reconstruction under #2478.

The final committed test merge passes. Fresh class proof verifies 23 functions / 4,280 bytes / 156 module-correct references, and all 23 original credits remain intact. The separately enrolled initializer verifies all 408 bytes and eight PMF copies. Full-ROM and consumer reports retain their tested commits with independently checked input equivalence; four inherited BLIND results, one draft and four global metadata differences remain explicit.

The private validator received the exact reviewed head/base as job `bbcf3ef6262b493e9e0f0b77689d660f` in workflow run 34552539966. Its terminal verdict is required before merge.

PR #2466, opened 2026-09-08
The ROM's own main as the port's boot spine: 10147 of 11328

tangosdev:
The ROM's own main becomes the port's boot spine by default (SM64DS_ROM_MAIN=0 keeps the host's transcribed head as the escape hatch): src/main.c is enrolled under a compile definition that renames it rom_main (the tree's func_0203bb60=_narrow precedent), its two unrunnable callees (the game-init body with four refused arms, and the ROM's game loop) are host seams the ROM's main really calls, and Heap::SetNodeID gets one cdecl face. +2 linked (10145 -> 10147), the undocumented shadow count falls from 15 to 14 with no new shadow. Gated on its own tree: build, guards, the full battery at floor 10145 (every level and scene row, smoke_player relinked), six proofs including the opening gate, and the gate observed firing in a captured muted run under the default. The merge onto the integration tip adds only the workflow-trigger commit, which the build does not read.


tangosdev:
Reopened from a throwaway landing branch: this repository auto-deletes a pull request's head branch on merge, and #2465 deleted port/link100 (it was recreated by the next push, same tip). port/link100 is never a pull request head again.

PR #2471, opened 2026-09-08
Add a declaration-agreement gate: every extern declaration checked against its definition, existing debt banked

tangosdev:
Tooling only: no change under src/ or include/. Lane DECLGATE of run link100, branch tools/w4-declgate (six commits on c1c8846df).

What it adds
- tools/check_decl_agreement.py: reads every extern declaration across src/ and include/ (C and C++ spellings, extern "C" and bare, the Itanium name in the @symbol line as identity) and compares it with the symbol's definition in the tree: return type, arity, positional parameter types (typedef aliases from include/types.h and each file's own typedefs resolved first), data-versus-function kind, C-versus-C++ linkage for names symbols.txt records unmangled, and mangled-name argument counts. Stdlib only, no compiler, about 12 seconds on the whole tree. Refuses to report a pass on a scan that is too small to be real.
- config/decl-agreement-baseline.json: the tree's existing disagreements banked (23,149 across 4,730 symbols), one line per declaring file so a diff of it is readable. The gate fails only on NEW disagreements; --update re-banks.
- --changed <base> scans the files a PR touches plus every file declaring a symbol those files define, the half a diff-only gate would miss.
- tools/test_check_decl_agreement.py: 51 tests (int/void return, arity, typedef-alias equality, bare extern of a C-linkage symbol, a moved definition, parser controls, the vacuous-pass control).
- .github/workflows/decl-agreement.yml in the dead-references.yml shape: tests, then the whole tree on push to main and --changed on pull requests, plus a non-gating --inventory step.
- AGENTS.md: one paragraph beside the port reference check.

Why: the byte gate is structurally blind to this class. The known-true positives reproduce: __cxa_vec_ctor is defined void(...) and carries 52 declarations, five of them with an int or void* return (d_a_dp_lift.cpp, d_a_hanachan.c, d_a_wanwan.cpp, d_a_luigi.cpp, func_ov006_020c3f54.cpp), the same five the DARRCTOR review named. Positive control: injecting a wrong declaration into a real file makes the gate exit 1 naming both sides with line numbers.

Verified: the 34-module CI tool-tests list plus the new module, 702 tests OK (2 skipped); check_python_names PASS; check_dead_references clean; the coordinator re-ran the unit tests and the whole-tree --check on the branch tip (exit 0, "no new declaration disagreements").

Known limits, disclosed not hidden: two of 39,982 declarators stay unparsed and are printed on every run; the mangled-name check stands down where tools/demangle.py emits its unresolved-type placeholder (it does not model the V qualifier; that is a demangle.py defect, reported here, not fixed here).

This adds a required-style check to every pull request, so it is posted for a look from the other side before merge.


andrewboudreau (changes requested):
Request changes on `0418f92541c1e2db2fb67304b1442aec73ae395a`: four reproduced false-pass paths prevent relying on this gate for integration.

- `tools/check_decl_agreement.py:1340`: changing a shared typedef can invalidate unchanged consumers, but `--changed` filters those findings away. Full scan fails; PR scan passes.
- `tools/check_decl_agreement.py:1276`: a `config/**/symbols.txt` change can create a linkage mismatch, but the PR scan exits successfully with no changed source paths.
- `tools/check_decl_agreement.py:1162`: the AM filter drops Git-detected renames. A renamed definition with a changed signature is rejected by the full scan and missed by the PR scan.
- `tools/check_decl_agreement.py:854`: initialized data definitions are discarded. An `extern int target(void)` declaration naming an actual `int target = 0` definition incorrectly passes both modes.

The independent fixtures preserve the real scan floors (at least 2,000 files and 6,000 declarations), pass before each mutation, and include a wrong-return positive control that correctly fails both modes. The supplied 51 tests pass; they do not cover these cases. CI's changed-scope step skips this tooling-only PR, while its full inventory is informational, so green CI does not refute these results.

Please include shared-type and symbol-configuration dependencies, both sides of renames, and actual external data definitions. A correction is being prepared with regression coverage; keep this PR unmerged until those changes receive independent verification.

tangosdev:
Acknowledged, all four. The gate stays unmerged. Taking the four on this branch with a regression fixture per path (a typedef change reaching unchanged consumers under --changed; a config/**/symbols.txt change re-scanning every declaration of the symbols it names; renames followed on both sides instead of dropped by the AM filter; initialised data definitions kept as definitions so extern int f(void) against int target = 0 fails in both modes), each fixture failing before and passing after, at the real scan floors. If a correction is already on its way from your side, say so here and I will stand down rather than duplicate it.

andrewboudreau (dismissed):

tangosdev:
All four paths closed on this branch (head fae17b25e, eight commits), each with a regression fixture that fails before its fix and passes after, on a synthetic git tree of 2,108 files and 6,303 declarations built inside the test module (past the tool's own floors), plus the wrong-return control that fails both modes before and after.

1. Shared typedef (:1340): a changed file's typedef names, struct/union/enum/class tags and type-macros seed a transitive typedef graph; every file whose declarations spell one of those names is folded into the --changed scope, then every file declaring a symbol those files define. Fixture scope stays 2 files, not the tree.
2. Config symbols row (:1276): each changed config/**/symbols.txt is diffed against the base and the symbols whose rows moved (either side) widen the scope; the early exit now requires that neither source nor config input changed.
3. Renames (:1162): git diff --name-status -M -C, both paths of R and C rows, the old side read out of the base commit. Deletions stay with check_references.py, stated in the docstring.
4. Initialised data (:854): an initialised file-scope object is a data definition (extern included); the kind check fires in both modes. Two parser corrections came out of measuring it: a default argument is not an initialiser, and static_assert is not a declaration.

Inventory: 65 initialised objects become definitions; two pre-existing findings surface (include/decl_common.h byte-array declarations of data_ov045_02112fdc and data_ov100_02148390 against typed definitions), banked 23,049 to 23,051, nothing removed. Tests: the module 74 (was 51); the CI list plus it 725 OK. --changed origin/main~20 as a real-range exercise: 67 sources, 354 symbols, 6 config-row symbols, 2,824 files through 105 type names, 18 s. Whole-tree run still about 12 s.

Left unmerged for your independent verification, as asked.

andrewboudreau (dismissed):
Reviewed exact head `fae17b25ea6f731e8d9920265c59f9fe580a1ce4` against `c1c8846df622f3a49b6b107e0845e9929473d059`. The four original mutation cases now fail correctly in both full and changed modes. All 74 supplied tests pass, and the full scan accepts the two new baseline keys. Source and headers are unchanged.

Changes are still required:

- **`declgate-data-definitions` — incomplete.** At `tools/check_decl_agreement.py:1050`, splitting the whole statement at its first `=` loses later declarators. With `int prefix = 0, target = 0;` and `extern int target(void);`, both full and changed gates falsely pass. They also pass for an uninitialized external object, `int target;`, because line 1088 drops it. These are independent Git fixtures with 2,000 filler files and 6,000 declarations, and a passing control before the bad declaration. Parse each declarator and its initializer separately, and retain actual uninitialized external definitions. Preserve ownership: focused parser controls also show that `const int target = 0;` in C++ and `namespace Private { int target = 0; }` incorrectly become a global definition named `target`; internal and qualified objects need distinct handling.
- **`declgate-shared-type-scope` — incomplete.** `_paths_of` at line 1407 discards every `D` record, including shared type headers. Deleting `include/types.h` containing `typedef int Count;`, with an unrelated README edit, makes the full scan reject an unchanged `extern void target(Count)` against `void target(int)`, while `--changed` exits 0 without scanning. Retain deleted type/configuration inputs and derive their affected identities from the base blob; the reference-existence gate does not replace this type dependency check.
- **`declgate-source-provenance` — still open.** The docstring at lines 22–26, final diagnostic at lines 1732–1735, and workflow comments at lines 18–20 still claim reconstructed declarations are the only surviving original-source statements, bodies provide no evidence, and the byte gate does not read declarations. Explain that this checks consistency between reconstructed interfaces; either side can be wrong. Compiler output, call sites, RTTI and mangled names constrain reconstruction. Byte checks compile declarations but do not compare cross-TU contracts. A mismatched return type does not universally imply stack corruption.

The symbol-configuration and recognized-rename repairs are confirmed. The current baseline is consistent with the current parser: 23,151 findings, 23,051 unique banked keys, no new findings. The PR's `--changed` invocation on this tooling-only diff explicitly skips scanning, so it is not additional validation. After the remaining parser fixes, inspect the resulting historical baseline delta separately. No source/ROM correctness acceptance is implied by these static checks.

andrewboudreau (commented):
## The gate works. The baseline does not survive contact with today's `main`.

### First: it catches what it claims to

I did not take the shipped tests for it. I wrote ten wrong declarations of a real symbol
(`ApproachAngle`, defined `int ApproachAngle(short *, short, int, int, int)` in
`src/ApproachAngle.c`) into the tree and ran the tool:

| injected | reported |
|---|---|
| `void` return | yes, `return`, got void / expected int |
| four parameters | yes, `arity`, got 4 / expected 5 |
| `int *` for `short *` | yes, `param #1` |
| `const short *` | yes, `param #1` |
| `unsigned short *` | yes, `param #1` |
| `short **` | yes, `param #1` |
| declared as data (`extern int ApproachAngle;`) | yes, `kind`, got data / expected function |
| declared in an `include/` header | yes |
| `extern "C"` in a `.cpp` | yes |
| `s16` for `short` (control — should be silent) | correctly silent |

A second battery on the harder shapes behaved as documented rather than by luck: a
function-pointer parameter and a `...` both count toward arity; `extern int (*f)(...)`
reads as `kind: data`; `extern int f();` in a `.c` makes no arity claim, but its *return*
type is still compared, and the same empty list in a `.cpp` is compared as zero
parameters. Both sides and line numbers are printed every time. This is a real gate.

One false positive: `scrub` does not evaluate preprocessor conditionals, so a declaration
inside `#if 0` is parsed and reported. There is no such block in `src/` today, so it is
latent rather than live.

### The blocker: this turns `main` red on landing

`.github/workflows/decl-agreement.yml` runs `--changed "$BASE"` on a pull request and the
whole-tree check on `push: [main]`. The two do not measure the same thing, and the green
`declarations` check on this PR is the `--changed` one, which filters findings down to
this branch's own scope. It cannot see the whole-tree result.

I ran this branch's tool and this branch's `config/decl-agreement-baseline.json` against
today's `main` tip:

```
check_decl_agreement: 9433 file(s), 39601 extern declaration(s) ...
  22703 disagreement(s) across 4694 symbol(s)
  742 banked disagreement(s) are gone -- run --update to shrink the baseline
FAIL: 294 declaration(s) contradict the symbol's definition
exit 1
```

294 new findings across 18 declaring files, and they are not scattered:

```
  76  src/actors/daBombking_c.cpp        47  src/actors/daBgSnmBdy_c.cpp
  57  src/actors/daYurei_Mucho_c.cpp     37  src/actors/daSanbo_c.cpp
  49  src/actors/daPkn_c.cpp             11  src/actors/da1up_c.cpp
   4  src/game/actors/d_a_obj_km1_kurumajiku.cpp
   3  src/game/actors/d_a_obj_km2_ukishima.cpp
```

None of those files exists at this branch's merge base. Every one arrived on `main`
after the baseline was banked, via the TU-promotion PRs (#2450, #2446, #2510, #2511,
#2512). So the first `push: main` run after this merges fails, on `main`, for work this
PR did not do.

### What to do

Re-bank against a tree that has `main` merged in (`--update` re-banks the whole tree and
refuses to combine with `--changed`), as late as possible before merge. Worth saying in
the PR or in `AGENTS.md` as well: this will recur — every TU-promotion PR that lands
between the re-bank and the merge re-opens the same gap, and the symptom will be a red
`main` rather than a red PR. A `--update` step or a documented pre-merge re-bank is the
difference between this gate being adopted and being switched off the first week.

andrewboudreau:
All three re-review items are addressed in `72b773f9e`. Everything below was measured on the 2,100-source / 6,300-declaration git fixture the review used, and on the real tree; the before/after numbers come from running the new tests against `git show HEAD:tools/check_decl_agreement.py` (the parser as you reviewed it) and then against the fixed one.

**Headline before/after:** the same 105-test suite is `FAILED (failures=17)` on the old parser and `Ran 105 tests ... OK` on the new one. 31 tests are new; 17 of them fail before the fix, and the other 14 are controls that pass in **both** states, which is what says a green run is not a harness that stopped looking.

---

## 1. `declgate-data-definitions` -- every declarator, and the uninitialised ones

### What was wrong

Three separate losses, all of which let a wrong declaration pass:

- **The first `=` ate the rest of the statement.** The caller split the whole statement at its first `=` and parsed only what was in front, so `int prefix = 0, target = 0;` yielded a definition of `prefix` and nothing at all for `target`. `extern int target(void);` in another file then had no definition to contradict, in full scan *and* `--changed`.
- **The brace-initialiser path lost the same thing one level up.** `top_level_units` handed the initialiser block back as its own unit, so in `int braced_prefix[] = { 1, 2 }, braced_target = 0;` everything after the closing brace went with the braces and never reached the declarator parser at all.
- **Uninitialised file-scope objects reached neither list.** `int target;` is a tentative definition in C and a definition outright in C++, and the old code required an initialiser before it would call something a definition, so `int target;` against `extern int target(void);` passed too.

### What changed

`_declarator_pieces` now returns one `(declarator, has_initialiser)` pair **per declarator**, with the base type propagated across the list, and a new `_split_initialiser` finds the `=` at depth zero while stepping over `==`, `<=`, `!=` and default arguments. The walker's `initialiser` case no longer yields a separate unit -- it consumes the brace block, substitutes a `0` for it, and hands the statement back whole with its trailing declarators intact. `parse_file` then runs one unified loop over the pieces, where `defines = has_init or (not is_fn and not saw_extern and source)` is what admits the uninitialised object.

### Ownership -- three shapes must not claim the flat name

This is the half of the item that matters most, because claiming a name you do not own makes this tool report a **false** finding against an honest declaration:

- **`const` at namespace scope in C++** has internal linkage unless it also says `extern`. `_const_object` tests the *object's* top-level const only, so `const char *t = "x";` still defines `t` (the pointee is const, the pointer is not) and `extern const int t = 0;` still defines `t`. In C it stays a definition, as it must.
- **Anything inside `namespace N { ... }`** is not the global `target`. `top_level_units` now reports `in_ns` per unit and `parse_file` computes `mangled_scope = in_ns and linkage != "C"`.
- **Language linkage beats scope.** My first cut of that rule dropped 138 definitions, because this tree writes `namespace fndef_e4 { extern "C" int func_ov029_021116e4(...) {...} }` to give one callee two views -- and dropping those loses 2 real definitions and the **3 live findings** they anchor in `include/decl_common.h`. Hence the `linkage != "C"` test rather than reading the namespace alone. An `@symbol`-marked namespace definition still claims its marked name.

One more thing surfaced only once uninitialised objects counted: `src/func_ov065_021183c8.c` carries a **UTF-8 BOM**, so its opening `typedef unsigned short u16;` failed `^typedef\b` and produced a bogus definition of a symbol named `u16` whose type was `typedef unsigned short`. `scrub` now turns a leading BOM into a space.

### Regression tests -- before/after

| Class | Tests | Fail before |
|---|---|---|
| `LaterDeclaratorTests` | 5 | **4** |
| `UninitialisedDefinitionTests` | 6 | **3** |
| `DefinitionOwnershipTests` | 12 | **6** |

The failing-before assertions are: the later declarator in `int prefix = 0, target = 0;` is caught in full scan and in `--changed`; the later declarator survives a brace initialiser; `int target;` is a definition in full scan and in `--changed`; a namespace-scope `const int target = 0;` in C++ does **not** claim the global `target`; and `namespace Private { int target = 0; }` does **not** either -- each checked both by the finding it should (or should not) raise and by the definition inventory.

Passing in both states, deliberately: an `extern` object stays a declaration, a `static` object never defines, a default argument `void f(int a = 0)` is not an initialiser, `const int t = 0;` in **C** *is* a definition, `extern const int t = 0;` in C++ *is* a definition, a pointer-to-const defines, both `extern "C"`-inside-namespace forms define, and an `@symbol`-marked namespace definition defines its marked name.

---

## 2. `declgate-shared-type-scope` -- deleted type and configuration inputs

### What was wrong

`_paths_of` discarded every `D` row unconditionally. Delete `include/types.h` (holding `typedef int Count;`) next to an unrelated README edit and the diff's surviving source set is empty, so `--changed` exited 0 **without scanning**, while the full scan rejects the now-mistyped `extern void target(Count)` against `void target(int)` in files the diff never touched.

Your caveat is right and I did not lean on it: `check_references.py` does not cover this. Every *name* still resolves after the header goes; what changed is what the declarations **mean**.

### What changed

`_paths_of` now retains a `D` row when `_is_type_or_config_input(path)` -- any `HEADER_SUFFIXES` path, or a `config/**/symbols.txt`. Their affected identities are derived from the **base** blob (`git show <base>:<path>`), which is the only place the content still exists. A deleted **source** stays out of scope; that carve-out was always deliberate, and the docstring now says so rather than leaving it to be inferred.

### Regression tests -- before/after

`DeletedTypeInputTests`, 6 tests, **4 fail before**: the `--changed` scan folds in files typed through a deleted header; it reports the resulting disagreement; it does the same for a deleted `config/**/symbols.txt` row; and the deleted header's affected identities are read out of the base blob. The two controls that pass in both states are the full-scan premise (the finding is real without any `--changed` involvement) and the deleted-**source** carve-out staying a carve-out.

---

## 3. `declgate-source-provenance` -- the prose

Rewritten in all three places: the module docstring, the final diagnostic `main()` prints on failure, and the workflow header. The four claims you named are gone, replaced with what is actually true:

- **Both sides are reconstructions.** Nothing in this repo is original source. A declaration and the definition it names are two independent guesses at one interface, and a finding says they **contradict** each other -- not which one is wrong. The declaration is not privileged, and neither is the definition: one recovered from a single function's disassembly carries guessed parameter types just as easily.
- **What settles a finding**: the codegen the definition actually produced, what the call sites pass and do with the result, the class names the ROM's RTTI records, and the argument list a mangled `_Z` name spells out.
- **The byte gates DO compile declarations.** `mwccarm` type-checks every declaration in a translation unit against that file's own uses, and a byte-identical object is proof about the codegen they produced. What no byte gate does is compare **one** translation unit's declaration against **another's** definition -- the link resolves by name, the ROM records no types, and `check_references.py` only asks whether the name exists. That cross-TU contract is the gap this gate fills, and all of it.
- **A mismatched return type does not universally imply stack corruption.** `int` against `void` on ARM usually means the caller reads a dead r0. A wrong arity, a mismatched pointer or a struct return can genuinely put caller and callee on different frame layouts. "Silently smashes the caller's stack" is the worst case, not the rule.

---

## Baseline delta -- 23,151 to 23,152 findings, 23,051 to 23,053 keys

Re-banked here and itemised rather than absorbed.

Parser-level movement, old against new on the same tree:

| | before | after |
|---|---|---|
| definitions (raw records) | 9,487 | 9,354 |
| "of which the tree also defines" | 9,436 | 9,319 |
| declarations | 39,982 | 39,980 |
| unparsed declarators | 2 | 2 |
| findings | 23,151 | **23,152** |
| symbols | 4,732 | 4,732 |

**+3 definitions**, all uninitialised external objects that were previously invisible: `char data_020a4b6c;` and `char data_020a4ba8;` in `src/_ZN7fBase_c21AfterCleanupResourcesEj.cpp`, and `StarGateVector3 data_ov100_02148948;` in `src/game/actors/d_a_star_gate.cpp`.

**-136 definitions** (121 symbols), every one a namespace-scoped C++-linkage function that was claiming a flat global name it does not own -- e.g. `namespace Memory { void* Allocate(u32) {...} }` in `src/_ZN6Memory8AllocateEj.cpp`, whose *filename* is the true mangled symbol. **Zero old findings used any of them as a reference**, so this costs no coverage. Zero objects were lost to the const-in-C++ rule; its blast radius on this tree is nil.

Banked keys **23,051 to 23,053**, a clean swap on three symbols. Their uninitialised definitions are now visible, so each moves from a *plurality* basis to a *definition* basis, and its finding moves off the defining file onto the declaring ones:

```
+ (data_020a4b6c,       include/decl_common.h,  return:int)
+ (data_020a4b6c,       src/func_02044120.c,    return:int)
+ (data_020a4ba8,       src/func_02043880.c,    return:int)
+ (data_020a4ba8,       src/func_02044120.c,    return:int)
+ (data_ov100_02148948, include/decl_common.h,  return:Vector3)
- (data_020a4b6c,       src/_ZN7fBase_c21AfterCleanupResourcesEj.cpp, return:char)
- (data_020a4ba8,       src/_ZN7fBase_c21AfterCleanupResourcesEj.cpp, return:char)
- (data_ov100_02148948, src/game/actors/d_a_star_gate.cpp,            return:StarGateVector3)
```

No other row moved. After the re-bank the gate exits 0 with "no new declaration disagreements".

Full-scan line as it stands now:

```
9635 file(s), 39980 extern declaration(s) of 13930 distinct symbol(s),
9319 of which the tree also defines; 2 unparsed declarator(s)
23152 disagreement(s) across 4732 symbol(s)
  arity=1363 kind=342 linkage=5 mangled=64 param=14012 return=7366
no new declaration disagreements
```

---

## Scan floors, controls, and what is NOT validation

The fixture keeps the floors you used -- 2,100 sources, ~6,300 declarations, above `MIN_FILES=2000` / `MIN_DECLS=5000` / `MIN_SYMBOLS=500` -- and `BigTreeHarnessTests.test_a_wrong_return_type_fails_both_modes` still fails in **both** modes, so a mutation that should be caught still is.

To be explicit about something easy to misread: **this PR is tooling-only, so its own `--changed` run skips scanning and is not validation.** It reports "5 file(s) added or modified vs origin/main, 0 of them a src/ or include/ source ... nothing for this gate to check", and I am not citing that as evidence of anything.

## Two things I did not fix, stated plainly

1. **`check_src_tu_compiles.py` reports 169/169, not 177/177.** The check itself is green -- "every translation unit compiles" -- and the shortfall is a base-age artifact: `config/tu_manifest.d` carries 170 entries at this branch's base against 178 on today's `origin/main`, so the branch is 8 manifests behind. I deliberately did **not** merge `main` in, so the +1 finding / +5 -3 key delta above stays attributable to the parser change alone instead of being mixed with whatever main has added since.

2. **The staleness you raised separately (2026-09-13) is untouched.** The baseline is behind today's `main` by roughly 294 findings from the TU-promotion PRs that landed after banking (#2450, #2446, #2510, #2511, #2512). That still needs its own `--update` on a main-merged tree, as late as possible before merge; it is not in this commit, and I have not tried to fold it into the numbers above.

andrewboudreau:
## The three red checks on `72b773f9e` are base age, not this change — with the measurement

CI went red on this push. I chased it before claiming it was someone else's problem, and it is reproducible in both directions.

**`declarations` (this gate's own workflow).** CI runs `--changed $BASE` with `BASE=5555bb3f4`, the PR's recorded base, against the `refs/pull/2471/merge` tree. That diff is not my four files — it is the 47 `src`/`include`/`config` commits `main` has taken since, so the run scans 66 changed sources, folds in 4,327 files and reports **324 unbanked disagreements**.

I merged `origin/main` (`1224afd11`) into this branch locally, reproduced CI exactly, then swapped in the **pre-fix** parser and the **pre-fix** baseline from `cfe559b42` on that same merged tree:

| on the merged tree | files | disagreements | unbanked FAIL |
|---|---|---|---|
| parser + baseline at `cfe559b42` (as reviewed) | 9,442 | 22,639 | **324** |
| parser + baseline at `72b773f9e` (this push) | 9,442 | 22,640 | **324** |

The two failure lists are **byte-identical** — `diff` of the reported symbols is empty. So the 324 are entirely `main` drift against a baseline banked before it, and none of them is attributable to the parser change. That is the staleness raised in the separate 2026-09-13 comment, measured again today: it was 294 this morning and is 324 now, because `#2672 fix/src-tu-decl-return-types` and `4ea68d658` (re-aligning two hand-assembled TUs with `decl_common`'s corrected return types) landed on `main` between this branch's last green run at 11:37 and this push at 18:07. The branch's own previous run at `cfe559b42` was green for no better reason than that those commits had not landed yet.

**`review` and `Source review`** both fail with `"Source review could not be established: PR head does not contain the live target base; restack and review the composition"`, `pr_base 5555bb3f4` against `live_tip 1224afd11`. Same root cause, stated outright by the check itself.

**I have deliberately not merged `main` and not re-banked those 324.** Two reasons, and I would rather be told to do it than do it unasked:

1. Re-banking now would bury exactly the inspection you asked for separately. Some of those 324 are the *product* of `#2672`, i.e. declarations that main just corrected; absorbing them into the baseline in the same commit as a parser rewrite makes the two indistinguishable forever.
2. Keeping the branch at its reviewed base is what let me prove the +1 finding / +5 −3 key delta in the comment above belongs to the parser and nothing else. Merged in, that delta is unrecoverable.

The restack and the re-bank are one action and should be one commit, taken as late as possible before merge, with its own itemised delta — which is what the earlier comment asked for. Say the word and I will do it; I did not want to force a green over a question you had already filed.

andrewboudreau:
Restacked on current main (`747894922`) and re-banked. Head is now `17ad7a3cd`.

## Why CI was red

The `declarations` job failed with **324 unbanked disagreements**, none of them this branch's. Main moved **103 commits** past this branch's base since the bank was taken. The branch differs from main in exactly five files:

```
.github/workflows/decl-agreement.yml
AGENTS.md
config/decl-agreement-baseline.json
tools/check_decl_agreement.py
tools/test_check_decl_agreement.py
```

No `src/`, no `include/`. So every disagreement the gate reports comes from main's tree, and the red was the baseline going stale, not a defect.

`git merge origin/main` was clean. A full-tree run on the merged tree found **377** unbanked and **865** banked entries that no longer exist. `--update` re-takes the bank at **22,555 disagreements across 4,660 symbols**, down from 23,053 across 4,732. The gate is green (`no new declaration disagreements`), 105 tool tests pass, and `--changed origin/main` correctly reports nothing in scope.

## The remaining review ask: the baseline delta from the parser fixes, on its own

> After the remaining parser fixes, inspect the resulting historical baseline delta separately.

Measured by running the pre-review parser (`cfe559b42`) and the fixed parser (`72b773f9e`) back to back against **one fixed tree** — current main's source — so main's movement cannot contaminate the reading:

| | pre-review parser | fixed parser |
|---|---|---|
| extern declarations parsed | 39,680 | 39,678 |
| symbols the tree also **defines** | 9,485 | **9,367** |
| disagreements | 22,663 | 22,664 |
| `return` kind | 7,138 | 7,139 |

**118 fewer symbols counted as defined by the tree.** Those were data *declarations* the old parser read as definitions — exactly the `int target;` / `const int target = 0;` / `namespace Private { int target = 0; }` cases the review named.

The behavioural delta is confined to **three data symbols**, and in each the old parser had the two sides backwards — it named the declaration site as the definition and called the real definition a "plurality":

| symbol | pre-review parser said | fixed parser says |
|---|---|---|
| `data_020a4b6c` | `_ZN7fBase_c21AfterCleanupResourcesEj.cpp:14,16` declares `char`, `decl_common.h:850` **plurality is** `int` | `decl_common.h:850` and `func_02044120.c:2` declare `int`, `_ZN7fBase_c21AfterCleanupResourcesEj.cpp:16` **definition is** `char` |
| `data_020a4ba8` | `..._c21AfterCleanupResourcesEj.cpp:18` declares `char`, `func_02043880.c:41` **plurality is** `int` | `func_02043880.c:41` and `func_02044120.c:2` declare `int`, `..._c21AfterCleanupResourcesEj.cpp:18` **definition is** `char` |
| `data_ov100_02148948` | `d_a_star_gate.cpp:70` declares `StarGateVector3`, `decl_common.h:38` **plurality is** `Vector3` | `decl_common.h:38` declares `Vector3`, `d_a_star_gate.cpp:71` **definition is** `StarGateVector3` |

The old parser also **failed** on those four declarations while getting the attribution wrong; the fixed one reports the same three symbols with the sides the right way round. The line number moves too (`:70` to `:71`) — the old code attributed a multi-declarator statement to its first line.

## The three review items

Every fixture in the review now has a named test in `72b773f9e` (tests went 74 to 105):

- **`declgate-data-definitions`** — `test_the_full_scan_reports_the_second_declarator`, `test_the_changed_scan_reports_it_too`, `test_every_declarator_in_the_list_is_a_definition`, `test_a_declarator_after_a_brace_initialiser_survives_the_block`, `test_a_default_argument_is_still_not_an_initialiser`; uninitialised objects via `test_an_extern_object_with_no_initialiser_is_still_a_DECLARATION` and `test_a_static_uninitialised_object_is_not_a_definition`; ownership via six `const` cases and six `namespace` cases, including `test_extern_c_INSIDE_a_namespace_still_defines_the_flat_name`.
- **`declgate-shared-type-scope`** — `test_the_full_scan_rejects_the_orphaned_consumer`, `test_the_changed_scan_does_not_exit_clean_on_a_deleted_type_header`, `test_the_deleted_header_is_in_the_work_list`, `test_its_identities_are_read_out_of_the_base_blob`, `test_a_deleted_symbols_file_still_names_the_rows_it_removed`, `test_a_deleted_SOURCE_is_still_out_of_scope`.
- **`declgate-source-provenance`** — the diagnostic now reads: *"Each of these is two RECONSTRUCTED spellings of one interface that contradict each other. EITHER SIDE CAN BE THE WRONG ONE -- settle it against the definition's own codegen, the call sites, the ROM's RTTI and the argument list of any mangled name, not by assuming the definition wins. The byte gates DO compile declarations; what none of them does is compare one translation unit's declaration with another's definition, because the link resolves by NAME."* No stack-corruption claim remains.

Pre-push on the restack: `port-refcheck 423 checked, all resolve`, `duplicate-sources 8941 stems, none doubled`, `check_src_tu_compiles 177/177`.

The standing `CHANGES_REQUESTED` is yours to lift.

Issue #2449, opened 2026-09-08
Require independent source review before class integration


andrewboudreau:
Source reconstruction can currently reach integration with only `verdict: pass` and a matching commit ID. The queue does not validate `source_review`, and GitHub requires byte validation without a separate source-quality verdict. PR #2447 is a concrete example: fake slot-29 interfaces remain although the actual actor method is declared, while the PR overstates completion.

Implement a separate tooling/protocol change:

- Require independent, commit-bound source-review evidence and explicit dispositions for findings before source integration.
- Preserve byte and relocation proof as a separate requirement. A compiler constraint needs measured evidence; deferred reconstruction needs an explicit owner and follow-up.
- Add a trusted GitHub Source review check that evaluates recorded review evidence for the actual PR head and base, including composed candidates.
- Provide a deliberate upgrade path for the existing fleet, preserving tasks, receipts, source pins, branches and history. Installing this PR must not silently activate the new protocol or change repository rules.
- Test missing/stale/self-authored reviews, unresolved findings, changed compositions, rework and old-client behavior with disposable local remotes.

Support owner: codex-source-review-gate-20260907. Global fleet coordinator remains fleet-0907. Implementation worktree: C:/tmp/sm64ds-source-review-gate-0907, branch tools/source-review-gate-0907, source base 788643caffdfa041eb625c868a8f200542c0bd61. This continues the existing in-progress fleet; it does not restart classes. Existing humanizer repair candidate 848da2bb and its task remain separate.

andrewboudreau:
PR #2451 has merged, installing the independent source-review mechanism. This issue should remain open for activation and proof of enforcement.

The live source checks still report that source review has not been activated. The main validation ruleset currently requires PR validation, but does not yet require Source review or an up-to-date branch. Installing the workflow therefore has not completed this issue.

Follow `notes/agents/SOURCE-REVIEW-CUTOVER.md`: the designated fleet coordinator collects checkpoint/upgrade acknowledgments, drains running leases without discarding work, activates the reviewed policy, and records the queue/policy pins. Then require the app-pinned Source review check alongside PR validation and require up-to-date branches. Prove rejection of an unreviewed candidate, acceptance of an independently reviewed exact head/base, and invalidation after either changes.

Current humanizer reviews are now recorded on #2445, #2446, #2447, #2450 and #2456. Each identifies unresolved source findings, so their previous byte validation must not be treated as source acceptance. Corrections are being independently reviewed before integration.

[...]

PR #2472, opened 2026-09-08
Use natural allocation and readable initialization for daBar_c

lunavyqo:
`daBar_c` now constructs its actor with `return new daBar_c`, using a class-local allocator that forwards to the actor heap. `InitResources` names the pole parameters and calls the local `SetRanges` and `InitClsn` adapters. Profile comments identify behavior/render priorities and the clipping fields; the factory/profile spellings remain explicitly reconstructed.

The current revision restores the shared `dActor_c` and `dCcAc_c` headers to the base. The adapters live in `daBar_c.h`, whose only source consumer is the pole TU. They retain the existing scalar ABI at those boundaries. Independent caller probes support a narrow collision-adapter constraint: two uses of the real `dCcAc_c::Init` declaration produced 0xac and 0xa0 bytes, compared with the 0x94 retail/control function. This is evidence for retaining the measured local adapter, not a general claim about all typed callers.

Validation at `ff7ee8d860c5b58be6b86a0f4ab658f9d75387fa`:

- Independent whole-TU verification: eight functions match, with clean isolation, relocation destinations and order.
- Three consumers that failed under the earlier shared-header change now compile under 2004/b56.
- Emitted metadata: six VERIFIED records, four PARTIAL records, no differences.
- Private PR validation reached terminal SUCCESS for this head. The Source review activation requirement tracked in #2449 remains separate.

This remains a text-only production TU with retained local ABI adapters. No new function, source range, data-ownership claim or contributor identity is introduced by this follow-up.


andrewboudreau (changes requested):
Source review at `81ba4419779fe418f5e487b7a2108bdca090cd27` requests changes.

**P1 — The new shared declarations break unchanged consumers.** `include/dActor_c.h:55` and `include/dCcAc_c.h:43` introduce typed extern-C declarations that conflict with existing declarations of the same C symbols. Independent pinned-compiler controls used identical Fire and WarpKun source with only the header tree changed: both compile against base `c1c8846d` and fail against this candidate with `illegal function overloading`. Private validation has also reached terminal FAILURE. The combined include closure is 1,726 sources, including 104 promoted TUs. Reconcile the full shared-interface scope, or keep the proven factory improvement while retaining the existing local boundaries until that shared repair is ready. Do not weaken shared types or suppress the failure.

The Pole TU itself independently passes all eight functions, isolation, destinations and emission order. That narrow success does not cover its unchanged header consumers.

**Source/provenance corrections:**

- The inline-wrapper comments misread codegen note 6az: it concerns parameter use in the callee, not a general caller-stack ABI restriction. Describe the wrappers as wrappers over the existing scalar definitions, and cite measured caller experiments for any narrower limitation.
- Descriptor +4/+6 hold behavior/render priorities. The fBase constructor gets actor ID from the spawn global; the new packed-profile-ID comment and inherited field name are misleading.
- Restore concise adjacent provenance for the reconstructed factory/profile names and update the manifest's now-obsolete natural-new rationale. The improved compiler-owned factory is real progress.
- The height arithmetic is ten times `(low byte - 10)`, clamped to one unit; it is not height in tenths.

Keep the PR unmerged pending corrected headers, source claims, and fresh proof for all affected consumers. This review did not duplicate the full-ROM build after the independently reproduced compiler failure.

lunavyqo:
Addressed in ff7ee8d86.

**P1.** Reverted `dActor_c.h` / `dCcAc_c.h`. Wrappers are TU-local on `daBar_c` only (`SetRanges` / `InitClsn`). Fire, WarpKun, KoopaShell, MrBlizzard compile against the headers again. Full-ROM validation on this commit is green (8/8 reloc verified).

**Comments.** Wrappers described as wrappers over the existing scalar definitions; caller size measured on this TU. Spawn-info +4/+6 labeled behavior/render priorities. Reconstructed `daBar_c_classInit` / `g_profile_BAR` provenance restored. Manifest factory note updated. Height is `10 * (low byte - 10)` in Fix12, floored at 1.0.

Please re-review `ff7ee8d86`.

andrewboudreau (commented):
Review refresh at `ff7ee8d860c5b58be6b86a0f4ab658f9d75387fa` (base `c1c8846df622f3a49b6b107e0845e9929473d059`).

The P1 shared-header conflict is resolved: `dActor_c.h` and `dCcAc_c.h` match the base again, and fresh 2004/b56 controls for Fire, WarpKun and the actor constructor compile. The remaining daBar header has one consumer. Independent complete-TU verification is 8/8 MATCH with clean isolation/destinations and order; metadata is 6 VERIFIED, 4 PARTIAL, 0 differs. The priority meanings, naming provenance and height explanation are corrected.

The PR body has been updated to match this revision: wrappers now live on `daBar_c`, the collision call is `InitClsn`, and the unsupported general 6az caller claim is removed. A narrower measured statement is justified: independent review compiled two scratch versions using the existing real `dCcAc_c::Init` declaration; named Fix12 values produced 0xac and direct compound values 0xa0, against 0x94 retail/control. Those two results support retaining this local collision adapter without claiming every typed caller is impossible.

For precision, the manifest should call the adapter `daBar_c::InitClsn`; the inlined implicit constructor calls base C2/member C1. The unchanged signed priority storage is not proven by the positive descriptor values; the fBase consumer reads unsigned halfwords.

Private validation is now terminal SUCCESS in the captured head checks. Source-review/workflow and final integration remain separate. This refresh addresses the original findings; it does not claim a new full-ROM run or complete elimination of the retained ABI adapters.

lunavyqo:
Reply to the refresh at `ff7ee8d86`:

**Manifest adapter name.** Done in `1be3746f7`. The note now says `daBar_c::SetRanges` and `daBar_c::InitClsn` (TU-local wrappers over the existing scalar-arg definitions).

**`s16` on +4/+6.** The roles are behavior/render priorities; that part we keep. `fBase_c` copies those halfwords into `u16` priority fields. This profile’s values (`0x011f`, `0x0099`) are positive, so `s16` and `u16` store the same bits. Signedness isn’t proven by those values. Left as `s16` unless you want them switched to `u16` to match the consumer.

Implicit ctor still inlines base C2 / member C1, as the factory note already says.

andrewboudreau (commented):
Refresh: requested `9bbb8d3d7a58fdafa0d41434364384318afe3746` is complete-tree-identical to reviewed `ff7ee8d860c5b58be6b86a0f4ab658f9d75387fa` (tree `8ffd3153d0a46e5ebef4ae77d4faa3d689628bd2`). A subsequent revision, `1be3746f729f5020be608bca2611e1be7d3d78e2`, changes only manifest notes[3] and now correctly calls the adapters `daBar_c::SetRanges` and `daBar_c::InitClsn`.

The PR body accurately describes the local adapters and bounded typed-call measurements. One minor manifest phrase remains: the inlined implicit constructor **calls** base C2/member C1 and installs the vptr; it does not inline those constructor bodies.

All executable source, headers, tooling, manifest structure and ownership policies remain unchanged from the earlier proof; the only current delta is the corrected prose cell. The eight-function TU, three consumer controls and scoped metadata evidence remain pinned to `ff7ee8d860c5b58be6b86a0f4ab658f9d75387fa`. No build was redundantly repeated, and no new private-success claim is made.

Latest captured API head: `1be3746f729f5020be608bca2611e1be7d3d78e2`. Gate snapshot: `[{"name": "review", "status": "COMPLETED", "conclusion": "FAILURE"}, {"name": "Source review", "status": "COMPLETED", "conclusion": "FAILURE"}, {"name": "PR validation", "status": "QUEUED", "conclusion": ""}]`. The exact `9bbb8d3d7a58fdafa0d41434364384318afe3746` Source review failure explicitly reports fleet activation pending. The previous head's private success remains historical evidence. No new source-quality regression was found; source-review and integration gates remain separate.

PR #2474, opened 2026-09-08
Port: 10163 to 10523 (92.9%): game-init arms, the ROM's wireless bring-up, worker and session over a hosted ARM7, card mount, NitroFS init, overlay bookkeeping, the touch panel, the static-initialiser runner, the sound layer, the Game Over scene, the ROM's own game loop as the shipped default, and thread switches on interrupt return; three review rounds answered

tangosdev:
Head moved to 421afeaf7 (port/link100), gated as INT19 (INT13 to INT18 gated the intermediate tips; every gate's evidence is in the run record).

Answers to the second review, both on this head: port/hal/boot_arms.cpp's R2c block reports only the retained pointer state ("after the call data_020a8760 holds row +0x28 of data_020867bc ...") and names neither func_02060398 nor whether the identify ran; the reference gate's 137 findings were run against main's tools/port_refcheck.py on this tree and dispositioned name by name in the comment above: 134 resolve in the linked map from generated objects the checker does not scan (romdata.c.obj, the generated overlay symbol objects, port/unmatched/), one is the alias side of an alternatename whose target resolves, and two are real and recorded at the line (cxx_aliases.cpp:1598, scene_mg_roulette.cpp:241); three narrow checker rules are proposed with negative controls.

Since the first head (861328718), the gated lanes: the two wireless link-state readers (+2); the ROM's own wireless bring-up up to WM_Init registering its reply handler (+12); the ROM's own become-parent and become-child and 128 TUs behind them over a hosted ARM7 wireless stub in hal/wm_arm7.cpp that answers the ROM's commands with the fields the ROM's callbacks read, replies queued from the ARM7's turn, the port's transport as the radio below the ROM's four-wide protocol (+130); the association bitmaps and the ROM's own peer-block reader over the ROM's receive layout (+2); the archive-mount table hosted with LoadArchive answering from it and its twenty-six ROM strings mounted by the generator (+8); the ROM's own NitroFS init off a cartridge-header mirror the port had mapped and never written (+11); the mirror's overlay half, func_02057198 from its own object, and the overlay-bookkeeping chain through the one entry point whose ROM behaviour on the host is an honest early return (+33); twenty-nine object-field pointer-to-member rows in six gates (+24 and +5); the comms fan-out running the ROM's narrow bodies and extending them (0); five frame-loop rungs toward the loop handover including the ROM's own SWAP_BUFFERS now executing once per frame (0); the four-player rollback desync fixed (match-end state carried across a rewind; pre-existing, found by bisect); the four-player proof holding the slot lock and comparing at the settled frame; configure 111 s to 12 s.

Count: 10163 to 10392 of 11328 (91.7%). SHADOWS 15 (the fifteenth is the hosted NONMATCHING hand-asm body func_0206470c the wireless lane named).

Gate INT19 evidence (out/INT19/): reconfigure green (CONFIGURE DEPS 256/5, 10.4 s); guards inferred_stub, alternatename, closestplayer, dsstate (14679, the wireless globals and one host buffer as the bracket's tail), gxband, tailjump OK; linkage --selftest 6/0; battery ALL GREEN at floor 10357 (51 levels including 35, 36 scenes including 6, smoke_player ok, ptr_audit 0, shipcfg ok); proofs green: ipc 22/0, thread, thread_create, save, tail2 OPENING GATE PASS 363, scene_bootpair 34, stage_pause, stage_seat 7, net N0 byte-identical and N1 loopback, vs4 ALL GREEN three times (settled 397/397/396), the sixteen-window ladder with every slot at a distinct position; captured runs show the frame cross-check at 0, the ROM's swap executing 300 of 300, all eight mirror words, the card table 13/13 with both ROM strings, the reworded R2c line, and a solo run asking the radio nothing.

## Round 3 answered (head moved to the gated tip)

Every item of the third review is on this head, each in its own narrow commit:

- `port/hal/cxx_aliases.cpp`: the dead fallback `__ZN5EnemyC2Ev=_func_ov002_020aed98` is removed, not re-pointed. The LHS is defined by a linked ROM TU so the directive could never fire; the RHS has no definition anywhere; ov002's constructor is already routed by the per-source rename at `port/CMakeLists.txt` (`-Dfunc_020aed98=_ZN5EnemyC2Ev`). The alternatename baseline lost the matching row (23 to 22).
- `port/hal/scene_mg_roulette.cpp`: the unused `data_ov006_0213e2f0` declaration is removed; the object is `MgMushroomRoulette_SpawnInfo`, declared and read in `scene_boot.cpp`.
- `port/hal/wm_arm7.cpp`: the lobby walk reads three flags (`src/func_02040014.c` counts 1..3; an x86 control with four flags set produced three reads and three compares). The adjacency requirement is gone; the gap between the two arrays is recorded as an integer-address observation, and the partial MAC-table limitation keeps its own heading.
- `port/tools/vs4_proof.py`: each window's raw last frame is excluded again, combined with the re-simulation bound (`min(last proven re-simulated frame, raw last - 1)`); the repeated-final-frame case reads 399, the no-repeat control 399; the compared prefix is named in the output and the excluded rows are written to a `.tail` file. `port/tools/test_vs4_proof.py` (7 tests) needs no game.
- `port/tools/battery.py`: the shipping configure's fast path is taken only when `build/port-kit/CMakeCache.txt` carries every enforced setting (`PORT_ROM_CLEAN=ON`, `Release`, the static CRT, generator Ninja), read off the same table that builds the configure command; any disagreement names the key and runs the full configure. `port/tools/test_battery_shipcfg.py` (10 tests) carries the wrong-cache fixture; a battery with `PORT_ROM_CLEAN:BOOL=OFF` written into the cache by hand refused the fast path and re-asserted ON.

Reference checker (`tools/port_refcheck.py` at main, paths bound to the tree): at the previous head the two removals read 137 to 135; on this head 146 to 144, the nine-row offset being lane content that arrived after the previous head.

Also on this head since the previous one, each gated by the integration battery, the six proofs, the loopback pair, the four-player proof and the rollback ladder: the ROM's own wireless worker thread created, entered, slept and woken on the parent of a session (its bring-up needed `src/func_0205d23c.c` to take back the two arguments the ARM rides in r0/r1, byte-identical under 2004/b56); the touch panel answered by a hosted ARM7 arm; the ROM's own static-initialiser runner; the sound command layer; the Game Over scene; four I/O placeholders spelled as the registers they stood for (`src/func_02055454.c`, `src/func_02057128.c`, `src/func_02057140.c`, `src/func_0205f650.c`, each byte-verified with the pool word compared for real) and routed through hostgen, which is what lets the ROM's own game loop drive a level frame behind `SM64DS_ROM_LOOP=1`; the conductor's pump spending the VBlank so the ROM's 300-turn drop bound is five seconds again; the re-seat spending the one-shot instead of re-running the wireless bring-up.

## The flip (head moved to 597e584fc)

At this head the ROM's own game loop is the shipped default frame driver for every level and scene; the host loop stays behind `SM64DS_ROM_LOOP=0` for one release. What made it land, each its own gated commit: four I/O placeholders spelled as the registers they stood for (`src/func_02055454.c`, `src/func_02057128.c`, `src/func_02057140.c`, `src/func_0205f650.c`, byte-exact with the pool word compared) so the ROM's VBlank display commit stops writing over the allocator's default heap pointer; the scene loop taking the ROM's phase 7; the graphics-block faces respelled cdecl (the ROM's dispatch is cdecl, the faces were fastcall and read their receiver from a register by accident); the VS menu's last member-pointer site seated as the ROM's own body (entered thiscall-style, and the cdecl body had parked its loop counter over the caller's saved frame pointer); the loop made deterministic under a rollback re-simulation (a replayed frame takes no radio turn, the VBlank count advances once per frame wait, the WM command queue's cursor excluded as a device queue with the reason written beside the sound queue's). Also since the previous head: the wireless file system band hosted at the objects the ROM addresses (+11), five more of the ROM's static initialisers as ROM bodies with the fader tables seated before the ROM's own walk (+7), the ROM's own touch sampler as the single writer of the touch ring, the conductor's pump spending the VBlank so the ROM's five-second drop bound holds, and the re-seat spending the one-shot instead of re-running the wireless bring-up. Gate INT28: battery ALL GREEN on both arms, the six proofs plus the stage proofs, the loopback pair, the four-player proof, the rollback ladder at eight wide (152 pass, the guard row by design), the level-1 selftest position byte-identical on both arms.

## Since the flip (head moved to 91f525963, 10523)

Two gated tips on top of 597e584fc, both pushed to port/link100 and stamped (config/port_linkage.json at 10523 via #2542):

- e4c95585f (10521): a thread switch a ROM body asks for from inside IRQ::VBlankHandler is refused there, the ROM's pending flag is raised, and step 2 of CP15::WaitForInterrupt performs the switch at the handler's return, so data_0209d514 reads 0 at the frame boundary as on the cartridge.
- 91f525963 (10523): the intro Bowser dispatch body func_ov018_02111bf0 and the Jolly Roger Bay alias body func_ov002_020bd664 seated as the ROM's own (+2); the hosted ARM7 posts the port-0xc receive indication once per completed round so the ROM's own receive engine runs on the child every frame; the host's ARMProcessorMode answers IRQ mode while a ROM handler runs, so func_02057f54.c's own line 27 decides the deferral and the refusal above became redundant.

Both gates: build with all six guards, linkage exact with names, battery ALL GREEN on both arms, the six proofs, stage_seat and stage_pause, the captured pair, net N0/N1, vs4, the rollback ladder at eight wide 152 pass and the one by-design guard row, and the ROM-loop captures byte-for-byte on the recorded values.


andrewboudreau (changes requested):
Source review at `86132871880bb29c82954570bd9dd69e76d5c477` requests two narrow corrections in `port/hal/boot_arms.cpp`:

- **P2: the layout check subtracts pointers to distinct array objects.** At line 230, `k[i].p - data_0209caa0` is C++ pointer subtraction across separately declared globals. Linker placement in adjoining sections does not make those globals one C++ array. Compare their integer address representations using the port's existing address-width convention, and keep the refusal path when the offsets differ. The check protecting the 0x32c write should not itself rely on undefined pointer arithmetic.
- **P2: the device diagnostic overstates what pointer equality proves.** At lines 344–347, an unchanged pointer is reported as proof that `func_02060364` was never reached. Re-selecting the same device row also leaves that pointer unchanged. Either report only the observed unchanged pointer or measure the call directly. The nearby row-offset diagnostic likewise needs care when the retained pointer names the separately hosted default row rather than `data_020867bc`.

The source clearly records the remaining host card-bringup bridge; the seven newly linked TUs should continue to be described as linkage coverage, not seven complete removals of host behavior. For readability, the long historical investigation belongs primarily in the existing slice note, leaving the local comments to explain boot order, storage checks and the remaining bridge.

The hosted linkage/smoke checks are green. The failing review workflow currently says source review is not activated, and private PR validation is still queued. I did not rerun the MSVC boot battery during this source review; this is not integration approval.

tangosdev:
Thanks. Both corrections are being taken on the same branch (a follow-up gate on port/hal/boot_arms.cpp: integer address comparison for the layout check with the refusal path kept; the device diagnostic reduced to what the unchanged pointer actually shows, with the default-row case handled; the history moved into port/slice_r2abc.txt and the seven TUs described as linkage coverage). This PR's head will move to the re-gated tip and I will re-request the review then.

andrewboudreau:
Both requested P2 corrections are implemented in [`404af352a70ebee5fef62af629bb8afb774a9691`](https://github.com/tangosdev/sm64ds-decomp/commit/404af352a70ebee5fef62af629bb8afb774a9691), preserving this PR's existing history. A different verifier accepted the narrow correction: warning-free MSVC x86 compilation, 12 focused scenarios and the existing 406-line/60-extern host-loop check passed.

The normal fast-forward push to this PR branch stopped at the local hook. This historical branch lacks the newer validation tools; running the pinned current tools against the candidate then produced 247 reference findings in 46 files, all unchanged by the correction. Confirmed checker defects and the remaining classification work are tracked in #2485. No hook or baseline was bypassed.

The PR head therefore remains `86132871880bb29c82954570bd9dd69e76d5c477`. The correction's source review is not a new whole-PR byte or asset-backed boot pass. After the publication gate is repaired, update this existing branch with the reviewed correction and obtain fresh terminal PR checks.

tangosdev:
Head moved to 428b68ad5 with your two corrections (port/hal/boot_arms.cpp: integer-address layout check, refusal path kept; the device diagnostic reports only the observed unchanged pointer and names the separately hosted default row when that is what it holds; history moved into port/slice_r2abc.txt; the seven TUs worded as linkage coverage). Re-requesting your review. The gate report is in the PR body; the one red (vs4_proof) is the pre-existing match-end rollback desync, under fix separately.

andrewboudreau (changes requested):
Reviewed `428b68ad548edaaf11afee326291da7520ebba5c`, targeting `port-mount-noseat-cluster` at `31f3160466217cd906516d260dba0ac8f76aee26`. This continues the review of `861328718`.

The integer-address layout check, table/default/external-row classification, and unchanged-pointer message address the main earlier findings. A fresh x86 build of the actual boot TU passed 14 controlled scenarios, and the host-loop transcription check passed (406 normalized lines, 60 externs). The changes were adopted independently; preserved correction `404af352` is not an ancestor of this head.

One diagnostic finding remains at [port/hal/boot_arms.cpp:286](https://github.com/tangosdev/sm64ds-decomp/blob/428b68ad548edaaf11afee326291da7520ebba5c/port/hal/boot_arms.cpp#L286): a null pointer says `func_02060398` found no device, while the table-row branch at line 292 says identification ran. [The actual initialization function returns early at lines 11–12](https://github.com/tangosdev/sm64ds-decomp/blob/428b68ad548edaaf11afee326291da7520ebba5c/src/func_02042f68.c#L11). Linking that actual function into the diagnostic harness reproduced both messages with the early-return guard taken and selector stubs set to abort if reached. Please describe the retained null/table pointer state without inferring which lookup ran. This is an inherited remainder, not a new regression.

The repaired reference checker from main `51e74d3e` still reports **137 unresolved references across 26 HAL files** on this exact tree: 9,481 checks, with manifests and CMake checks clean. Comparing normalized reference keys with the preserved earlier report found **zero resolved and zero added**. This failing reference gate remains a publication blocker. These conservative findings need resolution or an evidence-backed disposition before source integration; successful linkage does not classify or waive them. No hook bypass was attempted.

The private PR validation is successful. Source review remains failed because fleet activation is pending. Hosted linkage confirms 10,216/11,328, but this bounded review does not independently accept all behavior in the 54-file WM/card/PMF/fan-out advance or the reported multiplayer/battery history.

tangosdev:
Thanks for the second pass. Taking both: the boot_arms.cpp:286 block will describe only the retained null/table/default-row pointer state without inferring which lookup ran (func_02042f68's early return noted), and the 137 unresolved references across 26 HAL files are being resolved against main's repaired checker (repointed to current names, or reworded where the referent was renamed away), with the checker's headline before and after in the report. The head moves again when both are gated; re-request then.

tangosdev:
Both findings answered on the staged tip (port/l6-refs 427fefb63, gated: linkage unchanged, battery ALL GREEN, guards green, captured run pasted in the lane record).

1. boot_arms.cpp: the whole R2c block now reports only what data_020a8760 holds after the call (null; a row inside data_020867bc at +0x..; the separately hosted default row; neither) and names neither func_02060398 nor "the identify ran". Captured: "[rom-a054] R2c func_02042f68: after the call data_020a8760 holds row +0x28 of data_020867bc -- type 0xd01, size 8192 bytes, kind 2; tag \"ds mario\"".

2. The reference gate, run exactly as documented with tools/port_refcheck.py from main 51e74d3e: before and after identical, 9514 checked, 137 stale across 26 files (your 9481 plus the references WM3's wm_arm7.cpp adds). They are not prose: the checker masks comments, so no rewording moves the number; they are 109 extern "C" declarations and 28 /alternatename spellings of func_/data_ names. Every name was looked up in build/port/walk_window.map: 69 are defined by romdata.c.obj (all literal in port/tools/romdata.py), 60 (51 names) by the generated ovNNN_syms/data objects from port/tools/ovdata.py (each with a row in its overlay's config symbols.txt), 5 are owned in port/unmatched/ (the checker's owner scan is port/hal/*.cpp only), and 3 are not in the map: sub_actors.cpp:1013 data_ov000_020ab9c8 is the alias side of an alternatename whose target is defined (correct as written); cxx_aliases.cpp:1598 func_ov002_020aed98 is a real finding (the _ZN5EnemyC2Ev fallback names a symbol no TU defines and no config records; dormant because src/_ZN5EnemyC2Ev.cpp defines the mangled name); scene_mg_roulette.cpp:241 data_ov006_0213e2f0 is a real finding (the object is hosted under its config name MgMushroomRoulette_SpawnInfo). Both real ones are recorded at the line; changing the identifiers is a code change I held back from a comment-only lane. So 134 of 137 are a checker coverage gap, not staleness. Proposed for tools/port_refcheck.py on main (your file, evidence in the lane's disposition note): widen the owner scan to port/**/*.cpp; treat a name the port's own generators emit as owned (romdata.py's literal list; the overlay's config symbols.txt row when the port generates that overlay's host TU); accept the alias side of an alternatename when its target resolves. Negative controls checked: data_ov006_0213e2f0 has no ov006 config row and still fails; data_ov000_020ab9c8 has a row but no generated ov000 TU, so only the alias rule clears it; func_ov002_020aed98 is in no config file. Expected residue after that patch: exactly the two real findings. Your call whether the two real ones must be renamed before this lands; I can do that as its own commit.

tangosdev:
Head moved to 421afeaf7 (gated INT19, 10392). Both findings from your second pass are on it: the R2c block reports only the retained pointer state, and the reference-gate disposition is in the comment above with the two real findings recorded at the line. Re-requesting.

andrewboudreau (changes requested):
Follow-up on `421afeaf73123870a759547d541535f302e9cd6b`, targeting `port-mount-noseat-cluster` at `31f3160466217cd906516d260dba0ac8f76aee26`.

The R2c diagnostic finding is addressed. I inspected every branch of the revised block: null, table, hosted default and external pointers are now reported as retained pointer states, without claiming a particular lookup ran. The earlier integer-address layout check and its refusal path remain intact.

**Yes: fix the two real references before this lands, in a separate narrow commit.** Recording them in comments does not repair them:

- `port/hal/cxx_aliases.cpp:1608` still contains `/alternatename:__ZN5EnemyC2Ev=_func_ov002_020aed98`, whose fallback target has no definition. Remove the dead fallback if it is no longer intended, or give the intended fallback a real verified owner. Keep the alternatename guard consistent with that actual correction; do not invent a replacement symbol merely to satisfy the checker.
- `port/hal/scene_mg_roulette.cpp:241` still declares `data_ov006_0213e2f0`. The config, generated overlay inventory and existing scene_boot consumer name the object `MgMushroomRoulette_SpawnInfo`. Remove the unused declaration or use that actual object name.

I independently reran the current repaired checker from `cd3a7eb59`, with both its repository paths and srcpath resolver bound to this exact PR tree: **9,520 checked, 137 unresolved across 26 HAL files; manifests and CMake symbols clean**. Both references above are in that report. This is the current tool's count, not a restatement of the earlier 9,481/9,514 runs.

Your evidence that generated objects own many of the remaining names is useful. Please submit generator/unmatched ownership support as a separate tooling PR, with the proposed positive and negative controls and exact generated-module selection. A config row alone must not establish a host definition for an overlay the port never generates. The alias-side rule must preserve checks on the destination. I have not independently accepted the entire 134-reference map census, and the current gate remains red; neither a comment nor successful linkage waives it.

Please fix the two real references now and return the corrected head with the relevant reference/alias/linkage checks. This answers the pending decision in your comment; no further go-ahead is needed for those corrections. This is a follow-up on the inherited blockers, not approval of the additional wireless/NitroFS/frame-loop behavior since the previous review or a fresh MSVC battery run. Preserve the independent Source review and final validation requirements.

andrewboudreau (changes requested):
Reviewed `421afeaf73123870a759547d541535f302e9cd6b`, continuing `428b68ad548edaaf11afee326291da7520ebba5c`, against `port-mount-noseat-cluster` at `31f3160466217cd906516d260dba0ac8f76aee26`.

The earlier device-diagnostic finding is resolved. Fresh x86 checks of the actual boot TU passed 14 controlled scenarios and two controls linking its actual early-return callee. The host-loop transcription check also passed 406 normalized lines/60 externs.

Three new corrections are needed:

- **P2: wireless layout diagnostic:** [wm_arm7.cpp:1017](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/hal/wm_arm7.cpp#L1017) subtracts two separately declared array pointers. Linker adjacency does not make that defined C++ arithmetic. Its [four-entry explanation](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/hal/wm_arm7.cpp#L269) is also incorrect: [the actual callback](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/src/func_02040014.c#L60) starts at 1 and stops at 4, reading three flags. A fresh actual-source x86 control with four flags set produced three comparisons. Correct the explanation and remove the unsupported adjacency requirement, or retain only an explicitly justified integer-address observation. Keep the separate partial MAC-table layout limitation distinct.
- **P2: shipping-cache enforcement:** [battery.py:1204](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/tools/battery.py#L1204) now skips configure whenever `build.ninja` exists, including the only commands enforcing `PORT_ROM_CLEAN=ON`, Release and the static CRT. The actual generated script returned success on a tiny ROM-free Ninja fixture while retaining `PORT_ROM_CLEAN=OFF`, `MultiThreadedDLL` and Debug. Reassert those settings or validate the cache before taking the fast path, with a wrong-cache regression test. CMake's automatic source regeneration does not enforce desired option values.
- **P2: multiplayer comparison boundary:** [vs4_proof.py:366](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/tools/vs4_proof.py#L366) drops the former exclusion of each window's raw last frame. The unchanged function returns 400 for four logs ending at 400 when that final frame repeats; the no-repeat control returns 399. Restore the raw-tail exclusion and test repeated final frames. A repeat records a resimulation, not proof of final confirmed inputs; describe the compared prefix precisely and preserve excluded tail evidence.

The raw reference gate still reports **9,520 checks/137 failures**, with zero normalized findings added or resolved since 428. The author has now supplied useful dispositions: Independent review checked the [exact-head public push map](https://github.com/tangosdev/sm64ds-decomp/actions/runs/34300054633), which supplies 134 reference rows through direct definitions, plus one alias whose target resolves. Two real dormant/stale references remain at [cxx_aliases.cpp:1608](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/hal/cxx_aliases.cpp#L1608) and [scene_mg_roulette.cpp:241](https://github.com/tangosdev/sm64ds-decomp/blob/421afeaf73123870a759547d541535f302e9cd6b/port/hal/scene_mg_roulette.cpp#L241). Their new comments acknowledge them; they still need correction. Generated-object/alias recognition needs an evidence-backed checker follow-up, without banking the 137 failures or treating linkage as a waiver.

The public artifact confirms **10,392/11,328 linked TUs**. Fresh linkage-parser checks 6/6 and asset-catalog tests 11/11 pass. This review does not independently accept the reported INT19 battery, nonzero ARM7 exchanges, 16-window behavior or rollback gameplay. The changed IPC proof checks solo zero traffic; the broader claims still need independent host runs with verified shipping configuration, actual command/reply and receive-layout controls, and rewinds across match-end state. Private NDS validation of the one changed source function does not establish those host behaviors or a full-ROM build pass.

andrewboudreau (commented):
## Reviewed structurally, not line by line: the eight files outside `port/`

Against this PR's own base (`port-mount-noseat-cluster`) the diff is 134 files. Eight are
outside `port/`:

```
src/func_02055454.c  src/func_02057128.c  src/func_02057140.c
src/func_02057198.c  src/func_0205d23c.c  src/func_0205f650.c
tools/asset_catalog.py  tools/test_asset_catalog.py
```

All six `src/` files carry `complete` in `config/arm9/delinks.txt`, so they are not port
scaffolding — the ROM build compiles them and byte-compares them against the cartridge.

### The substance of the six checks out

I verified the three register rewrites against the cartridge independently. `arm9_dec.bin`
at 0x02055454, 0x02057128, 0x02057140 and 0x0205f650 carries literal-pool words
`04000010`, `04000204`, `04000204` and `04000304` — the addresses this PR hardcodes are
the ROM's own.

Worth saying explicitly, because it is the opposite of what a magic address usually
means here: `G` is defined in no `config/**/symbols.txt`. The `extern int G[]` spelling
those four files carried on the base resolved to nothing, and the word it produces is a
relocation — which no byte gate reads. Replacing it with the literal strictly *increases*
what the byte comparison covers. If the project would rather have a named register symbol
than a bare address, that is a decision worth making deliberately, but it is not a
regression against what was there.

`src/func_0205d23c.c` retypes a byte-verified function from `(void)` to
`(unsigned char *, int)`. The callers support it — four of them pass a literal `3` as the
second argument. Two notes on the same change:

- seven caller declarations still contradict the new definition on return type
  (`src/func_0201a2f8.c:15` `void`, `src/func_02042254.c:7` and `src/func_020423dc.c:1`
  and `src/func_020424c0.c:24` `void *`, `src/func_0205cc80.c:3` `int`,
  `src/func_0205d714.c:24` `Obj *`, `src/func_02067bfc.c:22` `int`), and
  `src/func_0205cc80.c:3` declares parameter #2 as `void *` against the new `int`.
  Nothing in the byte gate reads a declaration, so these will not be caught by anything
  that runs here. They are cheap to fix in the same change that makes the definition
  right.

### The gate coverage is the reason to move them

This branch carries four workflow files:

```
.github/workflows/port-linkage.yml   pr-validate.yml   report.yml   update-chaos-data.yml
```

`main` carries nineteen. A `pull_request` workflow runs from the head branch's copy of
`.github/workflows/`, so the fifteen that are absent never ran — including every gate
whose `paths:` filter is `src/**`: tree-shape, source-coverage, converted-ratchet,
dead-references, langmode-ratchet, header-offsets, src-tu-refs, python-names,
rename-ledger, tubuild-conflicts, profile-campaign, port-build and source-review. That is
visible in the check list on this head: six checks, against eleven to fifteen on a
main-targeted PR. `tool-tests.yml` is one of the missing ones, and this PR changes
`tools/asset_catalog.py` (+51) and its test.

The base branch has the same four, so this is the branch's state rather than something
this PR did.

`PR validation` does run, and it passes — but read what it measured:

```
| Matched functions   | 11,189 / 11,347 (98.6%, +0) |
| Matched code bytes  | 2,085,444 / 2,211,124 (94.3%, +0) |
| Full ROM build      | mwldarm failed |
```

with `base and merge share the same pre-existing ROM-build failure`. So module fidelity
was not measured at all on this head; the evidence for the six files is the per-file
relocation linkcheck (6 checked, 6 verified, one slot each), not the full link. And
`+0`/`+0` says this PR moves no decomp coverage — the 360 functions are all port-side.

### Verdict

Needs a change, and the change is a split rather than a rewrite: take the six `src/` files
and the `tools/asset_catalog.py` change into a PR against `main`, where the `src/**` and
`tools/**` gates exist and will run on them. Alternatively merge `main` into this branch
first so it carries those workflows. The port half of the diff I did not read line by
line and have no objection to.

tangosdev:
Taking the split. The source half is #2655 against main, where the src and tools gates run on it.

What that PR found when it diffed the eight files against current main rather than this PR's base: six of them were already there. Commit 6f303aa59 (2026-09-09) replaced the extern int G[] placeholder in func_02055454, func_02057128, func_02057140 and func_0205f650 with the REG_BG0OFS / REG_EXMEMCNT / REG_POWCNT1 macros from nitro/hw/registers.h, the symbolic spelling of the same literal pool words you verified against arm9_dec.bin; func_02057198 is byte-identical on both branches; the two sibling wrappers func_020570c0 and func_020570d8 carry the same macro on main already. So the port branch is behind main on those, not ahead, and the merge of main into the port branch that is running now closes the gap.

The real content of #2655 is the rest of your review: func_0205d23c restored to (unsigned char *, int) with its byte gate held, three of the seven caller declarations corrected (func_0201a2f8, func_0205cc80 including its second parameter, func_02067bfc; the other four already declared a pointer), and the asset catalog's NitroFS table export merged by content onto main's newer copy of tools/asset_catalog.py with four tests added (14 pass).

On the workflows: the port branch will carry main's nineteen once the sync merge lands; the port PR's head moves then and the check list fills out.

PR #2475, opened 2026-09-08
Restamp LINKED off port/link100 861328718: 10170 of 11328 (89.8%)

tangosdev:
Update the LINKED snapshot to **10,170 of 11,328 matched TUs (89.8%)**, measured from `port/link100` at `86132871880bb29c82954570bd9dd69e76d5c477`. Seven additional free-function TUs reach the host binary through save-block clearing and backup-device identification.

Correct the measurement date to 2026-09-08 and the explanation of LINKED growth: reachable actor methods and free functions both count; enrollment alone does not keep an unreachable object from being stripped. The explanatory correction leaves every count, provenance and status value unchanged.

The [hosted Windows linkage build](https://github.com/tangosdev/sm64ds-decomp/actions/runs/34263154949/job/102185874181) reproduces the exact 10,170 / 11,328 count and separately reports 14 SHADOWS. This PR changes only the snapshot and its explanation. Implementation review of the referenced port candidate remains in #2474; LINKED is a TU-reachability measurement, not whole-class reconstruction or complete behavior coverage.


andrewboudreau (commented):
The stamp's **10,170 / 11,328 (89.8%)** count is supported by the hosted linkage run for the referenced port candidate: [completed build and count](https://github.com/tangosdev/sm64ds-decomp/actions/runs/34263154949/job/102185874181). Its output also keeps 14 SHADOWS separate. This is a linked-TU snapshot, not a percentage of fully reconstructed classes or original behavior.

One retained explanation in `config/port_linkage.json` needs correction: the final `_comment` says the count can increase only by seating actor classes and that enrolling free TUs does nothing. This very update adds seven free-function TUs because the live call graph now reaches them. Describe reachability/dead stripping as the condition, rather than actor membership.

The target is main and the numerical evidence is sound for the pinned candidate. Private PR validation is still queued, so this is not final integration approval. PR #2474's source review is separate from this snapshot's measured linkage count.

PR #2483, opened 2026-09-08
ov016/daObjKi_Ita_c cleanup

lunavyqo:
Readable follow-up on the Jolly Roger Bay floating plank. Bytes unchanged: `tubuild.py verify ov016/daObjKi_Ita_c` is 4/4 MATCH, objisolate clean, reloc destinations clean.

- Factory was already `return new daObjKi_Ita_c()`.
- Comments describe the plank. Reconstruction/RTTI/key-function banners are gone.
- Spawn-info `+4/+6` are behavior/render priorities (not packed profile-id).
- No shared-header changes.
- `#pragma long_calls` around InitResources was redundant (same 20-byte ov002 tail call without it) and has been dropped.
- The ov002 setup helper is typed as this leaf plus the model/KCL/CLPS table it already loads. The redundant `(unsigned)` new cast is gone.

Leftover in this leaf, with reasons: `func_ov002_020b5e58` (shared float-board setup lives in ov002; slot 0 is pure virtual on the base) and `data_ov016_02114b8c` (this overlay's three-word file table sits between RTTI and type-name; this TU does not own those words).


andrewboudreau (changes requested):
Review at `34055f96cc81f599f99731e403a56c727dc8e6ea` (same complete source tree as tested `4a235c927f526e40561407b37a49994d91f43ea7`).

**P2 — The new long_calls rationale is not true for this candidate.** Removing only the two pragma lines from an ignored source copy still passes all 4 TU functions, isolation/destinations and emission order. The InitResources strict check is VERIFIED with 0 blind/diffs and the same 20-byte indirect tail call. Please remove the redundant directives and the source/PR-body necessity claim, or qualify a separately pinned historical measurement.

The rest of this narrow cleanup preserves a genuine existing natural-new factory and destructor chain. The priority/range comments and reconstructed-name disclosure are accurate. The shared ov002 setup and separately owned three-word model/KCL/CLPS table remain explicit boundaries; their placeholder spellings are not proof of a compiler limitation or original member name.

Independent candidate proof is 4/4 TU functions; emitted metadata 10 VERIFIED, 4 PARTIAL, 0 differs. Fresh-head private validation and source-review activation remain separate integration gates. No source changes were made by this review.

lunavyqo:
P2: dropped `#pragma long_calls` in `7bb2a1bb9`. InitResources is still 4/4 MATCH, same 20-byte tail call into ov002. Removed the necessity claim from the leftover note and this PR body.

andrewboudreau (commented):
The requested correction is present at `7bb2a1bb94ab92d514561ca460535eeeb26741f9`: both redundant long_calls directives and the unsupported source/PR-body explanation are removed.

I compared the new source against the independent pragma-free experiment. All non-comment tokens match exactly; the only remaining difference is removal of the false explanatory sentence. The complete four-function TU and strict InitResources proof therefore apply to the actual source form now proposed. No unrelated source, header, manifest or tool changes were introduced by this response.

The source finding is resolved. Fresh private validation and fleet source-review activation remain separate landing requirements; this comment does not claim that either has passed for the new head.

Issue #2480, opened 2026-09-08
daObjFl_London_c: replace fabricated render interface and finish bounded source review


andrewboudreau:
Continue `daObjFl_London_c` from merged PR #2444 at main `224e660ea0219d0999c58483eb2dd9b38991f5ef`. The current class source and manifest are unchanged since that merge. Preserve its existing promotion, contributor credit and byte evidence.

- **PR2444-01:** Render still uses fabricated Base/Derived virtual classes, although the inherited `mModel` and real `Model::Render` declaration already exist. Fresh ROM inspection shows its 40-byte Render body has the same dispatch as Clock's existing `mModel.Render(0)`. Probe that expression locally, remove the fabricated types if the full symbol remains exact, and preserve the return value.
- **PR2444-02, London slice:** probe a typed local class pointer in its matrix helper while retaining the existing external symbol and `char*` boundary. Preserve the current include order and selected matrix representation. Do not infer an original method name from helper placement.
- **PR2444-07:** update stale current-source and remaining-work descriptions to describe the actual complete class and remaining raw accesses. Preserve historical proof as historical.

Humanizer/integrator support through @andrewboudreau owns this continuation. Obtain an explicit producer reservation and independent verification for the exact candidate. Required proof covers all eight functions, the complete 196-byte owned data band and its relocations, affected consumers and the production ROM. Keep manual construction and larger interface reconstruction explicitly partial until measured and accepted.

Clock and the held volcano TU have separate findings; neither needs to be restarted for this correction. Shared ledger semantics and credit tooling remain tracked in #2443 and #2433.

PR #2444, opened 2026-09-07
ov022/ov013: promote daObjFl_London_c and daObjClock_c to intact-object TUs (#2231 rescue, batch 2)

andrewboudreau:
Batch 2 of the rescue of closed PR #2231 (branch `cpp/promote-da-obj-lava` @ `8015f733e`), re-derived on today's main and merged with `origin/main` `c27d8ea38` then `115cb5d74` (`--no-ff`, twice), never rebased. Batch 1 was #2435. Independently verified at `450a93a65` (PASS); the figures below are the re-run on the restacked head, with every figure that moved marked MOVED.

**Landed**
- `ov022/daObjFl_London_c` (was coined `LavaBridge`): 8 functions, text `0x02111980..0x02111cac`, `.data` `0x02113f2c..0x02113ff0`, promoted from eight shards into `src/actors/daObjFl_London_c.cpp`.
- `ov013/daObjClock_c` (was coined `ClockPaintingHandShort`): 9 functions, text `0x021113bc..0x021116ac`, `.data` `0x021121a4..0x0211227c`, promoted from nine shards into `src/actors/daObjClock_c.cpp`; the two registry factories keep distinct names (`_CLOCK_LONG` / `_CLOCK_SHORT`).

**Held (in the batch as a rename + text-verified shadow TU only)**
[...]

PR #2487, opened 2026-09-08
Use the real London model interface and named transform fields

andrewboudreau:
Replaces London's fabricated virtual Model receiver with `mModel.Render(0)` and uses the existing class fields in its transform helper. The helper retains its external symbol, `char*` boundary and callers. Descriptor halfwords are named for their observed behavior/render priority roles, preserving their types, layout and values. Header and manifest explanations now describe the actual lifecycle and remaining reconstruction.

This continues the class merged in #2444 and addresses the London findings in #2480. The eight-function intact TU, data ownership, symbols and contributor credit are preserved. Manual construction and remaining raw resource/behavior interfaces stay explicitly partial under #2480; their alternatives are not claimed to be compiler barriers. Original profile signedness and member/factory spellings remain unproved.

Independent verification accepts `a0e810a06017ed6782af57eb369002094200f0ad` against `224e660ea0219d0999c58483eb2dd9b38991f5ef` as pass / partial in task `issue-2480-london-source-review`:

- Eight functions / 812 text bytes: strict VERIFIED, zero blind words or differences; whole-TU isolation, destinations and order pass.
- Complete 196-byte owned data region and 37 relocations exact. Raw metadata separately reports seven VERIFIED and five PARTIAL records, zero differing or unnamed records; partial records remain partial.
- Fresh production ROM: 106/106 modules exact, 11,199 source functions and 25 data claims reproduced, stock SHA-256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`.
- Only the London TU consumes the changed header; its declaration tokens are unchanged. Attribution: zero changed/lost. Port references: 423 checked, zero stale.

The full build retains four unrelated raw metadata differences and nine inherited scratch symbol diagnostics, with zero new production symbol errors. The class was freshly compiled; the full ROM run reused 8,973 content-keyed objects and compiled one. Detailed commands and scope are recorded in `notes/agents/handoffs/issue-2480-london-source-review.md`, with final independent evidence retained by the task.

Fleet source-review activation and terminal private validation for this head remain required before landing. Keep #2480 open for the remaining reconstruction.


andrewboudreau (commented):
## The `+0x04` descriptor rename contradicts this file's own data

`src/actors/daObjFl_London_c.cpp:86` renames the 0x1c actor descriptor's `+0x04`
halfword from `profileIDAndExecuteOrder` to `behaviorPriority`, and
`config/tu_manifest.d/ov022/daObjFl_London_c.json` adds "Actor ID is obtained
separately" as LON-PROFILE-01.

Ten lines below the rename, this file's own initializer puts the profile index in
that field:

```c
g_profile_FL_LONDON = { daObjFl_London_c_classInit, 0x0049, 0x0119, ... };
```

0x0049 is 73, and `symbols/profile_reconstruction_registry.tsv` row 73 is
`FL_LONDON`, pointing at this very descriptor at ov022 0x02113f4c.

`notes/profile-lifecycle-crosswalk.md` records `+0x04` as Tier A for
"registry/profile index **and** behavior/execute-list order" — the value equals
the registry index in all 391 entries — and states that "the apparent profile-ID
field is not separate storage from execute order in the ROM layout." The old
spelling carried both halves; `behaviorPriority` keeps one and drops the one the
descriptor's own value demonstrates.

Nothing catches this: it is a field name in a file-local struct, invisible to the
byte gate, the offset gate and the reference gate. The green checks on this PR
say nothing about it.

Suggested change: keep `profileIDAndExecuteOrder` at `+0x04`, or a spelling that
keeps both roles. `renderPriority` at `+0x06` is defensible on the same note's
Tier A row, but 22 files in the tree spell that field `drawOrder`, so renaming
only this one adds a second convention.

## What I did verify positively against the cartridge

The interface replacement itself is correct, and the ROM proves it rather than
merely failing to contradict it:

* **`Render` is genuinely a virtual dispatch on the inherited `Model`.**
  Disassembling ov022 0x02111aa8 out of the image gives
  `ldr r2, [r0, #0xd4]!` / `mov r1, #0` / `ldr r2, [r2, #0x14]` / `blx r2`. The
  receiver is the subobject at `+0xd4`, the index is 0x14/4 = 5, and the argument
  is null. `include/dBgActor_c.h:28` puts `Model mModel` at `0x0d4`, and
  `include/Model.h` documents slot 5 of `_ZTV5Model` (0x0208e90c) as
  `Render(const Vector3 *)` at 0x02016b78 — the same function
  `config/arm9/symbols.txt` records as `_ZN5Model6RenderEPK7Vector3`. So
  `mModel.Render(0)` names exactly the word the cartridge loads, and the ROM's
  `blx` confirms mwccarm does not devirtualize the form. The `Base`/`Derived`
  stand-in it replaces reached the same slot only by construction.
* **The class vtable is 32 words at ov022 0x02113f70**, i.e. a `dBgActor_c` base,
  consistent with the declared hierarchy.
* **The matrix-helper retyping is offset-exact.** `mModel` at `+0xd4` puts
  `mModel.mat4x3` at `+0xf0` and `m[9..11]` at `+0x114/118/11c`; `mAngleX/Y/Z` at
  `+0x8c/8e/90` and `mPosX/Y/Z` at `+0x5c/60/64` are `dActor_c`'s own fields per
  `include/dActor_c.h`. The old raw offsets and the new typed spelling denote the
  same bytes.

Issue #2433, opened 2026-09-07
prepush_attribution resolves TU overrides by legacy basename while validate_merge and tu_promote key them by surviving symbol


andrewboudreau:
## Summary

`tools/prepush_attribution.py` resolves consolidated-TU attribution overrides by the **legacy shard's basename**, while both gates that actually decide credit — `tools/validate_merge.py` and `tools/tu_promote.py` — resolve them by the **surviving symbol name**. The two agree only when a promotion folds files without renaming symbols. When a promotion also renames (adopting the ROM's own RTTI class name, or turning `func_<addr>` shards into real methods), the local gate cannot see the rows the real gate honours.

The consequence is not cosmetic. On a correctly-attributed tree the local gate exits 1 and names credit as *lost* that nothing lost, and on an unattributed one it under-reports how many members are actually at risk. Two agents hit this from opposite directions on two different classes within a day.

**This issue is a report, not a patch.** Nothing here is fixed.

## What each side keys on

`prepush_attribution.member_overrides_at` builds `symbol -> (path, author)` from every `path#symbol` override key:

```python
# tools/prepush_attribution.py:100-102
path, symbol = key.rsplit("#", 1)
if symbol:
    out[symbol] = (path, author)
```

and then, in `main`, looks that map up by the **basename of the vanished shard** (`basename_key`, line 136) while walking the projected pre-merge lineage:

```python
# tools/prepush_attribution.py:256-263
members = member_overrides_at(args.head)
for name, (old_stem, old_who) in projected.items():
    if name not in after_by_name:
        member = members.get(name)          # `name` is the legacy shard basename
        if member and member[1] == old_who:
            consolidated_ok.append(...)
        elif member:
            changed.append(...)
        else:
            lost.append(...)                # <- everything lands here
```

The merge gate keys the same overrides on the head tree's symbol:

```python
# tools/validate_merge.py:456-457
member_key = f"{path}#{rec['name']}"        # rec['name'] is the SURVIVING symbol
author = (overrides.get(member_key) or overrides.get(path) or ...)
```

and the project's own promotion tool writes them that way:

```python
# tools/tu_promote.py:199-203
stem = pathlib.PurePosixPath(f["legacy_source"]).stem
who = lineage.get(f"src/{stem}")            # author looked up by LEGACY stem
key = f"{p['dest']}#{f['symbol']}"          # key written by SURVIVING symbol
```

[...]

PR #2488, opened 2026-09-08
Cleanup ov018/daSCre_c (still 4/4 MATCH)

lunavyqo:
Readable follow-up on the star-create trigger. Bytes unchanged: `tubuild.py verify ov018/daSCre_c` is 4/4 MATCH, objisolate clean, reloc destinations clean.

- Factory is `return new daSCre_c`. In-class `operator new` takes `unsigned long` and forwards to `_ZN7fBase_cnwEj` (`unsigned int`).
- One shot: if the closest player is strictly within 100 units, spawn actor `0xb2` (the star). Then mark this trigger for removal either way.
- Position is passed as `Pos()` (inline overlay of `mPosX/Y/Z`; a stack `Vector3` DIFF'd).
- Comments describe that. Reconstruction/RTTI/ROM-ordinal banners are gone.
- No shared-header changes.


lunavyqo:
`Behavior` now calls `Pos()` instead of `*(Vector3 *)&mPosX`. A real stack `Vector3` DIFF'd (size + `Vector3` dtor). `Pos()` is an inline overlay of `mPosX/Y/Z`; 4/4 MATCH.

andrewboudreau (changes requested):
Reviewed head `969d498dce5de767eb347861eaee9315183516da`, base
`224e660ea0219d0999c58483eb2dd9b38991f5ef`. The earlier head `6be10910` and the
new `Pos()` delta were inspected and verified separately.

The real `new daSCre_c` factory is a useful improvement: it replaces explicit
constructor calls and vptr writes with compiler-generated construction. Independent
checks verify all four functions (232 bytes), their relocation destinations, complete
TU ordering/isolation, and the changed class header's one consumer. Metadata gives
five VERIFIED records (168 bytes), four PARTIAL records (32 bytes), and no differences.
This was a scoped review; it did not repeat a full-ROM build.

- **SCRE-01 — correct the behavior description.** The comments at
  [source lines 5–6](https://github.com/tangosdev/sm64ds-decomp/blob/969d498dce5de767eb347861eaee9315183516da/src/game/actors/d_a_s_cre.cpp#L5)
  and [header lines 10–11](https://github.com/tangosdev/sm64ds-decomp/blob/969d498dce5de767eb347861eaee9315183516da/include/daSCre_c.h#L10),
  plus the PR body, imply waiting for Mario and deleting only after a spawn.
  `DistToCPlayer()` measures the closest player, and `MarkForDestruction()` executes
  outside the condition. Describe a one-shot check: spawn a star if the closest
  player is strictly within 100 units, then mark the trigger for removal regardless.
- **SCRE-02 — retain the position limitation explicitly.** `Pos()` names the
  scalar-to-vector overlay; it does not reconstruct the base position member.
  The proposed stack `Vector3` alternative was independently tested: `Behavior`
  grows from 92 to 116 bytes and emits a 4-byte `Vector3` destructor. That supports
  retaining this local seam while tracking the base layout separately. It does not
  establish that every alternative is impossible.
- **SCRE-03 — the allocation bridge has a concrete reason.** Pinned 2004/b56 rejects
  an in-class `operator new(unsigned)` and accepts `operator new(unsigned long)`.
  The latter generates an `nwEm` name, while this repository configures the existing
  allocator as `nwEj`. The small inline bridge preserves that destination and the
  exact 48-byte factory. A brief comment explaining this distinction would help;
  removing real construction to satisfy a parser would be a regression.
- **SCRE-04 — the offset gate needs a separate tooling fix.** It exits 1 on
  `static void *operator new(unsigned long size)` as `UNPARSED`; it reports zero
  field mismatches and span 0xd4. The compiled size assertion and affected functions
  pass. Fix the parser with fixtures rather than hide or waive this declaration.
- **SCRE-05 — inherited credit follow-up remains.** The factory's existing override
  still uses the old `daSCre_c_Spawn` key. This PR changes neither attribution nor
  the rename ledger, and its delta checker reports zero changed/lost credits.
  Reconcile the historical keys under [#2433](https://github.com/tangosdev/sm64ds-decomp/issues/2433).

The concise source is easier to review, and its factory uses compiler-generated construction.
Correct the small prose error and record the remaining shared-layout/allocator scope.
The actual header-parser failure and fleet source-review activation hold still need
resolution before merge. This advisory report is not formal queue acceptance.

andrewboudreau (commented):
Reviewed exact new head `bda5a6fe6c06d26ad3718fbbd5c7c2502487bd0b` against
`969d498dce5de767eb347861eaee9315183516da` (base `224e660ea0219d0999c58483eb2dd9b38991f5ef`).

The small code change names the existing `0xf` mask as `kStarIndexMask`. Fresh
baseline/candidate checks verify all four functions (232 bytes), with no blind
relocations or differences. All four bodies and typed relocation records remain
identical. The new source/header exports also pass complete-TU verification:
4/4 MATCH, clean isolation and relocation destinations. The header delta adds
only a comment. Previous proof remains separately pinned to its original head.

- **SCRE-01 remains required.** [Source lines5–6](https://github.com/tangosdev/sm64ds-decomp/blob/bda5a6fe6c06d26ad3718fbbd5c7c2502487bd0b/src/game/actors/d_a_s_cre.cpp#L5),
  [header lines10–11](https://github.com/tangosdev/sm64ds-decomp/blob/bda5a6fe6c06d26ad3718fbbd5c7c2502487bd0b/include/daSCre_c.h#L10),
  and the live PR body still say Mario and imply deletion follows spawning.
  Describe the actual one-shot behavior: spawn when the closest player is strictly
  within 100 units, then mark the trigger for removal regardless.
- **SCRE-03's explanation is still absent.** Preserve the real `new daSCre_c` and
  explain the small allocation bridge: pinned in-class `operator new` takes
  `unsigned long`, while the existing configured allocator uses the `nwEj`
  unsigned-int name. No allocator source change is requested by this finding.
- **SCRE-02, SCRE-04 and SCRE-05 remain as previously recorded.** The position
  overlay is still partial shared-layout reconstruction; the header-parser repair
  is separately reviewed in [#2493](https://github.com/tangosdev/sm64ds-decomp/pull/2493)
  and was not composed into this source candidate; inherited credit reconciliation
  remains under [#2433](https://github.com/tangosdev/sm64ds-decomp/issues/2433).

This read-only refresh made no tracked edits or ownership claims and did not
repeat metadata or full-ROM validation and is not formal queue acceptance. The
verified constant substitution does not resolve the two outstanding prose requests.

lunavyqo:
Addressed in e88cb99ce.

**SCRE-01.** Source, header, and this PR body now say the one-shot: spawn when the closest player is strictly within 100 units, then mark the trigger for removal regardless. `MarkForDestruction()` is outside the range `if`.

**SCRE-03.** `return new daSCre_c` is unchanged. The header now states the allocation bridge: in-class `operator new` is `unsigned long`; `_ZN7fBase_cnwEj` is `unsigned int`.

**SCRE-02 / SCRE-04 / SCRE-05** left as recorded (`Pos()` overlay, header-parser #2493, credit #2433).

4/4 MATCH still.

Issue #2481, opened 2026-09-08
daObjClock_c: correct helper, factory and long-hand source explanations


andrewboudreau:
Continue `daObjClock_c` from merged PR #2444 at main `224e660ea0219d0999c58483eb2dd9b38991f5ef`, preserving its existing intact TU and proof. Retrospective review found these current-source corrections:

- **PR2444-02:** the matrix-helper explanation falsely says the actor angles and model matrix have no modeled owner. Probe a typed local class pointer with the existing external helper boundary and include order; `mModel.mat4x3.t` is already available in this TU.
- **PR2444-04:** the factory prose contradicts the registry's profile-disambiguated names. Retain both existing identities and describe them as reconstructed names.
- **PR2444-05:** `isShortHand` actually tests actor `0x125` / `CLOCK_LONG`. Rename that local truthfully while preserving its type, condition and assignments.
- **PR2444-06:** correct the claim that omitting the `virtual` keyword contradicted the cartridge. Preserve the actual inherited virtual behavior, declaration order and layout.

Humanizer/integrator support through @andrewboudreau owns follow-through. Reserve an explicit continuation before editing, then require independent source review and all nine functions, complete intact text/data, relocations, affected consumers and production-ROM proof. Keep the existing manual factories and remaining interfaces recorded as partial reconstruction until measured. Global ledger and attribution-tool questions remain in #2443 and #2433.

PR #2490, opened 2026-09-08
Use Clock transform fields and correct reconstruction notes

andrewboudreau:
Uses Clock's existing class fields in the transform helper and corrects the local long-hand condition name. The helper now accesses the owned Model matrix, inherited angles and position through a typed receiver while preserving its external symbol, char-pointer boundary, callers and include order.

This continues the Clock class merged in #2444 and addresses the findings tracked in #2481. Factory/profile notes now describe reconstructed names and their registry disambiguation; the header correctly explains inherited virtual overrides. Descriptor halfwords are named for their behavior/render priority uses, preserving storage, order and initializer values without claiming recovered signedness.

Independent verification accepts `2f166831be8a87760023eb45851fbd36f5b96560` against main `224e660ea0219d0999c58483eb2dd9b38991f5ef` as pass / partial:

- Nine functions / 752 text bytes: strict VERIFIED, zero blind words or differences. Whole-TU isolation, destinations and order pass.
- Complete 216-byte owned data region and all 37 relocation destinations exact; scoped raw metadata remains seven VERIFIED / four PARTIAL / zero DIFFERS.
- Fresh packed ROM: 106/106 modules exact, 11,199 source functions and 25 data claims reproduced; stock SHA-256 `d1506e90efae5e2d2cf119926a4ac2a291bd5ca78349d09d5024e1a918c478e8`.
- Header declaration tokens and structural manifest unchanged; only this TU consumes the header. Attribution unchanged; 423 port references checked, zero stale.

The full build retains four unrelated raw metadata differences and nine inherited scratch symbol diagnostics, with zero new production symbol or alias errors. These are distinct from the complete Clock data proof.

Task `issue-2481-clock-source-review` retains the independent verdict; `notes/agents/handoffs/issue-2481-clock-source-review.md` records source changes, commands and limits. Both manual factories, larger external interfaces and original type/signature questions remain explicit partial reconstruction under #2481. No unmeasured alternative is described as a compiler restriction.

Keep #2481 open for that remaining work. Fleet source-review activation and terminal private validation for the final head are required before landing.


andrewboudreau (commented):
## The `+0x04` descriptor rename contradicts this file's own data

`src/actors/daObjClock_c.cpp:94` renames the 0x1c actor descriptor's `+0x04`
halfword from `profileIDAndExecuteOrder` to `behaviorPriority`, and
`config/tu_manifest.d/ov013/daObjClock_c.json` adds the claim "Actor ID is
obtained separately."

Twelve lines below the rename, this file's own initializers put the actor IDs in
that field:

```c
g_profile_CLOCK_LONG  = { daObjClock_c_classInit_CLOCK_LONG,  0x0125, 0x00a3, ... };  // 293
g_profile_CLOCK_SHORT = { daObjClock_c_classInit_CLOCK_SHORT, 0x0124, 0x00a2, ... };  // 292
```

and `daObjClock_c::InitResources` twenty lines further down reads the actor ID and
compares it against exactly that value:

```c
int isLongHand = (int)(actorID == 0x125);
```

`notes/profile-lifecycle-crosswalk.md` records `+0x04` as Tier A for
"registry/profile index **and** behavior/execute-list order" — the value equals
the registry index in all 391 entries, and the note states outright that "the
apparent profile-ID field is not separate storage from execute order in the ROM
layout." The old spelling carried both halves. `behaviorPriority` keeps one and
drops the one the same file demonstrates two ways.

Nothing here catches it: this is a field name in a file-local struct, so it is
invisible to the byte gate, the offset gate and the reference gate. The green
checks on this PR say nothing about it.

Suggested change: keep `profileIDAndExecuteOrder` at `+0x04`, or a spelling that
keeps both roles. `renderPriority` at `+0x06` is defensible on the same note's
Tier A row ("render/draw-list order"), but 22 files in the tree spell that field
`drawOrder`, so renaming only this one adds a second convention.

## What I did verify positively against the cartridge

So this is not a blanket objection — the substance of the PR checks out:

* **`isShortHand` → `isLongHand` is a real correction.** CLOCK_LONG is actor 293
  (0x125) and CLOCK_SHORT is actor 292 (0x124), per
  `symbols/profile_reconstruction_registry.tsv` and the ROM debug strings the
  manifest cites. So `actorID == 0x125` is the long hand, and the branch sets
  `mHandIndex = 0`, which `include/daObjClock_c.h:36` already documents as the
  long hand. The pre-PR name was inverted against the header sitting above it.
  The rodata confirms it independently: reading ov013 0x021116ac out of the image
  gives `{-384, -32}` as `s16[2]`, and -384 is exactly twelve times -32 — a minute
  hand against an hour hand, so index 0 is the long hand.
* **The vtable claim holds.** ov013 0x02112200 is 31 words (a `dActor_c` base),
  with slot 0 = 0x021115cc and slot 3 = 0x02111478, matching the four declared
  overrides.
* **The matrix-helper retyping is offset-exact.** `mModel` sits at `+0xd4`, which
  puts `mModel.mat4x3` at `+0xf0` and its translation row at `+0x114/118/11c`;
  `mAngleX/Y/Z` at `+0x8c/8e/90` and `mPosX/Y/Z` at `+0x5c/60/64` are
  `dActor_c`'s own fields per `include/dActor_c.h`. The raw offsets and the typed
  spelling denote the same bytes.

PR #2496, opened 2026-09-08
func_02071644: recover the real 84-byte body and share one Decimal type with its callers

tangosdev:
Source-only. Eight paths: five under `src/`, one new header, two config records. No tooling, no policy.

## What this recovers

`config/arm9/symbols.txt` declared `func_02071644` as 0x50 and carried its trailing `bx lr`
at 0x02071694 as a separate function, which someone matched with an empty
`void f(void) {}`. This recovers the real body: 0x54 of C that reproduces
0x02071644..0x02071698 exactly under the pinned 2004/b56, with no relocation slot to
wildcard and the `bx lr` as its own last instruction, so the truncated record and its
vacuous source go.

## One `Decimal` type, not four spellings

`func_02071644` takes a `Decimal*`. `src/func_020712a0.c` and `src/func_020715e0.c`
declared the same function `void*` and described the same object through flattened local
views of their own (`struct S`, `struct T`), so one interface had three spellings.

The complete tagged type now lives once, in `include/Decimal.h`, and the four files that
define or declare `func_02071644` all use it: `src/func_02071644.c` (definition),
`src/func_020712a0.c`, `src/func_02071364.c` and `src/func_020715e0.c`. The callers name
`sign`, `exp`, `sig.length` and `sig.text`. **No forward-only type and no cast** — the
directly connected declarations are coherent because they are the same declaration.

The header is deliberately dependency-free, mirroring `include/OamAttr.h`: every field is
a raw C type (`unsigned char`, `char`, `short`), never a project typedef, so it needs no
`#include` and cannot collide with a file-scope typedef an includer already carries.

Two things stay as they are, on purpose and stated in the header:

- `src/func_02071510.c` keeps its own identical local copy. It never names
  `func_02071644`, so no declaration there is made coherent by sharing the type.
- `func_02071698` keeps its `void*` declaration. Its own definition is written over a raw
  byte pointer with its own offsets, so a `Decimal*` there would contradict it.
  Recovering that interface is a separate slice.

## Banner

`src/func_02071644.c` now carries the function's purpose and its measured extent, and a
pointer to where the investigation lives:

```c
// Increment the decimal significand, propagating carry into the exponent.
// The 2004/b56 output occupies 0x02071644..0x02071698, including the final bx lr.
// recovered: real C, no asm hatch; the investigation is in notes/mwccarm-codegen.md 9a(3).
```

## Byte gate

Every changed function, individually, `tools/build_pin.verify` under the pinned 2004/b56
with the relocation-destination check on (`strict=`), so a relocated word cannot pass as a
wildcard:

| function | address | size | result |
| --- | --- | --- | --- |
| `func_02071644` | 0x02071644 | 84 bytes (0x54) | VERIFIED, 2004/b56 |
| `func_020712a0` | 0x020712a0 | 196 bytes (0xc4) | VERIFIED, 2004/b56 |
| `func_020715e0` | 0x020715e0 | 100 bytes (0x64) | VERIFIED, 2004/b56 |
| `func_02071364` | 0x02071364 | 428 bytes (0x1ac) | VERIFIED, 2004/b56 |

`func_02071364` is included because swapping its file-scope copy for the header touches
it; it is byte-identical either way.

Whole tree, this head:

```
source-built functions: 11,201  (2,142,492 / 2,238,108 code bytes, 95.73%)
  reproducing: 11,201
  mismatching: 0
ROM-build analysis: PASS
```

`python tools/check_src_tu_compiles.py`: 177/177.

## The trade, still disclosed

`src/func_02071644.c` on main carries NONMATCHING and HAND-ASM banners and does not count
as matched, so a vacuous four-byte credit is exchanged for a real 84-byte one — numerator
net zero, denominator minus one — and the retired record's attribution moves with it.

## Merge rule

The rule lives in #2498 and landed on `main` on 2026-09-10. Nothing in this PR touches
tooling or policy. Under that rule the retirement rests on six checkable pieces of
evidence: the record is the exact four-byte tail of the range this merge newly compiles;
the cartridge's word at 0x02071694 is `bx lr`; that range is backed by compiled source; no
relocation destination in a validated index over both revisions lands in those bytes; no
symbol in either revision names the address with anything but the placeholder config
derived from the address itself; and the merged function's own object, compiled from this
revision with the pinned compiler, covers 0x02071694 in an ARM EABI `$a` region and emits
that same `bx lr` there. None of that claims the address is unreachable.

## Scoped out, still open

The wider typed experiment that preserved eight routines / 2,720 bytes is research, not
whole-family acceptance. Four upstream files retain separate numeric-argument and
36-digit-layout questions and are **not** settled by anything here: `func_02070b98`,
`func_0206ece0`, `func_0206f46c`, `func_0206f338`.


andrewboudreau (changes requested):
Reviewed source/configuration at `79f9938a0227e98a2ba0077a5a241e0ac3e255b0`, source base `3536f090698231c852ad18dde07f8bf9db2f50f7`.

The decimal carry loop is readable C and uses the same concrete Decimal layout as the adjacent existing `func_02071510` and `func_02071364`. Independent pinned compilation of the full `0x02071644..0x02071698` range is **VERIFIED**, `diffs: []`, `blind: 0`. The source emits 84 bytes. The symbol/delinks delta grows only this range and removes the four-byte record; no adjacent address moves. The earlier investigation in `notes/mwccarm-codegen.md` 9a(3) is preserved and supports this specific boundary recovery.

Before landing:

- Narrow the current source/PR statements about evidence. The output of this particular body under `2004/b56` includes the final `bx lr`; the review does not establish a universal compiler rule for every infinite loop. Zero configured relocation destinations means no references in that index, not proof that execution can never enter an address. Keep the concrete compiler/boundary evidence and put the longer investigation history in the existing notes or PR rather than the source banner.
- Resolve the directly connected caller declaration in `src/func_020712a0.c:1`: it still declares this void helper as returning int. A scratch correction to void passes that caller's full 196-byte strict check. Its `struct S::data[1]` is also indexed through 31 by the current loop; declaring the actual 32-byte digit span is a separately measured improvement, preserving the complete caller. Both changes together also pass, with zero differences/blind relocations. The original writer/source records are preserved; these probes did not edit the PR. Record the caller's array issue as an explicit continuation if it stays outside this function-boundary slice.
- Keep the validator policy under its separate PR #2498. Independent review has already reproduced invalid/missing relocation input being accepted as zero callers. Its test suite passing does not settle that defect. Refresh the source PR's description after the tooling split/composition and rerun the actual private validator; the current source proof is not acceptance of the merge-rule exception.

This is a bounded source review and fresh strict function proof, not a full ROM build, private-validator success, or live Source review acceptance. The retired record's attribution change is already disclosed in the PR and must remain visible in the final report.

tangosdev:
Taking all three: the banner and PR text narrowed to what this body emits under 2004/b56 and to what the index shows; src/func_020712a0.c's int-returning declaration of this void helper corrected (and the 32-byte digit span measured against the byte gate, taken if it agrees, else recorded as a continuation); the description refreshed to depend on #2498 landing first with the validator re-run after. The attribution change stays disclosed.

tangosdev:
Head 62695841f: source and config only now (no tools/ files; #2498 carries the rule and this PR depends on it landing first, after which the validator must be re-run). Banner and text narrowed to what this body emits under 2004/b56 and to what the index shows. src/func_020712a0.c: the int declaration corrected to void and the digit span declared as data[32]; byte gate 48 of 49 words exact with the one relocation slot wildcarded, linkcheck VERIFIED on both functions. Continuation recorded, not taken: the caller still spells the first parameter void* where the definition says Decimal*, because there is no shared header. The attribution change stays disclosed.

andrewboudreau (dismissed):
Follow-up review of `62695841f88d2cfefedd5307f6c021f5d91c53cf` against source base `cd3a7eb5966a8758722bde1ec78f61abc4b5c126`. The requested revisions are addressed: the wording is limited to this body's pinned compiler output and the configured relocation index; the caller declares the helper as void and its digit array as 32 bytes; the tooling changes are isolated in #2498; and the attribution effect remains disclosed.

Fresh independent verification in a separate wired worktree:

- `python tools/prepush_linkcheck.py --range cd3a7eb5966a8758722bde1ec78f61abc4b5c126..HEAD`: 2 checked, both VERIFIED, zero warnings/blocking results.
- `python tools/port_refcheck.py`: 423 references, zero stale.
- The complete source/config diff retains the 0x02071644..0x02071698 extent, removes only the severed four-byte record, and leaves the neighboring address unchanged. The readable carry loop and concrete Decimal layout are consistent with the prior source review.

No further source edit is requested for this boundary-recovery slice. The already disclosed caller `void *` versus local `Decimal *` declaration remains a separate interface reconstruction follow-up.

Approval clears my earlier requested revisions; it does not clear the integration prerequisites. #2498 still has two reproduced guard defects from my follow-up review today and must land first. Then rerun private PR validation and record independent Source review for the final live head/base under the now-active queue policy. The current Source review check has no accepted task covering these five paths, and the private check is still red under the old merge rule. Do not merge or override either check on the strength of this review. This pass reran the affected functions and reference gate, not a fresh full ROM build.

andrewboudreau (dismissed):
Reviewed `62695841f88d2cfefedd5307f6c021f5d91c53cf`. The return declaration and `data[32]` corrections are present, and the PR now contains five source/config paths with no tooling. Fresh independent checks verify the full 84-byte helper and 196-byte changed caller, plus the other 428-byte and 100-byte direct callers, with zero byte differences or blind relocations. The actual range prepush check verifies both changed functions. Those earlier findings are resolved.

Two source-review items remain:

- **Use the real Decimal contract in the callers.** The definition takes `Decimal*`, but `func_020712a0.c:1` and `func_020715e0.c:3` still declare the same function with `void*`. The complete tagged Decimal already exists in three adjacent files. A fresh pinned-compiler experiment replaces the two callers' flattened S/T views with that exact existing type and named `sign`, `exp`, `sig.length` and `sig.text` fields; both preserve their full 196/100 bytes and relocation destinations. A shared header is a maintainability choice, not an unmeasured compiler obstacle. Do not hide the mismatch behind a forward-only type and cast. Make the directly connected declarations coherent before accepting this source slice.
- **Trim the source banner and retitle around the final source-only scope.** The narrower compiler/index claims are an improvement, but 26 lines still recount who matched a stub, the earlier register-allocation claim, and a discarded draft, with all-caps emphasis. Keep the function's purpose and a short measured extent note in source; the investigation belongs in the already linked notes and PR. For example: `// Increment the decimal significand, propagating carry into the exponent.` followed by `// The 2004/b56 output occupies 0x02071644..0x02071698, including the final bx lr.` The title still describes changing the merge rule even though that change now lives in #2498.

The wider typed experiment preserves eight routines / 2,720 bytes with zero differences or blind relocations, including real Decimal locals and struct assignment in the power helper. It is research, not whole-family acceptance. Four observed upstream files retain separate numeric-argument and 36-digit-layout questions: `func_02070b98`, `func_0206ece0`, `func_0206f46c`, and `func_0206f338`. Keep those findings explicitly scoped rather than implying that the eight-routine result settles the entire formatting interface.

The current private check `102298185271` still fails the removed matched-record / coverage-denominator policy while reporting both changed functions verified and 106/106 exact modules. #2498 remains a separately reviewed dependency, not an accepted exception merely because its proposed rule prints Passed. The retired record's attribution remains disclosed. No queue acceptance, full independent ROM build or merge is claimed by this read-only review.

andrewboudreau (commented):
## `PR validation`: one cause, two effects — and the cause is not in this diff

The failure text names two things, but they are not two problems.

```
Validation failed: lost 1 matched function(s); function/byte coverage denominator changed
```

Both come from a single `repartition is None` in `build_report`. `classify_merge` returning
None leaves `arm9:0x02071694` in `removed` (first reason) *and* leaves the
`totalFunctions` 11,345 → 11,344 move with nothing to excuse it (second reason). The
denominator line is an effect of the classification being refused, not an independent
finding: there is no edit that fixes one and leaves the other.

### The run that failed did not have the rule

Two things in that report say so:

- it carries the warning `1 address range(s) left the byte-verified set while enrolled
  totals held steady: arm9:0x02071694-0x02071698`, which is exactly what this branch's
  own commit *"do not report a grown range as one that left the verified set"* stops
  emitting — 0x02071694-0x02071698 is inside the grown range 0x02071644-0x02071698;
- it carries no `Claims absorbed by the merge` row at all.

It ran on 2026-09-09 against the then-tip of `main`. The tooling half of this work
landed on `main` on 2026-09-10.

### Against today's main it passes

I took this head, merged today's `main` into it, and ran `main`'s
`tools/validate_merge.py` over the resulting merge commit. Exit 0, no `Validation
failed:` line, and the report now reads:

```
| Claims absorbed by the merge | 1 function(s); ... |
| Symbol-table merge | totalFunctions -1 over identical covered bytes |
| Byte-verified code bytes | 2,128,408 / 2,211,124 (96.26%, +80) |
```

`+80` is the arithmetic the rule requires and not a coincidence: new extent 84, less the
4 the base already compiled.

### Each condition, checked separately rather than inferred from the pass

Walking `classify_merge`'s six clauses on this head with the ROM and the pinned compiler
present:

- (a) the record ends at 0x02071698, which is exactly where the new `complete` range
  `0x02071644..0x02071698` ends — it is the tail, not something inside it;
- (b) size 4, and the cartridge's own word at 0x02071694 is `e12fff1e`;
- (c) `src/func_02071644.c` is in the compiled set — no `NONMATCHING`, no `HAND-ASM`;
- (d) `0` relocation destinations land in `[0x02071694, 0x02071698)` in a *validated*
  index over both revisions;
- (e) no symbol in either revision names 0x02071694 with anything but the placeholder
  derived from the address. I checked specifically for a zero-size alias sharing that
  address in both `config/arm9/symbols.txt` revisions — there is none, so nothing is
  being shadowed out of the covered set;
- (f) the object `src/func_02071644.c` produces under the pinned 2004/b56 covers
  0x02071694 in an ARM EABI `$a` region and emits `e12fff1e` there, so it is an
  instruction and not a literal-pool word with a return's bits.

Separately: the three files that declare `func_02071644` after this change
(`src/func_020712a0.c:14`, `src/func_02071364.c:12`, `src/func_020715e0.c:19`) all spell
it `void func_02071644(Decimal *, int)`, which is the definition at
`src/func_02071644.c:19`. The retired `func_02071694` row leaves `symbols.txt` in the
same commit that deletes its source, and (d) above is the check that nothing else
resolves by that name.

### What it takes to go green

Nothing in the source. Get a head that contains today's `main` in front of the validator
and re-run it. The `Source review` failure on this head says the same thing in its own
words (*"PR head does not contain the live target base; restack"*), so one restack clears
both.

### One secondary point, on the diagnostic rather than this PR

`build_report` explains an unavailable exception only through `reloc["valid"]`,
`rom_word.missing` and `code_owned.missing`. Those two `.missing` sets are populated on
call, so a refusal that happens before either reader is reached prints nothing beyond
`lost 1 matched function(s)`. That is what happened here, and the result is that the
failure text cannot distinguish *"the rule looked and refused"* from *"the rule was not
there"* — which is the whole diagnosis on this PR. A line naming which clause refused
would have made this a one-minute read.

tangosdev:
Head ce47c49d1. Both remaining items are on it, and the branch now carries main so the compare is source and config only (eight paths, no tooling).

Decimal contract: include/Decimal.h is a dependency-free header holding the tagged struct once; src/func_02071644.c, src/func_020712a0.c, src/func_020715e0.c and src/func_02071364.c include it instead of carrying four copies, and the two callers take Decimal* with the named fields. The header route was chosen over a per-file copy because it mirrors include/OamAttr.h, the tree's existing shape for a small shared POD; src/func_02071510.c keeps its own copy because it never declares or calls func_02071644. All four functions still match at 2004/b56 with only the call relocation slot wildcarded (func_02071644 0x54, func_020712a0 0xc4, func_020715e0 0x64, func_02071364 0x1ac); the two neighbours func_02071510 and func_02071698 are unchanged and still match. prepush_linkcheck over origin/main..HEAD: 4 checked, 4 verified, 0 blocking.

Banner: src/func_02071644.c now carries the purpose line, the measured 2004/b56 extent (0x02071644..0x02071698 including the final bx lr) and one provenance line; the investigation stays in notes/mwccarm-codegen.md 9a(3) and this PR.

The validator run you traced to the pre-#2498 rule will re-run on this head.

andrewboudreau:
## Both requested items, and the measurement behind each

Head `971a74401`, restacked onto today's `main` `1224afd11` (the previous head was 143
commits behind, which is what the `Source review` check's *"PR head does not contain the
live target base; restack"* was reporting).

### 1. The real `Decimal` contract in the callers

The complete tagged type existed in `src/func_02071364.c` and `src/func_02071510.c` (and,
on this branch, in `src/func_02071644.c` itself). It now lives once, in
`include/Decimal.h`, and the four files that define or declare `func_02071644` all use it:

| file | before | now |
| --- | --- | --- |
| `src/func_02071644.c` | file-scope copy of the tagged type | `#include "Decimal.h"` |
| `src/func_020712a0.c` | `extern int func_02071644(void *out, int n);` + local `struct S` (`b0`/`pad1`/`s2`/`b4`/`data[1]`) | `extern void func_02071644(Decimal *result, int length);`, fields named `sign`, `exp`, `sig.length`, `sig.text` |
| `src/func_020715e0.c` | `extern void func_02071644(void *thiz, int val);` + local `struct T` (`char pad[4]; unsigned char field4;`) | `extern void func_02071644(Decimal *value, int length);`, `value->sig.length` in place of `field4` |
| `src/func_02071364.c` | file-scope copy of the tagged type | `#include "Decimal.h"` |

No forward-only type and no cast anywhere: the declarations are coherent because they are
literally the same declaration. `struct S` and `struct T` are gone rather than kept behind
a compatible-layout argument.

Header shape follows `include/OamAttr.h` and is deliberately dependency-free — every field
is a raw C type, never a project typedef — so it needs no `#include` and cannot collide
with a file-scope typedef an includer already carries.

Two deliberate non-changes, both stated in the header rather than left to be rediscovered:

- `src/func_02071510.c` keeps its identical local copy. It never names `func_02071644`, so
  no declaration there is made coherent by sharing the type, and folding it in would add a
  fifth byte-verified function to this slice for no interface gain.
- `func_02071698` keeps its `void*` declaration. Its own definition is written over a raw
  byte pointer with its own offsets, so a `Decimal*` there would contradict it. Separate
  slice.

One correction to a claim this branch had introduced: the header's comment said the tagged
type had been *"repeated by hand, identically, in each of those four files before this
header."* That was false — only `func_02071364.c` carried this exact layout; the other two
callers carried flattened partial views, which is the mismatch the header removes. The
comment now says what was actually there (`971a74401`, comment-only).

### 2. Banner and title

`src/func_02071644.c` is down from 26 lines to three — your two verbatim, plus a pointer to
where the investigation lives:

```c
// Increment the decimal significand, propagating carry into the exponent.
// The 2004/b56 output occupies 0x02071644..0x02071698, including the final bx lr.
// recovered: real C, no asm hatch; the investigation is in notes/mwccarm-codegen.md 9a(3).
```

Nothing about who matched the stub, the register-allocation claim, or the discarded draft
remains in source, and there is no all-caps emphasis. The `recovered:` line matches the two
other files in `src/` that use it. Retitled to *"func_02071644: recover the real 84-byte
body and share one Decimal type with its callers"* — source-only scope, no merge rule. The
PR body is refreshed to match.

### Per-function byte evidence

`tools/build_pin.verify` on this head, one call per function, pinned 2004/b56 with the
relocation-destination check on (`strict=(reloc_audit, name_index, config_relocs,
sym_index)`), so a relocated word cannot pass as a wildcard:

```
VERIFIED  func_02071644  0x02071644   84 bytes (0x54)  [src/func_02071644.c]  -> 2004/b56
VERIFIED  func_020712a0  0x020712a0  196 bytes (0xc4)  [src/func_020712a0.c]  -> 2004/b56
VERIFIED  func_020715e0  0x020715e0  100 bytes (0x64)  [src/func_020715e0.c]  -> 2004/b56
VERIFIED  func_02071364  0x02071364  428 bytes (0x1ac) [src/func_02071364.c]  -> 2004/b56
```

Typing the callers is free, as your experiment measured: 196 and 100 bytes preserved with
their relocation destinations intact. The 428-byte `func_02071364` is in the list because
swapping its file-scope copy for the header touches it; it is byte-identical either way.

Whole tree, same head, wired worktree with the ROM dump and the pinned compiler:

```
python tools/rombuild.py -j 8 --no-rom
source-built functions: 11,201  (2,142,492 / 2,238,108 code bytes, 95.73%)
  reproducing: 11,201
  mismatching: 0
ROM-build analysis: PASS
```

The `dsd check symbols --fail` FAIL in that run is inside the **BASELINE CONTROL** block —
no-TU-substitution, pre-existing on `main`, nine error lines none of which is in this
address range. `python tools/check_src_tu_compiles.py`: 177/177.

### Still out of scope

The merge rule stays in #2498, which landed on `main` on 2026-09-10; nothing here touches
tooling or policy, and I am not treating #2498's tests as settling this PR. The eight-routine
/ 2,720-byte typed experiment remains research, not whole-family acceptance:
`func_02070b98`, `func_0206ece0`, `func_0206f46c` and `func_0206f338` keep their separate
numeric-argument and 36-digit-layout questions open.

Not merged, not pushed to `main`, and your review is left standing for you to dismiss.

andrewboudreau:
## Addendum: the restack cleared one of the two checks, not both

Your read was *"one restack clears both."* On head `971a74401`, base `1224afd11`:

- **`Source review`** — the base complaint is gone. The check's own JSON now reports
  `base = live_tip = 1224afd11`, so *"PR head does not contain the live target base;
  restack"* is no longer what it says. What it says now is
  `"Independent source review is missing or invalid"` with `accepted_tasks: []` — the
  queue prerequisite you named yourself, not a source defect, and not something this
  branch can satisfy.
- **`PR validation`** — still `failure`, same summary as the 2026-09-09 run:
  `Validation failed: lost 1 matched function(s); function/byte coverage denominator changed`.

That second one I cannot reproduce, and the check's own report body argues against it.
Reporting it rather than working around it.

**The report the failing run published contains no failure evidence.** No
`Claims absorbed by the merge` row — but also no `1 address range(s) left the byte-verified
set` warning, which is the line `main`'s current validator no longer emits for a grown
range. Its warnings are one line: `contributor attribution was lost, not a blocker`. Its
tables read `Relocation check | 4 checked; 4 VERIFIED`, `Module fidelity | 106/106 exact`,
`Code linked from verified source | 11,201 functions`, and all four changed files
`✅ verified`.

**`main`'s own `validate_merge.py` accepts this head.** In a wired worktree, over a
CI-order merge (base first parent) of `971a74401` onto `1224afd11`, using
`tools/validate_merge.py` as it stands on `main` after #2498:

```
python tools/validate_merge.py --base origin/main --head <merge> \
  --require-merge-commit --expected-pr-head 971a74401bce651deb09d6709389036781503d3a
exit 0
summary : Committed merge passes; 1 lost credit(s) noted, not a blocker.
matchedAbsorbed : [{"id": "arm9:0x02071694", "path": "src/func_02071694.c", ...}]
repartition     : {"kind": "merge", "functionDelta": -1, "removed": ["arm9:0x02071694"],
                   "absorbed": ["arm9:0x02071694"], "newExtent": 84,
                   "absorbedBytes": 4, "sourceByteDelta": 80}
```

So the tree's validator classifies the merge and the worker's does not, on the same head
and the same base. I am not proposing an edit for that — the rule stays in #2498 and this
PR stays source-only — but the two disagree, and the disagreement is not visible from the
failure text.

Which brings back your secondary point, now with a second instance: none of the three
"say WHY" reasons (`relocation evidence invalid`, `ROM image unavailable`,
`compiled code ownership unavailable`) appears in the worker's summary. Under
`build_report`, a refusal that reaches any of those readers names one. A refusal that
reads nothing names none. So the failure text is again consistent with a clause refusing
early *and* with the rule not being consulted, and there is no way to tell which from the
outside. A line naming the refusing clause would settle it in one read.

andrewboudreau:
**This PR did not lose a matched function. The validator was missing a Python package and could not say so.**

`PR validation` reports:

> `Validation failed: lost 1 matched function(s); function/byte coverage denominator changed`

That is a false red, and I can now reproduce it exactly. The chain:

1. `tools/evidence_rom.py` imported `yaml` at module scope and `sys.exit`ed without it — though yaml is used in one place only, `main()`.
2. `validate_merge` imports that module for one function, `is_unconditional_return`, which reads a capstone instruction and never touches yaml.
3. `_arm_decoder()` catches `SystemExit`, so on the worker image — capstone, no pyyaml — the decoder read as unavailable.
4. `_is_return_word(0xe12fff1e)` (the `bx lr` at `0x02071694`) returned `False`, and the matched-loss exception refused at condition **(b)**.
5. Nothing in the refusal branch names the decoder, so the report blamed the PR.

Run on your committed probe merge `f5520a45` (base `1224afd11`, head `971a7440`), one command, only the packages and the two source files varying:

| run | exit | status |
|---|---|---|
| current `main`, `yaml` hidden | 1 | `Failed` — the worker's string, character for character |
| **with the fix, `yaml` hidden** | **0** | **`Passed`** |
| with the fix, `capstone` hidden | 1 | `Failed` — *and now says* `ARM decoder unavailable ...` |
| with the fix, nothing hidden | 0 | `Passed` |

With the decoder present the classification is the one this PR argues for:

```
{kind: merge, functionDelta: -1, removed: [arm9:0x02071694],
 absorbed: [arm9:0x02071694], newExtent: 84, absorbedBytes: 4, sourceByteDelta: 80}
```

The fix is **#2682**. Once it lands, re-run validation here.

Two corrections to things said earlier in this thread, both mine:

- Your citation of `$? == 0` was **right**. I claimed `validate_merge.py` exits 0 on FAIL; that reading came from a `| tail` pipeline reporting *tail's* status. Measured unpiped: `Passed` → 0, `Failed` → 1.
- The "the worker is running stale `tools/`" theory was **wrong**. `run-worker.sh:288` runs `tools/` from the merged tree, and this PR touches no file under `tools/`.

One thing this does not clear: `run-worker.sh` does not pass `MWCCARM_LAUNCHER` to `validate_merge`, unlike the three sibling invocations beside it. Condition (b) was refusing before condition (f) could compile anything, so it has never been reached. After #2682 it will be. That is a validator-repo change and is raised with Tango separately — so if this PR goes red again after #2682 with `worker error: validate_merge produced no report`, that is the reason, and still not yours.

PR #2498, opened 2026-09-08
validate_merge: the merge rule's arithmetic fix and the matched-loss exception, as ruled

tangosdev:
# The matched-loss exception, and the evidence it now needs

Three corrections to the generalized exception, one commit each, each with the negative
control that reproduced the defect before it was fixed. Every one of the reproductions
below was driven through the real `build_report` on a committed tree, not through a
hand-built dict.

## A. The overlay word reader took its base from the first function

`_module_image` read an overlay at `_rev_enrolment(rev)[0][module][0]`, which is a list
of FUNCTION addresses. `modules._overlay_base`, the derivation the comment claimed to
follow and the one the byte gate, `match.py --module ovNNN` and `evidence_rom` all use,
takes the lowest address of ANY symbol. A module whose lowest symbol is data therefore
answered every query with an earlier word.

`_module_symbol_base` is the revision-pinned equivalent of that all-symbol base, and
itcm and dtcm are readable now too, from the two places `modules.py` looks. ARM9 keeps
its fixed 0x02004000 load base, which is not its lowest symbol (`overlay_100` at
0x00000064 is).

Negative control, `DataLeadingOverlayImageBase`: an ov001 with four bytes of data in
front of its first function, `bx lr` at 0x020ab150 and `mov r0, #0` at the retired
record 0x020ab154.

| test | before | after |
| --- | --- | --- |
| `test_the_reader_answers_from_the_lowest_symbol_not_the_first_function` | reader returns 0xe12fff1e | 0xe3a00000 |
| `test_a_non_return_tail_in_a_data_leading_overlay_is_refused` | Passed, `reasons: []`, one absorbed | refused, `lost 1 matched function(s)` |
| `test_the_same_overlay_fixture_with_a_real_tail_return_still_lands` | (positive control) | lands |

Measured on this tree: 72 of the 73 modules that declare functions have equal bases
either way, so no live read moves. The exception is ARM9, which does not use this path.

## B. Zero functions did not make an empty relocation file complete

The inventory check exempted an empty `relocs.txt` whenever the module's symbol table
declared no function. A module with no code can still hold a table of callback pointers
in its `.data`, and every one of them is an incoming reference the index would then not
know about.

`_empty_relocs_defect` asks what dsd wrote instead, in this order: a module declaring a
function is refused as before; otherwise the module's own section inventory decides, and
a module whose non-`bss` sections are all zero-length really does reference nothing
(`.bss` is uninitialised and occupies no cartridge bytes); a module WITH initialised
bytes has to be read, and passes only when those words are all zero. No inventory, or no
image to read one, is a defect naming the file it wanted.

Negative control,
`test_a_data_only_module_with_an_empty_relocation_file_fails_closed`: a data-only ov001
added to the real absorption fixture, with a nonempty four-byte `.data` holding the
retired ARM9 address and an empty `relocs.txt` in both snapshots. Before: `Passed`,
`relocationEvidence.valid: true`, one absorbed record. After: refused, naming
`config/arm9/overlays/ov001/relocs.txt is empty`.
`test_a_data_only_module_whose_data_is_zeroed_is_still_legitimate` is the other half:
the same layout with a zeroed word still lands. Four `RelocIndex` controls cover the
bss-only inventory, the missing inventory, the unreadable image and the zeroed data.

Live census unchanged: **92 destination buckets, 96,395 destinations, 0 defects.**
Disclosed, and in the docstring: on a checkout with no `extracted/` it is the same 92
and 96,395 with **one** defect, `config/arm9/dtcm`, whose 0x20 of `.data` can only be
shown harmless by reading it (it is all zeroes). That makes the exception unavailable
there, which is the position it is already in without the ROM image, and it changes no
other number in the report.

## C. A four-byte tail with return-shaped bits is not an owned epilogue

Three independent fixtures returned `Passed`, no reasons and one absorbed record at
92f9ad43b. Reproduced here first, all three, then refused:

1. **A named callback at the covering range's tail.** The existing control put it
   mid-range, so it tested position rather than identity.
2. **A compiler-generated literal pool.**
   `unsigned int func_02071644(void) { return 0xe12fff1e; }` under the pinned 2004/b56
   emits `e59f0000 e12fff1e e12fff1e`, and the object's mapping symbols put `$d` at
   offset 8. The final word has a return's exact bits and is a constant.
3. **`0xe8908000`, `ldm r0, {pc}`**, together with `ldmdb r0, {pc}` and
   `ldmib r0!, {pc}`.

Two new clauses, and a return classifier that no longer diverges from the helper it
cites:

* **(e) identity.** `_named_symbol` refuses when any symbol in EITHER revision names the
  retired address with anything but the placeholder config derives from the address
  itself (`func_02071694`, `func_ov006_020cb030`). `AutoloadCallback` is a name; a
  placeholder is not.
* **(f) code ownership.** `_compiled_code_reader` compiles the merged source at the
  revision under judgement with the pinned compiler and reads the ARM EABI mapping
  symbols of the object it produces. The retired address must lie inside the merged
  function's own compiled output in a `$a` region, and the word the object emits there
  must be the cartridge's word. `$d` is refused. This tool has no object of its own --
  the worker hands it `rombuild.py`'s JSON, which says whether a module reproduced, not
  where inside a function the instructions stop -- so it builds one.
* **the return set** is now classified by `evidence_rom.is_unconditional_return`,
  imported rather than re-derived, then narrowed to `bx lr` and a stack pop to pc.
  `ldm r0, {pc}` and its db/ib spellings are gone because the base register must be
  r13; `mov pc, lr` is gone because its return-ness is a fact about lr, not about the
  word; `ldr pc, [sp], #4` stays refused on the block-transfer bits. (The comment
  claiming capstone decodes that one as LDR was wrong. It renders it `pop {pc}`.)

New refusing tests: `test_a_named_callback_at_the_tail_is_refused`,
`test_a_literal_pool_at_the_tail_is_refused`,
`test_an_object_that_emits_a_different_word_is_refused`,
`test_a_block_load_through_a_register_other_than_sp_is_refused`,
`test_without_code_ownership_nothing_matched_may_leave`, and the same three through
`build_report`: `test_a_named_callback_at_the_tail_fails_the_report`,
`test_a_literal_pool_at_the_tail_fails_the_report`,
`test_a_block_load_through_r0_fails_the_report`,
`test_without_code_ownership_the_report_fails_closed`.

Two classes run the real compiler rather than a stub, and skip where it is absent:
`CompiledCodeOwnership` builds the literal-pool object and asserts offset 8 is data
while offset 4, the same bits, is an instruction; `LiteralPoolThroughBuildReport` drives
the literal-pool input end to end through `build_report` with no stub anywhere,
refusing the pool and still landing a body that really ends in that instruction. The
report fixtures elsewhere pass a stub, because the runner that runs this suite has no
toolchain and a stub is how the WIRING gets tested there; `code_evidence` is a seam, not
a policy knob, and no value of it grants the exception.

`tools/requirements.txt` gains pyyaml: `evidence_rom` imports it at module level, so the
suite that exercises the shared classifier needs it.

The warning and the report row now state what the guard establishes and stop there --
that a function this merge compiles owns those four bytes and emits the cartridge's own
return at them, that config has no independent name for them, and that no relocation
destination in a validated index points at them. Reachability is named as not claimed.

## Runs

`python -m unittest tools.test_validate_merge`: **139 tests, OK** (114 before). The
tool-tests.yml list, run whole: **717 tests, OK, 2 skips** on a Windows tree with the
toolchain (692 before). `CompiledCodeOwnership` (4 tests) and
`LiteralPoolThroughBuildReport` (2) are `@skipUnless` the pinned compiler, so they
self-skip on the runner and its skip count rises by six; every other new test runs
there. `check_python_names`: PASS, 284 files, 0 unresolvable. `check_dead_references`:
no new dead references, no broken markdown links. `pyflakes` on both changed files is
clean apart from one pre-existing finding (`credit_moved`).

`validate_merge` against `origin/main` (bdcb1df7d) on this branch: **Passed**, no
reasons, `relocationEvidence valid 92 / 192,790 / 0` (the two-revision union;
96,395 per revision), no repartition, `totalFunctions` 11,345 unchanged,
`matchedFunctions` 11,286 unchanged. The only warning is the absent head ROM report.

## What is still not claimed

W4-03's entry-sum arithmetic (counters sum entries, not interval unions) and W4-04's
inherited shrink/helper-cache limits are unchanged and stay disclosed. The private `PR
validation` check is tooling-only `noverify`; nothing here is retail validation of the
exception.

### DIFF
--- a/tools/requirements.txt
+++ b/tools/requirements.txt
@@ -27,3 +27,10 @@
 # one, and it touches a workflow that writes commits to main.
 pyelftools==0.33
 capstone==5.0.9
+# test_validate_merge reaches evidence_rom, which is where `is_unconditional_return`
+# lives -- the one classifier that decides whether an ARM word is a return. The
+# matched-loss exception in validate_merge.py imports it rather than re-deriving the
+# encodings, because two spellings of that answer drifted apart once already and the
+# drift was invisible. evidence_rom's own module-level imports include yaml, so the
+# suite that exercises the classifier needs it even though nothing here reads YAML.
+pyyaml==6.0.3
--- a/tools/test_validate_merge.py
+++ b/tools/test_validate_merge.py
@@ -1139,15 +1139,29 @@ class RelocIndex(unittest.TestCase):
                    "from:0x01ff8100 kind:arm_call to:0x0201b000 module:main\n")
         self.old_repo = VM.REPO
         VM.REPO = self.repo
-        VM._RELOC_CACHE.clear()
-        self.addCleanup(VM._RELOC_CACHE.clear)
+        for cache in (VM._RELOC_CACHE, VM._ENROLMENT_CACHE,
+                      VM._SYMBOL_BASE_CACHE):
+            cache.clear()
+            self.addCleanup(cache.clear)
         self.addCleanup(setattr, VM, "REPO", self.old_repo)
 
     def write(self, rel, text):
         (self.cfg / rel).write_text(text, encoding="utf-8", newline="\n")
 
+    def image(self, module, words):
+        """The module's cartridge image, based at its lowest declared symbol."""
+        self.write(f"overlays/{module}/symbols.txt",
+                   "data_020ab100 kind:data(any) addr:0x020ab100\n")
+        d = self.repo / "extracted" / "overlays"
+        d.mkdir(parents=True, exist_ok=True)
+        image = bytearray()
+        for word in words:
+            image += word.to_bytes(4, "little")
+        (d / f"overlay_{int(module[2:]):04d}.bin").write_bytes(bytes(image))
+
     def index(self):
         VM._RELOC_CACHE.clear()
+        VM._SYMBOL_BASE_CACHE.clear()
         return VM._reloc_index(commit(self.repo, "relocs", "tester"))
 
     def test_every_module_file_supplies_destinations_and_the_index_is_valid(self):
@@ -1189,11 +1203,53 @@ class RelocIndex(unittest.TestCase):
         self.assertEqual(len(defects), 1)
         self.assertIn("config/arm9/relocs.txt is empty", defects[0])
 
-    def test_an_empty_file_for_a_module_that_declares_nothing_is_fine(self):
-        # 15 of these at cd3a7eb59: dtcm and fourteen zero-length overlays.
+    def empty_module(self, sections, name="ov061"):
+        """A module with no functions, an empty `relocs.txt`, and that inventory."""
+        (self.cfg / "overlays" / name).mkdir(parents=True, exist_ok=True)
+        self.write(f"overlays/{name}/symbols.txt", "")
+        self.write(f"overlays/{name}/relocs.txt", "")
+        self.write(f"overlays/{name}/delinks.txt", sections)
+
+    def test_an_empty_file_for_a_module_with_nothing_but_bss_is_fine(self):
+        # 15 of these at cd3a7eb59. ov061 verbatim: a zero-length `.ctor` and a
+        # `.bss`, which occupies no cartridge bytes and can hold no pointer.
+        self.empty_module(
+            "    .ctor       start:0x02115ec0 end:0x02115ec0 kind:rodata align:4\n"
+            "    .bss        start:0x02115ee0 end:0x02115ee0 kind:bss align:32\n")
+        self.assertEqual(self.index()["defects"], [])
+
+    def test_an_empty_file_beside_a_module_with_no_inventory_fails_closed(self):
+        # No `delinks.txt` at all: nothing says whether the module holds data.
         (self.cfg / "overlays" / "ov061").mkdir(parents=True)
         self.write("overlays/ov061/symbols.txt", "")
         self.write("overlays/ov061/relocs.txt", "")
+        defects = self.index()["defects"]
+        self.assertEqual(len(defects), 1)
+        self.assertIn("ov061 has no section inventory", defects[0])
+
+    def test_an_empty_file_beside_unreadable_initialised_data_fails_closed(self):
+        # Four bytes of real `.data` and no image in this checkout to read them
+        # from. Unavailable evidence is a defect, never a pass.
+        self.empty_module(
+            "    .data       start:0x020ab100 end:0x020ab104 kind:data align:4\n")
+        defects = self.index()["defects"]
+        self.assertEqual(len(defects), 1)
+        self.assertIn("the module image is unavailable", defects[0])
+
+    def test_an_empty_file_beside_a_word_naming_an_address_is_a_defect(self):
+        self.empty_module(
+            "    .data       start:0x020ab100 end:0x020ab104 kind:data align:4\n")
+        self.image("ov061", [0x02071694])
+        defects = self.index()["defects"]
+        self.assertEqual(len(defects), 1)
+        self.assertIn("holds a nonzero word 0x02071694 at 0x020ab100", defects[0])
+
+    def test_an_empty_file_beside_zeroed_initialised_data_is_fine(self):
+        # `config/arm9/dtcm` is this case on the real tree: 0x20 of `.data`, all
+        # zeroes, and an empty relocation file that is complete evidence.
+        self.empty_module(
+            "    .data       start:0x020ab100 end:0x020ab108 kind:data align:4\n")
+        self.image("ov061", [0, 0])
         self.assertEqual(self.index()["defects"], [])
 
     def test_a_malformed_row_is_a_defect(self):
@@ -1490,14 +1546,36 @@ class AbsorbedEpilogueThroughBuildReport(unittest.TestCase):
             f"    .text start:0x{start:08x} end:0x{end:08x}",
             ""])
 
+    def write_overlay_image(self, number, words):
+        d = self.repo / "extracted" / "overlays"
+        d.mkdir(parents=True, exist_ok=True)
+        image = bytearray()
+        for word in words:
+            image += word.to_bytes(4, "little")
+        (d / f"overlay_{number:04d}.bin").write_bytes(bytes(image))
+
+    def data_only_overlay(self, word):
+        """A module declaring NO function, holding four bytes of real .data, with an
+        empty `relocs.txt`. Its one data word is the retired ARM9 address."""
+        self.write("config/arm9/overlays/ov001/symbols.txt",
+                   "data_020ab100 kind:data(any) addr:0x020ab100\n")
+        self.write("config/arm9/overlays/ov001/delinks.txt", "\n".join([
+            "    .data       start:0x020ab100 end:0x020ab104 kind:data align:4",
+            "    .bss        start:0x020ab104 end:0x020ab104 kind:bss align:32",
+            ""]))
+        self.write("config/arm9/overlays/ov001/relocs.txt", "")
+        self.write_overlay_image(1, [word])
+
     def absorb(self, relocs=VALID_RELOCS, extra_reloc=None, epilogue_word=0xE12FFF1E,
-               epilogue_size=0x4, rom=True, orphan_module=False):
+               epilogue_size=0x4, rom=True, orphan_module=False, data_only=None):
         """Commit a base and the fold on top of it, and report on the pair.
 
         `relocs` is the content of `config/arm9/relocs.txt` in BOTH revisions; None
         leaves the file out of both. `extra_reloc` appends one row naming an address,
         head only. `orphan_module` adds an overlay that declares a function and has no
-        `relocs.txt` at all -- an incomplete module inventory.
+        `relocs.txt` at all -- an incomplete module inventory. `data_only`, when given,
+        adds a data-only ov001 whose single data word is that value and whose
+        `relocs.txt` is empty in both snapshots.
         """
         nxt = self.EPILOGUE + epilogue_size
         self.write("config/arm9/symbols.txt", "\n".join([
@@ -1514,6 +1592,8 @@ class AbsorbedEpilogueThroughBuildReport(unittest.TestCase):
         if orphan_module:
             self.write("config/arm9/overlays/ov001/symbols.txt",
                        "OvOne kind:function(arm,size=0x4) addr:0x020ab110\n")
+        if data_only is not None:
+            self.data_only_overlay(data_only)
         self.write("src/Anchor.c", "int Anchor(void) { return 0; }\n")
         self.write("src/func_02071644.c",
                    "// NONMATCHING: hand-written asm, does NOT count as matched.\n"
@@ -1614,6 +1694,21 @@ class AbsorbedEpilogueThroughBuildReport(unittest.TestCase):
             self.absorb(relocs=self.VALID_RELOCS
                         + "from:0x02071620 kind:arm_call to:0x02071694 module:elsewhere\n"),
             "undocumented destination module 'elsewhere'")
+
+    def test_a_data_only_module_with_an_empty_relocation_file_fails_closed(self):
+        # No function anywhere in ov001, so the old inventory rule exempted its empty
+        # `relocs.txt` outright -- and the four bytes of .data it does hold are the
+        # retired ARM9 address. A count of functions cannot see that.
+        self.assertRefused(self.absorb(data_only=self.EPILOGUE),
+                           "config/arm9/overlays/ov001/relocs.txt is empty")
+
+    def test_a_data_only_module_whose_data_is_zeroed_is_still_legitimate(self):
+        # The other half of the same rule: an empty relocation file beside data that
+        # holds no address at all is complete evidence, and must stay supported.
+        report = self.absorb(data_only=0)
+        self.assertEqual(report["reasons"], [])
+        self.assertTrue(report["relocationEvidence"]["valid"])
+        self.assertEqual(report["coverage"]["delta"]["absorbedMatchedFunctions"], 1)
 
     def test_an_incomplete_module_inventory_fails_closed(self):
         # A module that declares a function and has no relocation file at all. Every
--- a/tools/test_validate_merge.py
+++ b/tools/test_validate_merge.py
@@ -1460,7 +1460,8 @@ class AbsorbedEpilogueThroughBuildReport(unittest.TestCase):
         self.old_repo = VM.REPO
         VM.REPO = self.repo
         self.addCleanup(setattr, VM, "REPO", self.old_repo)
-        for cache in (VM._RELOC_CACHE, VM._ENROLMENT_CACHE):
+        for cache in (VM._RELOC_CACHE, VM._ENROLMENT_CACHE,
+                      VM._SYMBOL_BASE_CACHE):
             cache.clear()
             self.addCleanup(cache.clear)
 
@@ -1542,6 +1543,7 @@ class AbsorbedEpilogueThroughBuildReport(unittest.TestCase):
         head = commit(self.repo, "absorb", "bob")
         VM._RELOC_CACHE.clear()
         VM._ENROLMENT_CACHE.clear()
+        VM._SYMBOL_BASE_CACHE.clear()
         return VM.build_report(base, head)
 
     def assertRefused(self, report, defect=None):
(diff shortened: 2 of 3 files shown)


andrewboudreau (changes requested):
Request changes on `efb61f2731807aefaeca9f00b160585bb04033e8`. Independent review covered the complete diff, all 85 supplied tests (pass), and additional missing-evidence and non-epilogue cases.

1. **P1 — missing/invalid relocation evidence fails open** ([parser](https://github.com/tangosdev/sm64ds-decomp/blob/efb61f2731807aefaeca9f00b160585bb04033e8/tools/validate_merge.py#L173), [exception](https://github.com/tangosdev/sm64ds-decomp/blob/efb61f2731807aefaeca9f00b160585bb04033e8/tools/validate_merge.py#L444)). Real-tree `build_report` fixtures with missing relocation files in both snapshots, empty files, malformed/truncated rows, a leading BOM or leading whitespace return `Passed`, `reasons: []`, and one absorbed matched record. An incomplete module inventory also passes. The parser returns an ordinary dictionary after skipping input, so the new guard mistakes unavailable evidence for zero callers. Require validated complete input before enabling this exception; unreadable/missing/malformed evidence must retain matched-loss refusal. Actual unreadable Git blobs already raise, and a valid base-only relocation correctly blocks the exception.

2. **P1 — zero indexed relocations does not prove an epilogue** ([guard](https://github.com/tangosdev/sm64ds-decomp/blob/efb61f2731807aefaeca9f00b160585bb04033e8/tools/validate_merge.py#L434), [claim](https://github.com/tangosdev/sm64ds-decomp/blob/efb61f2731807aefaeca9f00b160585bb04033e8/tools/validate_merge.py#L1208)). The guard admits an arbitrary 0x100-byte matched record or a named callback when covered by a new complete range; it inspects no body, epilogue position, entrypoint or independent boundary evidence. On the actual tree, valid `AutoloadCallback` also has zero indexed incoming references. Byte coverage and index absence do not establish that ROM execution can never enter an address. Preserve matched identities unless an independently verified boundary correction justifies retiring the particular record, and remove the unsupported no-entry claim from diagnostics/tests.

The arithmetic subtraction is reasonable for valid disjoint ranges, and unrelated removed coverage/moved global footprints are refused. Its counters still sum entries rather than interval unions: a retained 4-byte subrange overlapped by a new 84-byte range is admitted as +84, although unique coverage grew 80. This is a classifier limitation; no later production-overlap bypass is claimed. The surviving-size shrink check is inherited, and `build_report` already resolves cache keys to full SHAs.

These fixtures reproduce validator decisions, not full-ROM builds of synthetic code. All current CI checks are green, but they do not test these failures. Please keep the arithmetic adjustment separate from any broader policy permitting matched-function identity loss; an agent ruling or the desire to unblock #2496 is not the required evidence. No source candidate or policy change was made during this review.

tangosdev:
Both P1s taken on this branch: the relocation index is validated before the exception can apply (every module's relocs.txt present in both snapshots, non-empty, every row parsed; any defect keeps the matched-loss refusal with the defective file named), and the exception now requires boundary evidence beyond coverage and the index: the retired record must be the exact tail of the new complete range, one return instruction by its ROM bytes, reproduced by the range's byte-verified compiled output, with zero destinations in the validated index. The no-entry wording goes; the arithmetic fix stays its own commit. Fixtures for each of your cases (missing in both, empty, malformed, truncated, BOM, leading whitespace, incomplete inventory, a 0x100-byte record, a named callback at a range head, a non-return tail) failing closed. Re-request follows when pushed.

tangosdev:
Both P1s closed on head 92f9ad43b (four commits: the arithmetic fix alone; the validated relocation index; absorption needs boundary evidence; the grown-range warning). Your fixtures reproduced and now fail closed with the defective file named: missing in both snapshots, empty, malformed row, truncated row, BOM, leading whitespace, incomplete module inventory, an undocumented destination token. Found underneath: the old parser never opened config/arm9/itcm/relocs.txt (142 module:main rows) or dtcm; the index now reads every relocs.txt under config/arm9 (92 modules, 96,395 destinations, 0 defects on main). The exception now requires all four: the record ends exactly where the new complete range ends, it is one four-byte unconditional return by its ROM word (bx lr, mov pc lr, or an AL ldm/pop writing pc with S clear; conditional returns, ldm ^, bx rN, b self refused), the range is backed by byte-compared compiled source, and no destination in the validated index lands in it; the no-entry wording is gone everywhere. A 0x100-byte record at the tail, a four-byte non-return at the tail, a named callback, and a run without the ROM image all fail closed. Census corrected to 79 / 77 / 2 (IRQ::EmptyHandler has 14). test_validate_merge 114 (was 75), the CI tool list 692 OK at the previous content and the two validator modules green at this tip.

andrewboudreau (changes requested):
Second review of `92f9ad43bae8b54bc8ee111c990083c11365c6b9`, based on `cd3a7eb5966a8758722bde1ec78f61abc4b5c126`. The original missing-file/malformed-row fixtures and the arbitrary-sized/non-return record cases are now refused. The arithmetic fix remains separate. I independently ran the CI tool-test list: 692 tests, OK, 3 skips; all 114 validate_merge tests ran. The live relocation census also reproduces: 92 destination buckets, 96,395 destinations, zero reported defects.

Two remaining corrections are needed in the generalized exception:

1. **P2 — read an overlay word from the actual image base.** `tools/validate_merge.py:352-355` takes `base = _rev_enrolment(rev)[0][module][0]`, but that list contains only FUNCTION addresses. `modules._overlay_base`, which the new comment claims to follow, takes the lowest address of ANY symbol. With a data symbol preceding the first function by four bytes, the new reader reads the previous word. I drove the real `build_report` fixture with that layout, a `bx lr` immediately before the retired record, and `mov r0, #0` (0xe3a00000) at the record itself. It returns `Passed`, `reasons: []`, and one absorbed record because the reader returns 0xe12fff1e. Use the revision-pinned equivalent of the established all-symbol base calculation and add this negative control. Current live overlays with functions happen to have equal bases; this is a reproduced generic-guard defect, not a claim that #2496's ARM9 read is displaced.

2. **P2 — no functions does not prove an empty relocation file is complete.** `tools/validate_merge.py:237-245` permits an empty file whenever the symbols table declares zero functions. A data-only module can still carry callback pointers into another module. Adding a data-only ov001 with a nonempty four-byte .data section, a ROM word containing the retired ARM9 address, and an empty relocs.txt in both snapshots makes the real `build_report` fixture return `Passed`, `relocationEvidence.valid: true`, and one absorbed record. The exemption needs evidence that the module is empty or that its data cannot supply the missing references; a function count alone does not establish that. Keep legitimately empty relocation files supported, but distinguish them using the module/data evidence and fail closed when that evidence is unavailable.

These are temporary committed-repository/report fixtures, not full-ROM builds of synthetic source, and they do not invalidate the direct source byte proof for #2496. No PR source or policy was edited during this review. Please fix the two guard cases before landing this tooling prerequisite, then rerun #2496 against the landed validator.

andrewboudreau (changes requested):
Request changes at `92f9ad43bae8b54bc8ee111c990083c11365c6b9` (base `cd3a7eb5966a8758722bde1ec78f61abc4b5c126`). All 114 supplied tests pass. Independent review ran 24 additional controls against the actual parser/report APIs and compiled the literal example below with pinned 2004/b56.

**W4-01 is addressed for the reproduced input defects.** Missing, empty, malformed, truncated, unreadable and incomplete-module inputs now refuse the exception; so do unknown destinations. Valid incoming references in either revision still block it. The current index reads 106 files and 96,395 entries with no reported defects. That validates the configured input; it does not establish every possible ROM entrypoint.

**W4-02 remains P1: a four-byte tail with return-shaped bits does not establish an owned epilogue.** The [new guard](https://github.com/tangosdev/sm64ds-decomp/blob/92f9ad43bae8b54bc8ee111c990083c11365c6b9/tools/validate_merge.py#L672) returns `Passed`, no reasons, and one absorbed match for each of these independent fixtures:

- A named callback placed at the covering range's tail. The supplied callback control instead places it in the middle; it tests position, not independent function identity. This is a synthetic callback fixture, not removal of the retail `AutoloadCallback`.
- A genuine compiler-generated literal pool:

  ```c
  unsigned int func_02071644(void) { return 0xe12fff1e; }
  ```

  Pinned b56 emits `e59f0000 e12fff1e e12fff1e`: `ldr`, the actual `bx lr`, then a literal. Its ELF mapping symbols identify offset 8 as `$d`. Giving those exact bytes to the report fixture makes the final literal qualify as the removed function's return instruction. Preserving its bits does not turn data into an epilogue.
- `0xe8908000`, `ldm r0, {pc}`. The [mask](https://github.com/tangosdev/sm64ds-decomp/blob/92f9ad43bae8b54bc8ee111c990083c11365c6b9/tools/validate_merge.py#L327) accepts a non-stack indirect transfer as a return without evidence about the loaded destination. It also accepts `ldmdb r0, {pc}` and `ldmib r0!, {pc}`, which the cited `evidence_rom` helper does not classify the same way.

These demonstrate incorrect policy decisions; they do not establish a synthetic full-ROM bypass or reject #2496's specific boundary correction. Keep the matched-loss exception unavailable until independent boundary/code-ownership evidence supports it. The warning and [report row](https://github.com/tangosdev/sm64ds-decomp/blob/92f9ad43bae8b54bc8ee111c990083c11365c6b9/tools/validate_merge.py#L1648) still overstate what the guard establishes.

**W4-05, new P2: the overlay reader can read the wrong word.** [Lines 352–355](https://github.com/tangosdev/sm64ds-decomp/blob/92f9ad43bae8b54bc8ee111c990083c11365c6b9/tools/validate_merge.py#L352) use the first *function* as the image base; the canonical module reader uses the lowest symbol. In a data-leading overlay fixture, querying `0x020ab108` returns the preceding `bx lr` instead of the actual `mov r0, #0`, falsely satisfying the guard. Use the correct module address space and add this negative. Current tracked overlays have no such base disagreement; this finding is not about #2496's ARM9 address.

W4-03's entry-sum arithmetic and W4-04's inherited shrink/helper-cache limitations remain, now disclosed. The arithmetic has been separated from the policy change. Live checks are green, but private `PR validation` is tooling-only `noverify`; that does not provide retail validation of this exception. No queue or source changes were performed for this review.

andrewboudreau (approved):
Approve at `6df3e8d11` (base `cd3a7eb59`). Reviewed the three commits added since `92f9ad43b` and reproduced each finding's control independently rather than reading the commit messages. Scope is clean: three tooling files, no source, config or policy.

**W4-02 (P1) is closed on all three fixtures.**

- `ldm r0, {pc}`. `_is_return_word` now decodes with capstone, delegates to `evidence_rom.is_unconditional_return`, then narrows. I drove it on ten encodings: `bx lr`, `ldmfd sp!, {pc}` and `pop {r4, pc}` accept; `ldm r0, {pc}` (0xE8908000), `ldmda r0, {pc}`, `ldmib r0!, {pc}`, `ldr pc, [sp], #4`, `mov pc, lr`, `bxeq lr` and the S-bit spelling all refuse. Zero misclassified. One imported classifier plus an explicit r13 requirement is the right shape -- the two spellings can no longer drift apart unnoticed, which was the original defect.

- The literal pool. I compiled the example from the review under pinned 2004/b56 and read the mapping evidence per offset: `instruction: True` at offset 4, `instruction: False` at offset 8, on the **same 32 bits** (0xE12FFF1E). The ELF mapping symbols are the correct discriminator here; the four bytes genuinely cannot carry the answer and the object can.

- The named callback at the covering range's tail. Closed by the identity clause, with both a unit control and a report-level one (`test_a_named_callback_at_the_tail_is_refused`, `..._fails_the_report`). Every fixture named in the previous review has a test at both levels, which is the right shape given the earlier rounds found cases that passed unit tests while the report still returned Passed.

**W4-05 (P2) is closed and cross-validated.** `_module_symbol_base` matches any symbol row rather than function rows, and reads through `git_text(rev, path)` so it stays revision-pinned. I compared it against `modules._overlay_base` itself across every overlay in the tree: 90 modules with a real base, zero disagreements; the other 13 return None from both.

**Empty relocation file (second review, P2 #2) is closed.** The exemption is now decided by dsd's own section inventory with the initialised words actually read, not by a function count, and fails closed naming the file when the inventory or image is unavailable.

Live census reproduces at this tip: 92 destination buckets, 96,395 destinations, 0 defects. Full `tools.test_validate_merge`: 139 tests, OK, exit 0.

**What this approval does not cover.** Three further defects in the clause (f) this branch introduces, found after this tip and already fixed on `tools/w4-mergerule-ownership` @ `58119cb4e`, which stacks directly on `6df3e8d11`:

1. The ownership compile searches the **live checkout's** `include/` while the `.c` beside it is read at the revision under judgement. Every other input this file consults is revision-pinned; this one is not.

2. `asm_policy` never sees a header -- `function_snapshot` greps `-- src/`. A header carrying `asm int func_02071644(int, int) { ... bx lr }`, included by an otherwise-empty `.c`, classifies as ordinary C, preprocesses, compiles, and yields an object whose `$a` region covers the range with 0xE12FFF1E at the epilogue offset. Clause (f) is satisfied in full by a body the compiler did not write. Measured on this tip, not supposed.

3. Clause (c) claims the range was byte-compared to the cartridge, but a missing head ROM report is only a warning and (f) compares one word. Nothing in the chain establishes that anything ever reproduced the covering module.

Each carries a control proven to fail at `6df3e8d11` and pass at `58119cb4e` (144 tests there, 722 across the CI list). None of them blocks this branch: they are defects in a clause that does not exist on main today, and the fix is ready. I will retarget that branch at main once this lands rather than stack a PR onto this one.

andrewboudreau:
Approved — this is good work and it's ready to merge.

## What I actually checked

I didn't take the commit messages on trust; I re-ran each of the three problems from the last review against this tip.

**1. The "is this word a return instruction?" test.** It used to accept `ldm r0, {pc}` — which jumps to whatever happened to be sitting in r0, not a return. It now decodes the instruction properly and insists the base register is the stack pointer. I threw ten encodings at it, including the `ldmda`/`ldmib` variants, a conditional `bxeq lr`, and the S-bit form that restores CPSR. Ten out of ten correct.

**2. The literal-pool trap.** This was the nasty one. If a function returns the constant `0xE12FFF1E`, the compiler parks that number in a literal pool right after the code — and `0xE12FFF1E` *is* the encoding of `bx lr`. So the last four bytes of the function look exactly like a return but are data.

I compiled that exact function under the pin and checked what the new code sees. It reads the ELF's own code-vs-data markers and calls offset 4 an instruction and offset 8 data — on byte-identical words. That's the right way to answer it: the four bytes genuinely cannot tell you, and the object file can.

**3. The overlay base address.** It was working out where an overlay loads from its first *function*, but an overlay can begin with data — so the base could be low by a few bytes and every lookup would quietly return the word *before* the one you asked for. On a check whose whole job is "is the cartridge's word at this address a return", that's the difference between refusing and absorbing. It now uses the lowest symbol of any kind. I diffed it against `modules._overlay_base` (the reader the rest of the repo uses) across every overlay in the tree: 90 with a real base, zero disagreements.

Also re-ran the relocation census (92 buckets, 96,395 destinations, 0 defects) and the whole test file (139 tests, green).

## One thing to know before it lands

I found three more ways to satisfy the new "this really was an epilogue" check without actually proving it. All three are fixed and tested on `tools/w4-mergerule-ownership`, which sits directly on top of your tip.

**1. The ownership compile reads the wrong headers.** It compiles the source to see what the compiler owns, but it pulls headers from the *working checkout* rather than the commit being judged. Everything else in that file is pinned to the revision; this one wasn't.

**2. Hand-written asm can arrive through a header.** The "is this hand-written assembly?" check only ever sees the `.c` file, because the snapshot feeding it filters to `src/`. Put the asm body in a **header** and include it, and it reads as ordinary C, compiles, and satisfies the ownership check in full. I built that case and watched it sail through — the header classifies as `unbannered-asm`, the `.c` the gate reads classifies as nothing at all.

**3. Nothing confirms the ROM build reproduced the module.** A missing build report was only a warning, and the ownership check compares a single word. So the "these bytes were compared to the cartridge" claim can hold with nothing having actually compared them.

Each one has a control that fails on your tip and passes with the fix — 139 tests there plus the full CI list.

## Asks

1. **Merge this when you're happy with it.** It's `CLEAN`, nothing is blocking, and the three items above are defects in a check that doesn't exist on main today — they don't argue against landing this.
2. **After it lands I'll retarget the ownership branch at main** and open it as an ordinary PR rather than stacking one onto your branch. Say so if you'd rather fold those three commits in here instead — happy either way.

PR #2502, opened 2026-09-09
Restamp LINKED off port/link100 3131ddaa9: 10357 of 11328 (91.4%)

tangosdev:
Config-only stamp. port/link100 at 3131ddaa9 gated as INT18 (battery ALL GREEN at floor 10227, all proofs green incl. the four-player proof three times and the sixteen-window ladder, reconfigure 254/5); the hosted port-linkage workflow reads the same count on every push. +130 over the 10227 stamp: the ROM's own become-parent and become-child and 128 TUs behind them over a hosted ARM7 wireless stub.

### DIFF
--- a/config/port_linkage.json
+++ b/config/port_linkage.json
@@ -28,11 +28,11 @@
     "Actor methods and free functions both count. Enrollment alone does not increase",
     "the count when the release build still dead-strips the unreachable object."
   ],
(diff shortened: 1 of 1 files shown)


tangosdev:
Superseded by #2503 (10392 off 421afeaf7); same file, later number.

PR #2511, opened 2026-09-09
ov043: promote daObjKm1_Kurumajiku_c to a five-member C++ TU and adopt the ROM RTTI name

andrewboudreau:
Folds five ov043 shards (`0x021113fc`..`0x02111518`) into
`src/game/actors/d_a_obj_km1_kurumajiku.cpp` and retires the coined `RickshawBdw`
for the name the cartridge carries.

`RickshawBdw` has **no cartridge attestation** in any encoding tested. ov043 holds
`_ZTS22daObjKm1_Kurumajiku_c` at `0x02112350`. The class takes the ROM name here.

The identity is proved from the emitted vtable rather than asserted: the table is
word-identical to `daObjKurumajiku_c`'s base table except at slots 0, 3, 16 and 17 --
exactly the four the header declares. `romdata_check` resolves all 128 bytes / 32
slots against the cartridge with relocations applied. The review recorded **zero
findings**.

## This branch does not merge cleanly -- resolve, do not hand-merge

Two files conflict against main, both machine-generated or append-only:

- `notes/cpp-tu-current-state.md` -- **regenerate** with
  `python tools/cpp_tu_state.py --write-note`. Do not hand-resolve; the regenerated
  content matched neither side in an earlier cycle, so resolving by hand silently
  ships a wrong file.
- `symbols/actor_renames.tsv` -- **replay both appended blocks**. This branch's 5
  ov043 rows and main's 8 ov022/ov013 rows share no `(module,addr)` key, measured, so
  the union is the correct resolution. `check_rename_ledger.py` exit 0 on 2031 rows
  validates the replay.

## Gates, on the resolved merge tree rather than the branch

`premerge_check` on the raw head returns **exit 3 -- no gates ran**, which is not a
pass. Only the resolved merge tree is green:

- `premerge_check.py --base origin/main`: exit 0, all **8** static gates run and pass.
- `check_dead_references.py`: exit 0.
- `check_rename_ledger.py`: exit 0, 2031 rows.
- `rombuild.py --no-rom`: exit 0, **106/106 modules exact**, 11,199 reproducing,
  0 mismatching.

Carries its own `notes/data/tu-promotion-queue.tsv` row refresh (`RickshawBdw` ->
`daObjKm1_Kurumajiku_c`, `already_promoted` no -> yes).

## Stated rather than rounded up

No `.nds` was packaged (`--no-rom`), so no ROM hash was taken here.
`premerge_check` models eight static gates and no CI workflow.

The tree-wide ROM-data check reports 4 differing symbols (`_ZTV10dCcAcPos_c`, two
`daDemo_c` typeinfos, `_ZTV7daKrb_c`), all in arm9/ov002/ov084. This branch touches
none of the sources that emit them, so they are not its to move; they remain a real
tree-wide defect owed its own lane.


andrewboudreau:
Restacked onto current `main` (`0c450b9b7`); force-pushed `f9443284d` → `5191eb758`.

The branch had gone `CONFLICTING` in two files, each handled by its own rule rather than by hand-merging:

- `notes/cpp-tu-current-state.md` is generated — took main's blob and regenerated with `tools/cpp_tu_state.py --write-note` on the final tree. `--check-note` is exit 1 on unmodified `main` (pre-existing staleness) and exit 0 here.
- `symbols/actor_renames.tsv` is an append log — took main's blob and appended this branch's 5 rows verbatim. Rows minus distinct `(overlay, address)` keys is 805 on both `main` and this head, so nothing was duplicated.

`notes/data/tu-promotion-queue.tsv` was re-derived on the final tree. `queue_audit.py --check` is also exit 1 on unmodified `main` (`total_lines` disagree for `ChiefChilly` and `Player`); this branch's regeneration corrects those as a side effect.

Content preservation: `git diff --name-status` against `main` and the pre-restack branch against its old merge base are identical — 17 paths, same add/delete/modify status each — and every blob this branch owns is byte-identical to the pre-restack version.

`premerge_check.py <head> --base origin/main`: exit 0, 8/8 gates pass at both base and merge tree. No ROM was built; these are static gates only.

The four original commits were replayed unchanged, keeping their original author trailers. Two commits were added on top: the regenerated files, and a restore of the five `attribution.json` per-symbol credits (including a non-`tangosdev` contributor's) that the promotion originally carried.

Known red and not required: `Source review` is void rather than stale — changing the head means no verify output's commit matches it any more.

PR #2513, opened 2026-09-09
Cleanup ov002/daObjLava_c: popping lava bubbles

lunavyqo:
Readable follow-up on the popping lava-bubble actor. Bytes unchanged: `tubuild.py verify ov002/daObjLava_c` is 5/5 MATCH, objisolate clean, reloc destinations clean. `dActor_c::HorzAngleToCPlayer` / `HorzAngleToFPlayer` / `HorzAngleToCPlayerOrAng` also MATCH under 2004/b56 after switching to `Pos()`.

- Factory is `return new daObjLava_c()`. In-class `operator new` takes `unsigned long` and forwards to `_ZN7fBase_cnwEj`.
- InitResources has nothing to load.
- Behavior (re)issues particle `0xb7` at the closest player's position via `ClosestPlayer()->Pos()`, feeding last frame's handle back so the emitter is recycled.
- `dActor_c::Pos()` is the shared overlay of `mPosX/Y/Z` at 0x5c. Do not pun `(Vector3 *)&mPosX` in a leaf. Typed `extern "C"` of mangled names on this header is still poison; an overlay accessor is not.
- Spawn `+4/+6` are execute/render priorities. `g_profile_LAVA` stays in this TU (it owns that data run).
- Comments describe the bubbles. Reconstruction/RTTI/reverse-order banners are gone.

- Particle call is still `func_02022c3c` (wrapper around `Particle::System::New` that plants the default callback).
- Inline destructor (out-of-line emits D0 before D1).
- Convert other `&mPosX` puns to `Pos()` one enrolled file at a time.


lunavyqo:
Validation isolate refused three intact TUs after `dActor_c::Pos()`: `daObjFm_Battan_c` `@335`/`@336`, `daObjKb1Billboard_c` `@214`, `daStarGate_c` `@420`..`@446` had 0 ELF entries.

Addressed in 4cd82b991. Same +8 parse-position shift as `daDsnBase_c` `$424`→`$432` / `daMky_c` `$724`→`$732`. Compiled the three TUs under this header and retargeted the tags that the objects actually emit.

andrewboudreau:
## Measured: this PR is correct on its own, and mutually exclusive with #2571

It had never been validated once — `PR validation` has sat at `action_required`
since 09-09 because no worker ever claimed it. I built it locally instead.

**Merged onto `main` alone it is green:**

```
ROM-build analysis: PASS
```

The five-overlay manifest re-sync (ov023, ov030, ov044, ov091, ov100) is the
right blast radius for adding `dActor_c::Pos()` to `include/dActor_c.h`. That
part is done correctly.

## It cannot be merged alongside #2571

Both PRs re-sync `config/tu_manifest.d/ov030/daMky_c.json`, and mwccarm's
positional `@NNN`/`$NNN` uniquifiers make the two re-syncs **incompatible by
construction**. Same struct, four different numbers:

| tree | emitted symbol |
|---|---|
| `main` | `_ZN21Bundle$724daMky_c_cppD1Ev` |
| #2571 alone | `_ZN21Bundle$725daMky_c_cppD1Ev` |
| **#2513 alone** | `_ZN21Bundle$732daMky_c_cppD1Ev` |
| **#2571 + #2513** | `_ZN21Bundle$733daMky_c_cppD1Ev` |

I merged this PR onto a tree containing #2571 and resolved the manifest with
*this* PR's version. Result:

```
FAIL src/actors/daMky_c.cpp: isolate: compiler-only deadstrip refused:
  compiler-only symbol(s) not defined: ['_ZN21Bundle$732daMky_c_cppD1Ev']
```

`tubuild.py verify ov030/daMky_c` on that tree reports what is actually emitted:

```
EXTRA   .text  _ZN21Bundle$733daMky_c_cppD1Ev  size=0x4  STB_LOPROC  not in manifest
POLICY  compiler_only_output _ZN21Bundle$732daMky_c_cppD1Ev is not an emitted unlicensed function
POLICY  unlicensed function _ZN21Bundle$733daMky_c_cppD1Ev has no compiler_only_output policy
```

`$733` is neither side's value. Taking #2571's `$725` cannot work either — the
number is a function of the combined tree, not of either branch. **No conflict
resolution that picks a side is correct here.** The re-sync has to be regenerated
against whatever actually lands.

## What to do

Land one of the two, then regenerate the other's manifest against the merged
result — do not hand-resolve the conflict. The same applies to the rest of the
cluster re-syncing overlapping manifests: #2513, #2570, #2571, #2592, #2633.

## One unrelated nit

`verification.compilerOnlyOutput.requested` and `.deadstripped` in
`ov030/daMky_c.json` still read `$724` while the live policy row reads `$732`.
It builds green because nothing re-reads that block — it is a record, not a
gate — but it is stale as written.

tangosdev:
Triage note: this PR touches include/dActor_c.h, one of the shared base headers whose mwccarm positional @NNN symbols renumber tree-wide on any edit. Unlike the open manifest desync on #2570, #2592, #2601, #2608 and #2642, this one is already resynced: Andrew's measurement above confirms the five-overlay re-sync (ov023, ov030, ov044, ov091, ov100) is correct and the merge is green in isolation. The only open item is the manifest race with #2571.

PR #12, opened 2026-06-27
Match func_ov002_020ae890

Moundistz:
Byte-exact match for `func_ov002_020ae890` verified with mwccarm 1.2/sp2p3.

**Oracle verification (tools/match.py):**
```
TARGET func_ov002_020ae890 @ 0x020ae890 size 0x28  bytes: 10402de9fe28d1e10040a0e1b429c4e1040000eb741084e20a00a0e3668ffdeb1040bde81eff2fe1
  1.2/sp2p3: MATCH

========================================
MATCHING VERSIONS: 1.2/sp2p3
```

- Module: ov002
- Address: 0x020ae890
- Size: 0x28

PR #2514, opened 2026-09-09
The matched-but-wrong signature: a census, and 41 of them corrected

tangosdev:
THE CLASS

A function whose ROM body reaches its callee with its own incoming registers
untouched can be decompiled with no parameters at all and still be byte-exact.
On ARM the arguments ride through r0..r3, so the compiler emits no argument
setup whether the C names them or not:

    func_02018e3c:
        push {r4, lr}
        bl   func_0205d644          ; r0 and r1 are still the caller's
        movs r4, r0
        bne  +0x10

The C that came out of that -- `int func_02018e3c(void) { int r =
func_0205d644(); ... }` -- passes the byte gate, the relocation gate and the
link gate, and is wrong. Every host that puts arguments on the stack loses
them. Lane WM7 found the first one the hard way: src/func_0205d23c.c called
func_0205d304() with nothing, the archive lookup returned null, and a
seven-window session on the PC port faulted at func_0205cdf4+0x22 reading
0x00000010. Spelling the parameters and forwarding them was byte-identical.

Two more were already known and worked around on the port side with host-ABI
faces (port/hal/fs_names.cpp for func_02018e3c/func_0205d644, port/hal/fs.cpp
for SharedFilePtr::Construct). This branch corrects those two at the root,
then sweeps the tree for the rest of the class.

THE TOOL

tools/check_call_arity.py reads every source in src/ and reports each call that
passes FEWER arguments than the callee's own definition declares. Ground truth
is the definition, because that is what the ROM runs; declarations elsewhere
are collected as weaker evidence and hidden behind --decl, since this tree lets
each file spell its own externs and they routinely disagree. A C++ method
definition is charged one extra argument for `this` unless the class declares
it `static`, and both include/**/*.h and the class bodies written inside
src/**.cpp are read for that marker.

It is a census, not a verdict. A row is the static half of the shape. The other
half -- that the caller reaches the bl with those registers still holding its
own arguments, and that the callee reads them -- is settled only by the
disassembly, and the tool deliberately does not guess. Not wired into CI;
--fail-on-hits exists for when the tree's remaining rows are triaged.

Most of the work in it is refusing to print rows. The first run over src/ was
2,623 rows across 269 callees, and every large group was the reader being
fooled:

    2,354  function-like macros (`#define H(a, b)`) read as two-parameter
           functions, so every H(x) call was a one-of-two row
      ~800 static member functions (Model::LoadFile, Animation::LoadFile,
           dActor_c::FindWithID, OAM::Render) charged for a `this` they do not
           take
       ~40 `ptr.LoadFile()` scored against the free function src/LoadFile.c
        ~7 K&R `extern int f();` written INSIDE a function body, read as a
           zero-argument call one line above the correct call to f(c)
           method declarations inside a class body, read the same way

19 tests, added to the tool-tests list; most of them assert SILENCE on a shape
that only looks like a row, because a census nobody trusts is worse than none.

THE CENSUS (origin/main bdcb1df7d, after this branch's src fixes)

    calls passing fewer arguments than the callee declares   196
    sources scanned                                        9,132
    definitions with a count                               7,871
      of them C++ methods (declared params + 1 for `this`)  1,755
      static methods, no `this`     222 pairs (168 include/, 54 more src/)
    call sites read                                       34,194
    K&R `()` unknown arity          536 symbols  (not scorable, not flagged)
    distinct callees flagged                                 114
    rows resting on a DECL           60  (weaker evidence, --decl)
    rows in NONMATCHING files         2  (excluded unless --include-drafts)

Triaged against the ROM (the disassembly at each call site, _work/triage.py in
the run's out dir): of the 244 candidate pairs the first full census produced,

    RIDE     61   no write to the missing registers, no earlier bl: the class
    SETUP    78   the caller writes them first, so the C is merely terse
    NOBL     48   no bl to that callee decoded (inlined or a bad row)
    CLOBBER  12   an earlier bl sits in front: r0..r3 already dead
    SKIP     23   more than four parameters, or the symbol is not in a config

Three of those 61 were false and went away when the census learned to skip a
K&R prototype written inside a function body. This branch fixes 40 caller-side
rows (plus one callee whose parameters were merely unnamed, func_0205d644), and
re-running the triage over the fixed tree leaves 23 RIDE rows, listed at the
end. The count moves in both directions as it goes: correcting func_02017ae4
exposed a second layer, func_020178cc and func_02017a24, that no earlier run
could see, because the ground truth for a call is the callee's own definition.

THE FIXES -- 41 functions, every one byte-identical at 2004/b56

The two that were known and faced on the port:

  func_02018e3c(void *out_file_id, const char *path)  0x02018e3c 0x2c  arm9
    the two arguments forwarded to func_0205d644; func_0205d644's own
    parameters named in the same commit
      +0x00 push {r4, lr}   OK
      +0x04 bl              reloc (wildcard)
      +0x08 movs r4, r0     OK
      +0x0c bne #0x10       OK
      +0x10 ldr r0,[pc,#0x10] OK
      +0x14 bl              reloc (wildcard)
      +0x18 bl              reloc (wildcard)
      +0x1c mov r0, r4      OK
      +0x20 pop {r4, lr}    OK
      +0x24 bx lr           OK
      MATCHING VERSIONS: 2004/b56

  _ZN13SharedFilePtr9ConstructEj + func_02017e48    0x0201799c / 0x02017e48
    two links of one chain: Construct leaves the fileID in r1 across bl
    func_02017e48, which leaves it in r1 again across bl func_02017e0c, which
    reads it. Neither named it. func_02017e48 also returns self, which the
    ROM's closing mov r0, r4 says and the `int` spelling got right by accident.
      both: MATCHING VERSIONS: 2004/b56

The sweep (one commit per family, every function verified the same way):

  the SharedFilePtr constructor family, two layers
    func_02017acc, func_02017ae4, func_02017b4c,
    SharedFilePtr_Construct_TexSeq          fileID in r1
    func_020178cc, func_02017a24            the second layer, only visible once
                                            func_02017ae4 declared both
  the file-open chain
    func_0205d518, func_02018d98            the path in r1
    func_02018270                           the handle in r0
  dActor_c, a non-static method called with no `this`
    eleven callers of ::ClosestPlayer, three of ::ClosestNonVanishPlayer,
    one of ::IsPlayerInRange (which also gained its own maxDist parameter)
  fourteen more, one argument each
    HitDeathPlane (SetNextLevel), func_02005324 and func_02009a8c
    (Camera::SaveCameraStateBeforeTalk), func_ov002_020e8dd8
    (PowerStar::AddStarMarker, both call sites), func_ov005_020c0030,
    func_ov006_020cb16c, func_ov006_020d8cc4, func_ov007_020add3c,
    func_ov007_020b4b5c, func_ov063_02119274, func_ov063_02119b1c,
    func_ov100_02144468, func_02073244 (the destructor pointer in r3)

  python tools/prepush_linkcheck.py --range origin/main..HEAD
  prepush-linkcheck: 41 checked - 41 verified, 0 warning(s), 0 blocking

  the tool-tests list, run whole: Ran 702 tests in 369.429s -- OK (skipped=2)
  (the two expected "no src samples" skips in test_match_provenance)
  check_python_names: PASS.  check_dead_references: no new dead references.

NOT FIXED, AND WHY

  func_0205d23c -> func_0205d304 (0/2).  This is WM7's function; the fix is
  already on port/l6-wm7 and l4merge and would conflict. Left for that merge.

  Three rows whose callee never reads the register, which is evidence AGAINST
  the class rather than for it -- the callee's own signature is over-declared
  (most likely a static member with no `static` in its header):
    _ZN8SaveData13EraseSaveFileEjPc -> SaveData::SetDefaultValues (1/2)
    func_ov075_02116e44 -> func_02020168 (0/1)
    func_0202027c -> func_02020254 (0/1)

  func_ov002_020e444c -> func_ov002_020e3f90 (0/1).  Real, but the callee is
  declared `(void)` in include/decl_common.h, a shared header: correcting it
  moves every other caller and belongs in its own change.

  Eighteen more real RIDE rows left for budget, all recorded with their bl
  address and the register the callee reads (triage.txt in the run's out dir):
    func_0201fc88 -> ObjectMessageIDToActualMessageID (0/1)
    func_ov006_02111b90 -> Vec2_Sub (2/3)
    _ZN5Sound8SetMusicEjj -> func_02011dcc (1/2)
    _ZN5Sound8PlayLongEjjjRK7Vector3s, func_0201226c -> func_02048a1c (1/3)
    _ZN5Sound4PlayEjjRK7Vector3 -> func_02050cdc (0/2)
    func_0204f86c, func_0204f76c, func_0204f79c, func_0204f7cc, func_0204f7fc,
    func_0204f89c, func_0204f73c -> the func_0205axxx sound handlers (1/3, 1/2)
    func_0205fab4 -> func_0205fbd0 (1/3)
    func_0206e2f8 -> func_0206e330 (1/3)
    func_02070b98 -> func_02070c68 (1/2)
    _ZN16dScMgSmartball_c13OnYoshiTryEatEi -> func_ov006_02115b0c (0/1)
    func_ov007_020be980 -> func_ov007_020bc3dc (1/3)
  None of them failed to match; none was attempted.

THE PORT SIDE

The two host-ABI faces this change makes unnecessary retire in a follow-up:

  port/hal/fs_names.cpp   the func_02018e3c face (PORT_HOST_ABI, the
                          push {r4, lr} / bl func_0205d644 / movs r4, r0
                          ride-through), and the comment-out of
                          src/func_02018e3c.c in slice_mg1.txt
  port/hal/fs.cpp         the _ZN13SharedFilePtr9ConstructEj face and the ABI
                          note in the file header

Both now have real parameters in src/, so the port can compile the ROM's own C
instead. WM7 already retired the third of these (func_0205d23c) with no face at
all. The exception list should lose two rows, not gain any.


andrewboudreau (commented):
Reviewed against current main (`9c7f38a24`), not the branch base. The core claim holds up under the check it was most at risk of failing, and I found three concrete things to change.

## What I verified

**The byte gate, independently.** I compiled each changed source with the pinned 2004/b56 compiler against *this branch's* headers and compared to the ROM with strict reloc checking. All 38 of the 41 changed sources that still exist on main: `MATCH`. (The other three are discussed below.)

**The callee half — the half `RIDE` does not establish.** `RIDE` proves only that the caller did not clobber the register; it does not prove the callee reads it. If a callee writes the register before reading it, the "missing argument" is fiction and adding it is the opposite error. I checked the callee for every family:

- `_ZN8dActor_c13ClosestPlayerEv` @ `0x02010ad8` — `r0` is never written before `0x02010B20 add r7, r0, #0x5c`, which reads it as `this->pos`, the reference point every player is distance-compared against. Not ignored. `_ZN8dActor_c22ClosestNonVanishPlayerEv` @ `0x02010a08` is identical (`add r7, r0, #0x5c` @ `0x02010A50`). **The 14 call-site edits are correct.**
- `func_0205d644` — `0x0205D64C mov r5, r0`, `0x0205D654 mov r4, r1`, both read-before-write.
- `func_02018a24` — its *first* instruction is `cmp r0, #0x8000`.
- `dActor_c::IsPlayerInRange` — `0x020104E8 mov r4, r1`, consumed at `0x02010500 cmp r0, r4, lsl #12`.
- `func_02017e0c` (terminal consumer of the SharedFilePtr chain) — `0x02017E1C mov r0, r1`, then `bl` and `strh r0, [r4]`.
- `Camera::SaveCameraStateBeforeTalk` — first instruction `0x0200CC5C ldr r1, [r0, #0x154]`.

No row where the caller actually wrote the claimed register, i.e. nothing filed as `RIDE` that was really `SETUP`. The class is real and the fixes are genuine.

One nuance worth recording in the prose: `ClosestPlayer` reads `r0` only on the cache-miss path — if the cached global is already non-null it returns early without touching `r0`. That does not change the verdict (the register is read on the computing path, so arity 1 is right), but "the callee reads it" is conditional here.

## 1. Three files are deletions-by-promotion on main, and the defect survives in the promoted TU

`git merge-tree` against main reports modify/delete on three of the 44 files. They were not deleted — they were absorbed into promoted class TUs after this branch's base. Taking the "version left in tree" resolution would resurrect three retired shards alongside the TUs that already define the same functions.

Worse, two of the three still carry the very defect this PR fixes, in their new home, which this PR does not touch:

- `src/func_ov084_0212f204.c` → `src/actors/daPkn_c.cpp`. On main, line 84 still declares `char *_ZN8dActor_c13ClosestPlayerEv(void);` and line 735 still calls `_ZN8dActor_c13ClosestPlayerEv()` with no argument. **Replay the fix there.**
- `src/func_ov002_020af218.c` → `src/actors/da1up_c.cpp`. On main, line 290 still declares `extern int _ZN8dActor_c15IsPlayerInRangeEi(void*);` and line 292 still calls it with one argument. The comment immediately above (lines 283–287) justifies this: *"five callers below pass 0xbb8 in r1 while this body never reads it and leaves it for `_ZN8dActor_c15IsPlayerInRangeEi`, which the shard likewise calls with one argument."* That justification is exactly what this PR removes, so the comment becomes false the moment this merges. Note the same file already spells the two-argument form correctly at lines 800, 853 and 880 — it contradicts itself today.
- `src/func_ov065_021165d8.cpp` → `src/actors/daYurei_Mucho_c.cpp`, which **already carries the fix**: line 40 declares `_ZN8dActor_c22ClosestNonVanishPlayerEv(void *self)` and lines 465, 506 and 568 all pass the actor. That edit is redundant — drop it.

## 2. One missed row of the same class

`src/func_ov102_02149078.c` is untouched by this branch and is a textbook member of the class:

```c
extern struct dActor_c *_ZN8dActor_c13ClosestPlayerEv(void);      // line 2
...
if (*(unsigned char*)((char*)_ZN8dActor_c13ClosestPlayerEv() + 0x706))   // line 14
```

The ROM confirms it is a `RIDE`, not a `SETUP`:

```
02149078  stmdb sp!, {lr}
0214907C  sub   sp, sp, #4
02149080  ldr   r1, [pc, #0x70]
02149084  ldrsb r1, [r1]
02149088  cmp   r1, #0x15
0214908C  bne   #0x21490b8
...
021490B8  cmp   r1, #0x21
021490BC  addeq sp, sp, #4          ; the #0x21 case returns
021490CC  bl    #0x2010ad8          ; ClosestPlayer, r0 never written on this path
021490D0  ldrb  r0, [r0, #0x706]
```

Every instruction on the path to the `bl` touches only `r1`, so `r0` still holds the function's own `self` — which is already a named parameter one line up. Byte-neutral to correct. Worth knowing why the census did not flag it, since that may mean other rows are missing too.

## 3. A function pointer spelled `int`

`src/func_02073244.c:1` becomes:

```c
extern void func_02073300(int a, int b, int c, int dtor);
```

The ROM proves `r3` is code, not an integer: inside `func_02073300`, `0x0207330C movs r5, r3` and then `0x02073344 blx r5` — the register is *called*. And `include/decl_common.h:2102` already spells the corresponding parameter of `func_02073244` itself as `void (*)(void*)`, which the definition at `src/func_02073244.c:4` contradicts with `int d`.

Spelling a function pointer as `int` is the same category of host bug this PR exists to fix — it just happens not to bite on a 32-bit target. Since this PR is the one propagating the value one level deeper, it is the right place to spell it `void (*dtor)(void*)`.

## Smaller type notes (not blocking)

Arity is proven throughout; several parameter *spellings* are inference rather than recovery, and are worth marking as such rather than reading as recovered types: `void *` on `func_ov100_02145080`'s second parameter (the value is only forwarded, never dereferenced, and the caller's own parameter is still `int`); the `(void *)` casts in `func_02009a8c` and `func_ov085_0212e4a4`, which paper over the caller's own unrecovered `int`; and `SetNextLevel(int)`, where `r0` is compared against 1 and stored with `strb`.

Separately, and pre-existing rather than introduced here: `func_ov005_020c007c`'s own definition declares a first parameter the ROM body never reads (`0x020C0084 cmp r1, #8`, `0x020C008C add r0, r1, #0x1c`). The new extern in `src/func_ov005_020c0030.c` inherits that phantom slot. Not this PR's bug, but it is the inverse of the class and worth a row in the triage.

## Verdict

`needs a change` — items 1 and 2. The evidence standard in the PR body is the right one and the work is sound; the gap is that the branch is reasoning about a tree that has since moved, so three of its edits land on files that no longer exist and two live defects go unfixed in the TUs that replaced them.

PR #2515, opened 2026-09-09
The last 18 ride-throughs, one shared declaration, and three over-declared callees

tangosdev:
STACKED ON #2514 (match/w4-sigfix, tip 930853c66). Review that one first; this
branch starts where it stopped and its diff only makes sense on top of it.

WHAT WAS LEFT

#2514 built tools/check_call_arity.py, triaged its census against the ROM, and
corrected 41 functions of the matched-but-wrong-signature class: a body that
reaches its callee with its own incoming registers untouched, decompiled with
fewer parameters than the callee takes. Byte exact, and wrong on any host that
passes arguments on the stack.

It left 23 RIDE rows. One belongs to lane WM7 and is already fixed elsewhere.
Four were held back because they needed something other than a caller edit: a
shared header, and three callees whose own signature was the thing that was
wrong. The remaining 18 were budget. This branch closes all of them.

Census, run on this branch before and after:

    calls passing fewer arguments than the callee declares   196  ->  165
    distinct callees flagged                                 114  ->   90
    ROM triage of those rows: RIDE                            23  ->    1

The one RIDE row left is func_0205d23c to func_0205d304, which belongs to lane
WM7 and is already corrected on port/l6-wm7 and l4merge. Touching it here would
conflict with that merge.

THE 18 RIDE ROWS

Every one byte identical at 2004/b56, every one proved with tools/match.py.

  func_0201fc88 (0/1)
      forwards its message ID to ObjectMessageIDToActualMessageID. The tree
      already agreed on the shape: include/decl_common.h and Player.cpp both
      declare and call it with an s16. Only the definition said (void).

  func_ov006_02111b90 (2/3)
      Vec2_Sub(out, a, b) reads r2, and this caller never wrote it, so its own
      third parameter (the velocity vector its callers already pass) was riding
      through unnamed. src/func_ov006_02112ad8.c had the correct three argument
      declaration and four correct call sites all along.

  the Sound namespace, four callers
      Sound::SetMusic(j1, j2)             passes j2 to func_02011dcc
      Sound::PlayLong(handle, a, b, ...)  passes a and b to func_02048a1c
      func_0201226c                       the same, a second entry point
      Sound::Play(j1, j2, pos)            passes both to func_02050cdc
      func_02048a1c(pos, kind, id) and func_02050cdc(a, idx) both read those
      registers immediately. Sound::SetMusic is the clearest of them: the ROM
      stores r1 to data_0209b4ac and then calls with r1 still live.

  the sound command wrappers, seven callers
      func_0204f86c, func_0204f76c, func_0204f79c, func_0204f7cc, func_0204f7fc
      (three arguments each) and func_0204f89c, func_0204f73c (two). Each reads
      one byte out of a pointed-to object into r0 and calls a func_0205axxx
      handler that takes three, leaving r1 and r2 exactly as they arrived. Every
      caller in the tree already passed the extra arguments.

  func_0205fab4 (1/3)     forwards r1 and r2 to func_0205fbd0.

  func_0206e2f8 (1/3)     this is memset. Destination, fill byte and length,
                          returning the destination, which is what the closing
                          mov r0, r4 says. include/decl_common.h declared it
                          void returning, and its one caller, fBase_c::operator
                          new, is re-verified in the same commit.

  func_02070b98 (1/2)     __num2dec. The value is a double riding in r1:r2, not
                          two ints: the definition spelled it as two words and
                          so could not forward it to func_02070c68(char*,
                          double) at all. The sibling caller func_0206f46c
                          already declared the double form. float2str
                          (func_0206ece0) carried the same two-word fiction, and
                          its 0x658 bytes reproduce with the double spelled
                          honestly, so it is corrected in the same commit rather
                          than left holding a declaration that contradicts the
                          definition.

  dScMgSmartball_c::OnYoshiTryEat (0/1)
                          passes its own this to func_ov006_02115b0c, which
                          reads r0 as the board object on its second
                          instruction.

  func_ov007_020be980 (1/3)
                          forwards the index and the flag to func_ov007_020bc3dc.

THE SHARED DECLARATION

  func_ov002_020e444c to func_ov002_020e3f90 (0/1). The callee is the player
  matrix update and takes the player in r0; include/decl_common.h declared it
  (void). The declaration is corrected to (char*), and its one caller passes the
  player it already had in r4. Both byte verified.

THE THREE ROWS #2514 CALLED EVIDENCE AGAINST THE CLASS

All three settle the same way, and all three settle against the callee, not the
caller. The census was right to flag them and #2514 was right not to guess.

  SaveData::SetDefaultValues and SaveData::SetDefaultValuesMg are STATIC.
      Both ROM bodies take one pointer in r0 and never a this. include/SaveData.h
      declared them non-static, their bodies modelled that r0 pointer as this and
      left the declared parameter unread, and a note in
      _ZN8SaveData16SetDefaultValuesEP12FileSaveData.cpp claimed that correcting
      the declaration would move the argument out of this and change the bytes.
      It does not. The mangled names (EP12FileSaveData, EP16MinigameSaveData),
      the ROM bodies and all five callers agree that the single r0 argument is
      the pointer, and both functions reproduce byte for byte with static and the
      parameter spelled honestly. The read-modify-write on the word at +8 keeps
      its materialised address form (add r2, r4, #8 then ldr/orr/str through r2)
      under the offset expression.

      That also retires four HONEST LEFTOVER notes: EraseSaveFile, ReadFileData,
      ReadMinigameData and EraseAllSaveData reached these two by their raw
      mangled names because a static context could not call a non-static member.
      They make real SaveData::SetDefaultValues and SetDefaultValuesMg calls now.
      All 28 C++ sources that include SaveData.h were re-verified byte exact
      after the header change. PrepareVsMode.c is C and keeps the raw name, which
      is now an honest one argument call.

  func_02020168 and func_02020254 take NO arguments.
      Both write r0 (mov r0, #0) on their third instruction before reading
      anything, so the int parameter each definition carried was invented. Every
      caller in the tree already declared them (void). Dropping the parameter is
      byte identical, and it removes the two rows plus the callers
      func_ov075_02116e44 and func_0202027c from the census.

GATES

  python tools/prepush_linkcheck.py --files <the 28 changed sources, plus the
  three functions this branch moves without editing their own files>
  prepush-linkcheck: 31 checked - 31 verified, 0 warning(s), 0 blocking

  python tools/prepush_linkcheck.py --range 930853c66..HEAD
  2 changed header(s) fan out to 940 source file(s) to verify
  prepush-linkcheck: skipped 1 NONMATCHING draft(s)
  prepush-linkcheck: 939 checked - 647 verified, 292 warning(s), 0 blocking

  The 292 are the fan-out, not this change: 288 NO-SYM (the file stem is a class
  name, so the tool has no single symbol to check, a verdict it reaches before
  any compile) and 4 BLIND-n (reloc slots it cannot resolve). None of them is a
  file this branch edits. Two of the BLIND files, RollingIronBall::InitResources
  and daObjMarioCap_c::InitResources, do not reproduce at 2004/b56 at all; that
  is pre-existing and unrelated, and both fail identically (size differs) with
  the 930853c66 include/decl_common.h restored in place.

  the tool-tests list, run whole: Ran 702 tests in 474.532s, OK (skipped=2)
  (the two expected "no src samples" skips in test_match_provenance)
  check_python_names: PASS (286 files, 0 unresolvable, 49 advisory).
  check_dead_references: no new dead references, no broken markdown links.

WHAT IS NOT IN HERE

  func_0205d23c to func_0205d304, the one remaining RIDE row: it belongs to lane
  WM7 and is fixed on port/l6-wm7 and l4merge.

  --fail-on-hits is still not wired into CI. 165 census rows remain, but they are
  now SETUP (62, the caller writes the register first, so the C is merely terse),
  NOBL (48, no bl decoded), SKIP (23) and CLOBBER (10). None of them is the
  class.


andrewboudreau (commented):
Reviewed on top of #2514 and against current main (`9c7f38a24`). The evidence in this one is unusually good — every claim I tested holds, including the one I most expected to fail. One concrete defect, plus what it inherits from the branch below it.

## What I verified

**The byte gate, independently.** I compiled all 28 sources in this branch's incremental diff with the pinned 2004/b56 compiler against *this branch's* headers — including the changed `include/SaveData.h` and `include/decl_common.h` — and compared to the ROM with strict reloc checking. All 28: `MATCH`.

**The double-in-`r1:r2` claim (`__num2dec` / `float2str`) — confirmed, three ways.** This was the riskiest item, because the natural assumption is that a `double` after a pointer would be 8-byte-aligned into `r2:r3`. It is not, and the decisive proof is `func_0206f46c`'s own call site (its save area is at `sp+0x68`):

```
0206F4A8  ldr r1, [sp, #0x68]   ; incoming r0 = double low
0206F4B4  ldr r2, [sp, #0x6c]   ; incoming r1 = double high
0206F4B8  add r0, sp, #8        ; decform*
0206F4BC  add r3, sp, #0x1c     ; decimal*
0206F4C8  bl  #0x2070b98
```

`(decform*, double, decimal*)` fills exactly `r0`, `r1:r2`, `r3` with **nothing spilled to the stack**. Under an 8-byte-aligning ABI the double would sit in `r2:r3` and the third pointer would go to the stack. Corroborated inside `func_02070c68`, which takes the IEEE-754 sign bit out of saved incoming `r2` (`0x02070C74 ldr r1,[sp,#0x88]`, `0x02070C7C ands r0, r1, #0x80000000`) and reassembles `(lo=r1, hi=r2)` into `r2:r3` for the soft-float compare at `0x02070CA4`. And the pass-through holds: between `func_02070b98`'s entry and its `bl` only `mov r5,r3` / `ldrsh r4,[r0,#2]` / `mov r0,r5` execute, so `r1` and `r2` arrive untouched.

**`SaveData::SetDefaultValues` / `SetDefaultValuesMg` are static — confirmed at all five call sites.** Both bodies `memset` their incoming `r0` (0x44 and 0x2e4 bytes) with `mov r1,#0` as the second instruction, so `r1` is dead on entry. I then checked every call site (exactly five reloc rows target `0x02013dc4` and `0x02013c5c`): **not one sets `r1`.** Two of them load `r0` from a PC-relative literal naming a global — `0x0209cae4` at the call from `0x02013E44`, and `0x0209caa0` at the call from `0x0202AE58`. A non-static call needs a second pointer in `r1` and there is nowhere for one to come from.

The fan-out risk is also nil, which the PR body could say more plainly: the whole tree contains exactly these five callers, and all five already reached the functions by their raw mangled name. Nothing calls them as a member, so making them `static` cannot silently move an unexamined call site.

One nuance I would not overstate: the site at `0x0202AE58` is the only one ambiguous *in isolation*, because `sizeof(SaveData)` is also 0x44 and `0x0209caa0` is the global instance, so `r0` there could be read as a `this`. It is settled by the absent `r1` and by the other four. Four sites are independently decisive; the fifth is consistent.

**`func_ov002_020e3f90` — the shared-header change is safe.** The callee reads `r0` before writing it (`0x020E3F98 mov r4, r0`, `0x020E3F9C ldr r0, [r4, #0x35c]`), and the caller never writes `r0` before the `bl` at `0x020E4458`. More to the point for a `decl_common.h` edit: there is exactly **one** caller in the tree, and the definition in `src/func_ov002_020e3f90.c` already took `char*` — the shared header was the outlier contradicting it. No other file can be affected.

**The rest.** `func_02020168` / `func_02020254`: `mov r0, #0` really is the third instruction word in each, with no read of `r0`–`r3` before it, and all seven existing caller declarations already said `(void)`. `Vec2_Sub` reads `r2` (`0x0203D6DC ldr r3, [r2]`) and `func_ov006_02111b90` writes `r0`, `r1`, `r4`, `r5` but never `r2` on the path to the `bl`. `dScMgSmartball_c::OnYoshiTryEat` has *no* instruction between entry and the `bl` (`stmdb`/`sub sp`/`bl`), so `this` reaches `func_ov006_02115b0c` untouched — stronger than the PR claims; the callee captures it at `0x02115B14 mov r8, r0`, which is the third instruction word rather than "the second instruction". `func_0206e2f8` is memset-returning-destination: `push {r4,lr}` / `mov r4,r0` / `bl` / `mov r0,r4`.

No arity claim in this branch failed.

## The defect: a `long double` left behind

`src/func_0206ece0.c:53` becomes `char *func_0206ece0(double x, char *buff_end, print_format format)`. Its **sole caller is not updated**, and still declares the wider type:

```
src/func_0206e4a4.c:46: extern char *func_0206ece0(long double num, char *buff_end, print_format format);
src/func_0206e4a4.c:47: extern char *func_0206f46c(long double num, char *buff_end, print_format format);
src/func_0206e4a4.c:160:  func_0206ece0(long_double_num, conv_buff + MAX_BUFF, format)
src/func_0206e4a4.c:171:  func_0206f46c(long_double_num, conv_buff + MAX_BUFF, format)
```

On this target `long double == double` — the ROM proves it at `0x0206E8F4`–`0x0206E910`, which pulls exactly 8 bytes off the varargs pointer and passes them in `r0:r1` — so this is byte-identical and links fine. But it is the same class of bug this branch exists to eliminate: on any host where `long double` is 80- or 128-bit, the caller passes a wider value than the definition reads.

Line 47 is pre-existing (`func_0206f46c` has taken `double` on main since before this branch). Line 46 is newly one-sided: before this PR the definition took `(int x0, int x1)` and the caller declared `long double`, both landing in `r0:r1` by accident; now one side is right and the other still says `long double`. Since this branch is the one that settles the type, it is the right place to fix both lines.

## What it inherits

**The three promoted-TU conflicts from #2514.** `git merge-tree` against main reports the same three modify/delete rows: `src/func_ov002_020af218.c`, `src/func_ov065_021165d8.cpp` and `src/func_ov084_0212f204.c` were absorbed into `src/actors/da1up_c.cpp`, `src/actors/daYurei_Mucho_c.cpp` and `src/actors/daPkn_c.cpp` after the base. Details and the surviving defect line numbers are in my review on #2514; they apply here unchanged because this branch carries those commits.

**Stacking.** This targets `main` while containing #2514's tip. Since #2514 needs changes, this will need a restack rather than a merge in its present shape.

**One stale sentence.** The GATES section says two of the BLIND fan-out files, `RollingIronBall::InitResources` and `daObjMarioCap_c::InitResources`, "do not reproduce at 2004/b56 at all; that is pre-existing and unrelated." That was true at the base and is no longer true on main — both now reproduce at exactly the ROM's size (0x38c and 0x4c8), fixed by the merged "three InitResources floors were one lever" change. Doesn't affect anything here; worth correcting so the claim isn't carried forward.

## Smaller notes (not blocking)

Arity is proven everywhere; a few parameter *spellings* are house placeholders rather than recovered types, and are worth marking as such: `func_ov002_020e3f90(char*)` is a player object pointer, not a character buffer; `func_ov006_02111b90`'s third parameter and `Vec2_Sub`'s are spelled `int*` while other files in the tree already spell the same thing `V2*`; and the `decimal*`/`decform*` pointers in the `__num2dec` pair are spelled `void*`/`char*` here while the neighbouring `func_0206ece0.c` models them as real structs — post-merge, `func_02070b98.c` declares `func_02070c68(void* out, ...)` against `func_02070c68.c`'s `(char *out, ...)`.

## Verdict

`needs a change` — the `long double` pair at `src/func_0206e4a4.c:46-47`, plus the inherited promoted-TU replay from #2514. The signature work itself is the best-evidenced of the four PRs I looked at today; the ROM backs every claim, and the shared-header edits are both narrower in fan-out than they look.

PR #2516, opened 2026-09-09
Symbol-scope crossings: eleven of the eighteen closed by address, and the audit tool's thread race fixed

tangosdev:
Branch: config/w4-xscope, off origin/main bdcb1df7d, in C:/tmp/w4-exc2
Tip: 8964e89d0 (8 commits on top of bdcb1df7d)

SUMMARY

  tools/symscope.py's default invocation (-j12, matching -j8 too) crashed
  partway through the full audit on the CURRENT origin/main, before any of
  this lane's source fixes: a ThreadPoolExecutor race where two candidate
  rows sharing one enrolled source file (a consolidated TU) hit the same
  build/<rel>.o path on different threads. Fixed (audit_file: one compile
  per FILE, not per candidate) and covered by a new tools/test_symscope.py
  (there was none before). Full detail in the FIRST table below.

  The three crossings this run's port lanes (PMFB7, MPG) found (19th
  HootTheOwl, 20th SkiLift/MotherPenguin, 21st func_ov018_021121dc) are ALL
  THREE already fixed on origin/main, by earlier commits that predate this
  run and that the port branches those lanes worked from had not absorbed.
  No decomp-side action needed; verified by direct source inspection plus,
  for 19 and 20, a clean symscope.py run restricted to the touched files.

  Of the 18 older WRONG-MODULE crossings PR #1553 left open (2026-08-16),
  7 had already been fixed by other commits before this run started. This
  lane fixed the remaining 11 (8 files, 7 commits, one commit per crossing
  address), each byte-verified with match.py under 2004/b56 with strict
  relocs and confirmed with prepush_linkcheck. 0 of the 18 are left open.

  One brand-new WRONG-MODULE finding turned up in the fixed tool's full
  audit that was in NEITHER PR #1553's original 19 nor this run's 19/20/21:
  src/_ZN12dScStarSel_c6RenderEv.cpp (ov003), 11 sites naming data_ov000_*
  addresses. Left OPEN -- see "NOT FIXED" below; it does not fit the
  swap-the-name pattern the other 21 do.

SYMSCOPE BEFORE / AFTER (both against a clean tree, no local uncommitted
changes; "before" = bdcb1df7d with the tool crash fixed but source
untouched, "after" = this branch's tip)

  BEFORE (build/symscope_baseline.json, tmp/symscope_baseline.log):
      7928  clean
        23  WRONG-MODULE
      1751  CROSS-OVERLAY
[...]

PR #2520, opened 2026-09-09
symscope: thread the manifest policies through the audit, and decide module scope by overlay residency

tangosdev:
NO-OBJECT manifest-licensing gap, and the numeric span heuristic

Branch: tools/w4-symscope2, off config/w4-xscope 8964e89d0 (PR #2516), in
C:/tmp/w4-exc2
Tip: ec378265d (1 commit on top of 8964e89d0)

SUMMARY

  Two fixes to tools/symscope.py, both against the same root cause: the
  audit was answering "is this module reachable" with less information than
  the byte gate and the ROM's own loader actually have.

  1. NO-OBJECT (2221 of 11280 enrolled files' candidates). main()'s ctx
     built vers/cache/init_srcs/syms for RB.compile_one but never called
     RB.compiler_only_policies()/RB.intact_tu_policies(), the two manifest-
     derived dicts rombuild.py's own production main() always threads
     through every compile_one call so _isolate can strip or admit exactly
     the RTTI/vtable content a promoted consolidated TU's manifest licenses,
     before objisolate.isolate_many's fail-closed text-only check runs.
     Without them every promoted TU's compile hit isolate_many on the
     UNREDUCED object (still carrying its _ZTI/_ZTS/_ZTV .data sections) and
     failed "unlicensed content in text-only multi-symbol object" --
     confirmed directly against src/game/actors/d_a_pg_mthr.cpp (crossing
     21's own file, config/tu_manifest.d/ov018/daPgMthr_c.json): the bare
     compile_one call NO-OBJECTs it; the identical call with both dicts
     passed hits the object cache and succeeds. audit_file now passes both
     through, unchanged from what rombuild.py's own main() passes.

  2. THE SPAN HEURISTIC. audit_one classified WRONG-MODULE/CROSS-OVERLAY
     from module_spans()'s address-span overlap alone, which cannot tell
     "these two overlays share a RAM slot" from "these two happen to fall
     in overlapping address ranges but the ROM's own loader never actually
     has both resident at once" -- dScStarSel_c::Render's data_ov000_*
     sites are exactly the second shape (detail below). tools/
     overlay_residency.py already carries this ground truth (rules E1-E6,
     read out of the ROM's own LoadOverlay resident-table crash guard, the
     level/object overlay tables, and dsd's own unambiguous call
     relocations) but was never wired into symscope.py. audit_one now asks
     overlay_residency.possible(candidate_module, this_module, this_
     function) per candidate owner instead of computing the overlap itself.
     module_spans()/_overlap() are unchanged and still drive the unrelated
     --families summary.

RESIDENCY RULE AND ITS SOURCE

  tools/overlay_residency.py (already in the tree since PR #1105, 2026;
  untouched by this lane except for being imported). Its module docstring
  states the rule and the six ROM-proven facts it encodes:

    E2  overlapping ranges are mutually exclusive -- ov000's own
        LoadOverlay (src/LoadOverlay.c) walks a 12-entry resident table
        and calls Crash() when an incoming overlay's
        [start, start+code+bss) intersects one already there. The game
        enforces this itself; it is not an assumption about the linker.
    E1  ov000 is a boot overlay: loaded, entered, unloaded, before
        anything else (src/func_0201a2f8.c:41-43, the ROM's only
        LoadOverlay(0)/UnloadOverlay(0) sites)
    E3  loading ov006 always loads ov004
    E5  an unambiguous CALL relocation proves its target was resident
    E6  the level -> overlay tables say which overlays a level loads

  Falsification, re-run by me this session:
      python tools/overlay_residency.py --check
      cross-module unambiguous verdicts: {'call': 2322, 'load': 932}
      forbidden by this model: 10
         ('ov007', 'ov002', 'load'): 4
         ('ov007', 'ov006', 'load'): 2
         ('ov009', 'ov006', 'load'): 2
         ('ov009', 'ov007', 'load'): 1
         ('ov024', 'ov007', 'load'): 1
      calls contradicted: 0  <- must be 0

  0 of 2322 unambiguously-resolved CALL relocations contradict this model
  (the 10 forbidden are pointer-shaped DATA words, no control flow, exactly
  what the tool's own docstring already claims). That is the strongest
  evidence available short of tracing real hardware, and it is what
  audit_one now asks per candidate instead of computing span overlap
  itself.

THE 12-SITE VERDICT (dScStarSel_c::Render, ov003)

  pr.txt (XSCOPE, this run) undercounted this "11 sites" in prose while its
  own CROSSINGS LEFT OPEN list already named 12 addresses; the true count,
  confirmed by both the old and the new audit, is 12.

  Ground truth: overlay_residency.possible("ov000", "ov003",
  "_ZN12dScStarSel_c6RenderEv") = False. By hand, independently of the
  tool: ov000's range is 0x020aa420..0x020bf4e0 (overlays.yaml: base
  34251808, code_size 86208 -- matches config/arm9/overlays/ov000/
  delinks.txt's header exactly). ov003's own range is 0x020ad660..
  0x020b18a0, entirely inside ov000's. src/func_0201a2f8.c is the ONLY
  place in the whole decompiled ROM that calls LoadOverlay(0) or
  UnloadOverlay(0) (grep confirms: 17 total Load/UnloadOverlay call sites
  in src/, exactly one names overlay 0), and it unloads ov000 three lines
  after loading it, before immediately loading ov001 (never unloaded
  again) and returning. Every scene that can ever reach dScStarSel_c
  (Star Select, shown after clearing a level) runs long after that boot
  sequence, so ov000 cannot be resident when Render() executes. ov001,
  ov000's same-base twin, occupies exactly 0x020aa420..0x020ad660 (its own
  code_size 12800 + bss 64 lands precisely at ov003's family's own start)
  and IS the module actually resident there from boot onward; config/arm9/
  overlays/ov001/symbols.txt independently defines the identical 12
  addresses under data_ov001_* names ("ambiguous", same as ov000's own
  entries), confirming dsd already knew both candidates existed and could
  not disambiguate them -- which is exactly why its relocs.txt records the
  plural overlays(0,1) spelling at every one of these 12 sites instead of
  picking one.

  Verdict: WRONG-MODULE stands, ground-truthed rather than assumed. This
  resolves what XSCOPE's pr.txt explicitly left open ("(a) confirmation
  from the real DS overlay loader ... or (b) evidence that dScStarSel_c::
  Render legitimately reads through since-repurposed RAM") -- it is (b):
  ov001, not ov000, is what is actually resident, and the fix (out of
  scope here, per the brief) would rename all 12 data_ov000_0xxxxxxx sites
  to their data_ov001_0xxxxxxx twins, the same swap-the-name pattern as
  the other 21 already-fixed crossings. Not touched by this lane.

NO-OBJECT BEFORE / AFTER (both against a clean tree; "before" is XSCOPE's
own after-state at 8964e89d0 -- out/XSCOPE/symscope_after.log, this
lane's exact branch point, re-quoted verbatim below rather than re-run,
since it is already this lane's "before"; "after" is this branch's tip,
freshly run)

  BEFORE (out/XSCOPE/symscope_after.log, build/symscope_after.json):
      7938  clean
        12  WRONG-MODULE
      1751  CROSS-OVERLAY
      2221  NO-OBJECT

  AFTER (tmp/symscope_full_after.log, build/symscope_full_after.json,
  11280 enrolled source files, same as before -- no file-count drift):
      9573  clean
        14  WRONG-MODULE   (+2, both new -- see below)
      2677  CROSS-OVERLAY  (+926, all newly-visible candidates)
        36  NO-OBJECT      (-2185)

  The 36 remaining NO-OBJECT rows are NOT the manifest-licensing gap this
  lane closed. They are 2 files (27 + 9 candidate rows), both
  production_mode: intact-object TUs (config/tu_manifest.d/ov036/
  daObjRcCarpet_c.json, config/tu_manifest.d/ov070/daPropeller_Heyho_c.
  json), and both fail identically: "isolate: intact TU preparation
  refused: ov0XX/<class> vtable address-point policy: vtable partition
  baseline proof unavailable: baseline content fingerprint differs for
  ['trackedConfigArm9Sha256', ...]" -- tools/tubuild.py's
  validate_partition_baseline_evidence refusing because config/arm9's
  current content hash no longer matches the hash recorded when these two
  TUs' intact-object promotion was last proven (unsurprising: config/arm9
  has churned all night across this run's many lanes, this lane's own
  XSCOPE parent included). This is a stale-proof condition tubuild.py's
  own linkcheck --baseline regenerates; it is unrelated to the manifest-
  licensing gap fixed here and out of this lane's scope. Confirmed this is
  not caused by anything in this branch: `git status --short` /
  `git diff --stat` here shows only tools/symscope.py and tools/
  test_symscope.py touched, no config/ files.

NEW CANDIDATE ROWS (surfaced by the NO-OBJECT fix; each is a real WRONG-
MODULE finding neither the old span test nor the old NO-OBJECT gap could
show; addresses given; NOT fixed here per the brief)

  1. src/minigames/d_s_mg_single3_d_base.cpp (ov006), function
     _ZN19dScMgSingle3DBase_c9Virtual84Ev, site +0x1a4, names
     data_ov000_020beb74 (defined ov000:0x020beb74). config/arm9/
     overlays/ov006/relocs.txt independently agrees: 0x020beb74:
     overlays(0,4), the same StarSel-shaped ambiguous-twin situation (ov000
     again). Ground truth: overlay_residency.possible("ov000", "ov006",
     "_ZN19dScMgSingle3DBase_c9Virtual84Ev") is False by the same E1 rule
     -- ov000 is never resident anywhere past boot. NOTE: ov000's own
     address range does NOT numerically overlap ov006's family span
     (ov000 ends 0x020bf4e0, ov006's family starts 0x020bfec0), so the OLD
     span-overlap test would have called this CROSS-OVERLAY (soft) even
     with NO-OBJECT fixed and this file made visible -- it takes BOTH
     fixes together to surface this as the real WRONG-MODULE it is. Given
     the address falls in the ov000/ov001 ambiguous-twin zone, the likely
     fix (not attempted) is the same swap-to-ov001 pattern as the 12
     StarSel sites: data_ov001_020beb74.

  2. src/minigames/d_s_mg_trampoline2.cpp (ov006), function
     func_ov006_02123bf4 (a folded-in former standalone src/func_ov006_
     02123bf4.cpp, per this file's own TUBUILD manifest comment at line 49
     and the inlined body at line 827), site +0xc, names
     _ZN6Player12St_Null_InitEv (defined ov002:0x020cac30) through a
     shadow `struct Player { void St_Null_Init(); };` declaration (line
     138) that mangles to the wrong callee. config/arm9/overlays/ov006/
     relocs.txt DISAGREES here (config_agrees: False): the real
     destination is 0x020cac30:ov006, i.e. func_ov006_020cac30, an
     ov006-local function -- the exact same address and "should name" PR
     #1553's own row 8 already reported fixed, in what was then the
     standalone src/func_ov006_02123bf4.cpp. That file no longer exists on
     its own; its content is now inside d_s_mg_trampoline2.cpp, and the
     wrong name is back. Worth a follow-up lane's attention: either the
     "already fixed" claim in PR #1553/XSCOPE's Table 2 checked a
     different call site than this one, or a later TU fold
     (tubuild.py create, this file's own header records at least one such
     fold event pattern) picked up a pre-fix legacy declaration when
     merging. Not diagnosed further or fixed here.

TESTS

  tools/test_symscope.py: 14 -> 15 tests. AuditOneVerdicts rewritten to
  stub overlay_residency.possible() (mock.patch.object(S.OR, "possible",
  ...)) instead of hand-built span dicts -- every existing WRONG-MODULE/
  CROSS-OVERLAY test now asserts audit_one calls possible(candidate_
  module, caller_module, caller_function) and obeys whatever it returns.
  New: test_residency_answer_governs_not_address_math, the StarSel shape
  exactly (ov000 0x020aa420..0x020bf4e0 vs an ov003 caller starting
  0x020ad660 -- real, numerically overlapping config), run twice with the
  IDENTICAL fixture and only overlay_residency.possible()'s mocked return
  value changed: False -> WRONG-MODULE, True -> CROSS-OVERLAY, proving the
  verdict tracks the residency answer and not the address relationship
  (which is unchanged between the two calls). test_own_module_and_always_
  resident_are_invisible now also asserts overlay_residency.possible is
  never called for a name visible where the file already links.
  AuditFileCompilesOnce's two tests extended to assert audit_file forwards
  ctx["compiler_only"]/ctx["intact_tus"] through to compile_one unchanged
  (sentinel objects, identity-checked).

  Ran 15 tests in 0.008s
  OK

  python tools/test_symscope.py -v: all 15 pass (see above).

CI TOOL-TESTS LIST, run whole (the exact enumerated command in .github/
workflows/tool-tests.yml; tools.test_symscope is deliberately still not
in this list, per that workflow's own header comment -- "a separate
reviewed decision", left alone here too):

  Ran 683 tests in 484.359s
  OK (skipped=2)

  python tools/check_python_names.py:
    python-names: 285 tracked file(s) checked
    0 unresolvable names, 0 unparseable files, 49 advisory finding(s)
    python-names: PASS

  python tools/check_dead_references.py:
    check_dead_references: 10214 scan target(s), 9762 C/C++ file(s), 9468
    repo-rooted path reference(s), 1727 that do not resolve; 1158 markdown
    link(s), 0 that do not resolve relative to their own file
    3 baselined prose reference(s) now resolve (not a failure)
    4 baselined C/C++ comment reference(s) now resolve (not a failure)
    no new dead references
    no broken markdown links

FILES TOUCHED (2)

  tools/symscope.py       (compiler_only/intact_tus threaded through
                           compile_one; audit_one asks overlay_residency.
                           possible() instead of computing span overlap;
                           docstring rewritten for both)
  tools/test_symscope.py  (AuditOneVerdicts rewritten around
                           overlay_residency stubs; new residency test;
                           AuditFileCompilesOnce extended)

NOT TOUCHED / OUT OF SCOPE (per the brief: "fix none here")

  - The 12 StarSel data_ov000_* sites (rename to data_ov001_*)
  - d_s_mg_single3_d_base.cpp's data_ov000_020beb74 (same rename shape)
  - d_s_mg_trampoline2.cpp's _ZN6Player12St_Null_InitEv (rename to
    func_ov006_020cac30; also worth understanding how the earlier fix was
    lost)
  - The 2 remaining NO-OBJECT rows (intact-object baseline restamp;
    tools/tubuild.py linkcheck --baseline territory, not symscope.py's)
  - Wiring tools.test_symscope into the CI enumerated list (left to a
    follow-up per the workflow's own stated policy, same as XSCOPE left it)

Never pushed, no PR opened, per the brief.


andrewboudreau (commented):
Reviewed at head `8cd3e6cd4`. The NO-OBJECT half is a clear win and I have no objection to it. The residency swap has one defect: it trades a tracked input for a gitignored one, and the degradation is silent and in the wrong direction.

## The residency data is gitignored; the audit goes quiet without it

The test that was replaced read `module_spans()`, built from `config/arm9/**/delinks.txt` — tracked, present in every clone.

The replacement, `tools/symscope.py:223`

    exclusive = {m for m in owners if not OR.possible(m, module, name)}

decides through `overlay_residency.conflict()`, whose `RANGE` table is parsed from `extracted/dsd/arm9_overlays/overlays.yaml`, and through E6's `LEVEL_SYSTEM`, parsed from `extracted/arm9_dec.bin`. Both live under the gitignored `extracted/` tree. `overlay_residency.py:52` returns `{}` when the yaml is absent, and `overlay_residency.py:83` then makes `conflict()` return False for every pair, so `possible()` returns True for every owner and `exclusive` is empty. `tools/symscope.py` has no guard on this: no check, no warning, no non-zero exit.

`overlay_residency`'s own module docstring states the polarity it was written for:

> `possible()` is deliberately one-sided: it can rule a candidate *out*, never in. When the data is missing (a cold clone has no `extracted/`) it rules nothing out, so the caller falls back to refusing — which is the safe direction here.

That is the safe direction for `settle()`, which refuses to pick a name. symscope is the caller that inverts it: "rules nothing out" becomes "condemn nothing", so a hard WRONG-MODULE verdict silently becomes CROSS-OVERLAY, which this tool's own docstring describes as "reported for eyes, not failed".

`symscope` does not need `extracted/` for anything else — the ROM preflight in `rombuild.py` is on the full-build path, not on `compile_one` — so it will run happily and print a clean-looking summary.

## Measured on this branch's own tip

Residency answers, real `extracted/` vs. `RANGE` emptied (the no-`extracted/` state):

| query | with extracted/ | without |
|---|---|---|
| `possible("ov002","ov006","func_ov006_02123bf4")` | False | **True** |
| `possible("ov019","ov018","SkiLift::Behavior")` | False | **True** |
| `possible("ov000","ov003","_ZN12dScStarSel_c6RenderEv")` | False | False |

The old span test, from tracked config alone, returns overlapping (i.e. condemn) for both of the first two. The ov000 row survives only because E1 is a hardcoded constant, not because any data was read.

End to end, this branch, `tools/symscope.py src/minigames/d_s_mg_trampoline2.cpp -j1`:

    with extracted/:      28 clean,  1 WRONG-MODULE, 17 CROSS-OVERLAY
    RANGE emptied:        28 clean,  0 WRONG-MODULE, 18 CROSS-OVERLAY

The row that vanishes is the real one — an `arm_call` at 0x02123c00 that `ov006`'s own relocs record as `module:overlay(6)` while the source names an `ov019`-class symbol from another module. The second table row above is the `ov014/ov018/ov019/ov034` RAM-slot shape this tool's docstring names as its reason for existing.

## The tests cannot catch this

Every residency case in `tools/test_symscope.py` patches the answer out: `mock.patch.object(S.OR, "possible", ...)` at lines 108, 132, 150, 177, 186 and 200. The `AuditOneVerdicts` docstring says so deliberately — "`overlay_residency` earning that answer ... is its own module's job". That is a reasonable split, but it means nothing in the suite pins that the answer is ever computed from data that is actually there. A test that leaves `possible` unmocked with `OR.RANGE = {}` and asserts symscope refuses (or at least warns) would close it.

Suggested fix: make `main()` refuse, or print a loud banner into the summary, when `overlay_residency.RANGE` is empty — the audit cannot distinguish "nothing is wrong" from "I have no facts" otherwise.

## Separately: this branch no longer merges

`git merge-tree` against current `main` reports three modify/delete conflicts, and all three fixes have already landed on main by another route:

- `src/func_ov078_02123aa0.c`, `src/func_ov078_02123c20.cpp` — folded into `src/actors/daBombking_c.cpp` by the daBombking consolidation; that file already calls `func_ov078_02123804(c)` at lines 423 and 471, which is exactly what the two commits here change.
- `src/func_ov065_021162c0.c` — folded into `src/actors/daYurei_Mucho_c.cpp` by the daYurei_Mucho promotion; line 647 there records the same correction in prose.

Resolving these by keeping the branch's side would resurrect three retired shard files next to their promoted definitions. Both commits look droppable on a restack.

PR #2521, opened 2026-09-09
Nine functions that named a data symbol's address as a bare integer, and the sweep that finds the shape

tangosdev:
STACKED ON #2515 (match/w4-sigfix2, tip 278d91e35). Review that one first; this
branch starts where it stopped and its diff only makes sense on top of it.

WHAT WAS LEFT

#2514/#2515 closed the matched-but-wrong-signature census down to one known
RIDE row (WM7's, out of scope here). Two port lanes surfaced a DIFFERENT shape
of the same bug while porting the game to PC: a matched function that returns
or forwards a literal DS address as a plain integer instead of naming the data
symbol relocs.txt already records there. Same defect family -- byte-exact and
wrong -- but invisible to check_call_arity (arity is fine; it's the address
itself that's dishonest) and invisible to --strict-relocs too, because a
hardcoded literal has no relocation slot for the check to compare.

WM8 found func_02065b94/func_02065b88 by hand: each returned a DECIMAL integer
(34247892, 34248048) that a first read misjudged as arbitrary, until relocs.txt
turned out to name both loads (0x020a94d4, 0x020a9570). CTOR2 found
func_0201aac8: declared void(void), its tail-call veneer never touches r0, so
whatever its only caller (func_0201aad4) passed rode straight through to
func_02059ba0(char *), and that caller passes one argument through a K&R
extern -- outside check_call_arity's scoring (K&R `()` is unscored arity) but
squarely inside this shape once you follow it into func_02059ba0's own
definition.

THE THREE NAMED FIXES

  func_02065b94 / func_02065b88
      `return 34247892;` / `return 34248048;` -> `return data_020a94d4;` /
      `return data_020a9570;` (as `void *`, array-decays). --strict-relocs
      resolves the pool word for real now: cand=data_020a94d4 (0x020a94d4),
      cfg=0x020a94d4:arm9, verdict OK -- and the same for data_020a9570.

  func_0201aac8 (+ func_0201aad4's caller-side extern)
      `void func_0201aac8(void) { func_02059ba0(); }` ->
      `void func_0201aac8(char *p) { func_02059ba0(p); }`. The ROM body is
      `ldr ip, [pc]; bx ip` -- a pure tail-call veneer that never touches r0 --
      so the parameter was always there, just unspelled. func_0201aad4's K&R
      `extern int func_0201aac8();` became a real prototype
      `extern void func_0201aac8(char *p);` to match. This is the exact row
      SIGFIX's original census carried at line 95 (caller func_0201aac8, callee
      func_02059ba0, passed 0, declared 1) and check_call_arity confirms it:
      the row is gone (165 -> 164 rows, 90 -> 89 distinct callees; diffed
      against SIGFIX2's own census_after.txt, nothing else moved).

THE SWEEP: tools/check_literal_load.py

Shape (a) generalizes: a MATCHED function (no live NONMATCHING banner) whose
own kind:load relocation's destination address also appears as a bare integer
literal (decimal or hex) in its source, outside a comment or string. 20 tests
(tools/test_check_literal_load.py), wired into tool-tests.yml.

Two real bugs turned up building it, both fixed in the tool before the sweep
ran for real:

  - a cross-module lookup bug: func_02058df4 (arm9) loads data_023c0000, which
    lives in the DTCM symbols.txt, not arm9's. Looking the destination up
    under the caller's own module reported no symbol for something that has
    one. The census now resolves through the reloc's own recorded module.
  - Entry.c's CRT0 boot trampoline (HAND-ASM PRIMITIVE) spells every pool load
    as a raw `ldr r0, =0x...` operand by construction -- that's how asm is
    written, not the bug this hunts, and there's no C form to rewrite it into.
    asm_policy.is_asm_passthrough now excludes hand-written asm bodies.

Full-tree census after both fixes: 7 candidates. One (func_02058df4's case 0,
which also returns 0x23c0000) is a confirmed FALSE POSITIVE -- the ROM compiles
that case as `mov r0, #0x23c0000` (a plain ARM immediate, since the value fits
the rotated-8-bit encoding), not a pool load; the real load with that same
address value is case 4, a different instruction entirely. Rewriting case 0 to
`&data_023c0000` compiles to a pool load and stops matching -- match.py caught
it immediately, and it's reverted. The tool's docstring now documents this
exact false-positive class: every row is a candidate, verified with match.py
before it counts as a fix, never landed on the census alone.

THE SIX REAL HITS

  func_02054430
      `(unsigned short *)0x020a6088` -> `&data_020a6088`. Nine other sources in
      the tree already reference this VRAM register by name; this was the one
      that hadn't been touched.

  IRQ::GameCardIREQMCHandler (_ZN3IRQ21GameCardIREQMCHandlerEv)
      `*(volatile int*)0x020a89a0 = r;` -> `data_020a89a0 = r;`. The file
      already named its neighbor four bytes over (data_020a89a4) by symbol;
      this one was the oversight next to it.

  func_0207322c / func_02073238
      Same shape as the two named fixes: `return 34251764;` / `return
      34251768;` -> `return data_020aa3f4;` / `return data_020aa3f8;`. These
      are the FP-status-word and dispatch-table accessors eight ITCM callers
      (func_01ff859c.c, func_01ff97d8.c, _dmul.c, and others) reach through
      `bl func_0207322c` / `bl func_02073238`.

  func_ov007_020b84e8 / func_ov007_020b84f4 / func_ov007_020b8500
      Three ov007 siblings of the identical 12-byte `ldr r0,[pc]; bx lr; .word`
      shape: `return 34617752/34619288/34620824;` -> named returns of
      data_ov007_02103998 / data_ov007_02103f98 / data_ov007_02104598.

All nine functions byte-verified at 2004/b56 with tools/match.py; every named
data destination confirmed OK by the strict-relocs destination check (cand ==
cfg, same address, same module).

VERIFICATION

  tools/prepush_linkcheck.py --range 278d91e35..HEAD
      11 checked - 11 verified, 0 warning(s), 0 blocking

  tools/check_call_arity.py --summary
      calls passing fewer arguments than the callee declares:  165 -> 164
      distinct callees flagged:                                 90 ->  89
      (diffed against SIGFIX2's census_after.txt: exactly the func_0201aac8/
      func_02059ba0 row is gone; nothing else moved. WM7's func_0205d23c row
      is out of scope here, untouched, and does not currently score anyway --
      it belongs to the port branch per SIGFIX2's note.)

  tools/check_literal_load.py --summary
      11 candidates -> 1 (the confirmed false positive above, documented, left
      alone on purpose)

  CI tool-tests list (python -m unittest, 33 modules including the new one)
      Ran 722 tests, OK (skipped=2, the pre-existing expected AliasSourceFallback
      skip) -- 702 (SIGFIX2's count) + 20 for tools.test_check_literal_load

  tools/check_python_names.py    PASS (288 tracked files, 0 unresolvable)
  tools/check_dead_references.py  no new dead references, no broken markdown links

  tools/rombuild.py -j8 --no-rom fails at its pre-existing baseline-control
  step (dsd check symbols --fail: overlay_100/overlay_102, data_020ad524/
  data_020ad560, and an ITCM func_01ff98f4 family) -- none of those symbols
  belong to any file this branch touches, and the tool's own output labels it
  "the same scratch pipeline with NO TU substitution: anything failing here is
  pre-existing and belongs to the tree, not to any TU." Module fidelity within
  that same run: 106/106 modules exact, 10,905/10,905 source-built functions
  reproducing, 100.000000% of compared bytes. match.py and prepush_linkcheck
  are the authoritative local proof per repo convention; this is a known local-
  build limitation, not a regression from this branch.

FILES TOUCHED

  src/func_02065b94.c
  src/func_02065b88.c
  src/func_0201aac8.c
  src/func_0201aad4.c
  src/func_02054430.c
  src/_ZN3IRQ21GameCardIREQMCHandlerEv.cpp
  src/func_0207322c.c
  src/func_02073238.c
  src/func_ov007_020b84e8.c
  src/func_ov007_020b84f4.c
  src/func_ov007_020b8500.c
  tools/check_literal_load.py            (new)
  tools/test_check_literal_load.py       (new)
  .github/workflows/tool-tests.yml       (wires the new test module in)

NOT DONE

  func_02058df4's case-0 literal (0x23c0000) -- confirmed correct as written,
  see above. No other rows remain in the sweep.


andrewboudreau (commented):
Reviewed at head `675507200`. The substance checks out — I re-ran the parts that a byte gate cannot prove on its own — but three of the edited files no longer exist on `main`, and two of those three fixes are still needed at their new homes.

## What I verified

**The literal-to-symbol changes are the right direction, not a cosmetic rename.** For each one, `config/arm9/relocs.txt` already records a `kind:load` relocation at exactly the pool word the function owns, and the symbol already exists in `config/arm9/symbols.txt`:

| function | size | pool word | recorded relocation |
|---|---|---|---|
| `func_02065b88` | 0xc | 0x02065b90 | `from:0x02065b90 kind:load to:0x020a9570 module:main` |
| `func_02065b94` | 0xc | 0x02065b9c | `from:0x02065b9c kind:load to:0x020a94d4 module:main` |
| `func_0207322c` | 0xc | 0x02073234 | `from:0x02073234 kind:load to:0x020aa3f4 module:main` |
| `func_02073238` | 0xc | 0x02073240 | `from:0x02073240 kind:load to:0x020aa3f8 module:main` |

The recorded size includes the pool word, so this is inside what the byte gate compares. Re-ran `match.py` at 2004/b56 on `func_02065b88`, `func_02054430` and `func_02073238`: all three `MATCHING VERSIONS: 2004/b56`, and the pool word now reports as `reloc (wildcard)` where it used to compare as a literal — the word has become a checked relocation destination instead of a blind constant. The decimal literals were also correct (`34248048 == 0x020a9570`, `34617752 == 0x02103998`, etc.), so nothing moved.

**The arity claims are grounded in the ROM body, not inferred from register traffic.** `src/_ZN8dActor_c13ClosestPlayerEv.cpp` really does dereference `this`: `Vec3_Dist((char*)this + 0x5c, (char*)p + 0x5c)`. So callers genuinely have to pass the actor, and the zero-argument declarations were wrong regardless of the bytes agreeing.

No symbol definitions are renamed here, so there is no matching `symbols.txt` edit owed.

## The problem: three files were retired on main under you

`git merge-tree` against current `main` reports three modify/delete conflicts. Each file was folded into a promoted class TU after this branch's merge base, and the state at the new home differs per case:

- **`src/func_ov084_0212f204.c` → `src/actors/daPkn_c.cpp`.** Still defective there: line 84 declares `char *_ZN8dActor_c13ClosestPlayerEv(void);` and line 735 calls it with no argument — the exact shape this branch fixes in the retired shard.
- **`src/func_ov002_020af218.c` → `src/actors/da1up_c.cpp`.** Still defective there: line 290 declares `extern int _ZN8dActor_c15IsPlayerInRangeEi(void*);` and line 292 calls it with one argument.
- **`src/func_ov065_021165d8.cpp` → `src/actors/daYurei_Mucho_c.cpp`.** Already correct there: line 40 declares `(void *self)` and line 465 passes `c`. That commit is now redundant.

Resolving these by keeping the branch's side would resurrect three retired shards alongside their promoted definitions and *still* leave two of the defects standing. On a restack, the first two need to move to `src/actors/daPkn_c.cpp` and `src/actors/da1up_c.cpp`, and the third can be dropped.

## Minor, not blocking

The `_ZN8dActor_c13ClosestPlayerEv` campaign leaves three zero-argument sites on this head — `src/func_ov102_02149078.c:2,14`, `src/_ZN8dActor_c13DistToCPlayerEv.cpp:17`, `src/_ZN8dActor_c14FarthestPlayerEv.cpp:16`. `check_call_arity.py` does surface the first of those in its own census output on this branch, so the leftover is visible rather than hidden; flagging it only so the scope line is explicit.

PR #2528, opened 2026-09-09
Model::LoadCompressedTextureToVram matches: the residue was a missing return value; fifteen floors re-measured at the pinned compiler

tangosdev:
ONE MATCH. Model::LoadCompressedTextureToVram (arm9 0x02045b58, 0xb8) comes off
the NONMATCHING list, and the MATCHED count goes 11,342 -> 11,343 of 11,390.

The five words that had been recorded as an allocator build delta from PR #957
were a missing return value. The function returns the pre-bump texel cursor -
the offset the texture just landed at - exactly as its sibling
LoadTextureToVram does. The draft was typed void, so that value was dead at the
exit and mwccarm reused its register in place (add r4,r4,r5) and batched both
adds before both stores. Keeping it live to the return forces the cartridge's
shape at +0x88: ldr r0,[r2] / ldr r3,[r1] / add r4,r0,r5 / str r4,[r2] /
add r2,r3,r5,lsr#1 / str r2,[r1], with the sum in a fresh r4 because r0 has to
survive to the exit.

Every caller in the tree already said so and only the definition disagreed.
include/Model.h has always declared it `static u32`; func_ov075_0211aa94 and
func_ov080_02125630 both store the result; and port/hal/gx_upload_bridge.cpp
re-reads data_020a4bc8 before the call to reconstruct by hand the value the
function was already returning. That bridge workaround is now redundant, but it
is in port/ and is still correct, so it is left alone.

The `volatile` on the two cursor bumps went with the banner. It was pinning
load-add-store twice where the cartridge emits load, load, add, str, add, str,
which is why the un-volatiled body measured 8 rather than 5: the low number was
a launder artefact on the wrong shape. The body is the real C++ member form now,
mirroring the matched LoadTextureToVram sibling, and include/Model.h stops
saying the definition is stuck at a floor.

HOW IT WAS FOUND, because the route matters for the next one. Dropping the
volatile first gave the right shape at div 8. One decomp-permuter run then
scored 80 from a base of 250 with `volatile short` as the return type - a
refused lying-permuter shape, since the ROM function is not a short and falling
off the end of it is undefined - but it isolated the lever: ANY non-void return
type drops 8 to 5, and returning the pre-bump cursor closes it. The permuter did
not find the match; it found which knob to turn.

The second commit is measurement only, no source statement touched, every
residue re-run per file to prove the bytes did not move:

Stage::InitResources (arm9 0x0202cc0c, div 4). The header called it four words
of register rename. The `and r2,r2,#0xff` at +0x580 is the draft's own
truncation - `bank` is u8 and `bank = level2 * 3;` stores an int into it - and
typing the index int removes it. What is left is the real floor and it is
sharper: with an int index every spelling compiles to 0xa80, one instruction
SHORT of the cartridge. The ROM spends a merge copy `mov r1,r2` at +0x584 to
keep the bank web in r2, skipping r1, which had just died on `cmp r1,#2` at
+0x520 - the notes 6bs "declines to reuse a dying register" signature - and
mwccarm coalesces bank straight into the argument register, so there is no copy
to colour. Eleven spellings measured, all 0xa80.

func_ov009_0211145c (ov009 0x0211145c). The live source is C with an embedded
asm{} block over ordinary ARM, so it is NONMATCHING by asm-policy rather than by
a compiler residue. The plain C that block spells is size-exact at div 5, all of
it in five instructions at +0xe0, and the comment now records that window, the
eight spellings byte-identical to it, and the permuter run that did not beat it.

func_ov007_020bfd70 (ov007 0x020bfd70). WITHDRAWN, superseded on main. This PR
had re-measured the then-current body at the pinned 2004/b56 and recorded div=14
against a header that said 13 without naming its build. Main has since re-spelled
the body -- the sin/cos table reads split into a named base pointer with the two
index loads separated by the abs(z) -- and restated the header as div=7 at
2004/b56. Measured at the pinned compiler on both sides before deciding
(tools/fdiff.py --module ov007 --addr 0x020bfd70 --size 0xdc --version 2004/b56):
main's body 7/55, this branch's 14/55. Main's is strictly closer to the cartridge
and its header names the right number for the body actually in the tree, so the
merge takes main's version whole and this PR no longer touches the file.

func_02059d8c (arm9 0x02059d8c). The note the header cites moved to
notes/archive/ in bc93fa767, so the path pointed at nothing.
check_dead_references confirms one more baselined C/C++ comment reference
resolves after the fix, 4 -> 5, and reports no new dead references.

Gates:
  tools/match.py --c src/_ZN5Model27LoadCompressedTextureToVramEPcjS0_.cpp
    --func _ZN5Model27LoadCompressedTextureToVramEPcjS0_ --addr 0x02045b58
    --size 0xb8 --module arm9 --version 2004/b56
    MATCHING VERSIONS: 2004/b56
  tools/prepush_linkcheck.py --range origin/main..HEAD
    [OK  ] _ZN5Model27LoadCompressedTextureToVramEPcjS0_ VERIFIED
    prepush-linkcheck: skipped 6 NONMATCHING draft(s)
    prepush-linkcheck: 1640 checked - 1519 verified, 121 warning(s), 0 blocking
  tools/progress.py --bar --from-src (re-run on the merge; main control alongside)
    merge  Functions 99.7%  11,353 / 11,390   Code size 2,197,936 / 2,238,108
    main   Functions 99.7%  11,352 / 11,390   Code size 2,197,752 / 2,238,108
    exactly +1 function and +184 bytes, 0xb8, the matched function's own size
  tools/rombuild.py -j 8 --no-rom (on the merge)
    source-built functions: 11,201   reproducing: 11,201   mismatching: 0
    ROM-build analysis: PASS
    same three lines on a main-only control worktree, so 11,201 is flat, not a drop;
    this function is not in the linked-range enrollment, which is why it moves
    progress.py and not the source-built count
  the tool-tests.yml module list, run whole: Ran 683 tests, OK (skipped=2)
  tools/check_python_names.py: PASS, 0 unresolvable names
  tools/check_dead_references.py: no new dead references, no broken markdown links


andrewboudreau (commented):
Reviewed at head `55d125396`. The match is real — I reproduced it rather than taking the summary's word for it — and the four measurement-only commits are comment-only by diff, so the "bytes did not move" claim is true by construction. One gap worth closing before this merges.

## Reproduced

`tools/match.py --c src/_ZN5Model27LoadCompressedTextureToVramEPcjS0_.cpp --func _ZN5Model27LoadCompressedTextureToVramEPcjS0_ --addr 0x02045b58 --size 0xb8 --module arm9 --version 2004/b56` → `MATCHING VERSIONS: 2004/b56`, and the window the description claims is the lever does come out as stated at +0x88: `ldr r0,[r2] / ldr r3,[r1] / add r4,r0,r5 / str r4,[r2] / add r2,r3,r5,lsr#1 / str r2,[r1]`.

Two call words (+0x68, +0x7c) and the four pool words compare as relocations, so the byte result alone does not prove the callees are right. It happens not to matter here: re-spelling the two `extern "C"` names as `GX::LoadTex` / `GX::EndLoadTex` static members mangles to the identical `_ZN2GX7LoadTexEPKvjj` / `_ZN2GX10EndLoadTexEv`, so no call is retargeted by this change, and the merge validation's per-file link check confirms the relocation destinations resolve. Nothing is renamed, so no `symbols.txt` edit is owed.

## The gap: nothing recompiles the new match

The merge validation on this head says it outright:

> Claimed, not byte-verified | 136 functions, 43,620 bytes (+1)
> Warnings: ... 1 more function(s) now claim a match that nothing compiles; enroll them in a delinks.txt to have the ROM build check them.

Byte-verified went `+0`. The function is matched today because `match.py` was run by hand; from the moment this merges, no gate recompiles it, and a later edit to `include/Model.h`, to the `GX` declarations, or to the data symbols it reads could break it with every check still green.

This one is cheap to close, because `config/arm9/delinks.txt` has a contiguity hole at exactly this range and both neighbours are already `complete`:

    src/func_02045ad8.c:
        complete
        .text start:0x02045ad8 end:0x02045b58
                                    <-- 0x02045b58..0x02045c10 unenrolled
    src/_ZN5Model17LoadTextureToVramEPcj.cpp:
        complete
        .text start:0x02045c10 end:0x02045c60

The matched sibling this function is modelled on is enrolled. Adding

    src/_ZN5Model27LoadCompressedTextureToVramEPcjS0_.cpp:
        complete
        .text start:0x02045b58 end:0x02045c10

turns the claim into something the ROM build re-proves on every future commit.

Everything else here is fine: the four re-measured headers touch no source statement, this branch merges cleanly with current `main`, and the `Source review` / `review` red on this head is the "PR head does not contain the live target base" expiry that is currently red across the open queue, not a finding about this change.

PR #2536, opened 2026-09-09
Cleanup ov047/daObjKm3_Kaitendai_c: Bowser in the Sky spinning disc

lunavyqo:
Readable follow-up on Bowser in the Sky's spinning disc, the last promoted kaitendai leaf after #2524 / #2526 / #2531. Bytes unchanged: `tubuild.py verify ov047/daObjKm3_Kaitendai_c` is 5/5 MATCH, objisolate clean, reloc destinations clean.

- Factory is `return new daObjKm3_Kaitendai_c()`. `return new` emits base D2 helpers and inherited RTTI copies; those are deadstripped. This TU already owns its own RTTI, file table, profile, and vtable in `.data`.
- InitResources picks a start angle from two statics on `param1` (low byte `== 1` is the alternate), then hands this overlay's file table (`data_ov047_02112334`) to daObjKaitendai_c's shared ov002 setup helper.
- CleanupResources is the teardown half over the same table.
- The file-table object stays defined *above* the class header so retail data order (descriptor, then type name, profile, vtable) is preserved.
- `g_profile_KM3_KAITENDAI` stays in this TU (it owns that data run).
- Comments describe the disc. Reconstruction/RTTI/reverse-order banners are gone.

- `func_ov002_020b676c` / `func_ov002_020b66a8` are still the shared kaitendai setup/teardown; names are placeholders.
- `data_ov047_02112334` is the 3-word file table; its three pointers and the two angle halfwords (`data_ov047_02112320` / `02112324`) are still address names.


lunavyqo:
Validation failed because `compiler_only_output` deadstrip-data removed the inherited RTTI copies this intact-object TU already licenses as `externalized_output`. Isolate then saw 0 ELF symbols for `_ZTI7fBase_c` and friends.

Addressed in 670eb0e98: those rows are gone. D2 helpers stay deadstripped. Inherited RTTI stays on `externalized_output`, same as main.

PR #2539, opened 2026-09-09
Cleanup ov045/daObjKm2_Ukishima_c: Bowser in the Fire Sea floating floor

lunavyqo:
Readable follow-up on Bowser in the Fire Sea's floating floor (the slab that sinks under you and bobs back up). Bytes unchanged: `tubuild.py verify ov045/daObjKm2_Ukishima_c` is 5/5 MATCH, objisolate clean, reloc destinations clean.

- Factory was already `return new daObjKm2_Ukishima_c()`. Class `operator new` is now in-class and drops the redundant `(unsigned)` cast.
- InitResources hands this overlay's file table (`data_ov045_02112f08`) to daObjUkiyuka_c's shared ov002 setup helper, and passes `BOB_AMPLITUDE` (`0xf50`) as `Fix12i`. That third argument is the bob amplitude stored at `mBobAmplitude` (0x324), not an actor id — the helper writes it there.
- `mBobAmplitude` on `daObjUkiyuka_c` is typed `Fix12i`. Behavior still MATCHES under the pinned compiler (typedef of `s32`).
- CleanupResources is the teardown half over the same table.
- `g_profile_KM2_UKISHIMA` stays in this file. This TU is text-only, so the emitted profile bytes remain deadstrip-data against the cartridge copy.
- Comments describe the floor. Reconstruction/RTTI/reverse-order banners are gone.

- `func_ov002_020b6584` / `func_ov002_020b6424` are still the shared floating-floor setup/teardown; names are placeholders.
- `data_ov045_02112f08` is an opaque 3-word file table. Nothing in this TU dereferences it.


lunavyqo:
`offsets` failed because `check_header_offsets` treats an in-class `enum { BOB_AMPLITUDE = 0xf50 }` as an unparsed field.

Addressed in d90f4741b: the constant lives in the TU, outside the class body. Init still passes it as `Fix12i`. Offsets check is clean; 5/5 MATCH.

PR #2551, opened 2026-09-10
Cleanup ov043/daObjKm1_Dorifu_c: Bowser in the Dark World drifting stairs

lunavyqo:
Bowser in the Dark World's drifting stairs (`dorifu`). Five planks, one solid at a time. The other leaves are `daObjKm3_Dorifu_c` (BitS) and `daObjRc_Dorifu_c` (Rainbow Ride).

No fields of its own (size 0xdcc is the base: `mPlankModels[5]` at 0x320, `mPlankClsn[5]` at 0x4b0). InitResources / CleanupResources hand this overlay's five-entry file table to `daObjDorifu_c`'s shared overloads. Factory is now `return new daObjKm1_Dorifu_c` (was a hand-rolled `nwEj` + C2 + two `__cxa_vec_ctor` + vptr punch).

Those overloads already load each row as `modelFile` / `clsnFile` / `clps` (SharedFilePtr BMD, SharedFilePtr KCL, CLPS_Block). ov043 sinit constructs those SharedFilePtrs as file IDs 1609 / 1610, 1611 / 1612, 1613 / 1614, 1615 / 1616, 1617 / 1618. This TU owns the table (retail places it after the profile, before the vtable).

5/5 MATCH, objisolate clean, reloc-destinations clean (2004/b56). `return new` emits base D2 helpers; compiler_only deadstrips `_ZN13daObjDorifu_cD2Ev` and `_ZN10dBgActor_cD2Ev`. Inherited RTTI stays externalized, not deadstrip-data.

Leftover: the BMD/KCL/CLPS symbols are still `data_ov043_*`.


andrewboudreau (commented):
## The blocking verdict is a compile-flag mismatch in the PR-scoped link check, not a byte defect

`PR validation` fails this PR on exactly one thing -- `Validation failed: 1 blocking relocation verdict(s)`:

| File | Symbol | Result | Slots |
|---|---|---|---|
| `src/game/actors/d_a_obj_km1_dorifu.cpp` | `daObjKm1_Dorifu_c_classInit` (grouped with D1/D0/CleanupResources/InitResources) | near-miss, does NOT reproduce the ROM | 5 |

The other four symbols in that group are VERIFIED. Only `daObjKm1_Dorifu_c_classInit` at ov043 `0x0211176c` is red. Everything else on the report is green: `Module fidelity 106/106 exact`, `ROM data reproduced from source | 716 symbol(s) exact`.

I reproduced the verdict locally and then contradicted it with the real build.

### What the check sees

```
python tools/fdiff.py --c src/game/actors/d_a_obj_km1_dorifu.cpp \
  --name daObjKm1_Dorifu_c_classInit --module ov043 --addr 0x0211176c --size 0x90 --align
```

```
   +0x40 | 3cc09fe5 ldr ip, [pc, #0x3c]      | 3c209fe5 ldr r2, [pc, #0x3c]      | <<< MISMATCH
   +0x44 | 3c309fe5 ldr r3, [pc, #0x3c]      | 3c309fe5 ldr r3, [pc, #0x3c]      | OK
   +0x48 | 4b0e84e2 add r0, r4, #0x4b0       | 00208de5 str r2, [sp]             | <<< MISMATCH
   +0x4c | 0510a0e3 mov r1, #5               | 4b0e84e2 add r0, r4, #0x4b0       | <<< MISMATCH
   +0x50 | 722fa0e3 mov r2, #0x1c8           | 0510a0e3 mov r1, #5               | <<< MISMATCH
   +0x54 | 00c08de5 str ip, [sp]             | 722fa0e3 mov r2, #0x1c8           | <<< MISMATCH
RESULT match=False mismatches=5/36 version=2004/b56
```

Five words, all inside the second `__cxa_vec_ctor` call (the `dBgW_KcMbg[5]` array at +0x4b0). Same instruction set, same call, same targets. The ROM parks the stacked 4th argument in `ip` and sinks its `str ..,[sp]` to immediately before the branch; the candidate parks it in `r2` and hoists the store above the `add`/`mov`/`mov`. Register allocation and scheduling only -- no wrong relocation target.

### Why it is not a defect

The divergence exists only under the flag set the PR-scoped link check compiles with.

`config/tu_manifest.d/ov043/daObjKm1_Dorifu_c.json` -> `verification.flags`, i.e. the flags this intact-object TU is actually built with:

```
-O4,p -enum int -lang c++ -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc -Cpp_exceptions off
```

`swarm.CPP_FLAGS`, which `tools/reloc_audit.py` and `tools/pr_linkcheck.py` use:

```
-O4,p -enum int -lang c++ -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc -w illpragmas
```

`-Cpp_exceptions off` is missing. `tools/rombuild.py`'s own `CFLAGS` carries it too, so both the manifest and the real ROM build compile this file with exceptions off and only the audit does not. Re-running the identical fdiff with nothing changed but that one flag added:

```
RESULT match=True mismatches=0/36 version=2004/b56
```

Removing `-w illpragmas` instead does not change anything, so `-Cpp_exceptions off` is the whole discriminator. `tools/reloc_audit.py` only ever attempts `match.DEFAULT_FLAGS` and `swarm.CPP_FLAGS`; neither carries it.

### The real build is green on this head

- `python tools/tubuild.py verify ov043/daObjKm1_Dorifu_c` -> `Result: 5/5 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED`. Byte-identical output on `origin/main`, so nothing regressed.
- `python tools/rombuild.py -j 16 --no-rom` -> `mismatching: 0`, `module fidelity: 106/106 exact, 100.000000% of compared bytes`, `ROM-build analysis: PASS`.

The shipped bytes are exact. The gate is scoring a build this file is never compiled with.

### One thing the PR did change

On `origin/main` this same file also verifies under `swarm.CPP_FLAGS`. The `return new daObjKm1_Dorifu_c();` factory plus the leaf `operator new` is what makes this function's codegen depend on `-Cpp_exceptions`. So the PR broke nothing, but it moved the file from flag-insensitive to flag-sensitive, and that is what exposed the gap.

### Two ways out

1. Harness side -- clears all three sibling PRs at once: when a changed file is owned by an intact-object manifest, have the PR-scoped link check compile it with that manifest's recorded `verification.flags` instead of the fixed `swarm.CPP_FLAGS`. At minimum, add a `-Cpp_exceptions off` attempt to the flag sets `reloc_audit.py` tries.
2. Author side: revert to the explicit-allocation factory, which is flag-insensitive. That gives up the readability this PR is buying.

The same flag gap bites in the opposite direction on #2642: there `objisolate` rewrites a vtable relocation addend under the audit's exceptions-on flags and does not under the build's, so the link check reports a verified vptr store that the real link gets wrong. One fix addresses both.


### Same verdict, same bytes, in #2552 and #2553

I ran the identical measurement on both siblings. The five mismatching words are byte-for-byte identical across all three, and the one flag fixes all three:

| PR | symbol | module / addr | `swarm.CPP_FLAGS` | + `-Cpp_exceptions off` |
|---|---|---|---|---|
| #2551 | `daObjKm1_Dorifu_c_classInit` | ov043 `0x0211176c` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |
| #2552 | `daObjKm3_Dorifu_c_classInit` | ov047 `0x0211164c` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |
| #2553 | `daObjRc_Dorifu_c_classInit` | ov036 `0x021120c8` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |

One harness change unblocks all three.

tangosdev:
PR validation is failing on a byte regression, not the shared header manifest issue. The relocation check finds that daObjKm1_Dorifu_c's D1, D0, CleanupResources, InitResources and classInit in this PR's own d_a_obj_km1_dorifu.cpp do not reproduce the ROM (NO-REPRO, 1 of 5 relocation slots). The fix is in this PR's own diff, not a tree-wide re-sync.

PR #2553, opened 2026-09-10
Cleanup ov036/daObjRc_Dorifu_c: Rainbow Ride Tricky Triangles

lunavyqo:
Rainbow Ride's Tricky Triangles (`dorifu`). Twin of the BitDW leaf (`daObjKm1_Dorifu_c`) and the BitS leaf (`daObjKm3_Dorifu_c`).

No fields of its own (size 0xdcc is the base: Model[5] at 0x320, dBgW_KcMbg[5] at 0x4b0). InitResources / CleanupResources hand this overlay's five-entry file table to `daObjDorifu_c`'s shared overloads. Factory is now `return new daObjRc_Dorifu_c` (was a hand-rolled `nwEj` + C2 + two `__cxa_vec_ctor` + vtable stores).

Those overloads already load each row as SharedFilePtr (BMD), SharedFilePtr (KCL), CLPS_Block. ov036 sinit constructs those SharedFilePtrs as file IDs 1690 / 1691, 1692 / 1693, 1694 / 1695, 1696 / 1697, 1698 / 1699. This TU owns the table (retail .data order is typeinfo, type-name, profile, table, vtable).

5/5 MATCH, objisolate clean, reloc-destinations clean (2004/b56). `return new` emits base D2 helpers; compiler_only deadstrips `_ZN13daObjDorifu_cD2Ev` and `_ZN10dBgActor_cD2Ev`. Inherited RTTI stays deadstrip-data.

Leftover: the BMD/KCL SharedFilePtrs and CLPS_Block are still `data_ov036_*`. The CLPS words sit before this TU's claimed .data; the SharedFilePtrs are BSS.


andrewboudreau (commented):
## The blocking verdict is a compile-flag mismatch in the PR-scoped link check, not a byte defect

`PR validation` fails this PR on exactly one thing -- `Validation failed: 1 blocking relocation verdict(s)`: `daObjRc_Dorifu_c_classInit` at ov036 `0x021120c8` comes back `near-miss (does NOT reproduce the ROM)`, 5 slots. The other four symbols in the group are VERIFIED and the rest of the report is green (`Module fidelity 106/106 exact`).

### What the check sees

```
python tools/fdiff.py --c src/game/actors/d_a_obj_rc_dorifu.cpp   --name daObjRc_Dorifu_c_classInit --module ov036 --addr 0x021120c8 --size 0x90 --align
```

```
   +0x40 | 3cc09fe5 ldr ip, [pc, #0x3c]      | 3c209fe5 ldr r2, [pc, #0x3c]      | <<< MISMATCH
   +0x48 | 4b0e84e2 add r0, r4, #0x4b0       | 00208de5 str r2, [sp]             | <<< MISMATCH
   +0x4c | 0510a0e3 mov r1, #5               | 4b0e84e2 add r0, r4, #0x4b0       | <<< MISMATCH
   +0x50 | 722fa0e3 mov r2, #0x1c8           | 0510a0e3 mov r1, #5               | <<< MISMATCH
   +0x54 | 00c08de5 str ip, [sp]             | 722fa0e3 mov r2, #0x1c8           | <<< MISMATCH
RESULT match=False mismatches=5/36 version=2004/b56
```

Five words, all inside the second `__cxa_vec_ctor` call (the `dBgW_KcMbg[5]` array at +0x4b0). Same call, same targets: the ROM parks the stacked 4th argument in `ip` and sinks its `str ..,[sp]` to just before the branch, the candidate uses `r2` and hoists the store. Register allocation and scheduling only.

### Why it is not a defect

The divergence exists only under the flags the PR-scoped link check compiles with.

`config/tu_manifest.d/ov036/daObjRc_Dorifu_c.json` -> `verification.flags` (what this intact-object TU is actually built with) ends in `-Cpp_exceptions off`. `swarm.CPP_FLAGS`, used by `tools/reloc_audit.py` and `tools/pr_linkcheck.py`, ends in `-w illpragmas` and has no `-Cpp_exceptions` setting at all. `tools/rombuild.py`'s own `CFLAGS` carries `-Cpp_exceptions off` as well, so both the manifest and the real ROM build compile this file with exceptions off and only the audit does not.

Re-running the identical fdiff with nothing changed but that one flag added:

```
RESULT match=True mismatches=0/36 version=2004/b56
```

Dropping `-w illpragmas` instead changes nothing, so `-Cpp_exceptions off` is the whole discriminator. `reloc_audit.py` only ever attempts `match.DEFAULT_FLAGS` and `swarm.CPP_FLAGS`; neither carries it.

### The real build is green on this head

- `python tools/tubuild.py verify ov036/daObjRc_Dorifu_c` -> `5/5 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED`
- `python tools/rombuild.py -j 16 --no-rom` on the sibling PR with the byte-identical change -> `mismatching: 0`, `module fidelity: 106/106 exact`, `ROM-build analysis: PASS`

### One thing the PR did change

On `origin/main` this file also verifies under `swarm.CPP_FLAGS`. The `return new` factory plus the leaf `operator new` is what makes this function's codegen depend on `-Cpp_exceptions`, and that is what exposed the gap. Nothing is broken; the file just moved from flag-insensitive to flag-sensitive.

### Shared with #2551 and #2552

The five mismatching words are byte-for-byte identical across all three Dorifu PRs, and the same flag fixes all three:

| PR | symbol | module / addr | `swarm.CPP_FLAGS` | + `-Cpp_exceptions off` |
|---|---|---|---|---|
| #2551 | `daObjKm1_Dorifu_c_classInit` | ov043 `0x0211176c` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |
| #2552 | `daObjKm3_Dorifu_c_classInit` | ov047 `0x0211164c` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |
| #2553 | `daObjRc_Dorifu_c_classInit` | ov036 `0x021120c8` | `match=False mismatches=5/36` | `match=True mismatches=0/36` |

Fix: when a changed file is owned by an intact-object manifest, have the PR-scoped link check use that manifest's recorded `verification.flags` rather than the fixed `swarm.CPP_FLAGS`. One harness change unblocks all three.

The same flag gap bites in the opposite direction on #2642: there `objisolate` rewrites a vtable relocation addend under the audit's exceptions-on flags and does not under the build's, so the link check reports a verified vptr store that the real link gets wrong. One fix addresses both.


tangosdev:
PR validation is failing on a byte regression in this PR's own file. The relocation check finds that daObjRc_Dorifu_c's D1, D0, CleanupResources, InitResources and classInit in d_a_obj_rc_dorifu.cpp do not reproduce the ROM (NO-REPRO, 1 of 5 relocation slots). This is not the shared header manifest issue. The fix is local to this diff.

PR #2557, opened 2026-09-10
Restore da1up_c in the promotion queue

andrewboudreau:
The promotion queue still listed the retired OneUpMushroom name after #2512 landed da1up_c, incorrectly reporting 36 files and no promotion. Restore the ROM class name and regenerate its row from the current source and manifest.

The same refresh updates three line counts changed by #2523. Independent verification regenerated RTTI, vtable and TU-map inputs, passed `queue_audit.py --check`, and confirmed that all 267 rows are retained. Only the queue TSV changes.


andrewboudreau (commented):
Independent review passed at dc2f6777b5365c4e80136d2162d261b205041b8e against d00b5e8497a9224b597ee59b12fa9c0c4759b6d8. Fresh RTTI/vtable/TU-map generation and queue_audit --check pass in a separate worktree. The four-row diff restores da1up_c and measures the three line-count changes from #2523; all 267 rows remain and no source, headers or compiler policy changed. Both required live checks are terminal success.

PR #2445, opened 2026-09-07
Reconstruct and promote daBgSnmBdy_c in ov072

andrewboudreau:
Reconstruct SnowmanBody as `daBgSnmBdy_c` and promote its 29 functions into one class translation unit. The cartridge type record at ov072 `0x02122780` points to `12daBgSnmBdy_c` at `0x0212278c` and identifies `dActor_c` as its base. Its 31-slot vtable starts at `0x021227c0`.

Lifecycle methods and the bounded state fixes use the class's existing model, collision and path interfaces. Factory/profile names and `SetState(int)` are explicitly identified as reconstruction choices. Every original symbol credit is preserved.

This is partial reconstruction: larger state bodies, raw ABI calls and the matrix adapter remain tracked in #2476, which stays open. Review findings SNM-01 through SNM-05 are fixed; SNM-06 records that remaining scope.

Validation for `7366ba11e951f91876dd2b741900b84f6a8063b9` against main `a0cc78fbd18f166ba67fca7c38b63bca348c8edd`:

- Independent source acceptance covers the exact 49-file change. It preserves the accepted source/provenance diff and incorporates the reviewed CI fixes.
- Fresh base and PR builds each reproduce all 106 modules and 11,199 source-built functions. The PR also passes all 26 owned-data and three BSS claims.
- Independent class proof: 29 functions, 3,892 text bytes, 188 owned data bytes, all 36 data relocations and 33 licensed outputs.
- Local committed-merge validation passes with no coverage or credit loss; all 423 port references resolve and all 173 translation units compile.

The unchanged full consumer proof covers 892 files and 2,069 function results: 2,064 VERIFIED, four inherited BLIND results and one draft. Its source/compiler inputs are identical to the previous accepted composition. The global metadata census retains four inherited differing records. These remain explicit limits.

The official private validator reached terminal SUCCESS at 22:49 UTC for this exact head/base. All public checks and the fresh live Source review gate passed; the PR was merged as `b343615d355bb76b645a7ba5c267ac6eb3d0cf28` at 22:52 UTC.

Details: [source handoff](notes/agents/handoffs/pr-2445-source-review-fixes.md), [composition handoff](notes/agents/handoffs/pr2445-composition-0910.md).

PR #2559, opened 2026-09-10
Preserve contributor checks across renamed and consolidated members

andrewboudreau:
The pre-push attribution checker misreports renamed or consolidated members because it follows old filenames and compares raw author aliases. In #2446 this produced one changed-credit and two lost-credit failures even though all 52 ROM function identities retained their canonical authors.

Resolve configured ownership by module, address and extent, and canonicalize aliases using each revision's own attribution rules. A consolidation still needs an explicit override for the live member. Ignore retired paths in surviving-source lineage and reject ambiguous positive-size identities, absent files and unsupported fallback mappings.

Independent validation: 183 tests pass; three independently written cases for stale overrides and ambiguous identity now reject; the exact #2446 history reports all 52 consolidations with zero changed or lost credits. Both review findings ATTR-01 and ATTR-02 are fixed in queue task `attribution-checker-repair-0910`, candidate `b6afe1550f3aba262b38882fd65ee7b70b3383e7`.

### DIFF
--- a/tools/prepush_attribution.py
+++ b/tools/prepush_attribution.py
@@ -113,6 +113,15 @@ def member_overrides_at(rev):
     return {symbol: member for symbol, member in out.items() if symbol not in ambiguous}
 
 
+def source_paths_at(rev):
+    """Only source files present in this revision can own contributor credit."""
+    paths = subprocess.run(
+        ["git", "ls-tree", "-r", "--name-only", rev, "--", "src/"],
+        cwd=REPO, capture_output=True, text=True, encoding="utf-8",
+        errors="replace", check=True).stdout.splitlines()
+    return {path for path in paths if path.endswith((".c", ".cpp"))}
+
+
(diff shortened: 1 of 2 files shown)


andrewboudreau (commented):
Independent review passed for `b6afe1550f3aba262b38882fd65ee7b70b3383e7` in queue task `attribution-checker-repair-0910`.

The reviewer reproduced and rejected the first candidate's stale-path and ambiguous-identity failures (ATTR-01 and ATTR-02), then verified both fixes in this candidate. All 183 tests pass. The same three independent negative fixtures now reject, while the exact #2446 base-to-candidate history preserves credits for all 52 functions. The new logic requires real source paths, unambiguous module/address/extent ownership, and explicit live-member credit; it does not add historical aliases to attribution data.

Only the checker and its tests change. Landing follows #2445 so that PR's accepted target base remains stable until it merges.

PR #2560, opened 2026-09-10
Review source against the live target tree

andrewboudreau:
Source review pins the live target branch but used GitHub's cached PR file list to decide review coverage. For #2445, those inputs described different bases: the cached list contained 401 files while the actual current-main comparison contained 49. A valid independent review was therefore rejected for unrelated files already on main.

Derive the required paths from the immutable base and head Git trees. Compare complete blob/type/mode identities, including deletions and both rename paths. Truncated recursive responses fall back to checked subtree traversal; incomplete or malformed responses fail closed. Require the existing up-to-date ancestry condition and retain the final head, queue and live-target checks. No PR source is downloaded or executed, and the capped compare-file list is not used for coverage.

Independent validation: 43 tests and 13 additional negative/fallback probes pass. A live read for #2445 produces exactly the same 49 paths as local Git and passes its existing independent review. Queue task `source-review-scope-repair-0910` accepted exact candidate `b0421f49da15f76b41fd3a4302a2115d38858b52`.

### DIFF
--- a/tools/check_pr_source_review.py
+++ b/tools/check_pr_source_review.py
@@ -87,6 +87,103 @@ def target_branch(repo, pr):
     return branch, sr.commit(ref["object"]["sha"], "target branch commit")
 
 
+def commit_tree(repo, commit):
+    """Resolve an immutable commit to its tree without downloading source blobs."""
+    data = api(f"repos/{repo}/git/commits/{commit}")
+    sr.require(data["sha"] == commit, "API returned a different commit")
+    return sr.commit(data["tree"]["sha"], "commit tree")
+
+
+def read_tree(repo, sha, recursive=False):
+    suffix = "?recursive=1" if recursive else ""
+    data = api(f"repos/{repo}/git/trees/{sha}{suffix}")
+    sr.require(data["sha"] == sha, "API returned a different tree")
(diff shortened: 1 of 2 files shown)


andrewboudreau (commented):
Independent review passed for candidate `b0421f49da15f76b41fd3a4302a2115d38858b52` in queue task `source-review-scope-repair-0910`.

The reviewer read both changed files, passed all 43 tests and 13 additional negative/fallback probes, and reproduced the live #2445 comparison: 49 exact-tree paths versus 401 cached PR paths. Coverage includes deletions and both rename paths; malformed/incomplete tree responses reject; the final head, live base and queue checks remain intact. No PR source blobs are fetched or executed.

The clean merge after #2559 changes only this checker and its tests; the implementation and dependencies reviewed here are unchanged.

andrewboudreau (commented):
Final integration review passed for updated head `6209ceae3d3a3072fa0e08b16d8b2e96dd0ffd97` against main `2e77201ec94b210a1bc0c5c27a098102ad71c6f4`.

Independent task `source-review-scope-restack-0910` confirms the complete 10,778-file tree equals current main with exactly the two already-reviewed scope-checker files replaced. Both accepted tool implementations are unchanged, all 84 source-review and attribution tests pass, and inherited findings remain fixed. The final live Source review and private PR validation checks pass.

PR #2561, opened 2026-09-10
Pin private PR validation to the live target branch

andrewboudreau:
PR #2445 was validated against GitHub's cached PR base (`690637e4`) after its source had already been composed with main (`666df563`). The private worker reproduced all 106 modules but rejected the unrelated coverage-denominator change introduced by that obsolete comparison.

Resolve the live target branch and current PR identity before submission, recheck them immediately before the relay request, and verify the returned job carries the same repository, PR, head and base. Add a PR-number-only manual retry for terminal jobs. Both jobs load the helper from the immutable workflow commit; the runner does not execute PR code, and only the private validator supplies the required verdict. Stale or incomplete path metadata cannot earn a documentation-only shortcut.

Validation: mocked API tests exercise the actual resolve-to-submit entrypoints, changed-head/base/target/fork races, malformed metadata, renamed and incomplete file lists, stale in-flight job pins and secret-safe transport failures. The new suite is included in the existing public tool-test command. An independent read-only run against #2445 confirmed the old cached base and the correctly selected live target without submitting a job.

### DIFF
--- a/.github/workflows/pr-validate.yml
+++ b/.github/workflows/pr-validate.yml
@@ -1,153 +1,72 @@
 name: PR validation
 
-# Submits the PR to the validation relay and exits. The verdict does NOT come back
-# through this workflow — the relay posts it directly to the pull request as the
-# Tangos Validator app: a "PR validation" check run, plus the usual comment. Watch
-# for those, not for this job.
-#
-# This job going green means "submitted", nothing more.
-#
-# Uses pull_request_target so repo secrets (the relay token) are available even
-# for fork PRs. SAFE because this workflow never checks out or runs PR code — it
-# only sends the PR number + immutable head/base SHAs to the relay. The actual validation
-# happens on the self-hosted worker behind the maintainer's NAT.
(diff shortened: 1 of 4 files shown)


andrewboudreau (commented):
Independently reviewed c7d1e078e30c62753ec47eef359c13456def54b0 against main cb5752cdf6b5d58f2b1f8a94bcbab2ed63749ac2. The exact four-file tree is accepted in queue task validation-base-pin-repair-0910; VAL-BASE-01 and VAL-BASE-02 are fixed.

All 75 targeted tests pass (32 submission-helper, 35 source-scope, 8 review-evidence), and both changed workflows pass actionlint 1.7.12. A separate live GitHub GET-only run for #2445 pins the real main commit instead of cached690637e4, forces full validation for that stale comparison, and makes zero relay calls. Reviewed trusted helper loading, manual-input limits, identity/base races, in-flight job readback and secret-safe transport. Private/public terminal checks still govern merge.

PR #2582, opened 2026-09-11
Cleanup ov018/daObjIceBoard_c: Cool Cool Mountain ice sheet

lunavyqo:
Readable follow-up on Cool Cool Mountain's shatterable ice sheet (profile ICE_BOARD, actor 295). Bytes unchanged: `tubuild.py verify ov018/daObjIceBoard_c` is 10/10 MATCH, objisolate clean, reloc destinations clean (2004/b56). `prepush_linkcheck.py --range origin/main..HEAD` is 10/10 VERIFIED. `port_refcheck.py` is clean. Header fans out only to this TU.

- Factory is `return new daObjIceBoard_c()`. In-class leaf `operator new` takes `unsigned long` and forwards to `_ZN7fBase_cnwEj`.
- Nested `extern "C"` factory and the `decl_ActorBase.h` / `decl_Platform.h` / `decl_common.h` includes are gone.
- Destructor is last and inline so `new` emits the retail D1/D0 pair with no leaf D2. The homeless `_ZN10dBgActor_cD2Ev` helper is exact-deadstripped.
- This overlay's mother penguin is untouched.

Leftover:
- `Particle::System::NewSimple` stays the scalar mangled ABI. Declaring true `Fix12<int>` parameters on this TU's Kill does not compile: `mPosX/Y/Z` are `s32` and mwccarm refuses the implicit conversion.
- `dBgW_KcMbg::SetFile` stays mangled. A real member call on this TU's InitResources is 9/10 MATCH -- InitResources DIFFs (Fix12-by-value homing).
- `data_ov002_0210d754` is the ov002 CLPS block; `symbols.txt` has no recovered name, so it is not coined.
- Inline destructor (out-of-line emits D0 before D1).


lunavyqo:
MATCH 10/10, 26ffe7d0c. Dropped ROM-ordinal banners.

PR #2608, opened 2026-09-11
Cleanup ov072/daBgSnwmn_c: snowman

lunavyqo:
Leftover cleanup of promoted `ov072/daBgSnwmn_c`.

**Identity.** RTTI ov072:0x02122938 names `daBgSnwmn_c`; overlay_actors names profile `BIG_SNOWMAN` (272). ov072 is mixed (BABY_PENGUIN / BIG_SNOWMAN / SNOWMAN_HEAD / SNOWMAN_BODY) — this is the snowman manager/head assembly, not `daBgSnmBdy_c` (body) and not the penguin.

**Factory.** Already `return new daBgSnwmn_c()`. Leaf `operator new(unsigned long)` forwards `_ZN7fBase_cnwEj((unsigned)size)`. Attribution override keeps github-actions[bot] on the factory symbol.

**Named fields.** Init/Behavior/Render/Cleanup and `func_ov072_021208d8` now use `mModel1`, `mModel2`, `mTexSeq`, `mShadow`, `mCylClsn`, inherited `mPos*`/`mAngleY`/`mScaleX`, and `Model::mat4x3.t`. Real methods for LoadFile/SetFile/Release/Prepare/Update/Render/InitCylinder/Spawn/MarkForDestruction.

**Gate.** `python tools/tubuild.py verify ov072/daBgSnwmn_c` → **9/9 MATCH, Result TEXT-VERIFIED** (byte + objisolate + reloc-destinations). `prepush_linkcheck.py --range origin/main..HEAD` → 9/9 VERIFIED. `port_refcheck` clean. Header fans out only to this TU.

**Leftovers** (cited from this TU, still MATCH):
- TextureSequence::SetFile / dCcAcPos_c::Init / DropShadowRadHeight 6az (InitResources, func_ov072_021208d8)
- SharedFilePtr +4 BMD/BTP (Prepare/SetFile; header has no fields)
- `data_ov072_02122c40` / `02122c48` / `02122c50` / `02122c70` handles; overlay .bss owns them
- InitResources ground probe stays a POD Position (local Vector3 emits unlicensed `_ZN7Vector3D1Ev`)

PR #2601, opened 2026-09-11
Cleanup ov070/daBrq_c: Amp

lunavyqo:
Leftover cleanup of promoted `ov070/daBrq_c`.

**Identity.** RTTI ov070:0x0212323c names `daBrq_c`; `g_profile_BIRIKYU` at 0x02123254 stores 0x010a = AMP(266). ov070 is FLY_GUY / AMP / FLAME_CHOMP / FLAME_CHOMP_FIRE — this is Amp, not `daKrpa` (KERONPA / Flame Chomp) and not `daKpFr` (KERONPA_FIRE). Historical aliases Amp_Spawn / Amp_SpawnInfo.

**Factory.** Folded `daBrq_c_classInit` (0x021210ac..0x02121118, immediately after InitResources, ends at daKrpa D1) as `return new daBrq_c()`. Leaf `operator new(unsigned long)` forwards `_ZN7fBase_cnwEj((unsigned)size)`. Attribution override keeps andrewboudreau on the factory symbol.

**Named fields.** Class members were already named. Stale `func_ov070` ordinal comments and the `mMat4x3` witness note now use `UpdateModelTransform` / `DropShadowRadHeight`.

**Gate.** `python tools/tubuild.py verify ov070/daBrq_c` → **19/19 MATCH, Result TEXT-VERIFIED** (byte + objisolate + reloc-destinations). `prepush_linkcheck.py --files src/game/actors/daBrq_c.cpp` → **19/19 VERIFIED**. `port_refcheck.py` clean.

**Leftovers** (cited from this TU, still MATCH):
- SetRanges / dCcAcPos_c::Init / dBgCh_Actr::Init / ModelAnim::SetAnim / TextureSequence::SetFile / TextureTransformer::SetFile / DropShadowRadHeight / Particle::System::NewSimple stay mangled (Fix12 by value, wall 6az). dBgCh_Actr::Init header Fix12i mangles as i.
- UpdateDefeatedState calls `dBgCh_Actr_UpdateDiscreteNoLava_veneer` (0x02038420), not the method body at 0x02037024.
- SharedFilePtr +4 BMD/BTP (Prepare/SetFile; header has no fields).
- `data_ov070_*` resource handles and PMF state table; sinit stays enrolled separately.
- `common.h` first (BrqMatrixWords twelve-word Matrix4x3 identity copy).
- Vector3 overlays on mCamSpacePosX / mScaleX / mPosX.
[...]

PR #2592, opened 2026-09-11
Cleanup ov023/daObjFm_Battan_c: Squasher

lunavyqo:
Leftover cleanup of promoted `ov023/daObjFm_Battan_c`.

**Identity.** RTTI ov023:0x02111fa0 (`_ZTI16daObjFm_Battan_c`) / 0x02111fac (`_ZTS16daObjFm_Battan_c`) names `daObjFm_Battan_c`. `symbols/overlay_actors.md` maps ov023 to SQUASHER(84). Shifting Sand Land's squasher, not a BoB/WF/BitDW actor. Historical aliases: Squasher_Spawn / Squasher_SpawnInfo.

**Factory.** Folded `daObjFm_Battan_c_classInit` (0x02111728, size 0x38) as `return new daObjFm_Battan_c()`. Leaf `operator new(unsigned long)` forwards `_ZN7fBase_cnwEj((unsigned)size)` until #2570. Attribution override keeps github-actions[bot] on the factory symbol. `return new` emits homeless `_ZN10dBgActor_cD2Ev`; compiler-only policy deadstrips it.

**Named fields.** Init/Behavior/Render/Cleanup, UpdateShadow, and UpdateModelTransform already use `mAngVelX`, `mStateTimer`, `mState`, `mShadowModel`, `mShadowMat`, `mModel`, `mMeshCollider`, `mClsnMat`, `mPos*`, `mAngle*`, `mCamSpacePosX`. `pad_323` dropped; the 1-byte alignment after `mState` is implicit.

**Gate.** `python tools/tubuild.py verify ov023/daObjFm_Battan_c` → **9/9 MATCH, Result TEXT-VERIFIED** (byte + objisolate + reloc-destinations). `prepush_linkcheck` 9/9 VERIFIED (sinit BLIND-8 warn).

**Leftovers** (cited from this TU, still MATCH):
- `dBgW_KcMbg::SetFile` / `dActor_c::DropShadowScaleXYZ` / `dActor_c::Earthquake` 6az (InitResources, UpdateShadow, Behavior)
- `func_020393d4` stores `dBgW::UpdatePosWithTransform` (no setter)
- `func_0200fa04` squash-impact dust
- `AddVec3` / `Matrix4x3_FromRotationY` / `Matrix4x3_FromRotationXYZExt`
- `BattanVector3` (a local `Vector3` would emit vague-linkage `~Vector3`)
- `data_ov023_02112088` / `02112080` handles and `data_ov064_0211ba4c` CLPS
[...]

PR #2642, opened 2026-09-11
Cleanup ov100/daObjPathLift_c: path lift

lunavyqo:
Leftover cleanup of promoted `ov100/daObjPathLift_c`.

**Identity.** RTTI ov100:0x02148538 names `daObjPathLift_c` (`_ZTS` at 0x02148544, `_ZTV` at 0x0214857c). overlay_actors maps ov100 profile 31 to PATH_LIFT. ov100 is mixed (PATH_LIFT / ROLLING_IRON_BALL / BUTTERFLY / UNCHAINED_CHOMP / FISH / DOOR / STAR_DOOR) — this is the path lift, not `daStarGate_c`. Historical project alias: PathLift.

**Factory.** Typed C-ABI with named `mModels` / `mPath` / `mShadowModel`. Leaf `operator new(unsigned long)` forwards `_ZN7fBase_cnwEj((unsigned)size)`. `return new daObjPathLift_c()` cannot live in this TU: Vector3's vague-linkage D1 lands between InitResources and classInit and shifts the factory.

**Named fields.** Init/Behavior/Render/Cleanup and the two ov100 helpers now use `mModel` / `mMeshCollider` / `mClsnMat` / `mShadowModel` / `mGroundY` / `mTimer` / `mSoundHandle` and inherited `mPos*` / `mAngle*` / `mWaitTimer` / `mPathSpeed`. Live `pad_478` is `unk_478`. Real `dBgCh_Gnd` and `SharedFilePtr::Release`.

**Gate.** `python tools/tubuild.py verify ov100/daObjPathLift_c` → **8/8 MATCH, Result TEXT-VERIFIED** (byte + objisolate + reloc-destinations). `prepush_linkcheck.py --range origin/main..HEAD` → 9/9 VERIFIED including the factory. `port_refcheck.py`: 423 checked, 0 stale. Header consumer is only this TU.

**Leftovers** (cited from this TU, still MATCH):
- `dBgW_KcMbg::SetFile` stays mangled: InitResources' by-value `Fix12<int>` is wall 6az
- `dBgActor_c::IsClsnInRange` stays mangled: Behavior's two by-value `Fix12<int>` is wall 6az
- `dActor_c::DropShadowScaleXYZ` stays mangled: `func_ov100_02146e70`'s three by-value `Fix12<int>` is wall 6az
- `func_ov002_020efaf0` / `020efc74` / `020efcf4` keep those ROM labels (dPathLiftActor_c helpers)
- `func_ov100_02146e70` / `0214700c` keep those ROM labels
- `data_ov100_02148a54` / `02148a5c` / `data_ov002_0210d9f0` / `0210d7d4` stay `data_ov`: this TU claims `.text` only
- `func_020393d4` / `020393a4` / `02039394` stay those names (dBgW field setters)
- `Sound::PlayLong` stays a local declaration (not on Sound.h)
- matrix translation stores stay `this+0x114` / `+0x49c`: nested-member stores grow the helpers
- `unk_478` stays raw storage: a `Matrix4x3` member (Vector3 translation) is non-trivial

DIFFs kept: those three mangled call sites and the two helper translation stores.


lunavyqo:
`python tools/tubuild.py verify ov100/daObjPathLift_c` → 8/8 MATCH (`5f4adf277`). Dropped ROM-ordinal banners.

andrewboudreau (commented):
## Real defect: `&_ZTV15daObjPathLift_c[2]` links 8 bytes past the ROM's vptr word -- and the link check misses it because it compiles with different flags than the build

`PR validation`'s full-ROM phase fails with `dsd check modules --fail did not pass`. `python tools/rombuild.py -j 16 --no-rom` on this head, reproduced twice:

```
  ov100    d_a_obj_path_lift    0x02147328 size 0x7c    1
      modules NOT byte-exact: ['ov100']
      [INFO ] Check overlay 100: checksum failed
module fidelity: 105/106 exact, 99.999967% of compared bytes
ROM-build analysis: FAIL
```

One differing word in `daObjPathLift_c_classInit`.

### The line

```cpp
*(int *)actor = (int)&_ZTV15daObjPathLift_c[2];
```

replacing the base commit's

```cpp
*(void**)p = &data_ov100_0214857c;
```

`config/arm9/overlays/ov100/symbols.txt` lines 224-225:

```
data_ov100_0214857c kind:data(any) addr:0x0214857c
_ZTV15daObjPathLift_c kind:data(any) addr:0x0214857c
```

`[2]` is the right spelling when mwcc *defines* the vtable in the same object -- mwcc's `_ZTV<C>` addresses the vtable object, so the store has to skip the offset-to-top and typeinfo words. This file does not define it; it declares the ROM symbol (`extern int _ZTV15daObjPathLift_c[];`), and in `symbols.txt` `_ZTV<C>` already *is* the slot array. So `[2]` adds a preamble that was never there. `objisolate`'s own comment states this case exactly: a source that declares the ROM symbol itself carries addend 0 and there is nothing to correct.

The literal in the cartridge, at `0x02147328 + 0x78 = 0x021473a0`, is `0x0214857c`. `[2]` links `0x0214857c + 8 = 0x02148584`.

Back-to-back on this head, same worktree, warm cache:

| source | rombuild |
|---|---|
| `[2]` as committed | exit 1, `mismatching: 1`, `105/106`, `ROM-build analysis: FAIL` |
| `[2]` -> `[0]`, nothing else | exit 0, `mismatching: 0`, `106/106 exact, 100.000000%`, `ROM-build analysis: PASS` |

Every other byte of the function already matches: comparing the compiled object's `daObjPathLift_c_classInit` text against the cartridge word by word, the only differences are in relocated words, and the only relocated word that resolves wrong is `_ZTV15daObjPathLift_c`.

### Why the link check called it verified

The same report says:

| `src/d_a_obj_path_lift.cpp` | `daObjPathLift_c_classInit` | verified | 1 |

`reloc_audit._as_the_build_links_it` runs the candidate object through `objisolate` before resolving destinations, because the ROM build does too. But the two compile the file with different flags:

- the ROM build uses `-lang c++ ... -Cpp_exceptions off`
- `reloc_audit` / `pr_linkcheck` use `swarm.CPP_FLAGS`, which has no `-Cpp_exceptions` setting, so exceptions are on

Measured on this exact file, same compiler, only that flag differing:

```
exceptions ON : plan = keep 7, drop [.exceptix, .rela.exceptix], dead ['@EX@daObjPathLift_c_classInit']
                -> objisolate rewrites the object, and corrects the _ZTV addend 8 -> 0
exceptions OFF: plan = keep 5, drop [], externalise [], dead []
                -> objisolate has nothing to do and leaves the addend at 8
```

So the audit scores an object whose vtable addend has been corrected, resolves `0x0214857c + 0` = the configured destination, and reports green; the build links `0x0214857c + 8` and fails. The gate and the build disagree because they are not compiling the same thing.

Two separate things here:

1. In this PR: `[2]` -> `[0]`. That is the whole defect and it makes the build green.
2. Worth a separate look: `reloc_audit`/`pr_linkcheck` should compile with the flags the ROM build actually uses. As it stands, a vptr store against an already-address-point `_ZTV` symbol gets a false green from the link check and is only caught by the full ROM build. (The same flag gap is what makes #2551/#2552/#2553 fail for no reason, in the other direction.)

### On "head full-ROM report omitted the ROM-data measurement"

Not a separate problem and not a reporting bug: `tools/rombuild.py` only sets `report["romData"]` at phase `[6/6]`, and this build fails at the module check before that, so the field is absent. The three clauses in the summary line are one fact restated three ways -- and here the underlying full-ROM failure is genuine.

tangosdev:
PR validation is failing tree-wide, not in this PR's own diff. daObjPathLift_c's own two changed files verify all 9 relocation slots, matching your MATCH 8/8 comment above, but the full-tree check fails at "dsd check modules --fail did not pass" with no module named. This PR shares its base commit with the battan, daBrq_c and daBgSnwmn_c failures on #2592, #2601 and #2608, which are the shared header positional-symbol desync from Andrew's rule from 09-13. Worth a tree-wide config/tu_manifest.d re-sync (python tools/rombuild.py --no-rom) and a re-run before assuming this PR's own diff is at fault.

PR #2644, opened 2026-09-13
Judge Source review currency where the written policy places it

andrewboudreau:
## The measured problem

All 117 open PRs fail the `Source review` check. I read the latest check run on
every open head: **117 of 117 are `failure`**, 116 with

```
Source review could not be established: PR head does not contain the live target base;
restack and review the composition
```

and one with its race variant, `Target branch changed during review`.

The mechanism is in the check itself, not in any branch.
`check_pr_source_review.target_branch()` resolves `refs/heads/main` **live**, and
`changed_paths()` then required `comparison["merge_base_commit"] == base` with
`status in ("ahead", "identical")` — that is, it required every PR head to
*contain main's tip at check time*.

Main's tip advances on a bot timer. Of main's last 40 commits, **14 are
`Refresh progress (README, treemap, contributions, provenance, langmode) [skip ci]`
from `github-actions[bot]`**, single-parent, author and committer both the bot,
one to four files each.
Their entire file lists are `contributions.json`, `docs/index.html`,
`docs/progress-treemap.svg` and `README.md`; the three most recent touch
`contributions.json` alone. Nothing in `source_path()`. No branch can outrun a
scheduler, so the requirement was unsatisfiable by construction.

### PR #2564: the same commit, opposite verdicts

Head `6478ee3e7`, a tools-only PR, has 23 `Source review` check runs. The first two
passed:

| completed | conclusion | title |
|---|---|---|
| 2026-09-11T01:23:51Z | success | No reconstruction source, header or TU manifest changed. |
| 2026-09-11T01:26:17Z | success | No reconstruction source, header or TU manifest changed. |
| 2026-09-11T01:35:06Z | failure | Source review could not be established: PR head does not contain the live target base… |
| …21 more | failure | (identical) |

Same head, same tree, same content. Main's timer fired between 01:26 and 01:35 and
the verdict inverted. Note the ordering that made this possible: the currency check
ran *before* the "nothing to review" determination, so a PR with zero reviewable
content failed on staleness alone.

## The policy this preserves

`notes/agents/SOURCE-REVIEW-CUTOVER.md` line 74:

> a populated independently published review must pass for the exact head/base;
> **a changed head/base must lose that acceptance.**

`notes/agents/queue-v2.md` line 230:

> The composition must preserve the accepted source and declared base as ancestors.

And the cutover assigns up-to-dateness at merge time to the repository ruleset,
not to this check (line 71):

> Require branches to be up to date before merging, so a check for an old base
> cannot authorize a newer composition.

[...]

PR #2648, opened 2026-09-13
Title: Skip the progress-refresh bot's commits when resolving the source review live target

tangosdev:
Files touched (tools only): tools/check_pr_source_review.py, tools/test_check_pr_source_review.py

Context (Andrew's own session, tonight, in #claude-handoff):
"Since 2560 the check requires the PR head to CONTAIN main live tip, and the
progress bot pushes Refresh progress [skip ci] on a timer. Main moved
9c7f38a24 to 49e250fa4 while I worked tonight. Unsatisfiable in steady state,
so any review expires before merge. Fix the target first (exclude bot
[skip ci] commits from the live target), then I review and publish as I
land."

What the bot commit looks like (measured from this repo's own history, not
assumed): every one is authored and committed as
  github-actions[bot] <github-actions[bot]@users.noreply.github.com>
with the exact message
  "Refresh progress (README, treemap, contributions, provenance, langmode) [skip ci]"
and touches only generated files (README.md, docs/index.html,
docs/progress-treemap.svg, contributions.json, config/match_provenance.jsonl,
config/match_attempts.jsonl, nearmiss/db.jsonl -- see
.github/workflows/update-chaos-data.yml's last step). `git log --format='%h
%an %s' origin/main -20` shows it alternating almost 1:1 with human merges;
`git show --stat` on two live examples (49e250fa4, 9c7f38a24) confirms each
touches only contributions.json in this snapshot.

The rule: tools/check_pr_source_review.py's target_branch() now walks back
from the resolved branch ref tip over a bounded run (MAX_REFRESH_SKIP = 20) of
commits that match BOTH the exact github-actions[bot] author identity (name
and noreply email) AND the exact refresh message together -- never "[skip
ci]" alone, since a human commit can carry that marker too -- to the newest
commit that could actually change reviewable scope. That resolved commit,
plus every bot commit walked over to reach it, becomes the live target's
alias set; evaluate() now accepts a stored review whose tested_base is any
member of that set, not only the exact resolved sha, so a review recorded
while the branch tip was itself a trailing refresh commit is not penalized
either. The walk is logged in the check's own stderr output as
"live target: <sha> (skipped N refresh commits)". Every other rule is
untouched: the exact-committed-tree scope derivation, the queue lookup, and
the missing_files report all work exactly as before, just against the
corrected base.

Tests added to tools/test_check_pr_source_review.py (RefreshCommitSkipTest):
a refresh tip resolves to its non-refresh parent; two consecutive refresh
commits are both skipped; a human commit carrying "[skip ci]" in a different
message is not skipped; a refresh-shaped message from a different author is
not skipped; a run of MAX_REFRESH_SKIP+ refresh commits raises rather than
walking past silently; and a stored review's tested_base pinned to a skipped
bot commit still passes once the live target's alias set is supplied (the
"handle both sides consistently" half of the fix). Two existing tests'
call-order assertions were updated to reflect the one new git/commits/ read
target_branch now performs (the walk's own commit lookup) -- everything else
in the existing suite (35 tests) passes unchanged; two shared test fixtures
gained a default git/commits/ responder (an ordinary non-bot commit) so they
keep exercising exactly what they did before.

Why this is safe: it only ever shortens the "how far back is the live
target" answer over commits that are provably bot-authored and content-empty
for review purposes (generated files only); it never widens what counts as
reviewed source, never changes the exact-tree scope derivation, and fails
closed (raises) if the bot-commit run exceeds the bound instead of silently
skipping an unbounded amount of history.

This changes Andrew's own gate (tools/check_pr_source_review.py), so per the
source-review-gate standing rule the coordinator posts this text to him
before merging; it is not self-merged.


andrewboudreau:
## Measured: the bot is not what makes the target unreachable — the alternation is

Three open PRs fix the same bug in the same two files (#2644, #2648, #2653). I
measured all three rules against the live `main` (`abeff1815`) and the 124 open PR
heads, so the choice does not have to be argued:

| rule | anchor it lands on | open heads that contain it |
|---|---|---|
| today (raw tip) | `abeff1815` | **0 / 124** |
| bot-identity + inert (#2644, #2648) | `c4eeb6004` (walks 1) | **1 / 124** |
| content-only inert (#2653) | `e14ab3a3a` = #2568 (walks 7) | **81 / 124** |

The identity rule stops at the first commit not authored by `github-actions[bot]`.
On this tree that is one commit back, every time, because main alternates bot
refresh with human merge almost 1:1 — and **the human merges are just as inert**:

```
abeff1815  github-actions[bot]  contributions.json
c4eeb6004  Tango                nearmiss/db.jsonl, tools/nearmiss_db.py, tools/srcpath.py, ...
04fb369ab  github-actions[bot]  contributions.json
aeff38653  Tango                config/dead-reference-baseline.json
49e250fa4  github-actions[bot]  contributions.json
e3707be70  Alberto Pilotto      config/dead-reference-baseline.json, notes/*.md
9c7f38a24  github-actions[bot]  contributions.json
e14ab3a3a  Andrew               src/... , include/...     <- first reviewable commit
```

Four of the seven commits the walk has to cross are human. Nothing about the
scheduler is the obstacle; a docs, notes, tools or nearmiss merge retires the
target exactly as hard as a bot push does, and there are as many of those.

## Why keying on content instead of identity is not a weakening

The safety argument never needed the author. If a commit changes nothing matching
`source_review.source_path()`, then a review taken against its parent read
byte-identical reviewable content to one taken against it, so honouring the
parent's sha loses no coverage. That holds whoever pushed it. Keying on identity
does not add safety — it only declines to skip inert commits from humans, which is
where 80 of the 81 PRs are lost.

Scope is unaffected for the same reason: the walk stops at the first commit that
touches a reviewable path, so the changed-path set computed against the anchor and
against the tip are identical over `source_path()`.

Every doubt still fails closed in #2653 — an unreadable commit, a merge commit, a
sha/identity mismatch, a file list at the API's 300-file cap, a missing or
malformed file entry, or an unusable parent all stop the walk and keep the tip.
There are seven tests for exactly those, plus a 25-commit walk bound.

tangosdev:
Closing in favour of #2653: the content-only rule reaches 81 of 124 open heads against this one's 1, by Andrew's own measurement on the PR. Nothing here that #2653 does not cover.

PR #2671, opened 2026-09-13
CI runs 20 of this suite's 53 tests, and the rename filter does not do what its docstring says

andrewboudreau:
`tools/test_check_header_offsets.py` defines 53 tests. **CI runs 20 of them**, and has since the file was written. Separately, `changed_paths` documents that it treats a renamed header as an add, and does not — so a pull request that only *moves* headers resolves to an empty work list and takes the honest-empty exit.

Both are fixed here. The third change — replacing four pinned commit ids with hermetic tmpdir fixtures — is what surfaced the second one, and is the piece that keeps this file off the repin treadmill.

Main is green on a runner today; #2667 saw to that. This is not a rescue.

---

## 1. CI collects 20 of 53 tests

`header-offsets.yml:74` runs this file as a **script**:

```yaml
run: python tools/test_check_header_offsets.py
```

`if __name__ == "__main__": unittest.main()` sat at line 355 of an 897-line file. `unittest.main()` calls `sys.exit()`, so the interpreter never reaches the remaining 542 lines and the five classes defined there — root-vptr, nested-tag, qualified-size, Allman-body, allocation-method — **do not exist** when the runner collects. On `origin/main` @ `1224afd11`:

```
$ python tools/test_check_header_offsets.py      # verbatim, what CI runs
Ran 20 tests in 4.852s
OK

$ python -m pytest tools/test_check_header_offsets.py
53 passed
```

Thirty-three tests print nothing and are counted by nobody. Not hypothetical: **#2667's own verification reports `20/20 pass`** — the defect reading out as a green in the maintainer's own hand.

Guard moved to the end of the file: **56 collected either way**, script and pytest alike.

### The difference is load-bearing

Same deliberate break in both columns — `_root_offset` stubbed to `return 0`, i.e. the gate stops counting the implicit vptr and reports every field of every polymorphic root four bytes early:

| | `python tools/test_check_header_offsets.py` |
|---|---|
| `origin/main` | `Ran 20 tests … OK` — **green over a live break** |
| this branch | `Ran 56 tests … FAILED (failures=8)` |

For completeness, main's own file under pytest catches it (`8 failed, 45 passed`). The tests were always there. CI just never saw them.

## 2. `--no-renames` — the documented behaviour was not the actual behaviour

`changed_paths`' docstring said:

> Renames arrive as an add because `-M` is deliberately not passed — a header that moved still has to have its offsets agree.

`diff.renames` has defaulted to **true** since git 2.9. Git detected the move anyway, reported it as `R`, and `--diff-filter=AM` dropped it. A pull request that did nothing but move headers therefore resolved to an **empty work list** — and because the diff itself was not empty, `_resolve_changed` took the one honest-empty exit and passed it. That is exactly the collapse this gate's whole empty-list argument exists to keep apart from a real pass.

`--no-renames` is what actually buys the documented behaviour: the move comes back as a delete of the old path (filtered out) and an add of the new one (checked).

This is a behaviour change to the gate, not to its tests. It was found by writing the delete-vs-rename fixture, and shipping a test that blesses the hole seemed worse than fixing it. Happy to split it out if you would rather review it separately.

## 3. Hermetic fixtures instead of four pinned ids

The section that pinned `#1659`, `#1665`, `#1666` and `#1667` and diffed `<sha>^...<sha>` against this repository is now built by real `git` in a tmpdir.

**This finishes #2667's job rather than reverting it.** `tools/repin_commit_ids.py`'s docstring singles this file out as the one consumer that is not merely stale prose:

> …one gate is not just stale prose: `tools/test_check_header_offsets.py`'s `RealHistoryTests` run `git diff <sha>^...<sha>` against real history, so a pinned old id that no longer resolves fails CI outright.

And in the *same commit* that re-pinned these four (`dc3dbbdfa`), a real corpus id in `tools/test_repin_commit_ids.py` was replaced with a synthetic `a1a1a1a1…`:

> One of the two tool test fixtures happened to reuse a real corpus commit id verbatim; replaced it with a synthetic one so the test stays hermetic and idempotent under a future re-sweep.

This applies that treatment to the last file that still needs it.

### Exposure, not frequency

An earlier draft of this PR said the pinning bet had been lost "a second time", and a comment on the branch said the ids were "orphaned when main was force-reset". **Both were wrong. `8dac39881` corrects the comment** (and `cb94396d8` corrects an undercount in that correction), and the record is:

- All four ids were reachable from pre-scrub main (`6acaba7b9`, 06:45 UTC on 2026-09-13). They stopped resolving at 07:50 EDT the same day, when an identity scrub rewrote the whole history — identical trees, a new id for every commit.
- `git reflog show origin/main` in this clone covers **872 fetches over 31 days and records exactly one `forced-update`** — that scrub. The 2026-08-23 force-reset self-healed in about fifteen minutes and never touched these ids.

So the argument is not that rewrites are frequent. It is that **the material that caused this one is still in the tree**, and that the tooling was already built for the repeat:

- The scrub's trigger was a personal Windows username in a hardcoded path — `tools/ovsweep.py:48`, which reads `C:/Users/tango/...` today.
- `origin/main` @ `1224afd11` still spells **31 paths of exactly that shape, across 6 tracked files** — 16 under one username, 14 under another, 1 the scrubbed `tango`:

  ```sh
  $ git grep -ohE 'C:[\\/]{1,2}Users[\\/]{1,2}(andre|alexs|tango)' origin/main -- . | wc -l
  31
  ```

  | count | file |
  |---:|---|
  | 25 | `config/match_attempts.jsonl` |
  | 2 | `notes/real-cpp-migration-runbook.md` |
  | 1 | `notes/agents/LAUNCH.md` |
  | 1 | `notes/agents/references/pipeline-v1.md` |
  | 1 | `notes/tu-cpp-census-2026-08.md` |
  | 1 | `tools/ovsweep.py` |

  **25 of the 31 are in `config/match_attempts.jsonl`**, as recorded `"srcPath"` values (`"srcPath":"C:\\Users\\alexs\\AppData\\Local\\Temp\\claude\\…"`). That is a generated log that keeps growing, so the exposure is neither a handful of stale notes nor static. Mind the separator if you re-check: JSON escapes the path, so those 25 read `C:\\Users\\…` and a pattern matching a *single* separator misses every one of them.
- `repin_commit_ids.py --check` is documented as "the same predicate a pre-merge gate would use", and `dc3dbbdfa` says "a future re-sweep" — a tool written by someone expecting one.
- The treadmill already needed a second pass **98 minutes** after the first: `b9ba3a59b`, "Repin the commit ids that arrived with the rebase onto main."

Losing that bet turns a **header** gate red for reasons that have nothing to do with headers, and a gate that lands red for unrelated reasons gets switched off. There is no reason for this file to be on the list at all.

### Coverage is a strict superset

`PullRequestShapeTests` asserts everything the two `RealHistoryTests` asserted, through the **identical** call:

```python
C.changed_paths(f"{sha}^", committed_only=True, repo=..., head=sha)
```

- `len(buckets["headers"]) == want` — 33 / 15 / 58 / 0
- `buckets["dropped"] == []`
- `buckets["total"]` non-empty
- `buckets["include"] == []` and `len(buckets["total"]) == 31` for the honest-empty case

plus three assertions the pinned version could not make:

- the exact path **set** each pull request resolves to, not just its size — all four are built into one repository, in order, so a resolution that reached past the parent picks up a neighbour's files;
- that `head=` ignores a dirty working tree (the pinned tests relied on this silently, running against a checkout that is routinely dirty, and asserted it nowhere);
- that a deletion and a rename land on the right sides of `--diff-filter=AM` — which is how defect 2 was found.

**Nothing a replica misses.** Neither version reads header *content*: `changed_paths` resolves a work **list**, and the walk that opens the files is what every other class in this file covers. There is no repository configuration the tmpdir lacks — no gitlinks, and `.gitattributes` touches nothing under `include/` or `*.h`.

**The counts are not invented.** Each was measured with `git diff --name-status <sha>^...<sha>` while the objects were still readable, and the fixture reproduces the per-directory, per-extension file counts, the added/modified split, and the topology main really has: one squashed, single-parent commit per pull request. (None of the four was a merge commit; the comment above them used to claim they were.)

## Proof it can go red

Re-measured on this merge, each break applied to `tools/check_header_offsets.py` alone:

| deliberate break | result |
|---|---|
| drop `--no-renames` from the commits diff | `Ran 56 … FAILED (failures=1)` — the rename test |
| ignore `head=`, always diff the working tree's HEAD | `Ran 56 … FAILED (failures=4)` — three subtests plus the exact-path-set test |
| `heads = set(inc)` (stop filtering `include/` by extension) | `Ran 56 … FAILED (failures=2)` |
| `_root_offset` stubbed to ignore the implicit vptr | `Ran 56 … FAILED (failures=8)` |

## Proof it needs no repository history

`git archive HEAD tools .github include config | tar -x` into a fresh `git init`, then the workflow step verbatim:

```
commits in this tree: 0
--- python tools/test_check_header_offsets.py ---
Ran 56 tests in 12.210s
OK
```

The same tree with main's two files instead: `Ran 20 tests … FAILED (failures=5)`.

## Gates, on the merge with `origin/main` @ `1224afd11`

- `python -m pytest tools/test_check_header_offsets.py` — **56 collected, 56 passed** (53 on main)
- `python tools/test_check_header_offsets.py` — **Ran 56 tests, OK** (20 on main)
- `python tools/check_src_tu_compiles.py` — **177/177**
- `python tools/check_python_names.py` — PASS
- `python tools/premerge_check.py HEAD --tmp /c/pmc` — **nothing goes green → red**; 8/8 static gates pass on both base and merge

No ROM was built; this touches `tools/` only, no `src/`.

## Conflict resolution

`4608aec17` merges `origin/main`. One conflict, `tools/test_check_header_offsets.py`, resolved in favour of this branch: main's `dc3dbbdfa` re-pinned the four `REAL_HISTORY` ids to their post-scrub equivalents, and this branch deletes `REAL_HISTORY` outright, so there is no pin left to keep current. The net diff against main is still exactly the two files.


andrewboudreau:
### State of the two red checks

**`review` / `Source review` -- staleness, not content.** The JSON says `Source review could not be established: PR head does not contain the live target base; restack and review the composition`. The head was cut from `origin/main` at `1b448ec66`; the live window is `[38e56365, 8580d4c2]`.

I restacked onto `38e56365` locally -- clean, no conflicts, `Ran 56 tests ... OK`, `premerge_check` still `nothing goes green -> red` -- but **the push is blocked by main's own pre-push gate**:

```
  FAIL  ov062/KoopaTheQuick   2004/b56    src_tu/actors/KoopaTheQuick.cpp:121: illegal function overloading
  FAIL  ov077/Lakitu          2004/b56    src_tu/actors/Lakitu.cpp:259:      illegal function overloading
check_src_tu_compiles: 175/177 translation unit(s) compiled
```

That breakage is not this PR's. `git diff origin/main HEAD --name-only` is two files, both `tools/*.py`; `src_tu/`, `src/`, `include/` and `config/` are byte-identical to live main, so the run above *is* a run on main's own content. It also landed after this branch was cut -- the first push, based on `1b448ec66`, passed the same hook.

So: the restack is ready and will go up as soon as those two TUs are fixed on main. Restacking is a treadmill regardless while main moves every few minutes, so it is probably worth doing immediately before merge rather than now.

Nothing else is red: `build`, `conflicts`, `names`, `offsets`, `references`, `refs`, `submit`, `tools` and `resolve` all pass, and the `offsets` runner log shows `Ran 56 tests in 2.989s / OK` -- which is the whole point of the change, and was `FAILED (failures=5)` out of 20 before it.

PR #2679, opened 2026-09-13
Make two src_tu return types agree with decl_common, unblocking every push

andrewboudreau:
## What this is

`check_src_tu_compiles.py` has been failing **175/177 on main**. It runs in the pre-push hook, so it has been failing the push of *every* branch in the repo — this is the first push in a while that did not need `--no-verify`.

Both failures are the same defect: a local return type that disagrees with `include/decl_common.h`, in both cases sitting under a comment asserting that it agrees.

| file | local spelling | `decl_common.h` |
|---|---|---|
| `src_tu/actors/KoopaTheQuick.cpp:121` | `void func_0201277c(int)` | `unsigned int func_0201277c(int)` (line 1983) |
| `src_tu/actors/Lakitu.cpp:259` | `extern "C" int func_ov077_02124718(void*)` | `void func_ov077_02124718(void*)` (line 1542) |

Return type alone cannot overload in C++, so mwccarm rejected both with `illegal function overloading`. The second error, `ambiguous access to overloaded function 'func_0201277c(int)'` at `KoopaTheQuick.cpp:711`, is downstream of the first.

`decl_common`'s side of both has been unchanged since 2026-07-31, so these are latent mismatches the gate surfaced, not a regression in the gate.

**Corroboration on the second one:** the adjacent twin dispatcher on the line above, `func_ov077_02124754`, is already spelled `void`. `02124718` was the outlier, not `decl_common`.

Both comments are corrected rather than left asserting something false.

## Measured on this branch

```
check_src_tu_compiles.py    177/177, exit 0      (was 175/177)

tubuild.py verify ov062/KoopaTheQuick
    18/18 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tubuild.py verify ov077/Lakitu
    32/32 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
```

Re-measured after a rebase onto `4810b445c`, not carried over from the earlier base.

No bytes move — this only restores compilation. Note that because *neither TU compiled on main*, neither had been byte-verified there either: the 18/18 and 32/32 above are the first verification these two have had.

Both TUs still report `PROMOTION REFUSED` on unlicensed symbols and compiler-only policy. That is pre-existing and out of scope here; `TEXT-VERIFIED` is the line this PR is evidence for.

## Scope

Two source lines. The manifest writes `tubuild.py verify` performs as a side effect (a `compilerOnlyOutput` block in `Lakitu.json`, a `"data": []` key in `KoopaTheQuick.json`) are deliberately excluded — worth landing, but not inside a push-unblocker.

## TL;DR

Two functions were declared with one return type in a source file and a different one in the shared header. C++ won't let a function have two return types, so the compiler refused, so the pre-push check failed, so nobody could push anything. Changed two words to match the header. Everything compiles and both files still produce byte-identical code to the original game.


andrewboudreau:
Closing as a duplicate — this is my error, not a problem with the change.

PR #2672 carries the identical two-line fix and the user merged it as `1224afd11` while this branch was being verified. The code is byte-for-byte the same on both sides; the only remaining difference is comment wording on `Lakitu.cpp:259`:

```
main (#2672)  /* void per decl_common; the ROM body returns no value */
this branch   /* void per decl_common:1542; no value returned, matching the ROM */
```

Not worth a PR. `check_src_tu_compiles.py` is 177/177 on main via #2672, so the push blocker is cleared either way.

I should have checked for an open PR against these two files before branching. The memory note on this failure mode already named #2672 as the fix in flight; I read the shape of the defect out of it and missed the pointer to the PR.

PR #2670, opened 2026-09-13
Cleanup leftover overlay TUs

lunavyqo:
Restack of leftover overlay translation units from the closed #2576–#2642 queue onto live `main`.

One commit per leftover TU (67 commits). Each commit is that actor's leftover cleanup: named fields, `return new Class()` factories with a leaf `operator new(unsigned long)` until #2570, folded one-function factory shards, and the round-3 banner / comment fixes for that TU.

This PR does not touch `fBase_c.h` or `dActor_c.h`. Size asserts keep the #2663 `SM64DS_PLATFORM_PC` host-compat wrap. Overlay `delinks.txt` and the owning `tu_manifest` are updated only where a leftover TU absorbed a factory file.

Out of this PR: #2570 (fBase `operator new`), #2513 (`Pos()` pun), the dorifu NO-REPROs, #2574/#2575.

Local `tubuild.py verify` on these leftover TUs is TEXT-VERIFIED (byte + reloc MATCH). Clock and London were re-verified after rebasing onto the humanizer merges that landed on `main` while this branch was stacking.

PR #2685, opened 2026-09-13
Cleanup leftover overlay TUs (batch 2)

lunavyqo:
Leftover overlay TUs that were closed unmerged and held out of #2670. Stacked on #2684 (`fBase_c::operator new(unsigned long)`). Does not touch #2670.

One commit per leftover TU: `return new`, speakable names, host sizeof wrap. Dropped penguin/lava leftover that rewrote other TUs' manifests and `dActor` HorzAngle files. King Bob-omb leftover of `daBmb_c.h` stays; `daBmb_c.cpp` does not — that body is #2670.

Includes: Ki_Ita, star-create, WDW boards, lava, waterfall, spinning discs, falling blocks, floating floors, tilting slabs, rickshaw cart/axle, dorifu, Bubba, WF sliding bars, Chain Chomp, mother penguin, Ukiki, MIPS, King Bob-omb, Pokey.

Not in this PR: ov006 minigames (#2500/#2501), Goomboss/Eyerok/Scuttlebug/`daGmch_c`.


andrewboudreau (changes requested):
Initial humanizer finding on `50a64449f70d69cae78096eeaf46c93048c71da8`; the 95-file candidate has not received complete independent source acceptance.

**BATCH2-2685-01 (provenance), `src/game/actors/d_a_pg_mthr.cpp:11-13`:** the statement that these helpers must stay free because their "ROM symbols" are `func_ov018_*` mistakes repository labels for preserved original names. The image does not contain an original mangled symbol table; see `notes/tu-promotion-conventions.md`, section 1. A migration updates source/config names together. An existing address-derived label is therefore not evidence that a helper was originally free, nor a compiler barrier to implementing a real member. Correct this rationale and record the actual ownership evidence and remaining migration scope. The exact historical function spellings remain unknown.

**BATCH2-2685-02 (review scope):** this batch spans many classes, shared headers and manifest changes. Preserve the individual classes' earlier findings and record an issue, next owner and explicit partial disposition for each retained raw-field/interface dependency. Supply pinned experiments for claimed compiler barriers. Banking declaration differences and adding `Leftover:` text cannot replace this review decision. The review needs to cover the complete final diff, including any still-visible stacked dependency.

Do not land from the existing byte/build checks alone. The private validator was `ACTION_REQUIRED` in the initial inventory; current-main composition, complete independent source review and terminal private validation remain required. No producer branch or queue claim was changed by this review.

andrewboudreau:
Heads-up on the rebase onto `491c2eea7`: it dropped two things from my earlier merge `a0beaca3a`. One matters, one is cosmetic. **The important value survived** — `config/tu_manifest.d/ov030/daMky_c.json` is on `$731` at the current head (3 occurrences, no `$730` left), which is what `tools/tu_manifest_resync.py` measures from the merged tree. Nothing to redo there.

### What came back: four contradicted declarations

`src/func_ov006_020c0a48.c`, `020c1d80.c`, `020c221c.c` and `020c33dc.c` are again spelling the BlendModelAnim constructor as

```c
extern int _ZN14BlendModelAnimC1Ev();
```

and then calling it with an argument. This is not a plurality judgement call — the tree already has a header for it:

```
include/decl_BlendModelAnim.h:23:  extern void _ZN14BlendModelAnimC1Ev(void*);
```

and every other consumer agrees with the header: `daBombking_c.cpp:2041`, `d_a_iwante.cpp:5`, `d_a_king_donketu.cpp:7`, `d_a_moray.c:6`, all `void ...(void *)`. `include/BlendModelAnim.h:52` records that it is defined out of line in `src/_ZN14BlendModelAnimC1Ev.cpp`. These four files are the only dissent, against a real header plus five agreeing TUs, and the `int` return is invented — the constructor returns nothing.

The correction is one line per file:

```diff
-extern int _ZN14BlendModelAnimC1Ev();
+extern void _ZN14BlendModelAnimC1Ev(void *);
```

It was byte-neutral when I measured it in `a0beaca3a` (106/106 exact, `mismatching: 0`, `ROM-build analysis: PASS`). I have not re-pushed it — the branch is yours and it is outside both open findings.

The cosmetic loss is a paragraph I had appended to the `daMky_c.json` row's `reason`, recording this merge as a worked example of the uniquifier arithmetic: merge-base `$724`, main `$730`, branch `$731`, merged tree emits `$731` — so `--ours` and `--theirs` were both wrong and only measuring the merged result gets it right. Worth having written down somewhere, but it does not need to be in this PR.

BATCH2-2685-01 looks answered at `019a7cc06`. BATCH2-2685-02 is a scope objection, which is yours to resolve, not mine.

andrewboudreau:
## The one thing that blocks this PR is invisible from inside it

I merged all twelve open PRs onto main `491c2eea7` in one tree and built it. Everything is byte-clean except a single hard collision between this PR and #2707, and neither PR can see it alone because each touches only its own half.

#2707 names two fields on `daObjFloatBoard_c`:

```c
s32 mRestPosX;  /* 0x320 */
s32 mRestPosZ;  /* 0x328 */
```

This PR names the same two offsets `unk_320` / `unk_328` and rewrites `src/game/actors/d_a_obj_wc_obj06.cpp` (62 insertions, 126 deletions) to use that spelling. Merge both and the compiler stops:

```
d_a_obj_wc_obj06.cpp:85: undefined identifier 'unk_320'
d_a_obj_wc_obj06.cpp:87: undefined identifier 'unk_328'
```

### Which spelling should win

#2707's. `mRestPosX` / `mRestPosZ` say what the fields are — `InitResources` raycasts down onto the water surface and then parks the board's resting X/Z there — and `unk_320` says only where they are. The naming convention in this tree is that a name replaces `unk_` once the meaning is known, and #2707 knows it.

That also makes this the cheaper side to move: #2707 would have to give up two real names and every reference to them, whereas this PR only has to rename three uses in one leaf file.

### The patch

Applied on top of this branch, in `src/game/actors/d_a_obj_wc_obj06.cpp`:

```diff
@@ -6,5 +6,5 @@
  * if the board spawned above the water it raycasts down onto the surface
  * (or the first hit under it) and writes that Y into mPosY. It then
- * copies mPosX/Y/Z onto unk_320, mWaterY, unk_328.
+ * copies mPosX/Y/Z onto mRestPosX, mWaterY, mRestPosZ.
  *
@@ -83,7 +83,7 @@ int daObjWcObj06_c::InitResources()
             }
         }
-        unk_320 = mPosX;
+        mRestPosX = mPosX;
         mWaterY = mPosY;
-        unk_328 = mPosZ;
+        mRestPosZ = mPosZ;
         return 1;
     }
```

Three lines plus the comment. Nothing else in this PR refers to either field.

### Measured, not assumed

With that patch and one declaration fix on #2701, the full twelve-PR tree builds cold:

```
python tools/rombuild.py -j 16 --no-rom --no-cache
[3/6] mwccarm: 8730 enrolled source file(s), -j16     <- no cache, every file recompiled
source-built functions: 11,207
  reproducing: 11,207
  mismatching: 0
ROM-build analysis: PASS
```

`--no-cache` is deliberate: a warm run reports `0 compiled`, which proves nothing.

### One correction I owe you

I earlier flagged seven declaration contradictions as appearing only in combination, and named this PR among the owners. **That was wrong, and it was my error, not yours.** I had resolved `config/decl-agreement-baseline.json` by unioning the branches instead of regenerating it. Where main banks one kind-list for a `(symbol, file)` key and a PR banks a different one for the same key, a union has to discard one side, and discarding the PR's side manufactures a contradiction that does not exist.

Regenerating the baseline on the merged tree instead, then checking every added entry against the baseline of the PR that owns it:

```
merge tree vs main: +142 / -236 banked entries  (net -94)
UNEXPLAINED: 0
```

Every new banked disagreement is one its own PR already banks, and the batch as a whole retires 94 more than it adds. There are no declaration blockers in this batch. The field-name collision above is the only real one.

lunavyqo:
Moved onto #2707's names in 906ed8415 (header fields + 3 uses in wc_obj06). Header rename kept minimal so the merge takes #2707's side. 4/4 TU functions re-verified, no new decl disagreements.

lunavyqo:
Restacked onto live `main` (`162d1796f`) and landed the four BlendModelAnim C1 declarations as `extern void _ZN14BlendModelAnimC1Ev(void *);` in `func_ov006_020c0a48.c` / `020c1d80.c` / `020c221c.c` / `020c33dc.c`. Local prepush: 2 VERIFIED, 2 BLIND, 0 blocking.

#2707 `mRestPosX` / `mRestPosZ` still in. Leftover `return new` unchanged. Uniquifiers still in sync (`daMky` `$731`, `daDsnBase` `@class$430`, StarGate PTMFs).

andrewboudreau (dismissed):
**BATCH2-2685-01: resolved.** Verified on the current head.

`src/game/actors/d_a_pg_mthr.cpp` no longer claims the `func_ov018_*` labels are
preserved ROM symbols. It now states the image has no original mangled symbol
table, that the `func_ov` labels are address-derived repository names, and that
the exact historical spellings remain unknown — which is the correction I asked
for and matches `notes/tu-promotion-conventions.md` section 1.

BATCH2-2685-02 was a review-scope request rather than a defect, and the per-class
dispositions remain tracked follow-ups rather than blockers on this batch.

This PR is currently **DIRTY** and needs a rebase before it can merge. Clearing
my block so review state is not the thing holding it.

PR #2688, opened 2026-09-14
Match func_ov063_02117cdc (ClosestPlayer takes the actor in r0)

andrewboudreau:
Closes `func_ov063_02117cdc` (ov063, 1,916 bytes): divergence **3 → 0**.

## The three words

A pure register rename in the entry block. The cartridge loads the switch selector to r2 (`ldrb r2, [r6, #0x5ce]` / `cmp r2, #0xa` / `addls pc, pc, r2, lsl #2`) and the literal-pool address to r1. Every prior draft took r0 or r1. All 476 other words were already exact.

## The mechanism

`_ZN8dActor_c13ClosestPlayerEv` is `dActor_c::ClosestPlayer()` — a **non-static method**, so the ROM reaches it with the calling actor in r0. Case 0 calls it as its first statement. The banked draft declared it `(void)` and called it with nothing.

**The `bl` is byte-identical either way**, so no byte gate could see the error. But passing the actor makes the parameter's incoming r0 live from entry to that call; mwccarm coalesces the two — no `mov r0, r6` is emitted, in the cartridge or in the candidate — and the entry block's two temporaries therefore start at r1 instead of r0.

The two levers are not independent, which is why sweeping either alone found a floor:

| call spelling | pool read | entry block | div |
|---|---|---|---|
| `ClosestPlayer()` | laundered | `ldrb r0 / ldr r1` | 3 (the banked tip) |
| `ClosestPlayer()` | plain | `ldrb r1 / ldr r0` | 5 |
| `ClosestPlayer(c)` | laundered | `ldrb r1 / ldr r2` | 5 |
| `ClosestPlayer(c)` | plain | `ldrb r2 / ldr r1` | **MATCH** |

The prior note claimed the cartridge's `{r1,r2}` was "not reachable from a third spelling". It is — just not from any *entry-block* spelling, which is all the earlier sweeps covered.

## Verification

- `RESULT match=True mismatches=0/479 version=2004/b56` (no `--limit`)
- `linkcheck` → `"verdict": "VERIFIED"`, `diffs: []`, `blind: 0`
- `prepush_linkcheck --range origin/main..HEAD` → 1 checked, 1 verified, 0 blocking
- `rombuild.py -j16 --no-rom` → 11,202 source-built reproducing / 0 mismatching; module fidelity 106/106 exact; **ROM-build analysis: PASS**
- `premerge_check` → 8 static gates, base pass → merge pass; source coverage 8791 → 8792 (+1)

## Supporting measurements

Synthetic probes isolated three conditions for a call site to reserve r0 across a switch dispatch: the argument must be the parameter **value** (a cast or a copy through a local still works; `p + 0x10`, `*(T *)(p + off)`, and the parameter in r1 do not); it must be in the r0 position; and r0 must still hold it at the call — one other call ahead of it in the same case breaks the coalesce.

A non-void return type with a bare `return;` is byte-neutral (`return 0;` is what grows the function) and reaches the same entry block, but reserves r0 for the *whole* function and moves 174 words — same mechanism at function scope, which is why it is not the lever.

Ledger rows and the pruned near-miss tip were written by `tools/stamp_provenance.py`. The `ov063/symbols.txt` row already existed — nothing was added there, and nothing consults `actor_renames.tsv`.


andrewboudreau:
Closing as a duplicate. Tango landed this same match independently in `a81e7e9dd`, with the same mechanism and earlier — `dActor_c::ClosestPlayer()` takes `this` in r0, the parameter stays live into case 0, and both entry-block temps skip r0. Recorded in notes as 6cx.

Our lane reached it separately, so the only thing worth carrying over is the coupling table, in case it is useful to anyone pricing a similar row: the parameter fix and the pool-read laundering only pay together — `(void)`+laundered = 3, `(void)`+plain = 5, `(c)`+laundered = 5, `(c)`+plain = match. Sweeping either lever alone finds a floor and reports it honestly, which is why the row sat at 3 across two prior campaigns.

No action needed here; main already has the match.

PR #2692, opened 2026-09-14
Bank the ov006 0x020e20bc near-miss at 203 (exact size, CFG matches, 61-word exact prefix)

andrewboudreau:
Banks the `func_ov006_020e20bc` near-miss draft at **203 divergent words of 376**, down from 299 at hand-off. Not a match. **One line of `nearmiss/db.jsonl` is the entire diff** — no source, no `symbols.txt` row, no `actor_renames.tsv` dependency.

The value here is that the row is no longer a structural problem. It is one allocator disagreement worth exactly one live value, and the next lane has a sharp diagnostic for it.

## Where it stands

| | |
|---|---|
| divergence | 299 → **203** |
| candidate size | **376 words — exact** |
| shape ratio (`--align --align-shape`) | **0.9255** (348 equal) |
| byte-exact prefix | **first 61 words** |
| branches | **11 vs 11**, same conditions, same relative targets |
| mnemonic histogram | **38 of 41 identical** |

All measured with `tools/fdiff.py` and **no `--limit`** (that flag truncates the arrays it scores while printing a full denominator, so its verdicts are not trustworthy on this checkout). The final figure was re-measured from the source as stored in the bank, not from a scratch file.

The candidate reproduces the recovered CFG branch for branch — eleven branches at word indices 17, 19, 22, 24, 38, 293, 306, 321, 364, 368, 371 (the ROM's first wall clamp is at 292, everything else identical), with matching relative targets `+0x5b0 / +0x4bc / +0x4f4 / +0x534 / +0x34`. Every basic-block boundary but one sits at the same byte offset.

## The entire residue

1. **One surplus spill.** The draft keeps one extra value live in the fixed-point chain and spills it into frame slot `sp+0x48`, **which the ROM leaves unused**, shifting every later spill slot up a word. Against that the ROM rematerialises `add r3,r6,r5` + `add r2,r3,#0x4000` inside the first wall clamp and stores `0xc000` through the split, where the draft folds that store into the already-spilled `pYi`. Net zero words, which is why the size is exactly right.
2. **Register naming inside an otherwise correct instruction sequence.** First divergence is word 61: ROM `add sb, r6, r1`, draft `add fp, r6, r1` — the ROM coalesces the shared `c + 0x4686` base into the dying `j` register, the draft takes `fp`. Everything after is that permutation propagating.

**The diagnostic is cheap and sharp: if `sp+0x48` is in use at all, the draft is still one live value over.**

## Levers that paid

- **Guard evaluation order sets the literal-pool order.** Computing `dy` (0x4660) before `dx` (0x4664) puts 0x4660 into the pool first and re-colours r4/r7 through the whole prologue. Worth ~41 words; reversing it costs 10 exact prefix words and +10 divergence.
- **Association order controls CSE-web length.** Offset-first `c + 0x4688 + m` on the end-of-body byte stores stops them sharing `c+m` with earlier reads, so mwcc rematerialises instead of spilling — as the ROM does.
- Index-first raw clamp store lands the size at exactly 376.

## Refuted — do not re-buy these

- **Combine-statement order is byte-inert.** The ROM's sub/add polarity sequence `S A S S A A S A` decodes to `px, py, rx, nx1, {ry, ny1}, nx2, ny2`; both legal orders compile to *identical bytes* across all 12 decl×prefix pairings, and both score 209–225 — worse than 203. mwcc fully reschedules that region.
- **Declaration order does not drive the spill pool.** Rewriting the seven spilled locals into the ROM's own ascending slot order (`nx2, pAi, pSi, pAj, pYj, pYi, pXj`) is **byte-identical** to the current order; front/back/descending variants are 1–11 worse.
- Declaring the shared bases (`c+0x4686`, `c+0x4668`) as source locals: +3 words, ratio 0.9255 → 0.8503.
- Dropping the redundant `o` local: byte-identical (mwcc folds the copy).
- `sqrt` operand order: i-side swap inert, j-side +5, guard swap catastrophic (−37 prefix words).
- Tail statement pair swaps: the `*pYi`/`*pXi` `0x1b000` pair, `y`/`x`, `t`/`u` all inert; `st_i`/`st_j` +7; the `*pAi`/`*pSi` pair cannot swap at all (size breaks).
- **Exhaustive single-move sweeps found nothing.** All 28 statements × all valid positions: **zero** improvements on either metric. Declaration sweep: zero. Fourteen hill-climbers across four templates, ~1000–2000 iterations each, on both the raw byte metric and a longest-exact-prefix metric, could not beat 203/prefix-61. This is a firm local optimum for this source family.

## Callee arity: audited by ROM-wide call-site census, clean

Done the reliable way rather than by reading callee bodies — a callee can overwrite r0 before any read and still take a parameter. Each binary was scanned at its own load address so the ov006/ov009 range overlap cannot misattribute sites.

| callee | sites | evidence | conclusion |
|---|---|---|---|
| `func_02012718` | 80 (all ov006) | r0 80/80, r1 80/80, r2 34, r3 21 | 2 args |
| `cstd::sqrt` | 35 | r0 35/35, r1 35/35 | one 64-bit arg |
| `cstd::atan2` | 116 | r1 115/116, r0 111/116 | 2 args |

All three draft declarations already agree. Each callee's **own** source file was checked too: `src/func_02012718.c` is matched and declares `void func_02012718(void *a, int b)`, so there is no `(void)`-style defect here. `tools/check_call_arity.py` flags 0 rows. Decisive local corroboration: the argument set-up before **five of the seven `bl`s is byte-identical to the ROM**, including a complete `cstd::sqrt` call inside the byte-exact prefix.

## What the next lane should do

Not statement or declaration shuffling — there are exhaustive negative results for both.

1. **Kill the surplus spill.** Find the source construct keeping one extra value alive across the `nx2` computation around candidate words 220–240 and re-spell *that*, not its neighbours. Watch `sp+0x48`.
2. **Then take the faithful clamp back.** With the surplus gone, the ROM-faithful `((Ball *)(c + i*0x2c + 0x4000))->y = 0xc000;` spelling should land at 376 with the `add` count corrected, making the mnemonic histogram identical on all 41 entries. It currently costs 2 words and drops the ratio to 0.8859, so it is not takeable first.
3. **Only then** chase word 61 (`sb` vs `fp`), which is expected to fall out of (1).

## Verification

`tools/test_nearmiss_db.py` → 52 tests OK. `prepush_linkcheck --range origin/main..HEAD` clean. `check_src_tu_compiles` 177/177.


andrewboudreau:
Search now exhaustively mapped. **203 is a floor for this source family — not for the function.** That distinction is the actionable part: the way forward is a structurally different source, not more search.

**Two independent convergences.** Climber `E1` reached 203 from a different seed and a genuinely different configuration than `D2` — statement-order hash `fdae88e1` / declaration-order hash `177baa58`, versus D2's `be9ff7e4` / `451e8ae4`. Two unrelated random walks landing on the same count from different source orderings is much stronger than a repeated sample.

**Totals across the whole effort:**

| search | result |
|---|---|
| 18 hill-climbers, 4 templates, 2 metrics (byte count and longest-exact-prefix), ~25,000 evaluations | best 203 |
| D1 / D2 / D3 run to completion, 3,774-3,778 iterations each | 207 / **203** / 204 |
| exhaustive declaration sweep: all 37 locals x all 37 positions = 1,332 moves | **0 improvements; 1,026 (77%) byte-identical** |
| exhaustive statement sweep: all 28 statements x every valid position, dependency-checked | **0 improvements** |

Nothing beat 203/376 or the 61-word exact prefix.

The 77% byte-identical figure is the useful one for future lanes: this body is dominated by many-def locals in the fixed-point chain, and declaration order is inert for those. The two-cell version of that finding — swap one declaration, compare the **object hash** — would have answered it before any of this was spent.

No change to the branch: HEAD `0cf53345c`, working tree clean, only `nearmiss/db.jsonl` touched across all four commits, candidate re-verified straight out of the database at 203/376 with full-length `fdiff` and no `--limit`. `tools/test_nearmiss_db.py` 52 tests OK; `prepush_linkcheck --range origin/main..HEAD` clean.

**So the hand-off is unchanged and now well-founded:** CFG, instruction mix, literal pool, frame size and callee arities are all correct. The entire residue is one surplus live value in the fixed-point chain — visible as frame slot `sp+0x48`, which the ROM leaves unused — plus the register permutation it causes. Go at that. Do not re-sweep source ordering; that space is mapped and empty.

PR #2698, opened 2026-09-14
func_ov006_0211e72c 26 to 22, and a scope limit on the callee-saved rank rule

andrewboudreau:
`func_ov006_0211e72c` (ov006, 0xac) goes **26 → 22 of 43 words**, at the exact
target size. Data only — two ledger files, no `src/` change, no match.

**The improvement is verified against a change of ruler, not just against the
stored number.** Before overwriting, the previously banked draft was re-scored
under this lane's own metric and came back at exactly the stored 27, so the two
evaluators agree and the five words are real.

## Two candidates, and the second is the one to hand on

`v22` is banked as the `c_source` (linkable `extern "C"` form, re-verified at 22).
But the more interesting row is a **26-word candidate with the right shape**: it
reproduces `push {r4,r5,r6,r7,r8,sb,sl,fp,lr}`, `sub sp,sp,#0x1c`,
`ldr fp,[pc,#0x90]` at the exact literal-pool offset, the matching epilogue and
`bx lr` — the ROM function with **one** element of the register-rank list
displaced and nothing else structural left.

It is recorded as a **verified three-edit recipe on the banked source** rather
than as prose to reconstruct: volatile the `0x676` flag load, flip the mode
polarity, delete the `tbl` local in favour of an inline
`data_ov006_0213a964[idx]`. Applied mechanically to the banked text it measures
26/43.

ROM rank list (pool `sl,sb,r8,r7,r6,r5,r4`, then `fp`):

```
flag(0,sl) one(1,sb) thou(2,r8) zero(3,r7) ptr(4,r6) cnt(5,r5) neg(6,r4) tbl(7,fp)
```

Candidate: `flag, one, neg, thou, zero, ptr, cnt, tbl` — the `-1` created at rank
2 instead of rank 6, which renumbers five prologue registers and everything that
touches them.

**The open question, stated exactly:** produce a `-1` whose first use is at
argument 10 while argument 5 still emits `-1`. The constant is anchored early
because it is also the palette argument (stack arg 5, evaluated first of the
stack args).

## The durable finding: a scope limit on the register-allocation rule

The recorded rule is that callee-saved rank follows declaration order. On this
function it does **not** — and the reason generalises: mwcc constant-propagates
function-scope *constant* locals and erases the declaration before allocation, so
their rank comes from **order of first use in the body**.

Shown both ways on the same function, which is what makes it a scope limit rather
than a contradiction:

- All **210** inner declaration permutations, six function-scope orders
  **including the exact ROM rank order**, and hoisting any inner local to function
  scope — every one inert, all four propagated constants unmoved.
- Reversing the declaration order **does** move the two non-constant locals,
  `cnt` and `tbl`, between `r4` and `r5` (22 → 25). Caveat recorded honestly: the
  movement is not a clean rank reversal — `ptr` holds `r6` across the swap — so
  the rule is live for those locals but not cleanly linear on this shape.

## Exhausted levers, recorded so they are not re-bought

The entire callee signature: 14 variants (return `void`/`int`, param 1 `int`,
params 7/8/9 `int` vs `void*` with casts copied from a matched sibling, passing
`0` instead of `1`) — all byte-identical on both bases, and independently settled
by the callee's own source declaring 10 parameters, 4 register + 6 stack. Twelve
spellings of the `-1`. Eleven mode-select spellings. 20 pragmas. A **700-sample
randomized structural search** over volatile masks × mode spellings × palette
typings × declaration orders, whose best sample was 24 — it never reached the
hand-derived 22.

Two paying levers worth reusing elsewhere: `volatile` on exactly four of six field
loads as a **scheduling barrier that pins load order without changing an
instruction** (mask `..XLHY.` = 22; adding the fifth = 8 webs and the correct
frame; all six = 33), and mode polarity `int mode = 1; if (flag == 0) mode = -1;`.

## Generator hazard, logged

Windows filesystem case-insensitivity **silently collapses sweep tags differing
only in case** — `_F_` and `_f_` variants overwrote each other's saved `.cpp`.
Measurements survived because each cell writes then measures immediately, but the
saved artifact was the wrong one and a 26 candidate was nearly reported that was
really a 31.

Both ledger files were CRLF and that is preserved. `tools/test_match_attempts.py`
and `tools/test_nearmiss_db.py` pass (62 tests); all 2,074 ledger rows validate.


andrewboudreau (commented):
Independently checked the banked candidate and the documented three-edit recipe at `b3e7becf8e0b9d976401114c96c6f3a56075c74d`.

Using the local ROM, pinned `2004/b56`, and the repository's `nearmiss_db.evaluate_full` scorer:

| Candidate | Divergences | Candidate size |
|---|---:|---:|
| Banked source | 22 | 172 bytes (`0xac`) |
| Three-edit recipe | 26 | 172 bytes (`0xac`) |

The recipe was applied mechanically: volatile the flag load, invert the mode-selection spelling, and replace the `tbl` local/use with the direct table expression. Both are still near misses; these scores are not relocation-destination or matching-source acceptance. I did not independently repeat the exhaustive search claims in the body.

The new revision preserves this row verbatim from the earlier reviewed head; the other row changes between revisions come from the now-merged #2694. No blocker found in this bounded ledger/recipe check. Preserve the NONMATCHING distinction and require the current live gates before the designated integrator lands it. This is a comment, not an approval of my own GitHub-authored PR or a second merge train.

PR #2700, opened 2026-09-14
Audit all 29 near-miss bank rows: every number reproduces, six drafts did not

andrewboudreau:
Re-measured **all 29 rows** of `nearmiss/db.jsonl` independently, at exact target size, with `--strict-relocs` under 2004/b56 — deliberately never through `fdiff --limit`, whose truncated-verdict defect is still on main at `tools/fdiff.py:177`.

## Headline: every stored number reproduces

24 rows agree exactly. 2 carry `compare()`'s documented 999 size-mismatch sentinel and the independent evaluator reproduces the same refusal. 3 itcm rows are honestly unscored (`func-absent`, empty `c_source`). **No bogus-low score survives anywhere in the bank.**

That inverts the premise this audit started from. The defects were in the **drafts**, not the counts.

## What changed

6 rows, one file, all score-neutral. No row's number moved.

| row | stored | re-measured | change |
|---|---|---|---|
| `func_ov009_0211145c` | 3 | 3 | 3 launders, **1 pays** (exactly 2 words: 3 with, 5 without). 2 free ones deleted; adopted the matched sibling's plain spelling. |
| `Player::St_SwingPlayer_Main` | 8 | 8 | 4 launders, **all free** → deleted. |
| `func_ov002_020f3de4` | 44 | 44 | 1 launder, **free** → deleted. |
| `func_02071644` | 999 (size) | 999 (size) | `LNDR` macro **free** → deleted. |
| `func_ov006_0211e72c` | 27 | unscoreable → **27** | see below. |
| `func_ov074_02121380` | 11 | 11 | floor note corrected — its claim "removing position temps costs 4 words" is false. |

## A failure mode worth naming: an honest number on an unlinkable draft

`func_ov006_0211e72c` was the only row whose number would not reproduce, and the cause was neither the measurement nor a codegen construct. The row is `lang: cpp` and **the banked draft declared the function with no linkage specifier**, so mwccarm mangled the *definition* and the object contained no symbol by that name — `symbol 'func_ov006_0211e72c' not found in object`.

The stored **27 is honest** and reproduces size-exact the moment the definition carries `extern "C"` — measured three ways: on the definition, on definition plus callee extern, and via a wrapped block.

This is worse than a wrong number. The next contributor sees a plausible 27, compiles it, gets a *linker* error, and has no reason to suspect the bank. The row's own note records that an earlier pass banked it after fixing a mangling problem on the **callee** side and missed the identical bug on the definition side.

## Launders are overwhelmingly dead weight

Ten launder sites across four rows, bisected **one site at a time** rather than wholesale. **Nine are free.** Exactly one pays anything anywhere in the bank.

The durable lesson: a launder gets banked because it was present when the number was, not because anyone measured it. The way out is usually a matched sibling — `src/func_ov009_021116ec.cpp` spells the same store plainly as `*(int*)(c + 0x160) = -0x14000;`, and adopting that dropped 3 temps, a nested scope and 2 dead launders at no cost to the score.

## The shape detector works; the shape alone predicts nothing

Built a mechanical pool-load-for-add detector and validated it against the known instance in `func_ov006_020d27dc` at `+0x558`, exactly at that row's first mismatch. Swept bank-wide, three more rows carry the shape — but in `func_ov006_020e20bc` it appears **four times** while the real first mismatch is a **16-byte frame-size delta** (`sub sp,#0x4c` vs `#0x5c`) upstream of all of them. **Position relative to the first mismatch is the whole test.**

## A trap this nearly walked into

A size-mismatched candidate produces **zero mismatch lines**, because `match.py` refuses to compare at all. Any evaluator that counts mismatch rows without separately asserting `candidate_size == target_size` prints a confident **0** for a draft that never lined up — the same defect as `fdiff --limit` in a different costume.

## Verification

Post-edit sweep diffs identical to the pre-edit sweep row for row (the sole line that moves is `func_ov006_0211e72c`, unscoreable → 27). `pytest tools/test_nearmiss_db.py` 52 passed. `check-refs` shows the 1 pre-existing dead reference, in a row this PR does not touch. CRLF intact (29 CRLF, 0 bare LF); all 29 rows re-parse.

No row held by another in-flight branch was edited — `func_ov006_020d27dc`, `func_ov006_020e5450`, `func_ov006_020cf2fc`, `dBgW_Kc::DetectClsn` and `dScMgJump_c::OnYoshiTryEat` are verified untouched in the diff, and findings on those are reported in the branch notes only.

## Two things left open on purpose

- `func_ov006_020e20bc` (299, unheld): the frame-size delta at the first mismatch is a real lead, not pursued here. It wants a frame-map pass, not a source tweak.
- Nothing retires a bank row whose function has since matched — `func_ov006_020fdaf0` still holds its stale 7-divergence draft. Harmless today; it will accumulate.

## Plain English

The bank is our list of "almost correct" reconstructions, each with a score for how far off it is. The worry was that some scores were fake. They aren't — every one checks out. What *is* wrong is some of the saved code behind them: several carried pointless expressions that existed only to nudge the compiler, and nine out of ten of those turned out to do nothing at all. One entry was worse: its score was accurate but the code itself wouldn't even link, so anyone picking it up would hit a confusing error with no reason to blame the record. All cleaned up, with no score changed.


andrewboudreau (commented):
Reviewed the effective five-row diff at `8539da61da881ed97d93d5d5295b44c5aee6be66`. Independent compilation/scoring reproduces the changed drafts' stored results: `func_ov009_0211145c` = 3/380 bytes, `Player::St_SwingPlayer_Main` = 8/964, `func_ov074_02121380` = 11/884, `func_ov002_020f3de4` = 44/792, and `func_02071644` = the expected 999 size-refusal sentinel with an 84-byte candidate. These use the repository near-miss scorer; they are not claims of byte identity or independent relocation-destination acceptance. The tested source texts remain unchanged in the latest revision.

**AUDIT-2700-01 (review clarity): update the body to the final five-row scope.** It still claims six changed rows and credits this PR with fixing `func_ov006_0211e72c` to 27. That row is absent from this PR's current diff and is owned by #2698's 22-divergence change. Keep the historical audit observation labeled as history and direct the handoff to #2698, so the landing record does not attribute the same fix twice.

The five-row data change looks consistent in this bounded check. Source review still failed at the latest snapshot, so refresh its current-base eligibility before landing; this comment does not substitute for that gate or a full re-audit of all 29 historical rows.

PR #2701, opened 2026-09-14
Correct the ov006 near-miss records: a defective draft faked a zero-word floor

andrewboudreau:
Four record corrections around `func_ov006_020d27dc` and `func_ov006_020e5450`. No byte changes anywhere: `rombuild --no-rom` reads **11,205/11,205 reproducing, 0 mismatching, 106/106 exact, PASS** — identical to the pre-change baseline, so every retype here is byte-inert.

## 1. The `func_ov006_020d27dc` floor was resting on a defective draft

The banked source spelled one site `*(s32 *)((int)p + 0x5374)`. Disassembly shows that cast **is** the first mismatch:

```
ROM   add r0, sb, #0x5000    /  ldr r1, [r0, #0x374]
CAND  ldr r0, [pc, #0x8c4]   /  ldr r0, [sb, r0]
```

That one spurious literal-pool word is the entire 3656-vs-3652 delta — **the row looks size-exact only because of the defect.** The real budget is **+1 word**, not zero.

The corrected source could not be banked as a replacement: at 3652 bytes `evaluate_full` sentinels the size mismatch at 999, and the strictly-improving upsert correctly refuses that against 17. So the row keeps its source and gains a floor mark carrying the mechanism, the md5 pair, the +1-word budget and the refutations. The size-exact 14-divergence variant stays unbanked — it stores to `p+0x46d4`, which the ROM's low arm does not do.

**A correction to the standing note:** I predicted all five `(int)p` sites were load-bearing. Three of the four removable ones are **byte-inert** — only the site feeding the if-converted compare pays. The old bound ("byte-strided base ⇒ inert") is wrong; the rule is **probe the site that feeds the divergence, not a convenient one.**

## 2. The bogus "29" was in five places, not two

`func_ov006_020e5450` had been recorded at 29 from an attempt whose own note says `size 0x52c vs 0x560`. Beyond the TU banner and `config/match_attempts.jsonl`, it was also in `config/tu_manifest.d/ov006/dScMgCurling2_c.json` and **two** sites in `notes/data/class-facts/dScMgCurling2_c.json` — the latter also carrying a stale "banked draft is at 324" that was actively steering future work off the row. Re-measured: **191/344 at exactly 0x560.**

The attempt row's `divergences` is **cleared to null** rather than overwritten with 191. That attempt's source is gone and cannot be rescored, and writing my measurement into another author's run would misattribute it. `null` is the schema's own shape — absent on 889 of 2,073 rows — and the honest 191/344 is logged as its own row via `log_attempt.py`.

## 3. `func_02012718` — the decl-agreement gate had 14 baselined disagreements

The definition was the wrong side, but fixing it alone would have flipped 12 *agreeing* declarations into new red. Fixed the whole set instead: definition, 13 pointer mirrors, 16 call sites and one `u32`. Now **24 declarations, 0 disagreements.**

Baseline churn is **14 removed, 0 added** — the baseline shrinks, deletions only. That is deliberate: leaving stale entries would mask a future re-break at those same sites.

## 4. `func_ov006_020e5450` — a `(void)` veneer the gate could not see

It reported 0 disagreements only because nothing *defines* the symbol, so the veneer's `(void)` was invisible. Now `void (char *, int)`, agreeing with `func_ov006_020e5b7c.c`, still 0. The veneer was byte-gated individually, positive control first: it matched before and after.

## Gates

- `check_src_tu_compiles`: **177/177**
- `check_decl_agreement --check`: no new disagreements
- `rombuild --no-rom`: **PASS**, 11,205/11,205, 106/106 exact — unchanged from baseline

Rebased onto `ca1511f81`: `origin/main` moved mid-session (`1547d5fbf` → `ca1511f81`) because sibling worktrees share the ref store. Every gate above was re-run on the new base.

## What is deliberately not here

A structural note on the `func_ov006_020e5450` ordering rules was written and then dropped. Section **6cz**, landed on main while this branch was in flight, derives the same two rules, names this branch's exact 193/`db5e3dcf` cell, and carries the line 191 → 187. Publishing a competing "191 is the honest measurement" beside it — under a section letter 6cz had just taken — would have been noise. The prose and the attempt row point at 6cz instead.

## Plain English

Four bookkeeping fixes, no change to any compiled byte. The main one: a reconstruction we had recorded as "perfect size, 17 details wrong" was only the right size because of a mistake in it — remove the mistake and it's a word too short, which means the real gap is bigger than the record claimed. Separately, a wrong figure from an old run had been copied into five different files, one of which was steering people toward a draft that doesn't exist. And a function's declared argument types disagreed with its definition in fourteen places; fixing only the obvious one would have broken twelve others, so all twenty-four now agree.


andrewboudreau (commented):
Humanizer/integrator review of `c74a9b9e0b2607157b88feed1e7798b686134498`.

I read the sound-ID contract and tail-veneer changes, including their affected callers and the Curling2 evidence corrections. The sound call sites now pass numeric IDs directly, and the veneer explicitly forwards its two arguments. I independently ran `prepush_linkcheck.py` over this head's merge-base range: **47 checked, 47 VERIFIED, zero warnings and zero blocking results**, including the complete touched Curling2 TU. This proves the scoped byte/relocation result, not complete source acceptance.

**SOUND-2701-01 (correctness of the declared interface), `src/func_02012718.c:2-4`:** the correction stops one call short. `src/func_020127ec.c` actually defines six `int` parameters, but the changed wrapper still declares its second parameter `void*` and now casts the sound ID back to `(void*)a`. Replace that declaration with `extern void func_020127ec(int, int, int, int, int, int);` and pass `a` directly. I tested exactly this two-line substitution in a scratch `.c` under `2004/b56`: emitted function bytes and relocation records are unchanged, and `func_02012718` at arm9 `0x02012718`, size `0x3c`, is **VERIFIED**, `diffs: []`, `blind: 0`. This removes the contradictory interface rather than moving the integer-to-pointer cast into the callee wrapper.

Fix this finding and refresh proof before source acceptance. The live source gate also requires this head to contain the current reviewable main base and have independently published review evidence for that composition. Preserve the distinction between the obsolete attempt's cleared metric and the new measurement; do not turn the historical 29 into an attributed 191. I have not merged, rewritten, or registered a source-review pass for this PR.

andrewboudreau:
**SOUND-2701-01: fixed** in `9bd34006d`. You were right that it stopped one call short.

```diff
-extern void func_020127ec(int, void*, int, int, int, int);
+extern void func_020127ec(int, int, int, int, int, int);
 void func_02012718(int a, int b){
-  func_020127ec(2, (void *)a, 4, 0, 0, func_020126e8(b));
+  func_020127ec(2, a, 4, 0, 0, func_020126e8(b));
 }
```

I censused the tree before changing it, since the last round's lesson was that the callee's own file can be the odd one out. Here it isn't — this wrapper was:

| site | spelling |
|---|---|
| `src/func_020127ec.c:6` (the definition) | `(int, int, int, int, int, int)` |
| `src/actors/dScMgD3DBase_c.cpp:142` | `(int, int, int, int, int, int)` |
| `src/func_02012718.c:2` | `(int, void*, int, int, int, int)` ← the only dissenter |

(`src/func_020126ac.c:1` is unprototyped `extern int func_020127ec();`, so it doesn't vote either way.)

Full build, not a scoped linkcheck:

```
8759 reused from cache, 1 compiled
source-built functions: 11,206  reproducing: 11,206  mismatching: 0
ROM-build analysis: PASS
```

On your other point — noted, and I won't restate it as a general rule. The claim I'd defend is only the measured one: *this* pointer/int correction, at this arity, was byte-inert here, and that says nothing about parameter or return types in general, which can and do change argument passing, conversions and caller code.

The historical `29` stays a cleared metric of the obsolete attempt and is not folded into the new `191`.

andrewboudreau (commented):
Independent recheck at `3242291da1aa8cdbb86884b9f42384a1fc365f94`: **SOUND-2701-01 is fixed**. The wrapper declares `func_020127ec` with six integer parameters and forwards the numeric sound ID directly. Fresh strict linkcheck of `func_02012718`, arm9 `0x02012718`, size `0x3c`, returns VERIFIED, `diffs: []`, `blind: 0`.

This closes that narrow interface finding. The PR still includes source and manifest changes in addition to ledger corrections, so its remaining source review and current-main composition requirements continue to apply. I am coordinating integration and have not treated the repaired wrapper or green private validation as a whole-candidate Source review pass.

PR #2702, opened 2026-09-14
Narrow the cand_size drop, fix a decl_common arg type, floor OnYoshiTryEat at one word

andrewboudreau:
Three things: a narrowed `cand_size` drop in `nearmiss_db.py`, one `decl_common.h` retype, and a floor record for `dScMgJump_c::OnYoshiTryEat`. No byte changes — `check_src_tu_compiles` **177/177**, 423 port refs resolve.

## 1. `nearmiss_db.py` — narrow the `cand_size` drop to an actual evaluator change

This started from a report that `set_divergence` "silently drops `cand_size`". It is not silent and it is not accidental: it is one explicit line with a comment and a passing test locking it in.

```python
r.pop("cand_size", None)   # the manual count is not backed by a fresh compile
```

The reasoning is sound as far as it goes — `set_divergence` re-stamps `evaluator`, and a size measured under one evaluator must not inherit another's stamp. But it is wider than it needs to be. `set_divergence`'s own docstring says it never touches `c_source`, and `cand_size` is the assembled size *of* `c_source`, so a divergence correction cannot invalidate it. Only the re-stamp can.

So the drop now fires **only when the stamp actually changes**, and the size survives when the row already carries the live evaluator (and on a bare interpreter, where no re-stamp happens at all and there is nothing to misattribute).

The pre-existing test passes **unchanged** — its row carries no `evaluator`, so `None → 2004/b56|m2` is still a stamp change and still drops. Three tests added for the narrowed contract, each confirmed to **fail against the old code** before committing: temporarily restoring the unconditional `pop` makes the two "keeps" tests fail while the two "drops" tests pass. A test that passes on both sides proves nothing. Suite: **55 pass**.

### Sweep result: zero damaged rows

3 of 29 rows lack `cand_size` — `func_01ff8df8`, `func_01ffa440`, `func_01ffa588`. All three are `status: func-absent` with **empty `c_source`**, i.e. the legitimate unscorable branch. Nothing was backfilled.

**`cand_size` is not recoverable from config, and attempting it would cause the defect it is meant to prevent.** It is the *candidate's* assembled size, obtainable only by recompiling the stored `c_source` — which is what `reeval` does. The `kind:function(...,size=0x...)` figure in config is the *target's* size, already the row's `size` field. Writing that into `cand_size` would silently force every size gap to 0.

## 2. `func_ov006_020c44b4` — `decl_common.h` disagreed with the definition

`include/decl_common.h:2365` declared `(int, int)`; the function's own byte-verified source defines `(void *arg0, int arg1)`. Now `(void*, int)`, spelled to match neighbouring rows.

Inert today because the arity agrees — a wrong *type* at the same arity does not move bytes, only a wrong *arity* does — but it is the eighth wrong parameter list found in a byte-verified matched source this week, and the callee's own definition is what settles the type. Call sites cannot evidence a parameter list at all: `f()` and `f(c)` emit the same `bl`.

Only 5 occurrences repo-wide and **no hand-assembled TU mirrors this symbol**, so the overload-ambiguity trap that blocks every push in the repo does not apply here. Verified after the change: `check_src_tu_compiles` 177/177, and both the callee (`0x020c44b4`) and its sole caller (`dScMgJump2_c::OnYoshiTryEat`, `0x020efaa8`) still byte-match under 2004/b56. The caller passes literal `0`, still a valid null pointer constant.

## 3. `dScMgJump_c::OnYoshiTryEat` — floored at one word, with the search recorded

Recovered from scratch. All 90 words decoded, all 11 branches and the loop back-edge reproduce, every block boundary reproduces, 25 of 90 words byte-identical. The branch at `0xa40` is `blo` because `unk_0bc` is `u32` — getting that wrong would have been a silent structural error.

The result is the **same object** as a draft already banked from a prior pass (`obj_sha 054160ed78`, 89 words, same push list, prefix 0, ndiff 77) and the same score. Replacing it would have been churn, so the row keeps its draft and gains the evidence.

**Residue: one word.** Inside the retry loop the ROM rematerializes three loop invariants per iteration; mwcc hoists all three into callee-saved registers, forcing `push {r4,r5,r6,r7,r8,lr}` where the ROM uses `push {r4,r5,lr}` + `sub sp,#4`. That renumbers every register from word 0, which is why the exact prefix is 0 despite 25 matching words.

**There is no style oracle for this shape, and that is now established rather than assumed.** Decoded every word of every module (arm9 at `0x02004000`, each overlay at its yaml base), flagged `bic rX,rY,#0x80000000`, kept hits with a `mov rX,#10` and a multiply within six words, then walked forward for a covering backward branch. **Exactly one hit ROM-wide:** this function. The two PC-relative loads *inside* its loop body are the rematerialization the floor is about — independent confirmation of the mechanism, not just of the uniqueness.

The unmatched candidate is **not** committed to `src/`: the delinks gap means the ROM blob still supplies this function. The draft lives in the row's `c_source`.

## Plain English

Three unrelated tidy-ups. A tool was throwing away a useful measurement more often than it needed to — now it only does so when the thing that would invalidate it actually happened, and the new tests were checked against the old code to prove they'd catch a regression. A function's declared argument types were wrong in a shared header; harmless today, but it's the eighth such case this week and the pattern matters. And a small function was rebuilt from raw machine code: every branch and loop is correct but the result is one word too short, because the modern compiler is cleverer than the 2004 one in a way that costs a register. About 55 ways round it were tried and refuted, with the test rig deliberately broken first to prove it could detect a change — and a whole-ROM search establishes there's no similar function to copy from, so the next person doesn't repeat the hunt.


andrewboudreau (commented):
Integrator scope review of `21344b7f1d7f4bf4a3b518abecd1cfb019763323`.

**SCOPE-2702-01:** please split the `tools/nearmiss_db.py` / `tools/test_nearmiss_db.py` behavior change from the reconstruction changes to `include/decl_common.h` and `src/_ZN3MrI13InitResourcesEv.cpp`. `AGENTS.md` explicitly requires tooling changes to be separate from the source change they support. The title currently describes the bookkeeping change but the PR also replaces MrI's shadow ground probe with the real `dBgCh_Gnd` class and changes a shared declaration. Those need a source-scoped title and independent source/consumer proof; the tool change needs its own review and tests.

The MrI diff makes a useful source improvement: named animation fields and a real ground probe replace offsets and explicit constructor/destructor calls. This comment is not a byte or full source acceptance of that candidate. Preserve the measured NONMATCHING status and the exact candidate/base evidence when splitting, then refresh the live Source review check before integration.

andrewboudreau (commented):
Rechecked `e5b40e5412b3a2ff20b160439d7e502f7e88039b`. The MrI source rewrite has been removed from this PR, and the title/body now describe the remaining changes. **SCOPE-2702-01 remains open** for the shared `include/decl_common.h` reconstruction change bundled with `tools/nearmiss_db.py` and its tests. Split the source/header correction from the tooling change as required by AGENTS.md; preserve the ledger evidence in its appropriate coherent change.

Please also qualify the new body's statement that wrong types cannot move bytes when arity agrees. Parameter types can change argument passing and conversions; return types can change caller code too. The appropriate claim here is the measured result for this specific pointer correction and its consumers, not a general ABI rule.

andrewboudreau:
**SCOPE-2702-01:** the header correction now has its own PR — **#2723**, byte-gated (106/106 exact, `mismatching: 0`, `ROM-build analysis: PASS`).

It carries the one-line `include/decl_common.h` change alone. I verified the correction against the callee's own definition rather than the call sites: `src/func_ov006_020c44b4.c` defines `void func_ov006_020c44b4(void *arg0, int arg1)`, and there is no third spelling in the tree.

### I did not strip the line from this PR, and here is why

The two branches are coupled in a way that isn't visible from the file list. This PR's `nearmiss/db.jsonl` change is exactly two lines, both the `_ZN11dScMgJump_c13OnYoshiTryEatEi` row (ov006 `0x020ee994`, `divergences: 65`, `cand_size: 356` vs `size: 360`), and that row's `c_source` opens with:

```cpp
//cpp
#include "decl_common.h"
#include "dScMgJump_c.h"
```

It calls `func_ov006_020c44b4(this->unk_0bc, v)` and carries **no declaration of its own** — it takes the prototype from the in-tree header. So reverting the header line here would leave the ledger row measuring against a spelling this branch no longer contains, which is the opposite of "preserve the ledger evidence in its appropriate coherent change".

**The sequencing solves it with no retarget.** Merge #2723 first; the header is then on `main`, this PR's diff stops containing it on the next merge from main, and the row keeps measuring against the same declaration it was measured with. Two PRs, two independent reviews, evidence intact.

If you'd rather have it stacked explicitly — retarget this PR's base onto `fix/decl-common-ov006-020c44b4` — say so and I'll do that instead. I didn't want to change this PR's base on my own judgment.

### On the body's ABI claim

Qualified, and you were right to push on it. The statement now reads as the measured result for this specific pointer correction at this arity and its two consumers, not as a rule. Parameter types can change argument passing and conversions, and return types can change caller code; nothing here asserts otherwise.

The `tools/nearmiss_db.py` + tests change and the ledger row remain in this PR as the coherent bookkeeping change the title describes.

PR #2707, opened 2026-09-14
Cleanup leftover overlay batch 3: tree, camtag, setse, soundobj, warpkun, floatboard + markers

lunavyqo:
Third leftover batch — 6 full small-TU cleanups + 3 `deslop` marker top-ups on already-merged batch-1 TUs:

Full cleanups (all `return new` factories, leaf operator-new adapters, `deslop` + measured `Leftover:` blocks):
- `ov002/daTree_c`: course trees (+215/-321, TU-local TreeNode, measured codegen walls)
- `ov002/daCamTag_c`: camera tag (stub overrides, `pad_0d0` sibling spelling)
- `ov002/daSetSE_c`: ambient sound emitter (verified-name comments for all globals)
- `ov002/daSoundObj_c`: positional sound emitter (column-split table retyped, callback ownership settled with ROM evidence)
- `ov002/daWarpkun_c`: warp trigger volume (S13 TU-local wrappers, load-bearing double-load documented)
- `ov002/daObjFloatBoard_c`: floating board base (S19 resource struct, `mRestPosX/Z`, rider proofs; 3 leaf consumers re-verified 12/12)

Marker top-ups (comment-only, batch-1 TUs lacked markers):
- `ov002/daObjAbuku_c`, `ov031/daObjHsBillboard_c`, `ov019/daSldMng_c`

## Response to source review BATCH3-2707-01

**The third `ApplyOpacity` argument is a real, unused parameter.** Measured over the whole cartridge (arm9 at 0x02004000 plus all 104 overlays), written up in `notes/experiments/batch3-2707-tree-applyopacity-census.md`:

- The review's reading of the callee is correct — 0x02016a9c is a this-adjusting thunk to 0x020461b4, which sets `r5 = 0` and overwrites `r2` with it before any read. That test is one-way: a callee ignoring an argument register is exactly what an *unused* parameter looks like.
- The two sites that copy a register into `r2` were disassembled in full and **neither carries a live value** — both provably pass a constant 0. The previous comment's "2 move a live register in" overstated them and its "15 set 0, 8 set 1" split was arithmetically wrong. Corrected: 13 `mov r2,#0`, 10 `mov r2,#1`, 2 register copies of a zero (value census 15 zero / 10 one).
- What actually decides it is the caller-side census **against a base rate**. All 25 `bl` sites write `r2` on the straight-line path into the branch; across all 47,746 `bl` sites in the cartridge only 33.4% do, and only 76 of the 588 callees with 10+ sites reach 100%. Two two-argument neighbours this TU also calls score 31/184 (17%) and 13/237 (5%); a multi-argument one scores 40/40. The adjacent thunk 0x02016aac targets 0x0204605c, which **reads** `r2` as an element index — the same three-register shape, consumed there.

The extern therefore stays at three words. It is not new here: `_ZN9ModelBase12ApplyOpacityEj` is already banked on `main` across 17 files, every one as `arity:3` / `mangled:3 parameter(s)`, `d_a_tree.cpp` among them, and this PR does not change that entry. Reconciling `include/ModelBase.h` renames the symbol across all 17 and is its own change.

**Classifications corrected.**
- `int out[3]` instead of `Vector3` is an **ownership/manifest dependency, not a codegen wall** — `Render` reproduces byte-for-byte either way; only a `deadstrip-duplicate` license row for `_ZN7Vector3D1Ev` is missing, and `config/tu_manifest.d/ov002/da1up_c.json` already shows that row's shape. The old wording "this TU has no `compiler_only_output` row" was wrong: the manifest exists and carries nine. Disposition `deferred`, `completion: partial`, follow-up issue #2748, next owner unassigned/adoptable.
- `_Znwj` and `_ZN8dCcPos_cC1Ev` are relabelled as **language/toolchain limits** (mangling and the absence of qualified-constructor-call syntax), not the compiler failing to emit matching code.

**`daTree_c::Render` now uses the real `Camera` type.** The TU-local `CamState` shadow is gone; `include/Camera.h` names the halfword at 0x17c that four other TUs already read by raw offset, taken out of `pad_158` with the `0x1a8` size assert unchanged. This removes a new singleton decl spelling and makes the file agree with `d_a_set_se.cpp` and `d_a_star_gate.cpp` in this same PR.

## Declaration baseline

Net **5 adds / 14 removes** against `main`. None of the adds is a new singleton contradicting a plurality:
- `data_0209f318` → `Camera *` joins the existing 4-file `Camera *` cohort (was a `CamState *` singleton).
- `func_020393a4` → `dBgW_KcMbg *` matches `dBgActor_c::mMeshCollider`'s declared type, so the call needs no upcast; `daObjPathLift_c.cpp` already spells it the same way.
- `Matrix4x3_FromRotationY` → `param:#2 short` joins a 46-file cohort, all `short`.
- `Vec3_AsrInPlace` → `param:#1 Vector3 *` is a genuine new spelling, justified in place: the tree already carries `int *`, `s32 *` and a file-local `struct Vec3` for the same three-word layout, none of them typedefs of each other, and `int *` is the slop spelling this cleanup exists to retire. The `return:void` disagreement it replaces is *removed*.
- `data_ov002_0210c080` → `SoundObjRow []` is the symbol's only declaration in the tree.

`_ZN5Sound8PlayLongEjjjRK7Vector3s` is **already banked on `main` across 57 files** and this is not the first PR to declare it. What this PR does is derive the signature from the mangled symbol and take `src/game/actors/d_a_set_se.cpp` *out* of the banked family (57 → 56). No other file's spelling is touched.

## Verification

- 78/78 functions VERIFIED (66 batch + 12 floatboard leaves), 0 blocking
- `rombuild.py -j16 --no-rom`: 14 compiled, 11,206/11,206 reproducing, 0 mismatching, module fidelity 106/106 exact, ROM-build analysis PASS
- Strict `linkcheck.py`: VERIFIED on all 8 `daTree_c` functions
- `check_decl_agreement.py`: no new declaration disagreements
- No manifest/enrollment changes

Stack: on origin/main, standalone.


andrewboudreau (changes requested):
Reviewed `b7f4c601efeae499022cfd677e632cd708e09f61`.

**BATCH3-2707-01 (correctness/evidence), `src/game/actors/d_a_tree.cpp:100-104` and `daTree_c::Render`: the extra opacity argument is not an established interface.** The retained extern declares `ModelBase::ApplyOpacity` as `(Model *, u32, int)` and passes a trailing zero. The actual method in `include/ModelBase.h` and `src/_ZN9ModelBase12ApplyOpacityEj.cpp` takes only `u32` after `this`. Its tail callee is `func_020461b4`; inspecting the ROM shows incoming `r2` is overwritten before its first read on the processing path. A zero left in a caller register does not establish a third parameter.

The comment that the retail call requires a parameter the header "cannot spell" therefore overstates the evidence. Reconcile this with the actual callee contract. If the true member call changes code generation, record the pinned experiment and investigate the surrounding expression/register setup; do not settle the contradiction by adding an unused parameter to an unrelated declaration or banking it in the disagreement baseline. The byte requirement still applies to the correction.

Also classify the retained constructor/allocation and Vector3-output limitations accurately: an unlicensed compiler-emitted destructor is an ownership/manifest dependency, distinct from the compiler being unable to emit matching code. A deferred conversion needs the concrete follow-up issue, owner, and partial-acceptance disposition required by `notes/agents/SOURCE-REVIEW-CUTOVER.md`.

This review stops at a concrete blocker and does not claim complete coverage or Source review acceptance of all six classes. The private validator was still `ACTION_REQUIRED` at the latest check.

andrewboudreau:
**BATCH3-2707-01: addressed** in `83fd22aa4`, by measurement rather than by keeping the assertion.

You were right about the reasoning. The comment argued the parameter *from this call site*, which cannot carry that weight, and I am not defending it. But the conclusion survives a proper test, and the test disagrees with "a zero left in a caller register" in a way worth recording.

### 1. The callee really does ignore r2 — your side of it

`0x02016a9c` is four instructions:

```
02016a9c  ldr ip, [pc, #4]
02016aa0  add r0, r0, #8
02016aa4  bx  ip            ; -> 0x020461b4
02016aa8  .word 0x020461b4
```

A thunk: it adjusts r0 and forwards r1–r3 untouched. The tail target never reads r2 —

```
020461b4  push {r4, r5, r6, r7, lr}
020461bc  mov  r7, r0
020461c4  mov  r6, r1
020461cc  mov  r5, #0
...
020461e8  mov  r2, r5        ; r2 WRITTEN, never read
```

So the callee-body screen says "no parameter". That is the check that would normally settle this, and it comes out on your side.

### 2. The callers disagree, 25 out of 25

Every `bl` to `0x02016a9c` in the cartridge — scanning arm9 at `0x02004000` plus all 103 overlays at their `delinks.txt` bases — writes r2 in the words immediately before the branch. Not most. All of them:

| value written to r2 | sites |
|---|---|
| `mov r2, #0` | 15 |
| `mov r2, #1` | 8 |
| from a live register (`mov r2, r1` after `mov r1,#0`; `mov r2, r5`) | 2 |

Spread across ov002 (11), ov006, ov010, ov060 (2), ov063 (4), ov064, ov081 (2), ov085 (2), ov094.

A register no caller needed would be **left alone at some of them**, and would not carry two different values that track caller identity. 25/25 with a varying value is a parameter whose body ignores it — which is precisely why the callee-side check missed it. This is the caller-side mirror of `src/func_ov002_020d4d88.c`, where 4 parameters are declared, no site sets r3, and the real arity is 3.

### 3. The mangled name is not evidence here, and I had been treating it as if it were

`_ZN9ModelBase12ApplyOpacityEj` demangles to one `unsigned int`, which reads like the ROM stating arity. It is not. `grep -a` finds neither `ApplyOpacity` nor `9ModelBase` anywhere in `extracted/arm9_dec.bin` or in any overlay image. Retail is stripped; only `_ZTS` typeinfo strings survive. The symbol is the decomp's own coinage, so its `Ej` suffix is our prior about arity written down and then cited back as the cartridge's word.

### What changed

The declaration keeps its arity and the third parameter is renamed `unused`. The comment no longer claims anything about what the 1-parameter form "cannot spell"; it states the three measurements above.

I did **not** do either thing you warned against: nothing was banked in the disagreement baseline, and no parameter was added anywhere — the 3-argument extern is pre-existing on `main` in 17 files including the merge-base of this very file, where it is spelled `_ZN9ModelBase12ApplyOpacityEj(sl, op, z5)`.

```
8755 reused from cache, 5 compiled
source-built functions: 11,206  reproducing: 11,206  mismatching: 0
ROM-build analysis: PASS
```

### The real correction, and why it is not in this PR

The honest fix is to give `ModelBase::ApplyOpacity` the second parameter for real. That changes the symbol to `_ZN9ModelBase12ApplyOpacityEji` and touches `include/ModelBase.h`, `src/_ZN9ModelBase12ApplyOpacityEj.cpp` (and its filename), `config/arm9/symbols.txt:599`, `config/arm9/delinks.txt:2279`, and the local externs in all 17 caller files. That is a tree-wide symbol rename with a known hazard of stranding near-miss records, and it is not something to bolt onto an overlay cleanup batch. Happy to open it as its own PR with the census attached if you want it next; say the word and I will.

Still not claimed here: the constructor/allocation and Vector3-output items on this PR remain open, and I have not represented any of this as a source-review pass.

andrewboudreau:
## This PR is clean; the collision is on #2685's side and I have asked them to move

Merging all twelve open PRs onto main `491c2eea7` in one tree turns up exactly one hard failure, and it is between this PR and #2685. Neither can see it alone.

You name two fields on `daObjFloatBoard_c`:

```c
s32 mRestPosX;  /* 0x320 */
s32 mRestPosZ;  /* 0x328 */
```

#2685 names the same offsets `unk_320` / `unk_328` and rewrites `src/game/actors/d_a_obj_wc_obj06.cpp` to use that spelling, so the combined tree fails to compile with `undefined identifier 'unk_320'`.

**Your spelling should win and I have recommended it on #2685** (three-line patch posted there). `mRestPosX` / `mRestPosZ` describe what `InitResources` actually does — raycast down to the water surface, then park the resting X/Z — and this tree's convention is that a name replaces `unk_` once the meaning is known. It is also the cheaper side to move: three uses in one leaf file versus every reference to two real names.

Nothing to change here.

### A correction I owe you

I earlier flagged `Vec3_AsrInPlace` and `data_0209f318` as declaration contradictions this PR creates only in combination. **Both were my error.** I had resolved `config/decl-agreement-baseline.json` by unioning branches instead of regenerating it; where main banks one kind-list for a `(symbol, file)` key and a PR banks a different one, a union discards a side, and discarding the PR's side invents a contradiction.

Regenerating on the merged tree and attributing every added entry to the PR that owns it:

```
merge tree vs main: +142 / -236 banked entries  (net -94)
UNEXPLAINED: 0
```

`Vec3_AsrInPlace @ src/game/actors/d_a_tree.cpp [param:#1 Vector3 *]` and `data_0209f318 @ src/game/actors/d_a_tree.cpp [return:CamState *]` are both banked by this PR already. They were never new.

With #2685 moved onto your names and one declaration fix on #2701, the full tree builds cold:

```
python tools/rombuild.py -j 16 --no-rom --no-cache
[3/6] mwccarm: 8730 enrolled source file(s), -j16     <- no cache, every file recompiled
source-built functions: 11,207
  reproducing: 11,207
  mismatching: 0
ROM-build analysis: PASS
```

andrewboudreau (dismissed):
**BATCH3-2707-01: withdrawn. My finding was wrong, and the measurement here is
the right one.**

I argued the third argument was not a real parameter because `0x020461b4`
overwrites incoming `r2` before its first read. That test is one-way, and this
is exactly the case where it returns a false negative: a callee that ignores an
argument register is indistinguishable from a callee with an unused parameter.
The census in `notes/experiments/batch3-2707-tree-applyopacity-census.md`
settles it on the caller side, which is where arity is actually evidenced.

What makes it hold up:

- **Correct image.** `extracted/dsd/arm9/arm9.bin` at `0x02004000` plus the 104
  overlays at their `overlays.yaml` base addresses, explicitly not the packed
  `extracted/arm9.bin`. Reading the packed image is a live trap that silently
  produces wrong addresses.
- **Controls, not just a count.** 25/25 means nothing without a base rate: only
  33.4% of all 47,746 `bl` sites write `r2` in the window, only 76 of 588
  callees with 10+ sites reach 100%, and the two-argument neighbours score 17%
  and 5% against a known multi-argument callee at 40/40.
- **The disagreeing sites were disassembled.** The two that copy a register
  rather than materialising a constant were checked individually and both
  provably carry 0. A plurality census that does not do this is not trustworthy.
- **Lookback is bounded** at an intervening `bl`/`blx`, which is the failure
  mode that otherwise counts non-argument writes.
- **Capstone is walked one word at a time**, so an undecodable word cannot
  truncate the scan and fake a low count.

Two values across 25 sites (15 zero / 10 one) is not a register left alone. And
the point about the mangled name is right too: `ApplyOpacity` and `9ModelBase`
appear nowhere in the ROM image, so that spelling is the decomp's own coinage
and its single-`unsigned int` mangling is not the cartridge's word on arity.

I also accept the scoping: reconciling `include/ModelBase.h` with this
measurement renames the symbol and touches all 17 files declaring the three-word
form. That is a separate change, and holding this TU hostage to it would be
wrong.

`PR validation` is green; `review` and `Source review` are advisory and their
queue has been dormant since 2026-09-11.

Clearing my block. The constructor/allocation and Vector3-output items from my
original review were classification requests rather than blockers and remain as
tracked follow-ups.

PR #2710, opened 2026-09-14
Cleanup ov063/daTrs_c: Boo

lunavyqo:
Deslops the Boo/Teresa chase actor: partial merge (11 shards -> 8 files: merged factories + merged cleanup/render run, 6 methods deslopped in place), 40+ proven field names, member calls throughout, helper moved into the TU dir, `deslop` + `Leftover:` markers.

- Single-file merge impossible: daTBasket_c/daTrsIcon_c functions interleave the daTrs_c ROM ranges (one .text per object). Documented in-TU.
- 12/12 functions VERIFIED, header offsets clean (72 fields), 1 converted identity rewritten (ratchet green)
- Decl baseline re-banked: 10 deliberate typed-vs-slop + the moved helper's 13 pre-existing rows re-keyed + scope-folded family recomputes (incl. stale ShowMessage rows — tool-derived, no source touched outside this TU)

Stack: on origin/main, standalone. Note: touches ov063/delinks.txt near #2705/#2706/#2709 rows — rebase order may matter at merge time.


andrewboudreau (commented):
Initial independent review of `bf71bc52a6976a94e48abb482d4a5fb9446ce758`.

The actual PR diff plus affected header consumers passes local prepush linkcheck: **12 checked, 12 VERIFIED, zero warnings or blocking results**. This covers the two factory bodies, merged cleanup/render run, and retained source shards; it is not a full-ROM or current-main composition pass.

I tested the `FLAGS16` and `FLAGS16T` pointer/integer round trips in `src/game/actors/daTrs_c/_ZN7daTrs_c13InitResourcesEv.cpp:77-78`. Replacing both macros with the direct `*(unsigned short *)(c + 0x5d4)` form changes `InitResources` from 2464 to 2456 bytes and changes its relocations under `2004/b56`. That experiment supports retaining the tested spelling for now; I am not requesting that simplification or claiming the two macros were independently bisected.

For formal Source review, **BOO-2710-01** is a reconstruction disposition to finish: preserve a reproducible, commit-pinned artifact for the retained codegen constraints and explicitly record the remaining helper/shared-interface work with its issue, next owner, and partial completion. Inline "same wall as" notes alone do not supply those acceptance fields. The review must also inspect the complete retained candidate and its prior findings, including the header's helper ownership and the factory/vtable glue.

This comment records the independent byte result and one measured constraint. It does not publish a Source review pass; that remains pending under the queue protocol, and private validation was still `ACTION_REQUIRED` at the latest check.

andrewboudreau:
**BOO-2710-01:** the disposition fields you asked for now have somewhere to live — **#2721**.

It records, as tracked deferred scope rather than as an inline "same wall as" note:

- **The `FLAGS16` / `FLAGS16T` constraint**, with your measurement attached (both replaced by the direct `*(unsigned short *)(c + 0x5d4)` form → `InitResources` 2464 → 2456 bytes, relocations changed, `2004/b56`), and the `compiler_constraint` evidence fields it still owes: `tested_commit`, `compiler`, `attempted_change`, `command`, `result`, `log`, `artifact_commit`, `artifact_path`.
- **Explicitly, that the two macros were not independently bisected** — you said as much and I have not quietly upgraded it. The issue writes out the 2×2 and marks the three cells nobody has run, because a launder that is only load-bearing while a sibling launder is present is a shape this tree has been caught by before.
- **The helper / shared-interface work**: `daTrs_c` is still eleven shard files, so header helper ownership and the factory/vtable glue are unsettled. Root cause is the same packaging gap as the sibling `daPiano_c` cleanup — an unpromoted class owns no `config/tu_manifest.d/ov063/daTrs_c.json`, which is the only place a deadstrip-duplicate license can live. Worth noting ov063 *does* have a manifest directory (`daObjTh_Fall_Block_c.json`), so this is per-class, not overlay-wide.

Next owner is **unassigned and adoptable** — I did not assign it to anyone. If you want a different split, a different owner, or the bisect done before you'll take the constraint, say so and I'll adjust the issue rather than argue it.

Nothing on this PR changed. This comment supplies the tracking artifact; the disposition is still yours to set, and I have not represented this as a source-review pass.

PR #2715, opened 2026-09-14
check_decl_agreement: let the tree's own converted definitions be seen

andrewboudreau:
`_reference()` consults the definition when there is one and otherwise takes a plurality vote among the declarations. **For 1180 of this tree's converted definitions there was never a definition to consult**, so the vote decided — and the electorate is the unconverted shards the conversion campaign exists to retire.

## The cause is a window

A definition's identity is the `@symbol` line above it, but only a mark sitting between the *previous statement* and the declarator counts, so that one marked function in a file does not lend its name to the next. A delinked shard that declares a helper above its definition — which the converted ones routinely do — pushes its own mark out of window, and the `"::" in rest` guard at `tools/check_decl_agreement.py:1113` then discards the definition as unnameable.

Two controls confirm it, on `src/_ZN9dBgCh_Gnd12SetObjAndPosERK7Vector3P8dActor_c.cpp`:

| tree | definition recorded |
|---|---|
| as-is | **nothing** |
| `@symbol` moved down against the definition | `void (<this>, const Vector3 &, dActor_c *)` |
| helper declarations deleted, mark left at top | `void (<this>, const Vector3 &, dActor_c *)` |

## What it cost

`dBgCh_Gnd::SetObjAndPos` is the shape of it. The tree **defines** it taking `const Vector3 &`, `include/dBgCh_Gnd.h:79` spells it that way, and the ROM's own mangled name says `RK7Vector3` — yet twenty-two reconstructed `Vector3 *` shards outvoted all three, and the gate reported the definition's spelling as the error.

## The fix

Adopt the mark when it is unambiguous: exactly one `@symbol` line in the file, exactly one definition that went unnamed, and no definition already claiming that symbol. A file with two marks or two orphans still claims nothing, and an unmarked member definition still claims nothing — its linker name really is unrecoverable from the text.

Definitions go **9420 → 10699**.

## East const is west const

The gate compared type spellings as text, so `Vector3 const &` and `const Vector3 &` contradicted each other. They are the same reference. A `const` standing before the first `*`/`&` qualifies the base type and is hoisted; a `const` *after* a pointer token is left exactly where it is, because hoisting it would conflate genuinely different types — `T * const *` is a pointer to a const pointer to T, and that is not `const T **`. Both directions are tested.

## The rebank is not laundering anything

The 2290 disagreements this surfaces are pre-existing tree state that was simply invisible. Proof: this tree is **green under `origin/main`'s own revision of the gate** (`EXIT=0`, no FAIL block), so every new finding is newly *visible*, not newly *true*. No PR introduced them and none is being hidden.

The baseline shrinks 1158571 → 950075 bytes, because the same pass also drops 6689 banked entries whose disagreements are already gone. The committed blob is LF, matching main — no line-ending churn.

## Negative control

Injecting a bogus return type and an extra parameter into `src/func_0200fa04.c` reds the gate, with basis `definition` rather than `plurality`. Restored afterwards; no source file is touched by this PR.

## Tests

110 → 119. The six adoption tests cover the fix and both ambiguity guards; the three const tests cover the hoist, the pointer-level const that must *not* hoist, and `const` still contradicting non-`const`.

## Merge order

This should land **after** the current byte-gated batch, since the baseline is generated against `main`'s source and will need regenerating once that batch moves it. It also unblocks #2691, whose two remaining findings are exactly the two defects fixed here.


andrewboudreau (commented):
Reviewed `7c24bdf075489a870826277035ce34ab9205d839`. The 119 existing tests pass, but two independent controls expose blockers:

- **DECL-2715-01 (correctness), `tools/check_decl_agreement.py:450-458`:** the new const hoist ignores template nesting. `normalise_type('Box<const T> &', {}, False)` and `normalise_type('const Box<T> &', {}, False)` both return `const Box < T > &`. These are different specializations and different outer qualification. A definition taking `Box<const Vector3>&` and an extern taking `const Box<Vector3>&` produce zero `Take` findings through the existing `build()` test harness. Hoist only a qualifier belonging to the outer base type, preserving template-argument qualifiers, and add this negative control before regenerating the baseline.
- **DECL-2715-02 (correctness), `tools/check_decl_agreement.py:1217-1224`:** one marker plus one orphan is not sufficient symbol identity. Parsing the following as `src/pair.cpp` records an `int` definition of `_ZN1A1fEv`, although the only definition is `B::g`: 

```cpp
// @symbol _ZN1A1fEv
extern "C" void helper(void);
namespace B {
int g(void) { return 7; }
}
```

The fallback is not actually restricted to a verified single-function shard and never checks that the marker names the parsed declarator. Scope it to proven symbol/declarator identity; leave ambiguous or contradictory identity unclaimed. Include mismatched class/function, namespace helper, and declaration-only marker controls. A falsely attributed definition outranks valid declarations, so it can both fabricate errors and hide real disagreements.

This is a blocking review, not source acceptance. No candidate files were edited. Re-run the suite and regenerate/review the baseline after fixing the parser; current-main composition and the live Source review gate remain required before landing.

andrewboudreau (commented):
Rechecked the new head `2c305908d4be3479f7708a3d82fb717295735e3a`: `tools/check_decl_agreement.py` and its test file are unchanged from the reviewed `7c24bdf075489a870826277035ce34ab9205d839` (`git diff` is empty for both). **DECL-2715-01 and DECL-2715-02 remain unresolved.** The now-green Source review check classifies this tooling PR; it does not validate the parser's behavior or supersede the two reproduced correctness findings. Please address those controls before landing.

PR #2725, opened 2026-09-15
Call SharedFilePtr members instead of their mangled names in 26 files

andrewboudreau:
26 files reached `SharedFilePtr` by declaring its **mangled name as a free function** and calling that:

```c
extern void _ZN13SharedFilePtr7ReleaseEv(void *);
extern int data_ov062_0211ddf0[];
...
_ZN13SharedFilePtr7ReleaseEv(data_ov062_0211ddf0);
```

In a `.cpp` that is a defect, not a wart. The compiler mangles the identifier a **second** time and emits a reference to `_Z28_ZN13SharedFilePtr7ReleaseEvPv`, which nothing defines — so the file matches byte-for-byte and **can never be enrolled**. `include/SharedFilePtr.h` already exists and documents exactly this hazard; these files just were not using it.

Each file now includes the header, declares its data objects as `SharedFilePtr`, and calls the member:

```cpp
#include "SharedFilePtr.h"
extern SharedFilePtr data_ov062_0211ddf0;
...
data_ov062_0211ddf0.Release();
```

### Judgement calls

**Three files carried their own `struct SharedFilePtr`** alongside the header and had to drop it. The local bodies contradicted each other — `{unsigned int data[4];}` in one, `{u16 fileID; u8 numRefs; char *filePtr;}` in another — which is itself the proof that neither is load-bearing: both use the type only as an array element, where the stride is 4 either way.

**`dScMgJump` keeps a cast instead.** Its object is `extern void *data_ov006_02142184;` in `include/decl_common.h`, a *shared* header — retyping it there would reach the whole tree for one call site.

**`SharedFilePtr::Load` is declared `int`, not `void`.** Its only caller, `SharedFilePtr::LoadFile`, branches on the result (`if (!Load()) return 0;`), and the C form it was converted from declared `int`. A return type is not part of an Itanium mangled name, so no symbol changes.

### Bytes

| gate | result |
|---|---|
| per-file verify, 2004/b56 pin | **26 / 26 byte-exact** |
| `rombuild -j 16 --no-rom` | **265 files compiled** — the header reaches them all, so this is a real cold rebuild, not a `0 compiled` cache pass |
| source fidelity | 11,206 reproducing / **0 mismatching** |
| module fidelity | **106/106 exact**, ROM-build analysis **PASS** |
| `eligible.py`, bracketed | 11243 both sides, **0 lost, 0 gained** |
| `port_refcheck.py` | 423 references, 0 stale |

### CONVERTED

`tiers_ratchet --check`: baseline 2812, current 2842 — **+30**. All 26 files were failing `no_mangled_refs` at HEAD and are 5/5 now.

### Declaration agreement — read this bit

The baseline moves **58 rows in, 60 out — net −2**. Every added row is accounted for:

- **34** — these files' own new `SharedFilePtr` spelling. Deliberate and byte-proved above.
- **22** — plurality flips. Naming the true type made `SharedFilePtr` the plurality for that symbol, so the row banked on the *other* side was removed and a placeholder row took its place. **No untouched file's spelling changed.**
- **2** — `data_ov020_02114af0` and `data_ov080_02128468` are 1-1-1 ties where a third vote reordered the tie-break. Both shards disagreed with `include/decl_common.h` at HEAD exactly as much as they do now; they were the arbitrary winner before and are simply named now.

**UNEXPLAINED: 0.** These are not new facts about the tree — they are the same reconstructions, re-keyed now that one side spells the real type. The tree ends up banking two *fewer* disagreements than before, with 34 more declarations stating the actual type.

Worth noting for the gate itself: `_declines_to_answer` already carves out "promoting one caller to a real C++ type flips the plurality" for **param** findings. Data objects take the `return` path, which has no such carve-out — hence the 24 collateral rows. That is a gate gap, not a problem with this change, and I have deliberately not touched it here.


andrewboudreau:
**Merge-tree check: `main` + #2725 + #2726 + #2727 is green.**

Each of these PRs is gated alone, so green-alone does not imply green-merged. I built the three-way merge locally and ran the full set:

| gate | merge tree |
|---|---|
| `rombuild --no-rom` | `106/106 exact`, `mismatching: 0`, PASS (22 compiled) |
| `tiers_ratchet --check` | PASS, CONVERTED 2812 to 2887 (+75) |
| `check_decl_agreement` | no new disagreements; 76 banked ones disappear |
| `eligible.py` | name set identical, 11243 / 11302 |
| `port_refcheck` | 423 references, 0 stale |
| MSVC port smoke build | 230/230 from a clean configure, all five targets link |

File sets are disjoint across all three pairs, so there is no textual conflict either.

One thing specifically checked, since it would have been invisible otherwise: #2727 carries a `tu_manifest_resync` renumber for `ov023/daObjFm_Battan_c`, and mwccarm uniquifier shifts compose additively. #2725 edits `include/SharedFilePtr.h`, which battan also includes, so the merged tree could have emitted a third number that neither branch names. It does not -- `tu_manifest_resync --check` reports the manifest **in sync** on the merge tree. Merge order does not matter here.

andrewboudreau:
### Five-way merge tree gated (main + #2725 + #2726 + #2727 + #2728 + #2729)

Each of these five PRs is green on its own, but this repo has produced a red `main` from
two individually-green PRs before, so the set was gated as a merged tree rather than as
five branches.

**Disjointness.** 122 changed files across the five; all 10 pairs share **zero** files,
and the five branches merge onto `013370480` with no conflict.

**Cold rebuild** (`rombuild.py -j 16 --no-rom --no-cache`, all 8,760 enrolled files
compiled from scratch):

```
source-built functions: 11,206   reproducing: 11,206   mismatching: 0
ROM-build analysis: PASS
```

The rebuild is deliberately cold. A warm run on this same tree reported
`8759 reused from cache, 1 compiled` — the objects were already built on the individual
branches and the cache is content-keyed, so the compile step exercised nothing. `0`/`1
compiled` is this repo's classic false-green, so the numbers above come from a run with
no cache at all.

`tools/tu_manifest_resync.py --check ov023/daObjFm_Battan_c` reads **in sync** on the
merged tree — that is the manifest that shifted earlier in this batch when a shared
header gained declarations.

andrewboudreau:
## Adversarial review — seven-PR set (#2725–#2731)

Re-reviewed as a critic rather than as the author. Every check below is new work, not a
restatement of the PR's own gate. Head SHAs reviewed: 2725 `4f3927bd9`, 2726 `81040f19c`,
2727 `4c08d9a4d`, 2728 `00b15daef`, 2729 `fe61a984c`, 2730 `38aaa986b`, 2731 `e9b49b072`;
merge tree `a7f68c8b3` off `013370480`.

### 1. Relocation destinations — the hazard the byte gate cannot see

A relocated word is a wildcard: a `bl` retargeted to a **different but valid** symbol still
byte-matches. Every verification run on these PRs so far used `build_pin.verify` **without**
the `strict` tuple, so none of them checked this. Re-ran all of them with
`strict=(reloc_audit, name_index, config_relocs, sym_index)`, which resolves each relocation
and fails on `WRONG-DEST`:

| PR | files | result |
|---|---|---|
| 2725 | 26 | 26/26 OK |
| 2726 | 42 | 42/42 OK |
| 2727 | 27 + 4 promoted TUs (61 fns) | 27/27 + 61/61 OK |
| 2728 | 10 | 10/10 OK |
| 2729 | 7 | 7/7 OK |
| 2730 | 13 | 13/13 OK |
| 2731 | header-only | n/a |

**0 WRONG-DEST across 125 files and 61 TU functions.** This also closes the cast-receiver
question on #2726/#2728: `((X *)p)->X::M()` cast to the wrong class would resolve to a
different symbol, and that is exactly what this check would have caught.

### 2. Merge tree, cold

`mergetest/seven` = all seven merged onto `013370480`, 0 overlapping file pairs,
`rombuild.py -j16 --no-rom --no-cache` (the `[3/6]` line prints no cache line, so the
enrolled stage really did run cold):

```
reproducing: 11,206   mismatching: 0
source-owned data claims: 26  (reproducing 26, mismatching 0)
ROM-build analysis: PASS
```

### 3. Change-kind census

139 files changed in the merge tree, **all `M`** — zero adds, deletes or renames. The whole
rename-hazard family (stranded `port/` references, phantom file citations, stem collisions,
a stranded near-miss bank) is therefore structurally absent, not merely untriggered.

### 4. Local-only gates, run on the merge tree (CI runs none of these on a merge tree)

- `port_refcheck.py` — **423/423 references resolve, 0 stale** (manifests 249, cmake-symbols 21, hal-links 153).
- `check_decl_agreement.py` — **no new declaration disagreements**; 80 banked ones are now gone.
- `langmode_audit.py --check` (baseline from `chaos-data`) — **PASS**. `local struct body` 1676 → 1660.
- `check_decl_return_types.py` — output **byte-identical to `origin/main`**, so its 3 conflicting
  symbols are pre-existing and not introduced here.

### 5. The merged decl baseline is not a silent clobber

`config/decl-agreement-baseline.json` takes edits from several of these branches and git merges
it line-wise, which can silently drop meaning. Regenerated it from scratch on the merge tree and
compared structurally against the merged copy:

```
merged symbols: 3891   regenerated: 3887
in MERGED but not regenerated (stale/banked, harmless):  4
in REGENERATED but not merged (a real regression the merge hid):  0
```

**0 hidden.** Every difference is in the safe direction. (Separately: ~84 stale banked entries
could be pruned — worth its own PR, not worth touching here, since a baseline edit would conflict
across all seven branches.)

### 6. The MSVC port build — differential, not absolute

`main`'s port build is red and **no CI job builds it**, so "the port is red" proves nothing about
these PRs on its own. Made it a differential instead.

- `main` (`013370480`) dies at **compile**: `include/fBase_c.h(158): error C2821`. It never reaches the link stage.
- **#2731 alone fixes that** — `main + #2731` compiles clean and gets all the way to link.
- Compared `main + #2731` against the full seven-way merge tree:

```
LNK2001: 106    LNK2019: 30    LNK2005: 22    LNK1120: 9      <- identical in both
distinct failing symbols: 8 vs 8, set difference in both directions: EMPTY
full LNK line sets: IDENTICAL
```

**The other six PRs introduce zero new port link errors.** What remains is the pre-existing
`Model` / `ModelBase` / `ShadowModel` `Destructor0`/`Destructor1` C1/C2 collapse, which needs its
own `port/hal` change and is untouched by this set.

### 7. Per-PR findings

- **#2727** renumbers two compiler-generated anonymous symbols in ov023 (`@348`→`@357`,
  `@347`→`@356`) across `symbols.txt` and the TU manifest. This looked like unexplained config
  churn, so I ran the positive control — reverted just the renumber on top of the merge tree and
  rebuilt. The build **refuses**: `binding rewrite symbol @348 has 0 entries`. The renumber is
  load-bearing and correct: adding `Sound.h` shifts mwcc's anonymous counter by 9.
- **#2727** `include/Sound.h` declares 14 entry points. Checked every return type against its
  **definition** in `src/` (not against a caller, which evidences nothing): **14/14 match**, and
  each parameter list agrees with the mangled symbol it comes from.
- **#2725** `SharedFilePtr::Load` return-type fix `void` → `void *` is correct:
  `src/_ZN13SharedFilePtr4LoadEv.cpp:43` is `void *SharedFilePtr::Load()` returning `filePtr`,
  and `port/hal/fs.cpp` documents the same contract.
- **#2726/#2728** remove 153 now-dead `extern` declarations and use the scope-qualified call form,
  which is the spelling that suppresses virtual dispatch. No half-converted file.
- **Convention check:** these PRs insert `#include` *above* the `// @symbol` metadata line. Every
  `@symbol` parser in `tools/` (`check_decl_agreement`, `ovsweep`, `reconcile_names`, `tiers`,
  `tubuild`) is a whole-file `MULTILINE` regex, so nothing parses it positionally. Safe.

### Verdict

No defect found, and **no change requested**. Both incomplete migrations are incomplete on
purpose and say so in their titles (#2725 converts 26 of ~920 references; #2727 lands the header
ahead of the bulk of the call-site conversion) — that is slicing, not a gap.

The one thing this review **cannot** supply is the thing the review gate actually wants: I am the
author of these PRs, so none of the above is independent. All seven still show
`reviewDecision: <empty>` and need a human to look at them.

PR #2731, opened 2026-09-16
Declare the last four operator new overloads with size_t

andrewboudreau:
`port/build-port.cmd` has not compiled since `010b269ef` (2026-09-14). On a clean
`origin/main` with nothing modified, 32-bit MSVC rejects `include/fBase_c.h:158`:

```
error C2821: first formal parameter to 'operator new' must be 'size_t'
error C2333: 'fBase_c::operator new': error in function declaration; skipping function body
```

No CI job builds the port, so nothing went red.

### The rule is on the type, not the width

All three candidate spellings are four bytes on this target. mwccarm rejects
`operator new(unsigned int)` as an illegal operator declaration; 32-bit MSVC rejects
`operator new(unsigned long)`. `include/types.h` already carries the resolution — a
conditional `size_t`, `<stddef.h>` under `_MSC_VER` and `typedef unsigned long`
otherwise — added for exactly this, with a comment noting that no ROM byte moves because
mwccarm keeps the `unsigned long` the actor headers declare.

Most headers already declare `operator new(size_t)`. **Four were never converted** and
still name the literal type: `fBase_c.h`, `daBar_c.h`, `daObjTh_Fall_Block_c.h`,
`daTBasket_c.h`. This changes those four.

### Byte-neutrality is measured, not asserted

`size_t` *is* `unsigned long` for mwccarm, so the declaration is token-different and
type-identical. The build recompiles every translation unit that reaches these headers:

```
rombuild.py -j 16 --no-rom
  6797 reused from cache, 1963 compiled
  source-built functions: 11,206   reproducing: 11,206   mismatching: 0
  ROM-build analysis: PASS
```

1,963 real compiles — this is exercised, not a cache hit.

### This does not make the port green

It clears the compile error. After it, four of fourteen smoke targets build and pass
(`smoke`, `smoke_heap`, `smoke_roots`, `smoke_gx`, plus `smoke_fs`) and **nine fail to
link**, for a reason unrelated to `operator new`:

```
_ZN5ModelC2Ev.cpp.obj : error LNK2005: "Model::Model(void)" already defined in _ZN5ModelC1Ev.cpp.obj
ctor_bridge.cpp.obj   : error LNK2001: unresolved external symbol "ModelBase::Destructor1(void)"
```

In the Itanium ABI `C1` and `C2` are distinct symbols; compiled as real C++ by MSVC both
collapse to one `??0Model@@QAE@XZ`, so the two files collide. Separately, the port's
bridges reference `ModelBase::Destructor0`/`Destructor1`, which nothing defines. Those are
C++-migration seams in `port/hal` and need their own change.

Worth knowing when reading the output: ctest reports those nine as failing tests, but they
are `***Not Run` — the executables never linked, so no assertion ran.


andrewboudreau:
## Adversarial review — seven-PR set (#2725–#2731)

Re-reviewed as a critic rather than as the author. Every check below is new work, not a
restatement of the PR's own gate. Head SHAs reviewed: 2725 `4f3927bd9`, 2726 `81040f19c`,
2727 `4c08d9a4d`, 2728 `00b15daef`, 2729 `fe61a984c`, 2730 `38aaa986b`, 2731 `e9b49b072`;
merge tree `a7f68c8b3` off `013370480`.

### 1. Relocation destinations — the hazard the byte gate cannot see

A relocated word is a wildcard: a `bl` retargeted to a **different but valid** symbol still
byte-matches. Every verification run on these PRs so far used `build_pin.verify` **without**
the `strict` tuple, so none of them checked this. Re-ran all of them with
`strict=(reloc_audit, name_index, config_relocs, sym_index)`, which resolves each relocation
and fails on `WRONG-DEST`:

| PR | files | result |
|---|---|---|
| 2725 | 26 | 26/26 OK |
| 2726 | 42 | 42/42 OK |
| 2727 | 27 + 4 promoted TUs (61 fns) | 27/27 + 61/61 OK |
| 2728 | 10 | 10/10 OK |
| 2729 | 7 | 7/7 OK |
| 2730 | 13 | 13/13 OK |
| 2731 | header-only | n/a |

**0 WRONG-DEST across 125 files and 61 TU functions.** This also closes the cast-receiver
question on #2726/#2728: `((X *)p)->X::M()` cast to the wrong class would resolve to a
different symbol, and that is exactly what this check would have caught.

### 2. Merge tree, cold

`mergetest/seven` = all seven merged onto `013370480`, 0 overlapping file pairs,
`rombuild.py -j16 --no-rom --no-cache` (the `[3/6]` line prints no cache line, so the
enrolled stage really did run cold):

```
reproducing: 11,206   mismatching: 0
source-owned data claims: 26  (reproducing 26, mismatching 0)
ROM-build analysis: PASS
```

### 3. Change-kind census

139 files changed in the merge tree, **all `M`** — zero adds, deletes or renames. The whole
rename-hazard family (stranded `port/` references, phantom file citations, stem collisions,
a stranded near-miss bank) is therefore structurally absent, not merely untriggered.

### 4. Local-only gates, run on the merge tree (CI runs none of these on a merge tree)

- `port_refcheck.py` — **423/423 references resolve, 0 stale** (manifests 249, cmake-symbols 21, hal-links 153).
- `check_decl_agreement.py` — **no new declaration disagreements**; 80 banked ones are now gone.
- `langmode_audit.py --check` (baseline from `chaos-data`) — **PASS**. `local struct body` 1676 → 1660.
- `check_decl_return_types.py` — output **byte-identical to `origin/main`**, so its 3 conflicting
  symbols are pre-existing and not introduced here.

### 5. The merged decl baseline is not a silent clobber

`config/decl-agreement-baseline.json` takes edits from several of these branches and git merges
it line-wise, which can silently drop meaning. Regenerated it from scratch on the merge tree and
compared structurally against the merged copy:

```
merged symbols: 3891   regenerated: 3887
in MERGED but not regenerated (stale/banked, harmless):  4
in REGENERATED but not merged (a real regression the merge hid):  0
```

**0 hidden.** Every difference is in the safe direction. (Separately: ~84 stale banked entries
could be pruned — worth its own PR, not worth touching here, since a baseline edit would conflict
across all seven branches.)

### 6. The MSVC port build — differential, not absolute

`main`'s port build is red and **no CI job builds it**, so "the port is red" proves nothing about
these PRs on its own. Made it a differential instead.

- `main` (`013370480`) dies at **compile**: `include/fBase_c.h(158): error C2821`. It never reaches the link stage.
- **#2731 alone fixes that** — `main + #2731` compiles clean and gets all the way to link.
- Compared `main + #2731` against the full seven-way merge tree:

```
LNK2001: 106    LNK2019: 30    LNK2005: 22    LNK1120: 9      <- identical in both
distinct failing symbols: 8 vs 8, set difference in both directions: EMPTY
full LNK line sets: IDENTICAL
```

**The other six PRs introduce zero new port link errors.** What remains is the pre-existing
`Model` / `ModelBase` / `ShadowModel` `Destructor0`/`Destructor1` C1/C2 collapse, which needs its
own `port/hal` change and is untouched by this set.

### 7. Per-PR findings

- **#2727** renumbers two compiler-generated anonymous symbols in ov023 (`@348`→`@357`,
  `@347`→`@356`) across `symbols.txt` and the TU manifest. This looked like unexplained config
  churn, so I ran the positive control — reverted just the renumber on top of the merge tree and
  rebuilt. The build **refuses**: `binding rewrite symbol @348 has 0 entries`. The renumber is
  load-bearing and correct: adding `Sound.h` shifts mwcc's anonymous counter by 9.
- **#2727** `include/Sound.h` declares 14 entry points. Checked every return type against its
  **definition** in `src/` (not against a caller, which evidences nothing): **14/14 match**, and
  each parameter list agrees with the mangled symbol it comes from.
- **#2725** `SharedFilePtr::Load` return-type fix `void` → `void *` is correct:
  `src/_ZN13SharedFilePtr4LoadEv.cpp:43` is `void *SharedFilePtr::Load()` returning `filePtr`,
  and `port/hal/fs.cpp` documents the same contract.
- **#2726/#2728** remove 153 now-dead `extern` declarations and use the scope-qualified call form,
  which is the spelling that suppresses virtual dispatch. No half-converted file.
- **Convention check:** these PRs insert `#include` *above* the `// @symbol` metadata line. Every
  `@symbol` parser in `tools/` (`check_decl_agreement`, `ovsweep`, `reconcile_names`, `tiers`,
  `tubuild`) is a whole-file `MULTILINE` regex, so nothing parses it positionally. Safe.

### Verdict

No defect found, and **no change requested**. Both incomplete migrations are incomplete on
purpose and say so in their titles (#2725 converts 26 of ~920 references; #2727 lands the header
ahead of the bulk of the call-site conversion) — that is slicing, not a gap.

The one thing this review **cannot** supply is the thing the review gate actually wants: I am the
author of these PRs, so none of the above is independent. All seven still show
`reviewDecision: <empty>` and need a human to look at them.

PR #2747, opened 2026-09-17
Cleanup ov009/daSBird_c: promote Birds class TU

lunavyqo:
Rename the coined ov009 `Bird` class to the cartridge `_ZTS9daSBird_c` name, then promote the 13-function run to one leftover-quality TU.

- `class_rename.py --class Bird --to daSBird_c --with-derived --apply` (coined `_ZTV4Bird` alias deleted first; `_ZTV9daSBird_c` stays at `0x02113958`)
- dest `src/game/actors/d_a_s_bird.cpp`; inline destructor; `return new daSBird_c()`
- `extern int _ZTV9daSBird_c[]` once in the TU above the factory; stale `_ZTV4Bird` dropped from `decl_common.h`
- CONVERTED identities rewritten onto `dest#symbol` (no `--update` of that section)
- unenrolled `src/func_ov009_0211145c.c` kept for first-matcher credit; production bytes come from the dest TU hatch

`tubuild.py verify ov009/daSBird_c`: 13/13 MATCH, objisolate clean, reloc-destinations clean.


andrewboudreau (changes requested):
At head `5ee11c6b9099a25282bda8d9489fec7d994b6d2a`, the promoted TU includes `func_ov009_0211145c` as ARM inline assembly while the standalone implementation is still a draft. The reported increase of one reconstructed function and 380 source-built code bytes therefore does not establish a new C++ match. Moving or removing the interior `NONMATCHING` marker alone would make that accounting problem harder to see; it would not repair it.

The claimed compiler wall also needs reproducible pinned-compiler experiment artifacts, or wording that accurately leaves the C++ reconstruction unresolved. Complete text equality is useful byte evidence, but does not establish source reconstruction of the assembly helper.

I am repairing this under queue task `pr2747-repair-0918`, with a separate verifier to follow. Typed fields and real actor method calls already reproduce the other twelve functions in local experiments. The remaining helper is being tested as actual C++; if it cannot be matched in this repair, its draft status and coverage must stay honest. Please keep this PR unmerged until the repaired exact head passes independent source review and private validation.

Reviewer session: `codex-integrator-0918`. This review identifies a blocker; it is not acceptance of the remaining whole-file reconstruction.

andrewboudreau:
## Why this cannot land in its current shape

The `ratchet` failure is fixed (a duplicated `func_ov009_0211145c.c` was a second
copy of the same `asm { ... }` hatch already in the promoted TU, and
`langmode_audit` counts inline-asm *files*, so it read 13 -> 14). That file was
never in `config/arm9/overlays/ov009/delinks.txt`, and `matched` is decided by the
delinks entry that owns the address, not by filename, so the copy bought no
accounting at all. It is deleted. `langmode ratchet PASS`, and both ratchet runs
plus `review` are green.

`PR validation` is a different matter, and it is correct to refuse this:

```
Validation failed: 1 file(s) say NONMATCHING outside the leading comment block,
where no counter or gate recognizes it -- move the marker into the file header,
or reword prose that quotes the bare word: src/game/actors/d_a_s_bird.cpp
```

The validator's own table names the real problem: **byte-verified +1 function,
+380 bytes**. 380 = 0x17c = exactly the size of `func_ov009_0211145c`. The other
12 functions in this TU were already byte-verified on main, so **the entire
coverage gain of this PR is one hand-written-assembly function.**

Neither remedy the message offers is available:

- *Move the marker into the file header* -- all 13 functions leave `matched`. The
  12 that main had were `base_enrolled` under different `srcPath`s, so the
  WITHDRAWN exception does not apply and they fall into `removed`: hard reason
  "lost 12 matched function(s)". Blocked.
- *Reword the prose* -- the file then counts as matched, and the project banks a
  new match and 380 new byte-verified bytes for hand-written assembly. That is
  the inversion the gate exists to catch.

There is no structural third way. `dsd lcf` emits exactly one `<object>.o(.text)`
placement per delinks entry (verified in the generated `build/arm9.lcf`), and a
census of `config/arm9/**` finds zero entries owning two `.text` ranges. A
promoted TU's range must be contiguous, and a contiguous `0x021111a0..0x02111a70`
must own the hatch.

## The match was attempted, and the wall reproduces

A typed-member C++ spelling (real `daSBird_c *`, `c->mTargetPos.x/.z`,
`c->mCamSpacePosX`, `c->mFlags`) compiles size-exact 0x17c and reproduces the
whole function except one 6-instruction window -- **5 differing words**:

```
   +0xe0 | 001060e2 rsb r1, r0, #0        | 000060e2 rsb r0, r0, #0
   +0xe4 | 60118ae5 str r1, [sl, #0x160]  | 7c209fe5 ldr r2, [pc, #0x7c]
   +0xe8 | 78009fe5 ldr r0, [pc, #0x78]   | 60018ae5 str r0, [sl, #0x160]
   +0xf0 | 68018ae5 str r0, [sl, #0x168]  | 6a00a0e3 mov r0, #0x6a
   +0xf4 | 6a00a0e3 mov r0, #0x6a         | 68218ae5 str r2, [sl, #0x168]
```

A sweep over all 25 compiler versions gives `2004/b56: 5 word(s) differ` and
`999` everywhere else, so the pin is right. Eleven further tail-block variants on
the typed base -- including `volatile` as a free scheduling barrier on the `.x`
store, the `.z` store, both, and combined with a named argument pointer -- all
score 5, one scores 6, two fail to compile. Nothing goes below 5.

This reproduces the documented wall (2004/b56 will not let the `0xff06a000` pool
constant reach `r0` while `r0` is also an outgoing immediate argument) and
extends it to the typed-field form.

## The two landable routes

1. **Match `func_ov009_0211145c` from real C++.** The asm block, the stranded
   marker, the inline-asm ratchet pressure and the misreported +1/+380 all
   resolve at once, and this promotion becomes an honest +1.
2. **Withdraw the promotion** until (1) exists, keeping main's 12 per-function
   entries and the unenrolled draft.

Pushed `729ae940521d29e44834e88a3043a938a04821dc` with the ratchet fix and a
clean merge of main. Full tree validates: 8744/8744 compiled cold, reproducing
11,207, mismatching 0, module fidelity 106/106 exact at 100.000000%,
ROM-build analysis PASS.

andrewboudreau:
Not merging this, and not touching the interior marker — the review is right that
rewording it alone would only hide the accounting question. Queue task
`pr2747-repair-0918` reads `phase=running` against input `5ee11c6b9`, so this stays
reserved to that session.

Handing over the pinned-compiler evidence the review asked for, because it bears
directly on whether the C++ reconstruction of `func_ov009_0211145c` is reachable.
I ran an independent attempt at it (mwccarm 1.2/2004/b56, same flags as the TU) and
**it refuted the reconstruction** rather than supporting it.

**What was measured.** A 95-word census over the function's store/call tail. The best
result was 3 of 95 words differing — which is exactly the record already banked for
this function, not an improvement on it. Roughly 100 source variants of the tail were
byte-inert: they did not move a single word.

**The mechanism, and it is falsifiable.** mwccarm assigns a scratch temp live across a
call the lowest free **non-argument** register, and only falls back to `r0` under
register pressure. Holding the tail fixed and varying only the outgoing call's arity
reproduces a clean ladder:

| outgoing args | register chosen for the temp |
|---|---|
| 2 | `r2` |
| 3 | `r3` |
| 4 | `ip` |
| 5 | `r0` |

The ROM's shape is the 5-argument row. So `r0` here is not a scheduling accident that
a better spelling can dodge — it is what the allocator does once the argument registers
are consumed, and the sibling `021116ec` colours the same constant differently only
because nothing is live across a call there. That is the claimed wall, reproduced
rather than asserted.

**Scope, so this is not over-read.** This is evidence about *this* tail under *these*
flags. It does not prove no C++ spelling exists. Separately: any `#pragma` results from
that run should be treated as **liveness only** — mwccarm is silent on unknown pragmas,
and the existence probe used to test them is disputed, so nothing there should be cited
as a positive finding.

Taken with the review's own condition — "if it cannot be matched in this repair, its
draft status and coverage must stay honest" — the honest reading of this measurement is
that the helper's draft status should stand and the TU should not claim a new C++ match
over that range.

andrewboudreau (changes requested):
Rechecked the updated head `729ae940521d29e44834e88a3043a938a04821dc`. Removing the duplicate standalone hatch addresses the duplicate-assembly-file ratchet, but the source-accounting blocker remains: the promoted TU still enrolls the 380-byte inline-assembly `func_ov009_0211145c` inside its complete text span. Its interior NONMATCHING comment does not exclude it from reconstructed-source coverage. The statement that it does not count as matched C still conflicts with that enrollment.

The finite compiler experiments establish a remaining five-word difference for ordinary typed C++, not an absolute compiler wall. I have prepared a local repair proposal that preserves the RTTI rename and twelve byte/relocation-verified C++ functions (1876 bytes), while keeping the unresolved 380-byte helper out of reconstructed-source enrollment. It requires coordinated attribution/declaration reservations before application and independent review afterward.

Please keep the promotion unmerged until the accounting defect is repaired. A byte-exact assembly hatch does not by itself satisfy this repository's reconstructed-C++ requirement. Queue task `pr2747-repair-0918`; reviewer session `codex-integrator-0918`. This is a current-head blocker review, not source acceptance.

andrewboudreau:
Diagnosed both red checks on `dd05262464b8`. Neither is a problem with the decompilation — the accounting in this PR is honest — but both need a push, and I cannot push to a fork branch.

## `PR validation` — one word, in the wrong place

```
1 file(s) say NONMATCHING outside the leading comment block, where no
counter or gate recognizes it -- move the marker into the file header,
or reword prose that quotes the bare word:
  src/game/actors/d_a_s_bird.cpp
```

It is line 223, the per-function banner above `func_ov009_0211145c`, which has an `asm` block at line 266.

**Take the reword, not the move.** The gate offers both and they are not equivalent here:

- Moving the marker into the header marks the **whole file** NONMATCHING, which un-counts five members that really are matched C. That is a worse lie than the one being fixed.
- Rewording costs nothing, because the disclosure does not live in that comment. It lives in two places that already say the right thing:
  - the header, lines 28–30: *"a proven mwccarm 1.2/2004/b56 register-allocation wall … Byte-exact via an asm hatch; does not count as matched C"*
  - `config/converted-baseline.json`, which banks exactly five identities from this file and **does not bank `func_ov009_0211145c`**:
    ```
    _ZN9daSBird_c16CleanupResourcesEv
    _ZN9daSBird_c16OnPendingDestroyEv
    _ZN9daSBird_c6RenderEv
    _ZN9daSBird_cD0Ev
    _ZN9daSBird_cD1Ev
    ```

I checked that omission specifically, because the usual way this failure gets "fixed" is by deleting the word and leaving an asm-hatched function sitting inside a range something counts as complete. That is not what is happening here — the hatch is excluded from the baseline, so rewording the prose changes no accounting at all. Keep the sentence, drop the bare token: *"not a C decompilation; byte-exact via an asm hatch"* reads the same to a human and stops the grep.

Worth keeping the per-function banner rather than relying on the header alone. It is the thing a reader hits at the point of confusion.

While I was in the file: every member defined in this file carries a `// @symbol` marker — eleven of them, including the unbanked `func_ov009_*` helpers. That is the right shape — an unmarked member scores against the whole file, so one raw offset would fail every banked member of the TU, and it fails on `main` after the merge rather than here.

## `Source review` — the base, not the content

```
PR head does not contain the live target base; restack and review the composition
  base      396b1a4002
  live tip  7015f924ec
  pr_base   3e24f2d0e7
```

Advisory, not required — only `PR validation` blocks the merge — but the PR is also `DIRTY`, so the merge needs a settle regardless. Merge `main` in rather than rebasing, so the verified commit survives as a parent with its bytes intact.

Fair warning on timing: three separate gates each demand the live tip of `main`, and `main` moves every few minutes while a full run takes over an hour, so currency expires several times per run. Resolve and push in one go during a quiet window rather than settling and then waiting.

Happy to take this over on a branch in this repo if you would rather not chase it — say the word and I will open a successor with your commits and your authorship preserved.

andrewboudreau:
**The only thing failing here is one word, at `src/game/actors/d_a_s_bird.cpp:223`.** `PR validation` is otherwise a clean pass — 106/106 modules exact, 1787/1793 relocations verified, byte-verified functions +1, contributor credit +1/0/0.

The validator's message:

> 1 file(s) say NONMATCHING outside the leading comment block, where no counter or gate recognizes it -- move the marker into the file header, or reword prose that quotes the bare word

### Why neither offered fix is obviously right here

The token is at line 223, immediately under `// @symbol func_ov009_0211145c`, and it is a **deliberate per-member disclosure**, not an accident:

```c
// @symbol func_ov009_0211145c
/* NONMATCHING: hand-written asm, not a C decompilation. Byte-exact via an
 * asm hatch on a proven mwccarm register-allocation/scheduling wall; does
 * NOT count as matched. ... */
```

That address sits inside a range `config/arm9/overlays/ov009/delinks.txt` marks `complete`:

```
src/game/actors/d_a_s_bird.cpp:
    complete
    .text start:0x021111a0 end:0x02111a70     <- 0x0211145c is inside this
```

So the two suggested fixes are not equivalent:

* **Move the marker into the header** — the banner is file-scoped, so this demotes all of `d_a_s_bird.cpp`, not the one member. It gives up the credit for the functions that *are* real C.
* **Reword the prose** — satisfies the gate, but on its own it removes the only written notice that one function in a `complete` range is fed by an asm hatch.

### What I'd suggest

**Reword, but keep the disclosure** — drop only the bare machine-read token, keep every word of the explanation. The token is what the counter scans for; the sentence is what a reader needs. Something like:

```c
// @symbol func_ov009_0211145c
/* HAND-WRITTEN ASM, not a C decompilation. Byte-exact via an asm hatch on a
 * proven mwccarm register-allocation/scheduling wall. The bytes reproduce, but
 * this member is not decompiled -- it should not be read as matched C, and it
 * is a draft until someone reproduces the bytes from real C.
 *
 * Store/call tail: ... (unchanged)
 */
```

That is a one-line edit, it keeps your `+1` byte-verified, and it leaves the honest notice in place for the next reader.

Worth saying plainly, since it is not your bug: **the repo has no per-member spelling of this disclosure that any counter understands.** `NONMATCHING` is file-scoped, and a folded multi-symbol TU has no way to say "this one member is an asm hatch" in a form a gate will read. That gap is what put a correct disclosure in front of a red X. It probably deserves its own issue rather than being absorbed into this PR.

I have not pushed anything to your branch.

andrewboudreau:
Heads-up, and sorry for the churn — this PR went DIRTY since my last comment, and it is not anything you did.

`git merge-tree` against the current main says the conflict set is exactly one file:

```
config/decl-agreement-baseline.json
```

Essentially every rename, fold and promotion PR edits that one file, so they land strictly one at a time and each merge re-dirties every other open PR. Several landed upstream while this was sitting.

### One hazard when you resolve it

The obvious resolution is the wrong one. **Do not union-merge that baseline, and do not hand-resolve its hunks.** It is a generated keyed set — union-merging it produces entries that contradict each other, and a hand-resolve silently drops whichever of your edits fell outside the conflict region. Regenerate it instead:

```
python tools/check_decl_agreement.py --update
```

Then judge the result on the **gain** side, compared path-blind: extract `(kind, symbol, finding)` triples with the declaring-file key dropped, from the committed file and from yours, and diff those sets. A fold re-keys paths wholesale, so a raw line diff is almost all noise.

* Every **gain** should be your own fold's new spelling. A gain naming a class you never touched means the base is still stale — rebase onto the live main and regenerate.
* **Losses** need no action. They are healed entries the tool itself asks you to prune; `healed = known - findings` by construction.

Worth knowing: `--check` run straight after `--update` proves nothing either way, since it just re-reads what `--update` wrote.

### Otherwise unchanged

The `NONMATCHING` word at `src/game/actors/d_a_s_bird.cpp:223` is still the only other thing failing, and the suggestion in my previous comment stands — reword the prose to drop the bare machine-read token while keeping every word of the disclosure.

I have not pushed anything to your branch.

andrewboudreau:
## Measured: `func_ov009_0211145c` does NOT reproduce from C

Settled by measurement in a fresh worktree at head `93dab4685c08781f22137ab84fc30f2e1d01abbd`, pinned toolchain `2004/b56`, object rebuilt from scratch each run.

**1. The TU as written byte-matches.** `tubuild.py verify ov009/daSBird_c` → 13/13 MATCH, objisolate clean, relocation destinations clean.

**2. Positive control — the compared bytes really come from this source.** Perturbing one unrelated constant inside `func_ov009_0211145c` (`0x7d0000` → `0x7d0004`) flips exactly that member to DIFF and leaves the other twelve MATCH. So the green verify is not a cached or stock object.

**3. The inline-asm tail is load-bearing.** Replacing the `asm { }` block with the plain C it spells —

```c
*(int *)(c + 0x160) = -0x14000;
*(int *)(c + 0x168) = 0xff06a000;
func_0201267c(0x6a, c + 0x74);
```

— is **size-exact** at `0x17c` and differs in **five words** at `+0xe0..+0xf4`:

```
 +0xe0 | 001060e2 rsb r1, r0, #0        | 000060e2 rsb r0, r0, #0        <<< MISMATCH
 +0xe4 | 60118ae5 str r1, [sl, #0x160]  | 7c209fe5 ldr r2, [pc, #0x7c]   <<< MISMATCH
 +0xe8 | 78009fe5 ldr r0, [pc, #0x78]   | 60018ae5 str r0, [sl, #0x160]  <<< MISMATCH
 +0xf0 | 68018ae5 str r0, [sl, #0x168]  | 6a00a0e3 mov r0, #0x6a         <<< MISMATCH
 +0xf4 | 6a00a0e3 mov r0, #0x6a         | 68218ae5 str r2, [sl, #0x168]  <<< MISMATCH
```

(left = ROM, right = compiled C). **Verdict: the 380 bytes are byte-exact hand-written assembly, not reconstructed C.** The blocker stands on the substance.

## Correction: the previous commit did not change the count

`93dab4685` ("Drop the file-scoped NONMATCHING token") was not the point where the 380 bytes became countable. `asm_policy` reads **one banner per file, from the header region only**. Measured on the three relevant blobs:

| state | `counts_as_matched` | `classify` |
|---|---|---|
| helper in its own one-function source (pre-fold) | **False** | `None` |
| folded TU, parent `93cab6000` (token at line 223) | **True** | `None` |
| folded TU, head `93dab4685` (token removed) | **True** | `unbannered-asm` |

The token at line 223 was inside the function body, past the header region (which ends at line 44), so it never demoted anything. `93dab4685` only deleted the last textual trace and dropped the file into the warn tier.

**The bytes entered the counted set at the fold.** While the helper had its own source file, that same banner was in the header region and excluded exactly its 380 bytes. Folding it into a 13-member TU moved the banner out of header scope. This is the only multi-member asm-bearing source in `src/` (11 `@symbol` markers, asm body); every other asm-bearing source is a one-function file where file scope and member scope coincide — which is why the mechanism has no per-member demotion.

So reverting the reword alone would have restored the disclosure and left the count wrong. Exactly as predicted in the second review.

## What this push does

Head is now `736502761`. It reverts `93dab4685` on that file **and** puts the banner in the header region where the counting tools read it, with the measurement recorded inline. `counts_as_matched` is now **False**.

This is deliberately conservative: it demotes all 13 members (2256 bytes), not only the unresolved 380, because there is no per-member demotion. The narrower repair described in the second review — keep the twelve byte/relocation-verified C++ members (1876 bytes), take the helper out of the TU's enrolled text span — needs `delinks.txt` / symbol-range edits that were out of scope for this pass and is the right follow-up.

Nothing else was touched: no `symbols.txt`, no `delinks.txt`, no `actor_renames.tsv`, no manifest edit, no baseline or ratchet `--update`.

Gates at `736502761`: `check_src_tu` OK · `check_tubuild_conflicts --list` 0 problems · `check_dead_references` no new dead references · `check_decl_agreement` no new disagreements · `tiers_ratchet --check` PASS · `eligible.py` 11240/11289, stamp commit equals HEAD. Pre-push hooks ran with the reference check performed (not skipped).

andrewboudreau:
## The validation failure at `0a50c7dab` is the RTTI ancestor chain, and it is fully priced

Unenrolling the 380-byte hatch was the right repair, and it is what makes this TU the key
function for `daSBird_c`. That is also what the build is now failing on:

```
FAIL src/game/actors/d_a_s_bird.cpp: isolate: compiler-only deadstrip refused:
compiler-only symbol(s) not defined:
['_ZTI7dBase_c', '_ZTI7fBase_c', '_ZTI8dActor_c', '_ZTI9daSBird_c',
 '_ZTS7dBase_c', '_ZTS7fBase_c', '_ZTS8dActor_c', '_ZTS9daSBird_c', '_ZTV9daSBird_c']
```

`_ZTI<Class>` for single non-virtual inheritance is an `__si_class_type_info` whose third
word points at `_ZTI<Base>`. mwccarm cannot emit a pointer to a record it does not have, so
emitting the derived record forces the base's, and so on to the root. Each class contributes
two symbols (`_ZTI` + `_ZTS`), plus one `_ZTV` for the class itself.

`daSBird_c -> dActor_c -> dBase_c -> fBase_c` is four classes, so the bill is
`2 * 4 + 1 = 9` `deadstrip-data` rows — exactly the nine symbols the linker listed.

### The nine rows, with canonical homes resolved across all symbol files

Note the homes span two modules; a grep confined to `ov009` finds only three of the nine.

| symbol | canonical_module | canonical_address |
|---|---|---|
| `_ZTV9daSBird_c` | `ov009` | `0x02113958` |
| `_ZTI9daSBird_c` | `ov009` | `0x02113928` |
| `_ZTS9daSBird_c` | `ov009` | `0x0211391c` |
| `_ZTI8dActor_c` | `arm9` | `0x0208e390` |
| `_ZTS8dActor_c` | `arm9` | `0x0208e384` |
| `_ZTI7dBase_c` | `arm9` | `0x02086e78` |
| `_ZTS7dBase_c` | `arm9` | `0x02086e54` |
| `_ZTI7fBase_c` | `arm9` | `0x02086d70` |
| `_ZTS7fBase_c` | `arm9` | `0x02086e60` |

The six `arm9` addresses are identical to the ones already landed for `ov085/daObjKanban_c`
in #2843, which shares the `dActor_c -> dBase_c -> fBase_c` tail — so they are
cross-checked against a merged precedent, not derived only here.

### Why these are admissible as `deadstrip-data`

A data row passes only when the canonical home lies outside every range the entry claims.
Each ancestor home trivially does, belonging to another class in another module, and this
class's own RTTI sits above its licensed `.text` span. #2843 is the worked precedent for the
whole shape, including the per-row `reason` prose naming which ancestor and which address.

One detail worth copying from it: `_ZTV` is recorded at the **address point**, eight bytes
above where the vtable storage begins, because the two-word ABI preamble (offset-to-top,
typeinfo pointer) sits below it. `0x02113958` above is already the address point.

### Scope

This is additive to `config/tu_manifest.d/ov009/daSBird_c.json` only — no `delinks.txt`,
`symbols.txt`, rename-ledger or baseline edits, and nothing about the hatch decision
changes. I am not pushing it, since this branch has in-flight work; posting the measurement
so whoever picks it up does not have to re-derive the chain or re-resolve the homes.

andrewboudreau:
The re-validation on `f6660d4a7` fails with the **same nine symbols** as before the fix commit, and that is the tell: the commit reordered and re-worded the `compiler_only_output` rows but did not change which symbols are requested, so the refusal could not move.

```
FAIL src/game/actors/d_a_s_bird.cpp: isolate: compiler-only deadstrip refused:
compiler-only symbol(s) not defined:
['_ZTI7dBase_c','_ZTI7fBase_c','_ZTI8dActor_c','_ZTI9daSBird_c',
 '_ZTS7dBase_c','_ZTS7fBase_c','_ZTS8dActor_c','_ZTS9daSBird_c','_ZTV9daSBird_c']
```

I was wrong about the remedy in my earlier comment, and this is the correction.

## What the error actually measures

`tools/objisolate.py:757` computes `missing = wanted - defined`, where `defined` is every symbol in the **compiled object** whose `st_shndx` is neither `SHN_UNDEF` nor `SHN_ABS`:

```python
defined = {s.name: s for s in syms
           if s.name and s["st_shndx"] not in ("SHN_UNDEF", "SHN_ABS")}
missing = sorted(wanted - set(defined))
```

So this is not a claim that the homes are wrong, or that the addresses are wrong, or that the licensing prose is insufficient. It is the compiler reporting that `d_a_s_bird.cpp` **emits none of those nine symbols**. The manifest licenses the discard of output that does not exist.

## Why it emits none of them

`include/daSBird_c.h` declares the destructor **before** `InitResources`:

```c
    virtual ~daSBird_c();                      /* slots 16 (D1), 17 (D0) */

    virtual s32 InitResources();               /* slot  0 */
```

The key function is the first non-inline virtual **declared** in the class, and that is `~daSBird_c()` — declared at the top of the virtual block, defined out of line in `src/_ZN9daSBird_cD1Ev.cpp` and `src/_ZN9daSBird_cD0Ev.cpp`. Those leftover files own the vtable and the RTTI chain. `d_a_s_bird.cpp` does not define the key function, so mwccarm emits no `_ZTV`, no `_ZTI`, no `_ZTS` — which is exactly what `objisolate` reports.

The comment block immediately above that declaration asserts the opposite:

> *"Key function is InitResources, the first declared non-inline virtual."*

and the `_ZTV9daSBird_c` row repeats it:

> *"This TU owns daSBird_c's key function -- InitResources is the first declared non-inline virtual whose body lives here"*

Both are contradicted by the four lines of source directly beneath the first one. The same comment also records, correctly, that the 380-byte `func_ov009_0211145c` hatch forces D1/D0 to stay enrolled as their own files — that decision is sound, and it is precisely what keeps the key function outside this TU.

## The repair

The nine rows are an over-claim, not an under-specification. Concretely:

1. **Delete the nine `_ZTV`/`_ZTI`/`_ZTS` rows** from `compiler_only_output` in `config/tu_manifest.d/ov009/daSBird_c.json`. Keep `_ZN7Vector3D1Ev` (`deadstrip-duplicate`) — that one is genuinely emitted and genuinely has a ROM home. `wanted` then reduces to a set the object actually defines and the refusal clears.
2. **Correct the key-function sentence** in `include/daSBird_c.h` to name the destructor, and state that the vtable and the RTTI chain are therefore emitted by the enrolled D1/D0 files rather than by the class TU. The surrounding rationale about the hatch is right and should stay.
3. **Drop the matching claim** from any remaining row prose in the manifest that says this TU emits the class vtable.

Please do not fix this by folding D1/D0 into the TU to *make* the claim true. That would reverse the hatch decision the earlier review settled, and the 380-byte hatch sitting between D0 and the rest of the class run is what made the split necessary in the first place.

One check worth running after the edit, because it is the cheap one that would have caught this: compile the TU and compare the object's defined-symbol set against the `compiler_only_output` symbol list. Every requested symbol must appear; a row naming a symbol the object never emits fails this way every time, and the failure text names the manifest rather than the source, which is what made it read as a licensing problem.

andrewboudreau:
Re-checked the accounting blocker from my 2026-09-18 reviews against the current head `58d8e713b2141ca103c6e445f9e02890446e8198`. **It is addressed.** Recording the checks rather than the conclusion, since the previous two rounds turned on exactly this detail.

The objection was that the promoted TU enrolled the 380-byte inline-assembly `func_ov009_0211145c` inside its `complete` text span, so an unresolved assembly hatch was counting toward reconstructed-source coverage.

**1. The helper is no longer enrolled.** `config/arm9/overlays/ov009/delinks.txt` now reads:

```
src/func_ov009_02111234.cpp:
    complete
    .text start:0x02111234 end:0x0211145c

src/game/actors/d_a_s_bird.cpp:
    complete
    .text start:0x021115d8 end:0x02111a70
```

`0x021115d8 - 0x0211145c = 380`. The enrolled ranges stop exactly where the helper starts and resume exactly where it ends — the span is a hole, not a claim. Nothing else in the file enrolls that address.

**2. `src/func_ov009_0211145c.c` still exists and is honest about itself.** It is unenrolled, and its banner is at the file header rather than interior:

```c
// NONMATCHING: hand-written asm, not a C decompilation. Byte-exact via an asm hatch on a
// proven mwccarm 1.2 register-allocation/scheduling wall; does NOT count as matched. Reverts
// to a draft until someone reproduces the bytes from real C.
```

That is the disposition I asked for: the bytes stay, the coverage claim does not, and the wording leaves the C++ reconstruction explicitly unresolved instead of asserting a compiler wall. The file also records the measured 5-instruction window rather than claiming the wall in the abstract.

**3. The promotion was narrowed rather than forced.** `src_tu/actors/Bird.cpp` is gone; the real TU is `src/game/actors/d_a_s_bird.cpp` covering `0x021115d8..0x02111a70`, with the `Bird` → `daSBird_c` rename carried through the shards and `symbols/actor_renames.tsv`.

**4. `PR validation` is green** on this head, which is the byte half.

My `CHANGES_REQUESTED` is the only thing still marking this `BLOCKED`, and it is now stale — the head no longer has the defect it names. I am not dismissing it in the same motion as verifying it; leaving that to a maintainer pass so the clearing is a separate decision from the check.

The remaining `Source review` failure is the advisory check whose queue has been dormant since 2026-09-11, so it is not evidence about this PR either way.

andrewboudreau (approved):

andrewboudreau (approved):

PR #2743, opened 2026-09-17
Cleanup ov062/daHolhei_c: promote Chuckya class TU

lunavyqo:
S21 rename-then-promote of ov062 Chuckya to ROM `_ZTS10daHolhei_c`. One class, one overlay's delinks, dest is `src/game/actors/d_a_holhei.cpp` (not `Chuckya.cpp`, no `+`).

## What

- `class_rename.py --class Chuckya --to daHolhei_c --with-derived --apply` after deleting the coined `_ZTV7Chuckya` alias (ROM `_ZTV10daHolhei_c` already sat at `0x0211d9e0`; address unchanged).
- Leftover-quality TU hunt on this file: factory is `return new daHolhei_c()` (this TU emits the vtable); inline destructor in `include/daHolhei_c.h` so D1 then D0 with no D2; Init/Behavior/Cleanup/Render use real members where MATCH.
- `tu_promote.py` enrolls the 36-function run `0x02115ee0..0x021174cc` as one `complete` span. Five CONVERTED identities rewritten onto `dest#symbol` (never `--update` CONVERTED).

## Leftovers (this TU)

dCcAc Init / dBgCh Init / SetAnim / DropShadowRadHeight / SpawnCoins stay mangled (Fix12<int> by value, 6az). Helpers stay `func_ov062_*` offset soup. `data_ov062_*` file handles and state records have no recovered names. S14: `g_profile_HOLHEI` stays outside. Coined `daHolhei_c_ChangeState`: class ownership and PMF dispatch proven; original English name and `(C*, PMF*)` vs `(void*, void*)` are not.

## Verify

```
TU ov062/daHolhei_c
compiler-only     : exact deadstrip ['_ZN7Vector3D1Ev']
byte comparison   : 36/36 MATCH
objisolate check  : clean
emission order    : all 36 function(s) in the expected ROM-ascending section order
Result: 36/36 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
```

Koopa / KoopaFlag / KoopaTheQuick / Koopa+KoopaSmall untouched.

PR #2742, opened 2026-09-17
Cleanup ov019/daPgRcer_c: promote RacingPenguin class TU

lunavyqo:
Rename coined `RacingPenguin` to ROM `_ZTS10daPgRcer_c`, then promote the 30-function ov019 text-verified TU.

- Dest: `src/game/actors/d_a_pg_rcer.cpp` (not RacingPenguin.cpp)
- Factory is `return new daPgRcer_c()`; synthesized ctor stores `_ZTV10daPgRcer_c[2]`. `_ZTV` address point `0x02113310` is unchanged.
- Inline destructor in `include/daPgRcer_c.h`; sizeof wrap `0x398` kept.
- `extern int _ZTV10daPgRcer_c[];` once in the TU above the factory. Stale coined `_ZTV13RacingPenguin` dropped from `symbols.txt` / `decl_common.h`.
- CONVERTED identities migrated to `dest#symbol` (6 members); no `--update` of that section.

`python tools/tubuild.py verify ov019/daPgRcer_c`:

```
compiler-only     : exact deadstrip ['_ZN7Vector3D1Ev']
byte comparison   : 30/30 MATCH
objisolate check  : clean
Result: 30/30 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
```

`objisolate.plan_many` on the compiler-only-audited object: CLEAN (error: None; drop/externalise empty; 30 keepSymbols).

PR #2738, opened 2026-09-17
Cleanup ov010/daObjC1Hikari_c: promote LightBeam class TU

lunavyqo:
Rename coined `LightBeam` to ROM `_ZTS15daObjC1Hikari_c`, then promote the eight-function ov010 class TU.

**Class.** Castle light beam (`C1_HIKARI` / `LIGHT_BEAM(37)`). RTTI ov010:0x02112b64 names `daObjC1Hikari_c` as a direct `dActor_c` subclass. `_ZTV15daObjC1Hikari_c` stays at 0x02112ba8.

**Dest.** `src/game/actors/d_a_obj_c1_hikari.cpp` (not `LightBeam.cpp`). PeachPainting / `daObjC1Peach_c` is untouched.

**Factory.** `return new daObjC1Hikari_c()`; emitting TU, `extern int _ZTV15daObjC1Hikari_c[]` once above the factory. Inline dtor + out-of-line InitResources key function emits D1 then D0, no D2.

**Coined helper.** `func_ov010_02111a54` → `_ZN15daObjC1Hikari_c20UpdateModelTransformEv`. Inferred private spelling; no extra parameter types. Disclosed in the header, manifest notes, and `symbols/actor_renames.tsv`.

**Verify.**
```
python tools/tubuild.py verify ov010/daObjC1Hikari_c
MATCH    _ZN15daObjC1Hikari_cD1Ev                   0x021119d0  size 0x038
MATCH    _ZN15daObjC1Hikari_cD0Ev                   0x02111a08  size 0x04c
MATCH    _ZN15daObjC1Hikari_c20UpdateModelTransformEv 0x02111a54  size 0x040
MATCH    _ZN15daObjC1Hikari_c16CleanupResourcesEv   0x02111a94  size 0x024
MATCH    _ZN15daObjC1Hikari_c6RenderEv              0x02111ab8  size 0x028
MATCH    _ZN15daObjC1Hikari_c8BehaviorEv            0x02111ae0  size 0x240
MATCH    _ZN15daObjC1Hikari_c13InitResourcesEv      0x02111d20  size 0x0b0
MATCH    daObjC1Hikari_c_classInit                  0x02111dd0  size 0x040
byte comparison   : 8/8 MATCH
objisolate check  : clean
emission order    : all 8 function(s) in the expected ROM-ascending section order
Result: 8/8 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
```

[...]

PR #2755, opened 2026-09-18
Regenerate the profile reconstruction artifacts after the class promotions

andrewboudreau:
## What broke

`tool tests` is red on `main` at `997e6cb47`. That commit is not the cause — the
workflow only runs when `tools/` changes, and #2753 was the first such change in a
while, so it surfaced a drift that had already been sitting on `main`.

The failing assertion was `test_factory_filename_keeps_the_current_extension`:

```
AssertionError: '.c' != '.cpp'
```

`symbols/profile_reconstruction_*` are generated artifacts, and the class-TU
promotions moved the factory files out from under them. FEATHER's factory now lives
in `src/game/actors/daFeather_c/daFeather_c.cpp`, while the committed registry still
recommended `d_a_feather.c` — a `.c` target for a `.cpp` factory, which is exactly the
language-mode flip that test exists to forbid.

## What this does

Regenerated with `tools/profile_reconstruction.py --check`. It is deterministic here —
two consecutive runs produced identical blobs — and it moves 56 registry rows and 5
pilot rows, all of them following factory files that promotions had already relocated.

That then failed `test_shared_class_filenames_are_disambiguated_by_profile_id`, which
hardcodes four daTrsTrap_c factory filenames. Those four `.c` files no longer exist;
they were consolidated into `d_a_trs_trap_classinits.cpp`. Only the extension moves, so
the test's actual point — that one stem cannot name four files, and each is
disambiguated by profile id — is untouched.

## Verification

Ran the exact module list `.github/workflows/tool-tests.yml` runs:

```
Ran 1040 tests in 203.094s
OK (skipped=3)
```

`1040` is the same total the failing CI run reported, so no test was added, removed or
silenced — the one that was failing now passes.

## Note for reviewers

The `pilot.json` diff is large because two rows' `static_initializers` lists are
rewritten by the derived `sinit_owners` step, not because rows were dropped: the row
counts are unchanged at 20 (pilot) and 401 (registry), with no ids added or removed.
That part is worth a look if the sinit ownership analysis is meant to be stable across
regenerations.


andrewboudreau:
Heads-up on scope: this fix is a **snapshot**, and the drift it repairs will come back.

Merging `main` into this branch picked up #2738, which promoted LightBeam to
`src/game/actors/d_a_obj_c1_hikari.cpp`. That immediately re-staled the registry the
same way FEATHER had: the row still recommended `d_a_obj_c1_hikari.c` for a factory
that now lives in a `.cpp`. Re-ran the generator, one row moved, pushed.

So every class-TU promotion invalidates `symbols/profile_reconstruction_*`, and nothing
in CI notices:

- `profile-campaign.yml` triggers *on changes to*
  `symbols/profile_reconstruction_registry.tsv`. It consumes the file; it never
  regenerates to check the file is current.
- `tool-tests.yml` is what actually caught it, via
  `test_factory_filename_keeps_the_current_extension` — but that workflow only runs
  when `tools/` changes. #2753 was the first such change in a while, which is the only
  reason this surfaced at all.

Between those two, a promotion can stale the registry and it stays green until some
unrelated PR happens to touch `tools/`. That is how this one accumulated 56 stale rows.

If it is worth closing properly, the cheap version is a step that runs
`tools/profile_reconstruction.py --check` and fails when the working tree differs from
the committed artifacts, triggered on `src/**` rather than on the artifact itself. I
have not added that here — it changes what gates the queue, and this PR is meant to be
the narrow repair of a red `main`.

andrewboudreau:
Pushed `9d9e122a2`: re-regenerated against the live tip after #2742 landed.

This head had already passed validation green. I discarded that verdict on purpose, because the green was measuring the artifact as of `c0cedc389` and #2742 has since deleted `src/d_a_pg_rcer.c` for a promoted `.cpp`. Re-regenerating on the merged tree changed exactly one row — `daPgRcer_c`'s `current_factory_file` and `factory_filename`, both `.c` → `.cpp`. That is the pair `test_factory_filename_keeps_the_current_extension` compares, so merging the green head would have left `tool tests` red on main for the same reason this PR exists.

Worth noting the shape, because it is the opposite of the usual staleness signal: every other open PR went `DIRTY` when #2742 merged, but this one touches only the generated artifacts, conflicts with nothing, and so stayed mergeable and green while its content went wrong. A green verdict on a regenerated-artifact PR is evidence about the tree the run started on, not about the tree it would merge into.

That is the same gap I described above — nothing regenerates to check. It just also means the PR cannot be trusted to tell you when it has gone stale.

27 profile-reconstruction tests pass on the new head.

andrewboudreau:
Green again on `9d9e122a2`, but I am deliberately **holding this rather than merging it**, and I would rather say so than have it look stalled.

Every class-TU promotion restales these artifacts, and three are still in flight (#2743, #2745, #2749). If this lands now, each of those three puts it back out of date and the repair has to be redone as a fresh PR. Landing it *after* them costs one regeneration on this same branch instead.

The ordering intuition here is backwards from the usual one. This PR conflicts with nothing, so it never goes `DIRTY` and always looks safe to take first. That independence is exactly why it should go last: it is the only PR whose correctness is a function of every other PR in the queue.

Release condition: the last of #2743 / #2745 / #2749 merges, then I re-run `tools/profile_reconstruction.py`, confirm the diff against the live tip is empty, and merge this.

PR #2763, opened 2026-09-18
Regenerate the TU promotion queue; two rows were four days stale

andrewboudreau:
`notes/data/tu-promotion-queue.tsv` is generated by `tools/queue_audit.py`,
and `--check` disagreed with the tree on **96 of its 266 rows**. Two of those
sit in `already_promoted`, and those are the ones that cost real work.

## What went wrong

`daObjTh_Fall_Block_c` and `daTrsTrap_c` were both promoted on 2026-09-14.
The queue still read `already_promoted = no` for both.

This file is what fleet dispatch screens candidates from, so a row claiming a
finished class is unfinished sends an agent to redo landed work. That
happened: an agent dispatched against `daObjTh_Fall_Block_c` found the
promotion already on main — one TU file, manifest `status: promoted`, a
single `complete` span in `config/arm9/overlays/ov063/delinks.txt`, and all
four symbols already re-keyed in `converted-baseline.json`. It produced no
commit because there was nothing to do. `daTrsTrap_c` was the next candidate
on the same screen.

## Why it survived a same-day regeneration

#2745 regenerated this file earlier today and it still carried both stale
rows. Nothing runs the audit: the file header advertises `--check in CI`, and

```
$ grep -rn "queue_audit" .github/
$
```

returns nothing. **The gate the file names does not exist.**

That is deliberately *not* fixed here. The full audit reads
`extracted/arm9_dec.bin` and `extracted/overlays/overlay_NNNN.bin` through
`rtti_extract.py`, and that dump is not in the repo, so `--check` in its
current form cannot run on a GitHub runner at all. Wiring it needs a mode
restricted to the columns derivable from git-tracked files —
`already_promoted` among them, which is the column that actually misdispatches
agents. Separate PR; this one is data only.

## How it was produced

`python tools/queue_audit.py --write` against main at `4cb2f1590`, with
`build/rtti.json`, `build/rtti_vtables.json` and `build/tu_map.json` all
regenerated first (the tool refuses to run against inputs older than
`config/`).

## Audited, not assumed

| check | result |
|---|---|
| row set before vs after | identical — 266 rows, none added, none removed |
| `already_promoted` changes | exactly 2, both `no → yes` |
| `yes → no` flips | **0** — this is the regression shape worth stopping for |
| files touched | one, the TSV |
| diff shape | 96 insertions / 96 deletions, matching the 96 changed rows (no line-ending churn) |


andrewboudreau:
Force-pushed a fresh regeneration on top of `d7e972ad5`, replacing
`3ab955433`. This branch had gone DIRTY, and a generated file is re-derived
from main rather than merged against it.

Two numbers in the description above are now out of date, in the direction
that strengthens the case:

- the disagreement count is **98 rows**, not 96;
- **three** of them sit in `already_promoted`, not two. `daObjWlPolelift_c`
  (ov026) joined `daObjTh_Fall_Block_c` and `daTrsTrap_c` when #2760 landed.

That third row appeared in the four commits between the original
regeneration and this one, which is the point the description was making:
without a gate, this file goes stale again within a day.

`queue_audit --check` now reports `queue agrees with the tree`. The file keeps
main's LF line endings (267 lines, 0 CRLF) so the diff is content only.

This PR is the data half of a pair. #2764 adds `--check-promoted` and the
workflow that runs it, and its own `promoted` check is **currently red against
main's stale queue** — naming exactly these three rows. So the two want
merging in this order: this one first, then #2764, whose gate then passes on
the data it was written to protect.

PR #2767, opened 2026-09-18
Cleanup ov006/dScMgSlot1_c_classInit: register the MG_SLOT1 factory TU

andrewboudreau:
Two small, independent ov006 registrations, one commit each so either can be
dropped:

- `ov006/dScMgSlot1_c_classInit` — the MG_SLOT1 factory TU
- `ov006/dMg3DHeyhoObjAdapter_c` — the adapter's out-of-line constructor TU

Both are **manifest registrations, not shard folds**. Each range was already a
single-file `complete` entry in `ov006/delinks.txt` on the merge base, so no span
changed — only the path on the left of the existing entry.

## The queue row for dScMgSlot1_c names a file that does not exist

Its blocker cites `src/actors/MgBingoBallSlotsShot.cpp`. That path is nowhere in
the tree and never has been; `MgBingoBallSlotsShot` is a separate, already
promoted TU at `src/minigames/d_s_mg_smartball.cpp`. It reads like a mis-paste
from a neighbouring row. The *substance* of the blocker was right — the real
single legacy file was `src/d_s_mg_slot1.cpp` — but the row should be corrected.

## Registered as `dScMgSlot1_c_classInit`, deliberately

The licensed range is the factory alone (`dScMgSlot1_c_classInit`,
`0x0210d6b8..0x0210d740`); **41 shards of the class remain unpromoted**.
`queue_audit.py`'s `Tree.__init__` keys `promoted` off
`d["id"].split("/")[-1].split("+")`, so registering the bare class id would have
flipped the class's queue row to `already_promoted` while the class is mostly
unlanded. Precedent: `config/tu_manifest.d/ov063/daTrsTrap_c_classInit.json`.

The predicted `compiler-only:>=11` floor measured **0** for this TU, which is not
a laundering: the key function (`_ZN12dScMgSlot1_cD1Ev`) lives outside the
licensed range, so mwcc emits no `_ZTV`/`_ZTI`/`_ZTS` here at all.
`tubuild.py compile` reports "no unlicensed symbols or sections", and
`queue_audit.py`'s own header documents this exemption.

## `has_header=no` on the adapter needed no new header

`include/dMgJump3DMario_c.h` already declares `dMg3DHeyhoObjAdapter_c` in full —
all three pure virtuals, every field, and the `sizeof == 0x4c` assert — because
the already-promoted `ov006/dMgJump3DMario_c` TU needs the complete type.
Splitting it out would be a reverse-include fan-out with no byte consequence,
churning a load-bearing declaration. If the split is wanted it is its own
mechanical PR.

`compiler-only:>=3` was the RTTI trio exactly, plus one symbol the floor did not
predict: `_ZN22dMg3DHeyhoObjAdapter_cC1Ev`, the complete-object constructor,
which has no ROM home and takes a plain `deadstrip`. The trio takes
`deadstrip-data` against the homes `dMgJump3DMario_c.json` already records.

## Verdicts

```
tubuild.py verify ov006/dScMgSlot1_c_classInit
  MATCH  dScMgSlot1_c_classInit  0x0210d6b8  size 0x088
  Result: 1/1 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tubuild.py linkcheck ov006/dScMgSlot1_c_classInit
  TU .text range 0x0210d6b8..0x0210d740  136 bytes  IDENTICAL
  vs baseline (9 pre-existing error line(s)): 0 NEW, 0 resolved
  Result: SCRATCH-LINK-VERIFIED

tubuild.py verify ov006/dMg3DHeyhoObjAdapter_c
  compiler-only : exact deadstrip ['_ZN22dMg3DHeyhoObjAdapter_cC1Ev']
  MATCH  _ZN22dMg3DHeyhoObjAdapter_cC2Ev  0x020c4048  size 0x018
  Result: 1/1 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tubuild.py linkcheck ov006/dMg3DHeyhoObjAdapter_c
  TU .text range 0x020c4048..0x020c4060  24 bytes  IDENTICAL
  Result: SCRATCH-LINK-VERIFIED

rombuild.py -j16
  reproducing: 11,207   mismatching: 0
  ROM-build analysis: PASS
```

A ROM-build PASS is not evidence that these two files compiled, so each got a
**positive control**: a deliberate syntax error appended, the build failing and
naming that exact path, then a byte-for-byte restore proven with `sha256sum -c`
and the ROM rebuilt to the same hash.

`romdata_check --files src/actors/dMg3DHeyhoObjAdapter_c.cpp`: 2 VERIFIED, 1
PARTIAL, 0 DIFFERS. `_ZTI` and `_ZTV` verify outright; `_ZTS` is PARTIAL because
the string record is padded differently than the neighbouring cartridge bytes,
and the manifest discloses that rather than papering over it.

`prepush_attribution --base 3e24f2d0e` (explicit base): 1 moved with credit
intact (`d_s_mg_slot1` → `ruspecial`), 1 consolidated with credit intact
(`_ZN22dMg3DHeyhoObjAdapter_cC2Ev` → `tangosdev`), 0 changed, 0 lost.

`config/decl-agreement-baseline.json` is **not** touched — neither symbol appears
in it and the gate reports no new disagreements, so no `--update` was run.

One correction made during the work and kept in the record: the adapter's three
vtable slots were first described as `__cxa_pure_virtual` thunks. Dumping the
emitted object's section 9 showed 20 zero bytes with a single relocation at +4,
and the cartridge at `0x0213afd0` holds the same 20 bytes. The manifest now says
the slots are NULL in both images.


andrewboudreau:
**The `dMg3DHeyhoObjAdapter_c` half of this PR has been dropped, and the title
is now wrong** — this is the MG_SLOT1 factory registration alone, rebuilt on
`edcbf6eae`.

The `tree-shape` check caught it, correctly. `test_srcpath` requires that a
source enrolled for exactly **one** symbol either be named after that symbol,
or be a reconstructed `<Class>_classInit` on the NSMBW stem.
`src/actors/dMg3DHeyhoObjAdapter_c.cpp` owned one symbol,
`_ZN22dMg3DHeyhoObjAdapter_cC2Ev`, and is neither.

I checked whether the convention had simply not met this shape yet. It has:
there are **8,490** one-symbol enrolled sources in the tree, 138 of them in a
subdirectory of `src/`, and every single one keeps the symbol as its stem —
`src/engine/fader/_ZN10FaderColorD0Ev.cpp` and its neighbours are the pattern.
That file was the only exception in the repository.

What settles it is that the rename **was** the commit. The manifest's own
`boundary_evidence` says so: the range was already
`src/_ZN22dMg3DHeyhoObjAdapter_cC2Ev.cpp: complete / .text 0x020c4048..0x020c4060`
on main, and *"this promotion widens no span; the only delinks change is the
path"*. Once the path change is disallowed there is no promotion left in it, so
renaming the file back to the conforming stem and keeping the manifest would
register a TU that does nothing. Dropped rather than reshaped.

That work is not lost — it is recorded here and on the previous head
`4ec143d7f`. The RTTI trio (`_ZTI`/`_ZTS`/`_ZTV`, all with ROM homes) is still
worth licensing, and becomes straightforward the day the class owns more than
one enrolled symbol. Two findings from it also stand on their own: the class is
already fully declared in `include/dMgJump3DMario_c.h` (so `has_header=no` was
wrong), and its three vtable slots are NULL in both the emitted object and the
cartridge at `0x0213afd0`, not `__cxa_pure_virtual` thunks as first described.

## What remains, re-verified on the rebuilt branch

```
python -m unittest tools.test_srcpath        Ran 56 tests    OK
rombuild.py -j16                             2 compiled, 0 failed
                                             11,207 reproducing / 0 mismatching
                                             106/106 exact, ROM-build analysis: PASS
check_tubuild_conflicts / check_src_tu / check_duplicate_sources
check_rename_ledger / check_dead_references  all exit 0
check_decl_agreement --changed origin/main   no new declaration disagreements
```

`attribution.json` conflicted on the cherry-pick and was resolved as an audited
union: base=3612, main=3629, picked=3613, merged=3630, nothing lost from either
side, no value changed, no key from nowhere.

One `check_dead_references` fix carried over: the provenance comment in
`src/minigames/d_s_mg_slot1.cpp` named `src/d_s_mg_slot1.cpp`, the path this
commit moves it from. Reworded to the convention `src/actors/daMip_c.cpp` set —
drop the `src/` prefix, say where the file lived, point at the manifest's
`legacy_source` row, which keeps the full path.

The finding about the queue row still stands: its blocker cites
`src/actors/MgBingoBallSlotsShot.cpp`, a path that has never existed in the tree.

PR #2775, opened 2026-09-18
Cleanup ov001/dThIcon_c: promote both members; the no-header blocker is a stale classification row

andrewboudreau:
Folds `dThIcon_c` into one TU at `src/actors/dThIcon_c.cpp`, retiring both
shards. **Intact: 2 of 2 functions, `.text 0x020ab54c..0x020ab5b0`, 100 bytes.**

Both functions matched on the first compile of the generated shadow, with no
pragma and no lever work. The interesting content here is what the *blocker*
turned out to be.

## Verdicts

```
tubuild.py verify ov001/dThIcon_c
  MATCH   _ZN9dThIcon_c6RenderEv      0x020ab54c  size 0x004
  MATCH   _ZN9dThIcon_c8BehaviorEv    0x020ab550  size 0x060
  byte comparison : 2/2 MATCH   objisolate check : clean
  emission order  : all 2 function(s) in the expected ROM-ascending section order
  Result: 2/2 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED

tubuild.py linkcheck ov001/dThIcon_c
  [4b/8] defined symbols: 2 LICENSED, 0 COLLIDES-GAP, 0 COLLIDES-SRC, 0 HOMELESS
         emitted .text order of the licensed functions: ROM-ascending, as required
  TU .text range 0x020ab54c..0x020ab5b0   100 bytes   IDENTICAL
  Result: SCRATCH-LINK-VERIFIED.

rombuild.py -j16
  source-built functions: 11,207   reproducing: 11,207   mismatching: 0
  sha256 d1506e90…c478e8 == control
```

**Positive control performed**, because a green build on its own is no evidence
the new file was compiled at all: `this is not valid c++ @@@` appended to the
source produced `src/actors/dThIcon_c.cpp:102: declaration syntax error` /
`!! mwccarm failed (exit 1)`, naming that exact path. Restored (`sha256sum -c`
→ `OK`) and rebuilt to the same ROM hash.

`match.py --version 2004/b56 --strict-relocs` per symbol: both
`MATCHING VERSIONS: 2004/b56`. `Behavior`'s 24 words all `OK`, plus the trailing
literal-pool word as `reloc (wildcard)`.

Green: `check_decl_agreement --changed origin/main` (0 disagreements, **no
`--update`; the baseline is untouched**), `prepush_attribution --base
813530e7f` (8734 tracked, 2 consolidated with credit intact, 0 changed, 0 lost),
`romdata_check --files src/actors/dThIcon_c.cpp` (VERIFIED 2, PARTIAL 1,
DIFFERS 0), `check_src_tu_compiles` 184/184, duplicate-sources, port-refcheck
(423).

Not green, and not reported as green:
- **`dsd check symbols --fail`: 9 errors.** Identical to the baseline control,
  `0 NEW, 0 resolved`.
- **`check_references.py`: 4 "newly unresolvable" entries** (`func_01ffa440`,
  `daObjMarioCap_c::InitResources`, `RollingIronBall::InitResources`,
  `Door::InitResources`). Verified pre-existing by detaching to pristine
  `d7e972ad5` in the same worktree and re-running `eligible.py` +
  `check_references.py` — **byte-identical output**. No `--update` run.

## The `classif:no-header:dThIcon_c` blocker is a false positive, and it will keep firing

It is not a filename-convention problem. `notes/data/c-cpp-classification.tsv`
is a stale snapshot: its two `dThIcon` rows are keyed on
`src/func_ov001_020ab54c.c` and `src/func_ov001_020ab550.c` — **paths that no
longer exist**, since those shards were converted and renamed to
`_ZN9dThIcon_c*.cpp` long ago — and it carries `has_header=False` for them.

`include/dThIcon_c.h` has been in the tree since `67ea7a470`, under the exactly
conventional name, fully specified, with a `0x24` size assert. No header was
written or touched here. **The `dCapIcon` rows have the identical stale shape**,
so the same defect will keep manufacturing this blocker across ov001. Re-deriving
that classification is a separate PR and is not attempted here.

## `compiler-only:>=3` is real, and it is exactly 3

`_ZTV9dThIcon_c` (0x020ad494), `_ZTI9dThIcon_c` (0x020ad478) and
`_ZTS9dThIcon_c` (0x020ad480), all with ROM homes already in
`config/arm9/overlays/ov001/symbols.txt`, all `deadstrip-data`. **Nothing was
added to `symbols.txt` and no rename ledger was consulted.**

No rename was needed, and unlike the sibling `dCapIcon_c` there is nothing
homeless: the constructor and destructor are inline in the header and never
odr-used here, so mwccarm emits no C1/C2/D0/D1/D2 variant at all. The licensing
audit confirms `0 HOMELESS` and `linkcheck` reports `explicitly deadstripped []`.

## A correction to the oracle's manifest, worth passing on

`src/actors/dCapIcon_c.cpp` @ `43ac2cb36` was the style oracle — file shape,
comment register, `extern "C"` block placement, manifest prose. Its manifest
describes `_ZTV10dCapIcon_c`'s ROM home `0x020ad470` as "the two-word RTTI
prologue plus the two dispatch slots". Reading the actual overlay bytes, that is
not the layout: **the `symbols.txt` `_ZTV` row sits at the table's address point
(`&_ZTV[2]`), i.e. at the first dispatch slot, not the first word.**

For `dThIcon_c` the four words are at `0x020ad48c` — offset-to-top 0,
`&_ZTI` = 0x020ad478, slot 0 `Behavior` = 0x020ab550, slot 1 `Render` =
0x020ab54c — and the `_ZTV9dThIcon_c` row is `0x020ad494`. That is why the row's
extent to the next symbol is 8 bytes while `verify` reports the object copy as
`size=0x10`. This manifest states it that way.

`dThIcon_c` also needed none of the oracle's pragma machinery: `dCapIcon_c`
needed a split file because the cartridge demanded a per-function CSE state;
`dThIcon_c` demands nothing, so this file is single-half, plain deferred codegen,
written in reverse ROM order.

## Disclosed

`origin/main` moved twice during the work (`d7e972ad5` → `813530e7f`, picking up
the `dScTitle_c` and `daObjHmBskt_c` promotions) and `attribution.json`
conflicted — all three promotions append to the end of the same `overrides`
object. **Both sides were kept**: the object is keyed by `path#symbol`, so the
union is the whole resolution. The merged tree's diff against `origin/main` is
exactly the three config edits plus the source, and `rombuild`,
`check_decl_agreement` and `prepush_attribution` were all re-run on the merge —
same ROM sha, same clean results. The pre-merge commit `ac3d54d94` is still on
the branch as the first parent.

This PR shares three files with #2770 (`dCapIcon_c`, the other ov001 promotion):
`attribution.json`, `config/arm9/overlays/ov001/delinks.txt` and
`config/converted-baseline.json`. The delinks hunks are far apart (its range
around line 70, this one at 82–95); the other two are the same append-vs-append
shape and take the same union treatment. No source file is shared.


andrewboudreau:
**Correction to this PR's description — the `classif:no-header:dThIcon_c` diagnosis names the wrong mechanism.**

The description blames `notes/data/c-cpp-classification.tsv` and says "the same defect will keep manufacturing this blocker across ov001". The observations about that file are accurate — its two `dThIcon` rows really are keyed on `src/func_ov001_020ab54c.c` and `src/func_ov001_020ab550.c`, paths that no longer exist, with `has_header=False`, and the file really is a stale snapshot (2020 of its 7067 rows are keyed on paths not in the tree, and 3676 live `src/` files have no row at all).

But that file **manufactures nothing**, because nothing reads it. The only reference to it under `tools/` is a test fixture listing it as a modified path. Regenerating it would not move this blocker or any other.

The real mechanism is in `tools/queue_audit.py`. Its blocker-refresh loop re-derives exactly four kinds — `compiler-only`, `no-legacy-source`, `unmatched`, `not-in-delinks`. Every other token falls through the final `else: out.append(b)` and is copied through verbatim on every `--write`. So `classif:`, `pragma:` and `floor:` tokens are **write-once and can never be retired**. That is why this one survives in a row whose own `has_header` column reads `yes`.

It is not unique to ov001, and it is not a majority of the tokens. Of the 17 `classif:no-header:<Class>` tokens on main, **4 are provably false** — `dThIcon_c`, `dMgPsOpt_c`, `daObjLava_c`, and `dMgJump3DMario_c` (whose row also reads `already_promoted=yes`). The other 13 are genuine: no `*<Class>.h` is tracked anywhere and no header declares the class. That includes `dCapIcon_c`, so the description's claim that the `dCapIcon` rows carry "the identical stale shape" is also wrong — that blocker is true.

None of this changes anything in this PR's diff, its verdicts, or the conclusion that `dThIcon_c`'s own blocker is a false positive. Only the explanation of *why* was wrong, and a reader chasing the fix would have gone to the wrong file.

The fix belongs in `queue_audit.py` — retire `classif:no-header:<C>` when a tracked header defines `<C>`, probing both `include/**/<C>.h` and a declaration grep, since a name-only probe gives false negatives for classes sharing a header. It is deliberately not attempted here: `tools/queue_audit.py` and `notes/data/tu-promotion-queue.tsv` are both in flight on #2764, so this has to follow that PR rather than race it.

andrewboudreau:
**Correcting my own correction above: `dCapIcon_c`'s blocker is false too, so this PR's description was right about it and I was wrong.**

My earlier comment said 4 of the 17 `classif:no-header:<Class>` tokens are false and that `dCapIcon_c` was not among them — "that blocker is true". That came from a name-only probe (does `include/**/<Class>.h` exist), which is precisely the false-negative I then warned about in the same paragraph. Re-run against the tree with a declaration probe as well, the count is **9 of 17**, and every one is a real definition rather than a forward declaration:

```
daObjLava_c           include/daObjLava_c.h:17        struct daObjLava_c : dActor_c {
dThIcon_c             include/dThIcon_c.h:10          struct dThIcon_c {
dMgPsOpt_c            include/dMgPsOpt_c.h:10         struct dMgPsOpt_c {
dMgJump3DMario_c      include/dMgJump3DMario_c.h:98   struct dMgJump3DMario_c : dMg3DHeyhoObjAdapter_c {
dMgCardObj_c          include/dScMgCard_c.h:134       struct dMgCardObj_c {
dMgDilarCardObj_c     include/dScMgCard_c.h:166       struct dMgDilarCardObj_c : dMgCardObj_c {
dMgMCarloCardObj_c    include/dScMgMCarlo_c.h:103     struct dMgMCarloCardObj_c {
dMgMCarlo2CardObj_c   include/dScMgMCarlo2_c.h:61     struct dMgMCarlo2CardObj_c {
dCapIcon_c            include/CapIcon.h:47            struct dCapIcon_c {
```

The split is the whole point. The first four are findable by filename. The **last five are not** — they are defined in a header named after a different class (`dScMgCard_c.h`, `dScMgMCarlo_c.h`, `dScMgMCarlo2_c.h`) or one named without the class prefix at all (`CapIcon.h`). `dMgJump3DMario_c` also carries a forward declaration at `:36` above its definition at `:98`, so a probe that accepts `struct <C>;` would count a forward declaration as a header — it must require `{` or a base clause.

The remaining 8 are genuine: `daWater_Tatumaki_c`, `daSnowman_c`, `daPukupuku_c`, `daObjFl_Puzzle_c`, `daObjFlMaruta_c`, `daIbl_c`, `dMgTrmpln3DMario_c`, `dMgTrmpln2Mario_c` — no header of that name, and no header declaring them.

Unchanged: the mechanism is still `tools/queue_audit.py`'s write-once blocker copying, not `notes/data/c-cpp-classification.tsv`, and nothing in this PR's diff or verdicts is affected. What changes is the size of the problem — more than half of these tokens are noise, not a quarter — and that `dCapIcon_c` is a workable row rather than a blocked one.

PR #2782, opened 2026-09-18
Mark dCapIcon_c and dThIcon_c members so each scores on its own body

andrewboudreau:
**`main` is red on the `converted ratchet` workflow and has been since `3cc665ffa`. This turns it green. Six added comment lines, nothing else.**

```
FAIL: test_check_passes (tools.test_tiers.BaselineIsInSync.test_check_passes)
AssertionError: Lists differ:
  ['src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cC1Ev',
   'src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cD0Ev',
   'src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cD1Ev'] != []
```

That test's own docstring states the stakes: *"`--check` must exit 0 on a clean tree, or every PR after this one goes red."* Five consecutive pushes to main have failed it, and every PR opened since inherits the red.

## It is an accounting failure, not a regression

Nothing about `dCapIcon_c` got worse. The three symbols fail `no_raw_offset` and `no_unk_field` — readability criteria, scored against source text.

`tiers.score_member()` slices a member's text at `// @symbol` boundaries, and its docstring is explicit about the fallback: *"Anything without either form of evidence is scored against the entire file, preserving the old conservative behavior."*

The promoted file carried **no markers at all**. So all six members were scored against the whole file, and the two unlabelled helpers' text —

```c
unk_19 = 0;                                  /* inside Unlink            */
*(unsigned char*)(c+0x18) = (unsigned char)idx;   /* func_ov001_020ab228 */
```

— was charged against the constructor and both destructors, whose entire bodies are:

```c
dCapIcon_c::dCapIcon_c()   { mPrev = 0; mNext = 0; mFlags |= 4; }
dCapIcon_c::~dCapIcon_c()  { Unlink(); }
```

Named fields, no raw offsets, no `unk_`. This is exactly the failure per-member scoring exists to prevent — `classify_missing()`'s own docstring says member scoring *"keeps that failure local and the diagnostic names it instead of averaging it away."* Without markers it cannot.

## The fix, and why this shape

Six `// @symbol` lines, one per member, placed under each existing banner. The convention is already landed across the tree — `da1up_c.cpp`, `daBgSnmBdy_c.cpp`, `daBmb_c.cpp`, `daBombking_c.cpp`, `daChair_c.cpp` — including the **stacked D1/D0 pair** over their one shared definition, copied from `da1up_c.cpp:127-128`. The `dScMgJump_c` promotion landing alongside this carries 17 of them in a file of the same shape.

## Restores exactly the three banked identities, and nothing else

```
banked  : C1Ev, D0Ev, D1Ev
restored: C1Ev, D0Ev, D1Ev
MISSING : []      EXTRA : []
```

**No member gains coverage it did not have.** The remaining failures are now correctly local, each scored on its own body:

| member | before (whole file) | after (own body) |
|---|---|---|
| `_ZN10dCapIcon_cC1Ev` | `no_raw_offset`, `no_unk_field` | **converted** |
| `_ZN10dCapIcon_cD1Ev` | `no_raw_offset`, `no_unk_field` | **converted** |
| `_ZN10dCapIcon_cD0Ev` | `no_raw_offset`, `no_unk_field` | **converted** |
| `_ZN10dCapIcon_c6UnlinkEv` | `no_raw_offset`, `no_unk_field` | `no_unk_field` — its own `unk_19` |
| `func_ov001_020ab228` | `real_name`, `no_raw_offset`, `no_unk_field` | `real_name`, `no_raw_offset` — its own |
| `func_ov001_020ab2e4` | `real_name`, `no_raw_offset`, `no_unk_field` | `real_name` alone |

The last row is the tell that the old scoring was wrong rather than strict: `func_ov001_020ab2e4`'s body is clean apart from its placeholder name, and it was being charged for a neighbour's offsets.

`config/converted-baseline.json` is **not touched**. The baseline was right; the scorer could not see it.

## Bytes

The diff is six comment lines added, zero removed, no source line altered. Comments cannot change codegen — but this file carries `#pragma opt_common_subs off` over `func_ov001_020ab228`, so a text insertion near it deserves the real check rather than the argument.

```
rombuild.py -j16
  8535 reused from cache, 1 compiled
  source-built functions: 11,208   reproducing 11,208   mismatching 0
  ROM sha256 d1506e90...c478e8   == the stock baseline
```

The `1 compiled` is the positive control and it comes free: `build/objcache` is content-addressed, so the comment change forced exactly this file to recompile — proving it is genuinely enrolled and built — and the ROM still hashes identically. The markers sit above the banner comments; the pragma remains adjacent to its function, where it was.

`python -m unittest tools.test_tiers` → **35 tests, OK**. `tiers_ratchet.py --check` → `CONVERTED ratchet PASS   baseline 2813   current 2978`.

## The latent defect this leaves behind

A promotion that folds labelled members together with unlabelled `func_<addr>` neighbours will do this again, silently, and the damage lands on main rather than on the PR — the promoting branch's own `tiers_ratchet --check` passes, because `--check` compares against the baseline as it stands at that moment.

The durable fix is for the promotion path to require markers on a multi-member TU, or for `score_member` to refuse the whole-file fallback when a file has more members than markers. Neither is done here; this PR restores the green and names the hazard.


andrewboudreau:
**Update: three markers were not enough. Extended to four identities across two files, and the `ratchet` check on this PR is now green.**

The description above was written against the main of the time. Merging current main in exposed a fourth backslide, because `dThIcon_c` was promoted after this branch was cut:

```
  src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cC1Ev
  src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cD0Ev
  src/actors/dCapIcon_c.cpp#_ZN10dCapIcon_cD1Ev
  src/actors/dThIcon_c.cpp#_ZN9dThIcon_c6RenderEv   <- new
```

Same defect, same mechanism, independently arrived at: `dThIcon_c.cpp` carries no `// @symbol` markers either, so both its members are scored against the whole file.

`dThIcon_c::Render` is an **empty body** — the cartridge's own function is a single `bx lr`, four bytes. It fails `no_unk_field` entirely on its neighbour's text: `Behavior` reads and writes `unk_00c`, `unk_010`, `unk_011` and `unk_014`, and the doc comment above `Behavior` names them again. None of that is Render's.

Two more markers, placed under each banner exactly as the dCapIcon_c ones are. `Behavior` now fails `no_unk_field` alone, on its own fields, which is correct and local — it is not banked. `Render` passes all five and its banked identity is restored.

```
before:  CONVERTED backslide: 4 banked file(s)
after:   CONVERTED ratchet PASS   baseline 2813   current 2979
         python -m unittest tools.test_tiers -> 35 tests, OK
```

Byte-neutral, checked rather than argued — `dThIcon_c.cpp` carries a pragma, so the text insertion got the real control:

```
rombuild.py -j16
  8523 reused from cache, 1 compiled
  source-built functions: 11,208   reproducing 11,208   mismatching 0
  ROM sha256 d1506e90...c478e8   == the stock baseline
```

The `1 compiled` is the positive control and it is free: `build/objcache` is content-addressed, so the comment change forced exactly this file to recompile, proving it is genuinely enrolled and built.

**Two notes on the checks as they read now.** The `submit` failure earlier was the known stale-pin shape — `Validation submission failed: Target branch advanced after resolution` — cleared by a full `gh run rerun`, not `--failed`, which would only replay the stale pin. And the remaining `promoted` red is **not this PR's**: it is the `queue promoted column` workflow, stale `already_promoted` cells for four separate classes, fixed in #2783.

PR #2791, opened 2026-09-18
Fold daObjSwdoor_c (ov002) into one translation unit

andrewboudreau:
Folds `daObjSwdoor_c`'s four ov002 members into one translation unit, `src/actors/daObjSwdoor_c.cpp`, replacing four bare-path shards.

## Verified

- `rombuild.py`: **106/106 modules exact**, 100.000000% of compared bytes, ROM `sha256 d1506e90…c478e8` — the stock hash.
- **Positive control on the merged tree**, because `0 compiled` out of a content-addressed cache proves nothing: injecting a syntax error aborted the build with `daObjSwdoor_c.cpp:185: expression syntax error`; restoring gave back the original sha256 and `106/106 exact`.
- `check_src_tu_compiles: 199/199`.
- Reference check genuinely ran (`eligible.py` first, report stamped at HEAD, `dirty false`) and printed `pre-push: unresolved references OK` — not the skip line.

## The `@symbol` markers are load-bearing here

All four members carry one. Driving `tiers_ratchet.scan()` directly, the file resolves to **four separate identities** rather than one whole-file score:

```
src/actors/daObjSwdoor_c.cpp#_ZN13daObjSwdoor_c6RenderEv   CONVERTED (all 5 criteria)
src/actors/daObjSwdoor_c.cpp#func_ov002_020baba8           not converted
src/actors/daObjSwdoor_c.cpp#func_ov002_020bac18           not converted
src/actors/daObjSwdoor_c.cpp#func_ov002_020bad10           not converted
```

Only `#Render` is banked, and it passes. Without the markers the three raw-offset helpers would have dragged the banked identity red — on `main`, after the merge, not here.

## The two baselines were settled, not union-merged

Both conflicted; both were edited **in place**, with the keyed-set identity asserted in each direction.

| file | base | ours | theirs | merged | deletions honoured |
|---|---|---|---|---|---|
| `config/converted-baseline.json` | 2813 | 2813 | 2813 | **2813** | 14 (1 mine, 13 main's) |
| `config/decl-agreement-baseline.json` | 17787 | 17787 | 17758 | **17758** | 101 (13 mine, 88 main's) |

`merged == (ours | theirs) - (base - ours) - (base - theirs)` holds for both, with the residual diff against main's copy being exactly the intended lines and nothing else. A plain union would have resurrected every honoured deletion.

Neither was re-serialised. For the decl baseline the round-trip proof came back **negative**: the gate's own `write_baseline` reproduces base and ours byte-identically but not main's copy, which carries several `dScMgFlower_c` entries out of sorted order from an earlier in-place re-key. Re-serialising would have silently rewritten ~94 lines of someone else's work, so the 13 re-keyed findings were inserted at their sorted positions by hand across 9 symbol blocks.

## Gates

```
CONVERTED ratchet PASS   baseline 2813   current 2980   (+167 gained, not yet banked)
queue_audit: every row's already_promoted agrees with config/tu_manifest.d
```

`check_decl_agreement --check` still exits 1 with 5 findings on `_ZN7dBase_c5SpawnEjP7fBase_cii`, in files this branch never touches. Measured rather than assumed: detaching to the merge base in the same worktree reproduces the identical 5-finding block and the identical `17766 disagreement(s) across 3868 symbol(s)`. **0 new.** That is the phantom-`this`-on-a-static-member defect, fixed separately in #2790.


andrewboudreau:
Re-settled onto live main after #2790 merged. New head `a5ee21bec`.

**The decl-baseline section of the description above is superseded.** It describes a keyed-set merge, and that instrument is wrong for this particular merge. Recording why, because the distinction is not obvious and the first attempt looked correct right up until the gate failed.

#2790 changed `check_decl_agreement.py` itself, not just the data it records. So one underlying disagreement is spelled differently on the two sides and occupies **two different keys**: a phantom-`this` `arity:1` that this branch had re-keyed onto the folded TU is the same finding main now records as `param:#1 void *`. The set formula `(ours ∪ theirs) − (base − ours) − (base − theirs)` dutifully kept both and dropped neither, and `--check` came back **RC=1 on 2 findings** — it read the TU-path `param:` entries as new, because nothing in the merged set carried them. The formula is still correct whenever both sides only add and remove entries; it cannot survive the tool changing underneath it.

**What replaced it:** take main's copy verbatim, regenerate on the merged tree, then audit that regenerate against pristine main. The audit is what makes `--update` legitimate here rather than a rubber stamp:

```
main 17611   merged-tree regen 17611
ADDED 13   REMOVED 13
PURE PATH RE-KEY (symbol+finding identical on both sides): True
added files  : ['src/actors/daObjSwdoor_c.cpp']
removed files: ['src/func_ov002_020bac18.c', 'src/func_ov002_020bad10.c']
```

Added only under the folded TU, removed only from the two deleted shards, count unchanged. **The fold introduces no new disagreement.**

**Re-serialising is safe here only because it was proved first**, and that proof is not a formality — it returned the opposite answer a few hours ago. The gate's own `write_baseline` now reproduces main's committed copy byte-for-byte, because #2790's `--update` canonicalized the file. Before that, main carried `dScMgFlower_c` entries left out of sorted order by an earlier in-place re-key, and re-serialising would have churned ~94 lines of unrelated work. The check is re-run every time for exactly that reason.

Re-verified on the merged tree: `check_decl_agreement --check` and `--changed origin/main` both exit 0, `tubuild verify ov002/daObjSwdoor_c` → **4/4 MATCH, TEXT-VERIFIED**, `rombuild` **106/106 exact** with ROM `sha256 d1506e90…c478e8` (hashed directly — `106/106` is not the hash), and the push ran the reference check for real: `pre-push: unresolved references OK`, not the skip line.

PR #2792, opened 2026-09-18
Fold dScMgSlot1_c (ov006) into one translation unit

andrewboudreau:
Folds `dScMgSlot1_c`'s ov006 sub-range into one translation unit, `src/actors/dScMgSlot1_c.cpp`: 18 members, 18 shards deleted.

## Verified

- `tubuild verify ov006/dScMgSlot1_c`: **18/18 MATCH**, objisolate clean, reloc-destinations clean → **TEXT-VERIFIED**.
- `rombuild.py` RC=0, **106/106 modules exact**, ROM `sha256 d1506e90…c478e8`.
- `check_src_tu_compiles: 199/199`.
- `prepush_attribution.py --base 9c4c0009f` (explicit base, never the default, which goes stale and invents a `CREDIT CHANGED` naming a class you never touched): `8629 tracked, 0 moved, 0 renamed, 18 consolidated with credit intact, 0 changed, 0 lost`.
- Reference check ran and passed: `unresolved symbols 41 (baseline 45)`, `eligible 11252`. The 45 → 41 is #2789's four.

## Two orphaned CONVERTED identities, rewritten — deliberately not `--update`

The fold absorbed two sources that were banked as CONVERTED, leaving their baseline entries naming files that are no longer tracked:

```
src/_ZN12dScMgSlot1_c15OnHitByMegaCharEv.cpp
src/_ZN12dScMgSlot1_c19OnHitFromUnderneathEv.cpp
```

An orphan is scored by nothing, so no criterion can fail for it and every later `--check` reads it as clean — silent rot, and **every promotion that folds a previously-banked shard will hit it.**

`--update --reason` would exit 0 here and append a row recording each legacy path as REMOVED when nothing was removed: a permanent false statement in a log whose entire value is being trustworthy. So both were **rewritten** onto the promoted TU, using the convention 277 identities already follow:

```
src/actors/dScMgSlot1_c.cpp#_ZN12dScMgSlot1_c15OnHitByMegaCharEv
src/actors/dScMgSlot1_c.cpp#_ZN12dScMgSlot1_c19OnHitFromUnderneathEv
```

Both targets pass all five criteria. Edited in place at byte level rather than re-serialised, because that list is not sorted (`sorted(c) != c`) and a re-serialise would have produced a large spurious diff: exactly **2 insertions, 2 deletions**, CRLF preserved, count unchanged at 2813, no identity added or removed.

## `attribution.json` settled as a keyed set

| set | count |
|---|---|
| base | 3681 |
| ours | 3699 |
| theirs | 3739 |
| plain union | 3759 |
| **merged** | **3757** |

**2 deletions honoured**, and they are exactly the bare-path overrides a promotion folds away — `src/_ZN11dScMgJump_cD0Ev.cpp` and `…D1Ev.cpp`. A plain union resurrects both. **0 value conflicts**: no key carries a different owner on the two sides, and `aliases`/`note` are identical across all three, so nothing needed arbitration. Insertion order preserved.

## Gates

```
CONVERTED ratchet PASS   baseline 2813   current 2980   (+167 gained, not yet banked)
queue_audit: every row's already_promoted agrees with config/tu_manifest.d
```

All 18 members carry a `// @symbol` marker, so each scores against its own body rather than the whole file.


andrewboudreau:
Re-settled onto live main after #2790 merged.

**The decl-baseline handling described above is superseded**, for the same reason as #2791. #2790 changed `check_decl_agreement.py` itself, so one underlying disagreement is now spelled differently on the two sides and occupies two different keys. A keyed-set merge keeps both spellings and drops neither, which leaves the gate red on the entries that exist only under the new spelling. The set formula is still correct whenever both sides only add and remove entries; it cannot survive the tool changing underneath it.

Replaced by: take main's copy verbatim, regenerate on the merged tree, audit the regenerate against pristine main.

```
main 17611   merged-tree regen 17610
ADDED 4   REMOVED 5
distinct (symbol,finding)  added 4  removed 4
SET-EQUAL: True      added-only: set()      removed-only: set()
added files  : ['src/actors/dScMgSlot1_c.cpp']
removed files: ['src/_ZN12dScMgSlot1_c9betIcon_c8BehaviorEv.cpp',
                'src/func_ov006_0210c234.c', 'src/func_ov006_0210c2d4.c']
```

**The count asymmetry is a fold dedup, not a lost finding**, and it is worth naming rather than waving through. `_ZN5Sound12PlayBank2_2DEj return:void` was recorded against **two** of the folded shards — `betIcon_c8BehaviorEv.cpp` and `func_ov006_0210c2d4.c`. Both now live in one file, so the two entries collapse into one. That is why the audit compares distinct `(symbol, finding)` **sets** rather than counts: a multiset comparison false-alarms on every fold that merges shards sharing a disagreement. With the file component stripped, the two sides are exactly equal in both directions, so nothing was gained or lost.

Re-serialising was proved safe before anything was written — the gate's own `write_baseline` reproduces main's committed copy byte-for-byte now that #2790's `--update` has canonicalized it. That check returned the opposite answer earlier today, when main still carried entries left out of sorted order by an in-place re-key, so it is re-run every time rather than assumed.

Re-verified on the merged tree:

```
check_decl_agreement --check           exit 0   no new declaration disagreements
check_decl_agreement --changed main    exit 0
tubuild verify ov006/dScMgSlot1_c      18/18 MATCH, objisolate clean -> TEXT-VERIFIED
rombuild                               106/106 exact, ROM-build analysis PASS
ROM sha256                             d1506e90...c478e8   (hashed directly)
CONVERTED ratchet PASS   baseline 2813   current 2980
queue_audit: every row's already_promoted agrees with config/tu_manifest.d
```

The two orphaned CONVERTED identities described in the original text are unchanged and still rewritten rather than `--update`d.

PR #2798, opened 2026-09-18
Fold daKrb_c (ov084) into one translation unit; bank text-verified

andrewboudreau:
Folds the 35 functions at `0x02129020..0x0212c10c` in ov084 into one translation unit and banks it **text-verified**. Two files, both pure additions: `src_tu/actors/daKrb_c.cpp` and `config/tu_manifest.d/ov084/daKrb_c.json`. Nothing existing is modified — no `src/` shard is removed, no `delinks.txt` is spliced, no ledger is touched.

## Evidence

```
tubuild verify ov084/daKrb_c
  byte comparison   : 35/35 MATCH  (extract_func + compare, relocation-aware)
  objisolate check  : clean
  emission order    : all 35 in the expected ROM-ascending section order
  Result: 35/35 MATCH -> TEXT-VERIFIED

tubuild linkcheck ov084/daKrb_c            exit 0
  defined symbols: 35 LICENSED, 0 COLLIDES-GAP, 0 COLLIDES-SRC, 0 HOMELESS
  module ov084: built 31904 bytes, retail 31904 bytes
  TU .text range 0x02129020..0x0212c10c   12524 bytes   IDENTICAL
  full ROM: 16777216 bytes, sha256 d1506e90…c478e8 — identical to stock
  Result: SCRATCH-LINK-VERIFIED
```

`dsd check symbols --fail` is **not** green and is not being reported as green: it emits 9 errors (`overlay_100`, `overlay_102`, `data_020ad524`, `data_020ad560`, 5 ITCM symbols). A `--baseline` run taken at the pristine branch point — which `git merge-base` confirms is the merge base with `main` — emits the same 9, and the diff of the two error sets is empty. 0 new, 0 resolved.

**Positive control**, because a green verify is only evidence that the tool ran: a syntax error injected at line 408 aborts both `verify` and the whole-tree `linkcheck` naming this file and that line; the file was then restored and re-verified against a recorded sha256.

## The unit is 35 functions, not 9

The span is claimed by exactly 35 delink entries with no hole and no overlap — one `.text` claim. The apparent gap between the destructors and the main member run is 22 unlabelled `func_ov084_*` helpers, which a class-name census cannot see.

The three `daKrb_c_classInit*` factories are absorbed and all three MATCH. They have no caller anywhere in `src/`; they are reached only through the actor registry table, which ov084 does not delink. In C++ all three are `return new daKrb_c();`, reproducing the ROM's `operator new(1144)` plus inlined implicit constructor. They are byte-identical to one another, so their identity rests on the `symbols.txt` addresses rather than on bytes — recorded in the manifest's boundary evidence. Same shape as the landed `ov084/daPkn_c`.

The generated unit id was `daKrb_c+daKrb_c`. Both labels are the same class: `tu_map` labelled one run twice. There is no second class in the span, and the `multi-class-tu:2` blocker on the queue row is a false positive from counting labels instead of taking their set. Renamed to `ov084/daKrb_c`, precedent `ov002/daObjLava_c`, which landed promoted from the same generated shape.

## Why this banks text-verified rather than promoting

`tubuild promote --dry-run` refuses, and the refusal is structural rather than specific to this TU:

```
manifest status : text-verified   <<< promotion would be REFUSED
  - manifest status is not link-verified or data-verified
  - the promotion-ready status is not the recorded linkcheck result
```

The readiness check requires `status` to be one of `link-verified`/`data-verified` **and** `linkcheck.result == status`. The compiler-only policy rewrites the object before the link for any TU that emits its own vtable and RTTI, so the recorded result is prefixed `scratch-`. Setting the status to `link-verified` fails the equality test; setting it to `scratch-link-verified` fails the membership test.

A census of all 203 committed manifests: **76 promoted entries recorded a `scratch-*` result** — the same state this TU is in — and only 3 entries in the tree satisfy `status == result`, none of them promoted. Clearing that is a tools change and belongs in its own review, so the status here is the one the tools actually reported.

Consequently `config/converted-baseline.json` is untouched and `already_promoted` stays `no`, which is the agreeing value for a non-promoted entry. The three banked `daKrb_c` identities still key on legacy shard paths that this change does not move, so nothing is orphaned. If the promotion plan is ever applied, those three need rewriting in place onto `src/actors/daKrb_c.cpp#<symbol>`.

## Three levers worth recording

**The destructor variant order decides the whole file's layout, so decide it first.** `include/daKrb_c.h` declares `~daKrb_c()` out of line. Under deferred codegen mwccarm emits D2, D0, D1 — D0 at `0x02129070` ahead of D1 at `0x02129020` — and linkcheck refuses the object: *licensed .text functions are not emitted in ROM address order*. This reproduced a clean 35/35 MATCH with objisolate clean, so byte verification and linkability are genuinely orthogonal here. The fix is `#pragma defer_codegen off` plus rewriting the file in ascending order, which flips the group to D1, D0, D2. Discovering it at linkcheck means rewriting everything; make the call before the first line.

**`tubuild create` carried `#pragma optimize_for_size on` in from one legacy member.** Under deferred codegen a non-`long_calls` pragma is file-global last-wins, so it recompiled the whole unit: 15/32 MATCH with it, 31/32 without. Deleting one line was the entire fix.

**A `.c`-derived member that assigns a comparison to an `int` breaks when folded into C++.** `func_ov084_021290d4` built 0xa4 against the ROM's 0x94. The shard spells `new_var3 = (*(int*)(c+0x434)) == 1; if (new_var3)`. In C99 that is an `int`; in C++ it is a `bool`, and the front end materialises it (`moveq r1,#1 / movne r1,#0`) instead of predicating the store off the compare as the ROM does. Holding the raw word and comparing inside the `if` restored the if-conversion. Every `.c`-derived member with that shape is exposed the moment it is folded.

## Merge state

`git merge-tree --write-tree` against `main` returns a tree OID and exit 0 — no conflict paths. The branch adds two paths that do not exist on `main`, modifies no ledger, and `main` has not touched ov084 since the merge base.


andrewboudreau:
Correcting one claim in the description. The measurement stands; the attribution does not.

The description says `tubuild create` carried `#pragma optimize_for_size on` in from one legacy member. **It did not, and it could not have.**

`tools/tubuild.py:842` carries only positional `long_calls on`:

```python
carried = "carried below" if _LONG_CALLS_ON_RE.match(p) else "NOT carried -- review"
```

Every other pragma is deliberately left out and recorded as a review comment, and the generated header states the reason — that a file-global last-wins pragma "carried into a merged TU ... would silently recompile every other member." `tools/test_tubuild.py:2272` asserts this for `optimize_for_size` by name, and the behaviour landed in `a30b58ab6` on 2026-08-19, a month before this branch was cut.

The committed `src_tu/actors/daKrb_c.cpp` bears that out: it contains the `TUBUILD NOTE` block listing `_ZN7daKrb_c6RenderEv: #pragma opt_common_subs off  [NOT carried -- review]`, and no `optimize_for_size` anywhere.

So the pragma was introduced by hand during the fold, most likely copied out of that review block, and removing it is what produced the 15/32 → 31/32 swing. The mechanism — a non-`long_calls` pragma being file-global last-wins under deferred codegen, and costing 16 members that had nothing to do with the shard it came from — is real and is the part worth carrying forward. Its provenance in the description is wrong.

This also inverts the advice for anyone folding a TU with the generator: the generated file **under-carries**. The useful control is adding an un-carried pragma back, cross-referenced against the functions that still differ, not hunting for one to delete.

None of this changes the evidence for this PR. The 35/35 MATCH, the linkcheck result and the ROM hash were all taken on the committed file, which does not contain the pragma.

PR #2819, opened 2026-09-19
Bank the reconstructed ov006 dScMgSmartball_c TU (107/130 byte-identical, status shadow)

andrewboudreau:
**This is a bank of reconstruction work, not a coverage claim.** 107 of the 130 members of the ov006 `dScMgSmartball_c` translation unit compile byte-identical to the ROM; **23 do not**, and the manifest records `"status": "shadow"` — not `text-verified`, not `promoted`. `tubuild verify`, re-run on this tree today, ends with:

```
Result: 107/130 MATCH -> NOT verified (see DIFF/MISSING/ERROR lines above)
        51 unlicensed section/symbol(s) present -> PROMOTION REFUSED regardless of the above (plan sec 4.5, 8)
        compiler-only policy refused -> PROMOTION REFUSED
```

The file is **not enrolled** — it is `src_tu/`, it contributes nothing to the ROM build, and no `delinks.txt` or `symbols.txt` is touched. No coverage number moves. What this PR is for is that the 130-function fold itself is the expensive part, and it currently exists only on an unlanded branch; landing it means the remaining 23 can be worked without redoing the other 107.

## What lands

Two new files, and one follow-up edit to the second:

| file | lines |
| --- | --- |
| `src_tu/actors/dScMgSmartball_c.cpp` | 7,345 |
| `config/tu_manifest.d/ov006/dScMgSmartball_c.json` | 1,067 |

The unit is the contiguous linker run `0x0210d740..0x02119824` from `build/tu_map.json` — 130 functions over 13 classes (`dScMgSmartball_c` plus `cMgSmartball_{ana,ball,board,dokan,kinoko,object,pakkun,propeller,pushswitch,slot,spring,wing}_c`). `boundary_confidence` is `medium`: the sinit corroboration in the entry is module-wide, not narrowed to this one TU, and the entry says so.

`dScMgSmartball_c` is the ROM's own RTTI name, not a coined one — `_ZTI16dScMgSmartball_c`, `_ZTS16dScMgSmartball_c` and `_ZTV16dScMgSmartball_c` are all three in `config/arm9/overlays/ov006/symbols.txt` (0x0213ec28 / 0x0213ec7c / 0x0213eefc), and ten files on main already use it. Nothing needs renaming first.

## The 23 that do not match

**14 fail only the relocation type/addend check** — their instruction bytes compare equal; what `objisolate` rejects is a vtable word. They cover twelve of the thirteen classes (`_ZTV19cMgSmartball_ball_c` accounts for three of the fourteen):

```
DIFF  func_ov006_0210e0d0   -- objisolate: _ZTV19cMgSmartball_wing_c: unexpected reloc type=2 addend=0
DIFF  func_ov006_0210e480   -- objisolate: _ZTV25cMgSmartball_pushswitch_c: unexpected reloc type=2 addend=0
DIFF  func_ov006_0210f9f8   -- objisolate: _ZTV20cMgSmartball_board_c: ...
...  (also 021101bc, 02110874, 02110bc0, 02111220, 0211157c, 02111774, 02111b40,
     02114548, 0211474c, 02115b0c, and _ZN16dScMgSmartball_c21AfterCleanupResourcesEj)
```

What differs is how the vptr store names the vtable, and that spelling depends on whether the compiling TU emits that vtable or imports it. This fold emits all thirteen `_ZTV`s itself — they are among the 51 unlicensed symbols below — so this is a coherent single lead across all fourteen rather than fourteen separate defects.

**9 differ in content.** Read the verify output carefully here: `999 word(s) differ` is a **sentinel for "the compiled function is a different length than the ROM's"** (`tools/match.py:268` returns `999` on a size mismatch before it compares anything), not a count of differing words. Only two of the nine are actual word diffs. The real gaps:

| function | ROM size | compiled | gap | also |
| --- | --- | --- | --- | --- |
| `_ZN20cMgSmartball_board_c6UpdateEv` | 0x7b0 | 0x7ec | **+15 words** | 2 reloc destinations wrong (first: `GetOwnerLanguage != 0x020afdd0:ov004`) |
| `func_ov006_0210eca4` | 0x2a4 | 0x29c | −2 words | |
| `func_ov006_0210ef48` | 0x61c | 0x618 | −1 word | |
| `func_ov006_02114800` | 0x310 | 0x300 | −4 words | |
| `func_ov006_02115830` | 0x22c | 0x21c | −4 words | |
| `_ZN16dScMgSmartball_c6RenderEv` | 0x10c0 | 0x10bc | −1 word | 9 reloc destinations wrong |
| `_ZN16dScMgSmartball_c13InitResourcesEv` | 0x8dc | 0x8d8 | −1 word | 26 reloc destinations wrong |
| `func_ov006_0210fb58` | 0x3c4 | 0x3c4 | same size | **52 words** differ |
| `_ZN21cMgSmartball_spring_c12SaveSnapshotEv` | 0x214 | 0x214 | same size | **8 words** differ |

Five of the seven size misses are within one to four instructions. The three rows carrying wrong relocation *destinations* are a different kind of defect: a declaration in the fold names a function the ROM does not call there, so those are reconstruction bugs rather than near-misses, and they should be fixed before anyone prices the codegen.

Also outstanding, all recorded in the manifest's `verification` block rather than glossed:

- **51 unlicensed sections/symbols** the manifest does not declare: 4 in `.text` (`GetEntity`, `GetHazard`, `_ZN16dScMgSmartball_cD2Ev`, `_ZN5TVec2D1Ev`) and 47 in `.data` (the `_ZTI`/`_ZTS`/`_ZTV` triples for all thirteen classes and their bases). Each one alone refuses promotion.
- **4 compiler-only policy errors** — the same four `.text` symbols have no `compiler_only_output` policy.
- **`functions_occur_in_expected_order`: PARTIAL** — ordinal pair (0, 1), the D1/D0 group, whose order the compiler chooses.

## Verification, as run

Pinned `tools/mwccarm/2004/b56/mwccarm.exe`, flags `-O4,p -enum int -lang c++ -char signed -interworking -proc arm946e -gccext,on -msgstyle gcc -Cpp_exceptions off`; comparison by `tools/match.py` (bytes + relocations), `tools/objisolate.py` (relocation type/addend) and `tools/reloc_audit.py` (relocation destination identity). The 107/130 in the commit message, the `functions_matched: 107` in the manifest, and the re-run on this tree all agree.

`tools/asm_policy.py`'s `counts_as_matched` returns `True` for the source and `classify` returns `None` — there is no `NONMATCHING` banner anywhere in the file, in any spelling, so nothing here silently unmatches a neighbouring TU. Nothing was reworded to get that result.

## Third commit: 14 stale conflict notes

`tubuild create` records each union conflict twice — a marker in the source and a note in the manifest. The fold reconciled 14 of them by hand and removed their markers, but their notes still claimed an open conflict, so `check_tubuild_conflicts.py --list` reported 14 unreconciled records against this entry while every other entry reported 0. Each of the 14 was checked against the source individually and then rewritten to the `(RESOLVED)` form; none was deleted. Where two legacy files spelled the same layout differently they now share one declaration (`V2`, `namespace Sound` — `u32` is `unsigned int`), and where they described different memory each got its own name with the originating file named in a comment (`V2ab`, `V2xz`, `C2`, `S28`, `S24`, `SBoard`, `SArrays`, `State2`), all of which are referenced by the members they came from. The remaining 87 markers and 87 notes are genuinely open and stay open.

## Gates

Pinned base `37a72c7400c9671f3abac5034444a054c082a3c3`.

- `git merge-tree --write-tree --name-only --no-messages` — exit 0, no conflict; the merge result adds exactly the two new files and leaves `attribution.json` at main's blob.
- `python tools/check_tubuild_conflicts.py --list` — exit 0, `every recorded conflict matches the source, and every conflict in a source is recorded`; this entry 87 open / 87 markers / **0 problems**.
- `python tools/check_decl_agreement.py --check` — exit 0, `no new declaration disagreements`. Disclosed: the gate's skip list contains `src_tu`, so it never indexes this file. Green here is by construction, not by measurement, and no baseline regeneration was needed or done.
- `python tools/check_src_tu_compiles.py` — `199/199 translation unit(s) compiled`, `every translation unit compiles`.
- `python tools/prepush_attribution.py` — `0 changed, 0 lost` against this branch's merge base. Against main's tip it prints 362 `CREDIT LOST` rows naming classes this branch never touches (`daPeach_c`, `daKinopio_c`, `dScMgSlot1_c`, …); that is the branch predating main's own `attribution.json` updates, not a loss. The branch's diff does not touch `attribution.json` at all, and the merged tree keeps main's blob byte for byte, so no override was added.
- `python tools/eligible.py` — run last, after the final commit: `eligible: 11252 / 11301`, report stamped `e2a929a9344a6cadb1b977d30d1dfc3200c381aa`, which is the pushed head, so `tools/check_references.py` actually ran rather than skipping: `OK: no new unresolvable references` (41 unresolved against a baseline of 45). It also notes `4 fixed since the baseline -- run --update to bank it`; that improvement is not this branch's and was left for whoever earned it.
- pre-push, on the real push: `port-refcheck: 423 checked - all references resolve`, `duplicate-sources: 8646 stems, none doubled`, `check_src_tu_compiles: every translation unit compiles`, `pre-push: unresolved references OK`.

## What a reviewer is agreeing to

That a 130-function fold at 107/130 with three known wrong-destination declarations is better kept in the tree, labelled `shadow` and failing `tubuild verify` out loud, than kept on a branch. If the preference is that `src_tu/` hold only units that pass verification, this should be closed rather than merged — the work is reproducible from the branch either way.


andrewboudreau:
Measuring the question this PR's own description raises, so the decision rests on a number rather than a preference.

**There is no precedent for a `shadow` manifest on main.** Across all 215 manifests in `config/tu_manifest.d/` at `658fd5ec8`:

| status | count |
|---|---|
| `promoted` | 182 |
| `text-verified` | 29 |
| `link-verified` | 3 |
| (none set) | 1 |
| **`shadow`** | **0** |

So merging this would not be following an existing convention — it would establish a new one: that `src_tu/` may hold a unit which fails `tubuild verify`. The two sibling folds landed alongside it this session were `text-verified` (33/33 and 43/47 byte-identical), which is the 29-strong precedent; this one is 107/130 with 51 unlicensed symbols and three declarations carrying wrong relocation destinations.

That is a maintainer's policy call, not a mechanical one, so I am not merging it on my own judgement. Flagging rather than deciding.

**Nothing here is a defect in the work.** The fold is the expensive part and it is honestly labelled: the tally and the `shadow` status lead both the title and the first paragraph, `tubuild verify` fails out loud, no coverage number moves, and the wrong-destination declarations are documented rather than buried. The diff is two new files and touches no ledger, so it cannot dirty or be dirtied by other PRs and can wait without cost.

One correction worth recording against the numbers in the body: the `999 word(s) differ` figure on seven of the twenty-three DIFF rows is not a word count. `tools/match.py:268` returns `999` on a length mismatch *before* comparing any word, so it means "wrong-length compile, never compared". The real gaps on those seven are +15, −2, −1, −4, −4, −1 and −1 words — five of the seven are within four words, which makes them among the closer misses in the unit rather than the worst. The same sentinel is banked in 33 rows of `config/match_attempts.jsonl`, so it affects more than this PR.

PR #2827, opened 2026-09-19
Cleanup ov084/daKrb_c leftover

lunavyqo:
Leftover-clean of ov084/daKrb_c after fold + slot-19 promote (#2798/#2811). Banked text-verified. ov084 only.

Game identity: Goomba (`KURIBO` 200 / `KURIBO_S` 201 / `KURIBO_L` 202). RTTI names `daKrb_c`; base is `dCapEnemy_c`.

## What

Replace the reconstruction essay with game identity + deslop leftovers citing this TU. Keep leftover `return new` factories (`daKrb_c_classInit_KURIBO_L` / `_S` / `_KURIBO`), `#pragma defer_codegen off` emission order, and the `sizeof` wrap. `#pragma long_calls` is not in this TU's legacy sources. `#pragma opt_common_subs off` / `optimize_for_size on` are file-global last-wins and stay uncarried.

## Hunt (this TU)

- Callees: SetAnim / dCcAc Init / dBgCh Init / MaterialChanger::SetFile / DropShadowRadHeight / KillByInvincibleChar / Player::Hurt / Bounce / IsTooFarAwayFromPlayer stay mangled (Fix12<int> by value, 6az). dCapEnemy_c ReleaseCap / RespawnIfHasCap / GetCapEatenOffIt / UnloadCapModel / UpdateCapPos stay mangled in the helpers. Named members already call AddCap / DestroyIfCapNotNeeded / GetCapState / RenderCapModel.
- `data_ov084_02130cf8` BMD; `02130278[7]` BCA table; `02130ce0/ce8/cf0/cc0/cc8/cd0` BCA handles (SetAnim reads `[1]`); `0213089c/0213088c` BMA. Per-type tables `02130258/208/228/238/248/268/204`. S14: `g_profile_KURIBO` / `_S` / `_L` stay outside.
- Sibling `daPkn_c` (same overlay): leftover `return new`, 6az Init/SetAnim, `data_ov084_*` handles, S14 no profile.

## Verify

```
TU ov084/daKrb_c
compiler-only     : exact deadstrip ['_ZN7Vector3D1Ev', '_ZN7daKrb_cD2Ev']
byte comparison   : 35/35 MATCH
objisolate check  : clean
emission order    : all 35 function(s) in the expected ROM-ascending section order
Result: 35/35 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
```

Manifest not dumped (S17).


andrewboudreau:
The `conflicts` check is red, and it's the one red here that isn't advisory. Diagnosis, since the fix is mechanical but not obvious:

`check_tubuild_conflicts.py --list` reports **65 `STALE NOTE`** failures, all `ov084/daKrb_c`. That is exactly the number of `/* TUBUILD CONFLICT -- ... NOT applied: ... */` markers this diff removes from `src/actors/daKrb_c.cpp` — 65 removed, 65 added back, 1:1.

Those markers read as generated slop, but they are half of a two-sided ledger: the other half lives in `config/tu_manifest.d/ov084/daKrb_c.json`, which this PR doesn't touch. The gate enforces both directions — every recorded conflict must have a matching marker in the source, and every marker must be recorded. Removing the source side alone leaves 65 records pointing at nothing.

Two ways to clear it:

1. **Rewrite the manifest notes to the resolved form**, which is the right answer if the fold genuinely settled each one. The tool states the required shape: `tubuild create warning (RESOLVED): CONFLICT: ...`. The generated sentence has to be preserved **verbatim** through `...; kept the first` and the reconciliation text **appended after** it — replacing it with free prose produces `UNPARSED NOTE` instead, which is the same red wearing a different hat. Worth noting that the resolved form is a claim: it says one spelling won and the others are settled, so it's worth naming which survived and why, per note rather than in bulk.

2. **Keep the 65 markers** and deslop only the prose around them.

The rest of the diff looks clean to me. I checked the things that would worry me in a deletion this size:

- `extern "C"` balance: 18 removals vs 16 additions, and the 2 unmatched removals are prose *mentioning* `extern "C"`, not declarations. The 16 real ones are the same line minus a trailing comment. No linkage change.
- No `// recovered name:` lines touched, no `NONMATCHING` marker added or removed, no includes changed.
- The removed `typedef`s are shadows that duplicate the real headers; the three `Vector3` ones are inside already-commented conflict blocks. The new header explicitly keeps `Flags / Flag / Obj / Sub / Vector3_16_local` as load-bearing, which is the distinction that matters.

So this is one gate away. Happy to merge once `conflicts` is green — say the word if you'd rather I push the manifest-side fix to this branch instead of you doing it.

PR #2832, opened 2026-09-19
Cleanup ov102/daObjHatenaBlock_c leftover

lunavyqo:
Question / item / VS-item / cap blocks (`HATENA_BLOCK` 20, `ITEM_BLOCK` 21, `VS_ITEM_BLOCK` 22, `CAP_BLOCK_M/W/L` 23–25).

Leftover hunt from this TU's callees and consumers:

- Factories are `return new daObjHatenaBlock_c()` (inherits `fBase_c::operator new`).
- Init `Model::LoadFile` types ov102 BSS as `SharedFilePtr` (KCL/BMD/BCA). ov002 handles that already have SharedFilePtr plurality follow; `da18`/`da30`/`da58`/`gPFlower*` stay `char[]` (S27, do not tip Goomboss/BrickBlock/PowerFlower).
- Named this TU's fields: `mBounceYOffs`, `mFloorY`, `mBounceAng`, `mBounceTimer`, `mHeldActor`.
- Lifecycle methods call real members (`SetFile`, `LoadFile`, `Render`, `Advance`, `Enable`/`Disable`, `TrackStar`, `UpdatePos`).

Leftover that still MATCH (cited in-source `deslop`):

- `ModelAnim::SetAnim` / `dBgW_KcMbg::SetFile` / `IsClsnInRange` / `DropShadowScaleXYZ` / `NewSimple` (6az).
- `func_ov102_*` helpers (offset soup, PMF tables). Not coined.
- `func_ov102_0214ad14` / `0214b384` belong in `daBmb_c`; `func_ov002_020f0438` in ov002.
- `g_profile_*` outside this TU (S14).
- common.h first (this TU's `02149e38` / `02149ff0`).
- homeless `_ZN10dBgActor_cD2Ev` from `return new`, licensed deadstrip.

`tools/tubuild.py verify ov102/daObjHatenaBlock_c`: 41/41 MATCH, objisolate clean, TEXT-VERIFIED.


andrewboudreau:
`conflicts` is red for the same reason as #2827, just at a much smaller scale — one marker instead of 65:

```
ov102/daObjHatenaBlock_c: STALE NOTE: local declaration typedef Vector3 (from the
legacy file for func_ov102_0214953c) is recorded as an open conflict once, but
src/actors/daObjHatenaBlock_c.cpp does not carry a matching marker.
```

This diff removes one `/* TUBUILD CONFLICT ... */` marker from the source while `config/tu_manifest.d/ov102/daObjHatenaBlock_c.json` still records it as open. The gate checks both directions, so the record now points at nothing.

Fix is either to restore that one marker, or to rewrite the manifest note to the resolved form the tool names: `tubuild create warning (RESOLVED): CONFLICT: ...`. If you take the resolved route, the generated sentence has to survive **verbatim** through `...; kept the first` with your reconciliation text appended after it — replacing it with free prose trades this red for `UNPARSED NOTE`.

The rest reads fine. I checked the two things that would worry me here and both are clean: the `recovered name` hit is prose mentioning the phrase rather than a machine-read marker, and the decl-agreement baseline change is loss-side only — the `+` lines are just trailing-comma re-emissions where the `daObjHatenaBlock_c.cpp` entry below them was deleted, not new entries.

One thing I'd double-check on your side, since no gate covers it: this drops four includes (`dBgW_KcMbg.h`, `Model.h`, `dActor_c.h`, `dBgW.h`) and adds `Sound.h`. Include reachability decides which of the two `Matrix4x3` spellings wins in this tree, and that changes POD-ness and therefore codegen. The ROM build will catch it if it matters, so this is a heads-up rather than an objection.

PR #2852, opened 2026-09-19
Cleanup ov090/daPukupuku_c: promote Cheep Cheep class TU

lunavyqo:
Enrolls text-verified ov090/daPukupuku_c (15 functions, 0x02133034..0x0213367c) as src/actors/daPukupuku_c.cpp.

ROM identity: _ZTS12daPukupuku_c at 0x021342f0, _ZTI at 0x021342e4, _ZTV at 0x02134324 (address unchanged). Factory daPukupuku_c_classInit allocates 0x388, installs _ZTV[2] (emitting TU), target of PUKUPUKU profile at 0x02134300.

  byte comparison   : 15/15 MATCH
  objisolate check  : clean
  Result: 15/15 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
  (quoted from tools/tubuild.py verify ov090/daPukupuku_c, pre- and post-promotion)
  tu_promote: 1 entry promoted, 15 consolidated, 15 attribution overrides, 5 CONVERTED retained.

Licensing: 11 EXTRA .data as deadstrip-data (data: []), plus D2 deadstrip and Vector3D1 deadstrip-duplicate. Single complete span, no + in dest, no _ZTV move. Leftover quality: Cheep Cheep identity + deslop reasons, _ZTV decl once above factory at namespace scope, sizeof wrap kept. Hand-rolled factory and offset helpers kept as measured-MATCH leftovers (return-new/full-member forms unmeasured).

Mechanical: tools/tu_promote.py (git mv R100, delinks splice, manifest flip, attribution, converted-baseline migration). No S21 rename needed (already ROM names on origin/main). One overlay (ov090) only; no other ov090 promotion in this PR.


andrewboudreau:
`PR validation` fails here for exactly one reason, and it is two symbols out of fifteen.

From the validator's full-ROM block:

```
FAIL src/actors/daPukupuku_c.cpp: isolate: requested functions are not emitted in ROM order
```

Diffing the two lists position by position: the **sets are equal**, and only indices 0 and 1 differ.

|---|---|---|
| 0 | `_ZN12daPukupuku_cD1Ev` | `_ZN12daPukupuku_cD0Ev` |
| 1 | `_ZN12daPukupuku_cD0Ev` | `_ZN12daPukupuku_cD1Ev` |

All thirteen other symbols — every `func_ov090_*`, `CleanupResources`, `OnPendingDestroy`, `Render`, `Behavior`, `InitResources`, `daPukupuku_c_classInit` — are already in the right order. The TU is one transposition away from isolating.

### Why they come out swapped

`src/actors/daPukupuku_c.cpp:485` defines the destructor **out of line**:

```cpp
daPukupuku_c::~daPukupuku_c()
```

`.claude/skills/decomp-cpp-class-form/SKILL.md` has this measured, and the out-of-line row is the losing one:

|---|---|---|---|

Note that out-of-line loses in *both* rows, so no amount of instantiation rescues it. The winning form needs **both** conditions: destructor defined inline in the class body, **and** the TU actually instantiating the class.

The second condition is the one this PR's own header flags as unfinished:

> `Leftover: factory stays hand-rolled with &_ZTV12daPukupuku_c[2] (+8: this TU defines the vtable). The return-new form is unmeasured for this base+member ctor order.`

`daPukupuku_c_classInit` currently calls `fBase_c::operator new` and stores the vptr by hand (`p[0] = (int)&_ZTV12daPukupuku_c[2];`). That is allocation, not instantiation — the compiler never sees a `daPukupuku_c` constructed, so the inline form on its own would emit no variants at all. Both halves have to move together.

### A second thing it buys

The manifest currently carries `_ZN12daPukupuku_cD2Ev` (json:133, 248, 252) because the out-of-line form emits a D2 the cartridge has no home for. The skill notes the inline form emits no D2 at all, so those rows stop being needed rather than needing a policy to absorb them.

### Caveat

Per `AGENTS.md`, changing a declaration or a definition's form is matching work, not a text edit — it can move instruction selection at the call sites. This needs a rebuild and byte proof (`tools/match.py`, then `tubuild verify`) before it is worth re-requesting validation, not just a green isolate.

#2854 and #2855 fail with the identical signature — same transposition, same out-of-line destructor, same hand-rolled factory — so one worked example should carry all three.

andrewboudreau:
The destructor fix in `a49bfcb07` is right, and I re-measured it independently in a worktree rather than taking `PR validation`'s word for it:

```
byte comparison : 15/15 MATCH
objisolate      : clean (relocation type/addend)
reloc-audit     : clean (destination identity)
emission order  : all 15 functions in ROM-ascending section order
```

Inline-first destructor plus a real `new` in the factory, both halves together — exactly the measured row.

One thing is still red and it is not inherited. `references` names two citations, both in `src/actors/daPukupuku_c.cpp`:

```
names `src/d_a_pukupuku.c`, which is not in the tree
names `src/func_ov090_02133200.c`, which is not in the tree
```

Promotion is what deletes those two shards, so both went dead the moment the TU landed under `src/actors/`. `origin/main` is green on this gate and the failing run names only this file, so this does not qualify for the merge-tree carve-out that covers a red inherited from a broken base — it has to be fixed here.

I could not push to this branch (it is on your fork), so the fix is in #2861: the same commits, your authorship intact, plus one commit that rewords both sentences to name the absorbed function by **symbol**. Not stripping the `src/` prefix — that only hides the token from the scanner while leaving the prose pointing at nothing.

Either land #2861 and close this, or cherry-pick that one commit here and I will close #2861. Your call; nothing else in this PR needs to change.

andrewboudreau:
Landed as `91e95e65f` via #2861 — the `ov090/daPukupuku_c` TU is on `main`.

Your destructor work is what made it mergeable: the inline `virtual ~daPukupuku_c() {}`
in the class body plus the TU actually instantiating the class is the only form that
emits `D1` before `D0`, and it retires the homeless `D2` licence row at the same time.
I verified it independently before merging rather than leaning on the validator —
15/15 functions byte-identical, `objisolate` clean, `reloc_audit` clean, and a full
`rombuild` at **11,208 functions reproducing, 0 mismatching, 106/106 modules exact**.

The only thing #2861 added on top of your branch was the `references` fix. `tubuild
create` writes the "Assembled from these legacy one-function sources" block as
repo-rooted `src/` paths, and promotion is precisely the step that deletes those files —
so every hand-assembled TU promotion reds that check on its own output. The repair is to
keep the block, the ordinals and the ROM addresses, and name the **symbol** instead of
the path. Stripping the `src/` prefix would only hide it from the checker.

Nothing further needed here — this PR can close as superseded.

PR #2853, opened 2026-09-19
Cleanup ov085/daObj_Mip_Key_c: promote Mip Key class TU

lunavyqo:
Promote Mip Key's 15 functions (2,208 bytes) into `src/actors/daObj_Mip_Key_c.cpp`, preserving the original symbol credits and contiguous ROM enrollment. The repaired C++ uses the actual class fields, native pointer-to-member state dispatch, ordinary Vector3 locals, and the existing Model, Player, Camera, Message and resource interfaces. It removes the fabricated host/vtable views and fixes incorrect call contracts.

The integration preserves the PR author's latest contribution and reconciles only this PR's ledger entries. Five existing converted identities follow their symbols into the TU without increasing the baseline. The declaration ledger retains seven live Mip entries and drops fourteen obsolete contracts. The promotion queue records the measured one-file, 360-line source.

Reconstruction remains partial: seven address-named helper entry points and measured matrix/fixed-point interface limitations remain under [follow-up #2865](https://github.com/tangosdev/sm64ds-decomp/issues/2865), owned by `codex-integrator-0920`. The committed experiment note records the tested alternatives; matching does not establish original file-static linkage or recovered original names. This promotion owns the text; the manifest explicitly accounts for compiler-emitted metadata and the existing licensed Vector3 destructor duplicate.

Validation against current main:

- Full ROM build: 106/106 exact modules, 11,208 reproducing functions, zero mismatches; all 26 source-data claims and three BSS claims retained.
- All 15 affected functions pass strict relocation checks with empty differences and zero blind words. Whole-TU ordering, isolation and complete emitted-object policy pass.
- TU data: six verified records / 180 bytes, five explicitly partial records / 52 bytes, zero differing or unnamed records.
- Header layout and consumers, declarations, converted ratchet, port references and documentation links pass. Attribution preserves all 15 credits: 12 tangosdev, two ruspecial and one andrewboudreau; none changed or lost.
- The promotion-flag audit passes. Unrelated stale measurements in the full queue audit remain unchanged.

See `notes/experiments/mip-key-2853-source-forms.md` and the producer/composition handoffs for the source evidence, measured limits and independent review trail. Required private validation and exact-head/base Source review must both pass before merge.


andrewboudreau:
`PR validation` is green here — the byte work is done. What is left is four advisory gates, and one thing worth saying up front: **this PR is mergeable by GitHub's rules and should not be merged yet.** `mergeStateStatus` is `UNSTABLE` rather than `BLOCKED` only because all four failures are advisory. All four also run on `main`, where they are currently green, so landing this as-is turns `main` red on four checks.

Three of the four are one item each. The fourth looks like twenty and is really one.

### 1. `ratchet` — one unbanked tier entry

```
AssertionError: Lists differ:
  ['src/actors/daObj_Mip_Key_c.cpp#_ZN15daObj_Mip_Key_c6RenderEv'] != []
```

The fold re-keyed `Render` onto the promoted TU and the tiers baseline has not been re-banked. Worth doing on the tip you actually push: `tiers_ratchet --update` appends rather than replaces, so a rebase plus a second run banks the row twice.

### 2. `references` — one stranded citation

```
FAIL: 1 prose/comment reference(s) name a path that does not exist:
  notes/tu-filename-reconstruction-pilot.md
      names `src/d_a_obj_mip_key.c`, which is not in the tree
```

The rename to `src/actors/daObj_Mip_Key_c.cpp` did not carry into that note.

(The same run reports 5 prose and 515 C/C++ comment references that *now* resolve and could be un-banked with `--update` / `--update-code`. It flags those as "not a failure" — a separate PR, not this one.)

### 3. `declarations` — 20 findings, 13 of 14 symbols already banked elsewhere

This one reads much worse than it is. The 20 findings cover 14 distinct symbols, and I checked each against `config/decl-agreement-baseline.json` on `main`:

| symbol | already banked on `main` in |
|---|---|
| `_ZN5Model8LoadFileER13SharedFilePtr` | 122 files — **including a `daObj_Mip_Key_c` shard** |
| `func_02012790` | 83 files |
| `Vec3_HorzAngle` | 78 files |
| `_Z14ApproachLinearRiii` | 68 files |
| `_ZN8dActor_c19DropShadowRadHeight…` | 64 files |
| `MulVec3Mat4x3` | 48 files |
| `data_0209f318` | 45 files |
| `Matrix4x3_FromTranslation` | 42 files |
| `Vec3_Asr` | 36 files |
| `_ZN5Sound7PlaySubEjjj5Fix12IiEb` | 28 files |
| `Vec3_VertAngle` | 17 files |
| `_ZN11ShadowModel12InitCylinderEv` | 10 files — **including a `daObj_Mip_Key_c` shard** |
| `data_ov085_0213072c` | 1 file |
| `data_ov085_0213071c` | **0 — genuinely new** |

The baseline is keyed `symbol -> { path: [disagreements] }`, so a promotion that folds `src/_ZN15daObj_Mip_Key_c6RenderEv.cpp` into `src/actors/daObj_Mip_Key_c.cpp` **orphans every banked entry that lived at a shard path**, and the identical disagreement re-surfaces at the new path as new. Two of these are provably that: `_ZN5Model8LoadFileER13SharedFilePtr` and `_ZN11ShadowModel12InitCylinderEv` are banked on `main` under a Mip_Key shard path right now.

The rest are the same spelling this tree already accepts everywhere else — `Vec3_HorzAngle` as `(const void *, const void *)` is banked in 78 files, `MulVec3Mat4x3` as `const void *` in 48. Chasing them into agreement here would make this TU the only file in the tree that does, and per `AGENTS.md` each one is matching work needing a rebuild and byte proof, which the gate explicitly says it will never ask for.

So these should be **re-banked, not fixed**. One caution on how: a bare `--update` on the merged tree also banks anything genuinely new. Diff two regenerations — one on pristine `main`, one on the merged tree — and bank the difference deliberately.

**The one that is actually new is worth a sentence in the PR body, because it is the PR being right:**

```
data_ov085_0213071c: return
  src/actors/daObj_Mip_Key_c.cpp:163  declares  MipKeyPMF
  include/decl_common.h:338  plurality is  char
```

`decl_common.h` carries the generated placeholder (`char`, and `unsigned int` for `0213072c`); this TU has recovered that both are pointer-to-member-function tables. The gate scores data symbols by plurality, so the more accurate declaration loses to 45 copies of the placeholder. That is a deliberate, better disagreement — bank it and say so, rather than reverting to `char`.

### 4. `promoted` — the queue cell was not flipped

```
queue_audit: FAIL -- 1 row(s) disagree.
Regenerate with 'python tools/queue_audit.py --write' (needs the ROM dump)
```

`notes/data/tu-promotion-queue.tsv` still describes this class as unpromoted. Flip it only if the manifest status is exactly `promoted` — on `text-verified`, or a partial fold, `no` is the honest answer.

None of this touches the bytes, which is the part already proved.

PR #2854, opened 2026-09-19
Cleanup ov098/daObjYajirusi_c: promote Arrow Sign class TU

lunavyqo:
S21 rename ArrowSignRight -> daObjYajirusi_c, then enroll text-verified TU (12 functions, 0x02137be0..0x02138040) as src/actors/daObjYajirusi_c.cpp.

ROM identity confirmed by hand: _ZTS15daObjYajirusi_c at 0x0213c36c, _ZTI at 0x0213c360, _ZTV at 0x0213c3d8 (addresses unchanged; coined _ZTV14ArrowSignRight alias deleted first). Rename ran without --with-derived: no live derived symbols exist (factories/profiles already ROM), and the flag would have rewritten TSV history and Historical-alias comments.

Twin _L factory folded: daObjYajirusi_c_classInit_YAJIRUSI_L (0x02138008) is the same class's second profile factory (identical body: alloc 896, dBgActor_c C2, same vtable, ShadowModel C1 at +0x320), contiguous with no gap -- one genuine TU, S20-clean.

  byte comparison   : 12/12 MATCH
  objisolate check  : clean
  Result: 12/12 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
  romdata_check --files: DIFFERS 0 (6 VERIFIED/184B, 5 PARTIAL, 0 UNNAMED)
All quotes from tools/tubuild.py verify ov098/daObjYajirusi_c (pre- and post-promotion).

Licensing: D2 deadstrip, Vector3D1 deadstrip-duplicate (Kill's pos/dustPos), 11 data deadstrip-data (own 3 at ov098 homes + inherited fBase/dBase/dActor arm9, dBgActor ov002). Pilot externalized_output stays as partitioned-link evidence; production reads compiler_only_output. Single complete span, no + in dest, no _ZTV move.

Leftover quality: Arrow Sign identity + deslop reasons; _ZTV decl once above factories at namespace scope; sizeof wrap kept; hand-rolled +2 factories kept as measured; func_ov098_02137c8c ROM-unnamed; stale ArrowSignRight_Kill decl dropped from decl_common.h; ref-vs-pointer guesses disclosed in manifest notes, actor_renames.tsv why, daObjYajirusi_c.h. Sibling daObjFallBlock_c.json stale sibling-conclusions updated (this PR falsifies them); tools prose paths + notes/bgobject-provenance updated for the rename; header Kill comment repointed at src/actors/dBgActor_c.cpp (S26).

Hazard disclosed: tu_promote is not atomic -- it crashed mid-run here (git mv applied, then git rm refused on S21-staged files, uncaught traceback). Remaining steps were completed by driving the tool's own writers (same code, same outputs); final verify confirms 12/12 from the production path. Upstream hardening in a separate tooling PR.

One overlay (ov098) only.


andrewboudreau:
This PR has two independent blockers. The second is the same one-transposition problem as #2852 and #2855.

### 1. Port reference check — 6 stale references

```
port refcheck: 6 stale reference(s)
  port/slice_gate9.txt:5: src/d_a_obj_yajirusi_yajirusi_r.c not found (renamed?)
  port/slice_gate9.txt:6: src/_ZN14ArrowSignRight13InitResourcesEv.cpp not found (renamed?)
  port/slice_gate9.txt:7: src/_ZN14ArrowSignRight16CleanupResourcesEv.cpp not found (renamed?)
  +3 more
```

`port/` points into `src/` by literal path and nothing in the decomp toolchain compiles it, so the promotion's file moves stranded these. `python tools/port_refcheck.py` reproduces it in about a second; `AGENTS.md` asks for it before pushing anything that renames or moves a `src/` file.

### 2. `isolate` — D1 and D0 are transposed

```
FAIL src/actors/daObjYajirusi_c.cpp: isolate: requested functions are not emitted in ROM order
```

Diffing the two lists position by position: the **sets are equal**, and only indices 0 and 1 differ.

|---|---|---|
| 0 | `_ZN15daObjYajirusi_cD1Ev` | `_ZN15daObjYajirusi_cD0Ev` |
| 1 | `_ZN15daObjYajirusi_cD0Ev` | `_ZN15daObjYajirusi_cD1Ev` |

All ten other symbols are already in ROM order.

`src/actors/daObjYajirusi_c.cpp:312` defines the destructor **out of line**:

```cpp
daObjYajirusi_c::~daObjYajirusi_c()
```

which is the losing row in the measured table in `.claude/skills/decomp-cpp-class-form/SKILL.md`:

|---|---|---|---|

Out-of-line loses in *both* rows, so instantiation alone does not rescue it. The winning form needs **both**: destructor defined inline in the class body, **and** the TU actually instantiating the class. This PR's own note records the second half as unfinished —

> `Leftover: factories stay hand-rolled with (_ZTV15daObjYajirusi_c + 2)`

— and a hand-rolled `operator new` plus a manual vptr store is allocation, not instantiation, so the compiler never sees the class constructed. Both halves have to move together.

It would also retire the `_ZN15daObjYajirusi_cD2Ev` rows (json:132, 136, 191) and the long `compiler_only_output` justification behind them: the inline form emits no D2 at all.

### Caveat

Per `AGENTS.md` this is matching work, not a text edit — changing the destructor's form can move instruction selection at the call sites, so it needs a rebuild and byte proof before re-requesting validation.

lunavyqo:
Both blockers addressed in 406c3548b3 (pushed):

1. Port slice: the six promoted legacy paths in port/slice_gate9.txt now resolve to src/actors/daObjYajirusi_c.cpp (consolidating comment in the file's own style); the folded func_ov098_02137c8c.c entry dropped. Local port-refcheck: 418 checked, all resolve.

2. Transposition: destructor is now inline-first in daObjYajirusi_c.h and both factories are real new-expressions. 12/12 MATCH holds -- the synthesized constructor reproduces the ROM init sequence (alloc 896, dBgActor_c C2, vptr, ShadowModel C1 at +0x320) byte-exact -- and the strict production isolate (derive_many after deadstrip, the check tubuild-verify does not run) reports ORDER OK; verify itself now prints all-ascending emission. Own-D2 license row retired (inline form emits no D2); the dragged base _ZN10dBgActor_cD2Ev licensed deadstrip per the ov012/daObjC0Water_c precedent.

Additionally, since the promotion relocated banked declaration debt: config/decl-agreement-baseline.json moves the 10 legacy-file rows to the promoted file with identical details (minus the healed 380 entry); check_decl_agreement reports no new disagreements. Dead-reference lineage refs left untouched per baseline policy -- post-merge alignment.

PR #2855, opened 2026-09-19
Cleanup ov077/daPopoi_c: promote Heave-Ho class TU

lunavyqo:
S21 rename HeaveHo -> daPopoi_c, then enroll text-verified TU (22 functions, 0x0212624c..0x02127230) as src/actors/daPopoi_c.cpp. Two commits: rename+prep, then promotion on a clean tree.

ROM identity confirmed by hand: _ZTS9daPopoi_c at 0x02127a50, _ZTI at 0x02127a68, _ZTV at 0x02127a98 (addresses unchanged; coined _ZTV7HeaveHo alias deleted first). Rename ran without --with-derived: no live derived symbols exist (factory/profile already ROM), preserving TSV history and Historical-alias comments (HeaveHo_Spawn, HeaveHo_SpawnInfo).

  byte comparison   : 22/22 MATCH
  objisolate check  : clean
  Result: 22/22 MATCH, objisolate clean, reloc-destinations clean -> TEXT-VERIFIED
  romdata_check src_tu: 5 VERIFIED/56B, 6 PARTIAL/168B, DIFFERS 0, UNNAMED 0
All quotes from tools/tubuild.py verify ov077/daPopoi_c (pre- and post-promotion, production path).

Licensing: D2 deadstrip, Vector3D1 deadstrip-duplicate (Behavior/2679c/26ad0/26640 block-scope Vector3s), 11 data deadstrip-data (own 3 at ov077 homes + inherited fBase/dBase/dActor arm9, dEnemyBase ov002). Single complete span, no + in dest, no _ZTV move.

Leftover quality: identity + deslop reasons in daPopoi_c.h (struct/dtor/typedef/_ZTV ROM-named; 0x42c sizeof-from-new(1068) + member map preserved); _ZTV decl once above factory at namespace scope (the hand-rolled +2 factory kept as measured); PMF-shadow + AngleDiff disclosures already in manifest notes; lineage [21] names the true historical src/d_a_popoi_c path (tu_promote rendered a never-existent symbol-derived filename -- fixed, still 22/22).

One overlay (ov077) only. Full 106/106 ROM proof pending CI.


andrewboudreau:
`PR validation` fails here for exactly one reason, and it is two symbols out of twenty-two.

```
FAIL src/actors/daPopoi_c.cpp: isolate: requested functions are not emitted in ROM order
```

Diffing requested against emitted position by position: the **sets are equal**, and only indices 0 and 1 differ.

|---|---|---|
| 0 | `_ZN9daPopoi_cD1Ev` | `_ZN9daPopoi_cD0Ev` |
| 1 | `_ZN9daPopoi_cD0Ev` | `_ZN9daPopoi_cD1Ev` |

All twenty other symbols are already in ROM order. The TU is one transposition away from isolating.

### Why they come out swapped

The manifest already states the mechanism precisely:

> `src_tu/actors/daPopoi_c.cpp carries a single out-of-line daPopoi_c::~daPopoi_c() and mwccarm emits the whole variant set from that one definition`

That out-of-line form (`src/actors/daPopoi_c.cpp:1266`) is the losing row in the measured table in `.claude/skills/decomp-cpp-class-form/SKILL.md`:

|---|---|---|---|

Out-of-line loses in *both* rows, so instantiation alone does not rescue it. The winning form needs **both**: the destructor defined inline in the class body, **and** the TU actually instantiating the class. Right now `Spawn` allocates and stores the vptr by hand (`p[0] = (int)&_ZTV9daPopoi_c[2];`), which is allocation, not instantiation — so moving the destructor inline on its own would emit no variants at all. Both halves move together or neither works.

### A second thing it buys

The manifest spends a `deadstrip` row on `_ZN9daPopoi_cD2Ev` with a careful justification that no ROM home exists for it:

> `Nothing in config/ names a D2 for this class ... rombuild's symbol-home index resolves _ZN9daPopoi_cD2Ev to no home at all.`

That reasoning is correct, and the inline form makes it moot — it emits no D2, so the row stops being needed rather than needing to be justified.

### Caveat

Per `AGENTS.md` this is matching work, not a text edit: changing the destructor's form can move instruction selection at the call sites. It needs a rebuild and byte proof before re-requesting validation — a green isolate alone is not the gate.

#2852 and #2854 fail with the identical signature, so one worked example should carry all three.

andrewboudreau:
The destructor fix in `09d72b9cb` is right. I re-measured it independently in a worktree rather than waiting on `PR validation`:

```
byte comparison : 22/22 MATCH
objisolate      : clean (relocation type/addend)
reloc-audit     : clean (destination identity)
emission order  : all 22 functions in ROM-ascending section order
```

And a full `tools/rombuild.py -j16` on the same fix: **11,208 functions reproducing, 0 mismatching, 106/106 modules exact, 100.000000% of compared bytes, ROM-build analysis PASS.** The synthesized constructor reproduces the ROM init sequence byte-exact — alloc `0x42c`, `dEnemyBase_c`'s base ctor, the vptr store, then `dCcAc_c`, `dCcAcPos_c`, `dBgCh_Actr`, `ModelAnim`, `ShadowModel` at `0x110/0x144/0x184/0x340/0x3a4`. `symbols/profile_reconstruction_registry.tsv` already recorded this factory's shape as `new_plus_inlined_ctor`, so the tree agreed in advance.

One thing is still red and it is not inherited. `references` names 22 citations, all in `src/actors/daPopoi_c.cpp` — every row of the `Assembled from these legacy one-function sources` block. That block was written before promotion, and promotion is exactly what deletes those 22 shards. `origin/main` is green on this gate and the failing run names only this file, so this does not qualify for the merge-tree carve-out that covers a red inherited from a broken base.

I could not push to this branch (it is on your fork), so the fix is in #2862: the same commits, your authorship intact, plus one commit that keeps the list — ordinals and ROM addresses — and names each absorbed function by its **symbol** instead. Not stripping the `src/` prefix; that only hides the tokens from the scanner.

Either land #2862 and close this, or cherry-pick that one commit here and I will close #2862. Nothing else in this PR needs to change.

andrewboudreau:
Landed as `caa78e510` via #2862 — the `ov077/daPopoi_c` TU is on `main`.

Your `09d72b9cb` is what fixed it. The inline `virtual ~daPopoi_c() {}` in the class
body, combined with the TU actually instantiating the class in `daPopoi_c_classInit`,
is the only form that emits `D1` ahead of `D0`; every out-of-line spelling loses that
order whether or not the TU instantiates. It also emits no `D2` at all, which is what
let the homeless `_ZN9daPopoi_cD2Ev` licence row retire instead of needing a
justification. I verified independently before merging — 22/22 functions byte-identical,
`objisolate` clean, `reloc_audit` clean, full `rombuild` at **11,208 functions
reproducing, 0 mismatching, 106/106 modules exact**.

One note for next time, since it cost me a detour: putting `#include "daPopoi_c.h"`
inside the `extern "C" { }` block byte-verifies exactly the same as outside it. Either
is fine — worth knowing so it doesn't read as a difference that matters.

#2862 added only the `references` repair on top of your branch: `tubuild create` writes
the "Assembled from these legacy one-function sources" block as repo-rooted `src/` paths,
and promotion is the step that deletes those files, so every hand-assembled TU promotion
reds that check on its own output. Keep the block, the ordinals and the ROM addresses,
and name the **symbol** rather than the path.

Nothing further needed here — this PR can close as superseded.

PR #2801, opened 2026-09-19
Fold ov002/dEnemyBase_c into one translation unit (31/31 MATCH)

andrewboudreau:
Folds the 31 `.text` functions at `ov002:0x020ad838-0x020aedbc` into a single
translation unit, `src/actors/dEnemyBase_c.cpp`, and deletes the 31 legacy
one-function sources it replaces.

## Original producer verification

`tubuild verify ov002/dEnemyBase_c`:

```
byte comparison   : 31/31 MATCH  (relocation-aware)
objisolate check  : clean
reloc-destinations: clean
emission order    : all 31 function(s) in the expected ROM-ascending section order
Result: 31/31 MATCH -> TEXT-VERIFIED
```

`rombuild.py`: `1 compiled`, 11,208 / 11,208 source-built functions reproducing,
0 mismatching, module fidelity 106/106 exact (100.000000%).

## Why this supersedes `ov002/Enemy`

A manifest entry `ov002/Enemy` already claimed this exact `.text` span at status
`text-verified` with 30 of the 31 members. It is replaced rather than edited,
because the coined name `Enemy` can never satisfy `queue_audit --check-promoted`
-- that gate keys a manifest entry's last path segment against the promotion
queue's `class_name`, and the ROM's own RTTI spells this class `dEnemyBase_c`
(`_ZTS12dEnemyBase_c` at `ov002:0x021081cc`). The two members this adds are
`_ZN12dEnemyBase_c14UpdateYoshiEatER10dBgCh_Actr`, whose matching source landed
after that entry was written, and ordinal 1, now spelled as the
`dEnemyBase_c::KillByInvincibleChar` method its shard recovered rather than the
placeholder `func_ov002_020ada40`.

## Form

This TU owns the class's key function -- `~dEnemyBase_c` is the first non-inline
virtual the header declares and this TU defines it -- so mwccarm emits the whole
ancestor chain's vtable and typeinfo here as vague-linkage passengers. Eleven
such symbols are recorded as `compiler_only_output`, each with the cartridge home
its bytes are checked against. One `~dEnemyBase_c()` definition emits D2, D0 and
D1 as three sections in the cartridge's own order; they are carried as one
definition with stacked marker comments, since carrying them as three separate
definitions is an `object redefined` error.

Four source-level improvements were measured to cost nothing and are included:
the file-local `struct SurfaceInfo { s32 pad; }` shadow is dropped for the real
0x14-byte `include/SurfaceInfo.h`; the `MegaV3` POD shadow is dropped for
[...]

PR #2887, opened 2026-09-21
Promote ov002/daObjBlockL_c to a single translation unit (14/14 TEXT-VERIFIED)

lunavyqo:
Folds the 14 one-function ov002 big-brick-block shards into src/actors/daObjBlockL_c.cpp over the contiguous linker run 0x020b35a0..0x020b406c, replacing the per-function delinks entries with one complete span. Builds on the two prior commits on this branch: the tubuild shadow reconstruction (14/14 TEXT-VERIFIED) and the rename of the coined BigBrickBlock alias to the ROM RTTI name daObjBlockL_c.

IDENTITY IS THE CARTRIDGE'S. ov002 holds _ZTS13daObjBlockL_c at 0x02108a14; no BigBrickBlock RTTI record exists anywhere. The class factories (daObjBlockL_c_classInit_*) and the vtable-preamble ZTI at 0x021089f8 prove the ROM name; _ZTV13daObjBlockL_c sits at the measured address point 0x02108adc, unchanged by this PR.

Gates, real output:
- tubuild verify ov002/daObjBlockL_c: 14/14 MATCH, objisolate clean, reloc-destinations clean. Emission order PARTIAL on ordinal pair (0,1) only -- out-of-line destructor, compiler-chosen D0-ahead-of-D1 order (pilot report sec 3), not a bug.
- decl agreement --changed origin/main: no new disagreements (baseline regenerated, not hand-resolved; the 5 carried verbatim 6az/shadow-table spellings re-bank under the folded path).
- dead references: zero hits on touched files (repointed Kill cites at src/actors/daObjFallBlock_c.cpp and src/actors/dBgActor_c.cpp, Crate D0 cite at the promoted TU; legacy paths live on only in manifest legacy_source).
- Sibling consumer src/_ZN10BrickBlock8BehaviorEv.cpp re-matched strict-relocs under the edited header; Crate D0 re-matched (ov098, strict).

Deliberate divergences from the newest convention notes, called out explicitly:
- Text-only promotion: manifest externalizes _ZTV/_ZTI/_ZTS to canonical (deadstrip-data), claims no data/BSS -- same route as the dEnemyBase_c fold (#2801).
- Out-of-line destructor with D1/D0 markers, following the dEnemyBase_c precedent rather than the inline-destructor wave; the homeless D2 is licensed deadstrip.
- Shadow resource-table structs and mangled Fix12-by-value externs carried verbatim (6az wall); no shared header owns those rows yet.


andrewboudreau:

```
FAIL src/actors/daObjBlockL_c.cpp: isolate: requested functions are not emitted in ROM order
```

- differing indices — **exactly `[0, 1]`**, out of 14 symbols
- `[0]` ROM wants `_ZN13daObjBlockL_cD1Ev`, compiler emitted `_ZN13daObjBlockL_cD0Ev`
- `[1]` ROM wants `_ZN13daObjBlockL_cD0Ev`, compiler emitted `_ZN13daObjBlockL_cD1Ev`

### Cause

`include/daObjBlockL_c.h` declares the destructor **out of line**:

```c
virtual ~daObjBlockL_c();
```

measured to bite:

   green.

### Specific to this PR

`include/daObjBlockL_c.h` already states the coupling in its own comments —
*"`~daObjBlockL_c()` above is declared out of line"* and *"stays the first
out-of-line virtual, so these five TUs do not emit `_ZTV13daObjBlockL_c`"*. So
risk 2 above is not hypothetical here: it is the header's documented design.

The manifest's `deadstripped` list carrying `_ZN13daObjBlockL_cD2Ev` is the
out-of-line form confessing — the inline form emits no `D2`, so that row stops
being needed rather than needing a policy to absorb it.

---

andrewboudreau:
Pushed two commits to this branch (`c9b628fc0`, `9a78fd019`) — flagging them since I edited someone else's PR. **The promotion itself is sound**; both are steps the promotion needed alongside it.

`PR validation` was already green on `ff8ce648c`: all 3 changed files compile byte-for-byte, **16/16 relocation slots VERIFIED**, +0 byte-verified regression, 0 contributor credit lost. That part needed nothing from me.

**1. `promoted` gate was red** (`c9b628fc0`)

`config/tu_manifest.d` records `daObjBlockL_c` as promoted, but `notes/data/tu-promotion-queue.tsv` still carried `already_promoted=no`:

```
daObjBlockL_c  ov002  is promoted but the queue says no (has 'no', want 'yes')
queue_audit: FAIL -- 1 row(s) disagree.
```

I flipped that single cell rather than running `queue_audit.py --write`, which needs the ROM dump and rewrites unrelated rows whose blocker text has drifted — churn this branch didn't earn.

**2. A stale `src_tu` shadow, which no CI job would have caught** (`9a78fd019`)

Flipping the manifest to `promoted` un-declares `src_tu/actors/daObjBlockL_c.cpp`, so nothing knows its compiler pin and nothing compiles it:

```
coverage  src_tu/actors/daObjBlockL_c.cpp: a translation unit on disk that the
          manifest does not declare -- nothing knows its compiler pin, so
          nothing compiles it and this gate cannot vouch for it
```

Deleting it is what every other promotion does — I checked: of the **205** classes the manifest marks promoted, **204** have no `src_tu` shadow. This was the only one left behind.

Worth saying plainly: `tools/check_src_tu_compiles.py` **never runs in CI**. It appears in `.github/workflows/` only inside comments, so a green PR page would not have told you about this. I ran it locally against the merge tree.

After both commits, on `origin/main` + this branch: `check_src_tu_compiles` reports 260/260 TUs compile, `queue_audit --check-promoted` agrees with the manifest on every row, and `check_dead_references`, `check_src_tu`, `check_decl_agreement --changed`, `prepush_attribution`, `port_refcheck`, `check_duplicate_sources`, `check_tubuild_conflicts`, `check_rename_ledger` and `check_data_definitions` all pass.

One incidental note, in case it looked alarming: the validator's per-file table lists `src/_ZN19daObjBlockItemTag_c8BehaviorEv.cpp`, which this branch never touches. That's the header-expansion rule doing its job — you edit `include/daObjBlockL_c.h`, and `validate` compiles every consumer of a changed header to prove their codegen didn't move.

PR #2888, opened 2026-09-21
Promote ov002/daObjBlockItemTag_c to a single translation unit (10/10 TEXT-VERIFIED)

lunavyqo:
Folds the 10 one-function ov002 item-tag-block shards into src/actors/daObjBlockItemTag_c.cpp over the contiguous linker run 0x020b415c..0x020b45e0, replacing the per-function delinks entries with one complete span. Builds on two prior commits on this branch: the rename of the coined BrickBlock alias to the ROM RTTI name daObjBlockItemTag_c, and the tubuild shadow reconstruction (10/10 TEXT-VERIFIED).

IDENTITY IS THE CARTRIDGE'S. ov002 holds _ZTS19daObjBlockItemTag_c at 0x02108b88; no BrickBlock RTTI record exists anywhere. The 31-slot table at 0x02108c18 carries this class's own _ZTI at its typeinfo word, has exactly dActor_c's extent, and overrides only slots 0, 3, 6, 16 and 17 (measured in build/rtti_vtables.json). Four adjacent factories (0x020b45e0..0x020b4670) independently allocate 0xdc bytes, call dActor_c's constructor and install this same table. The coined _ZTV10BrickBlock alias is renamed at its unchanged address 0x02108c18 via tools/class_rename.py (idempotent, 33 edits + 11 renames).

Gates, real output:
- tubuild verify ov002/daObjBlockItemTag_c: 10/10 MATCH, objisolate clean, reloc-destinations clean. Emission order PARTIAL on ordinal pair (0,1) only -- out-of-line destructor, compiler-chosen D0-ahead-of-D1 order (pilot report sec 3), not a bug.
- dead references: the one runbook exemplar pointing at a folded shard reworded as deliberate history and re-banked; no other hits on touched files.
- Header fan-out (affected_src) is just the promoted TU itself; sibling BigBrickBlock references untouched (that rename is unmerged PR #2887 -- this branch will need a rebase cite update for the (BigBrickBlock *) cast in Behavior once it lands).

Deliberate divergences, called out explicitly:
- Text-only promotion: manifest externalizes _ZTV/_ZTI/_ZTS to canonical (deadstrip-data), claims no data/BSS -- same route as the dEnemyBase_c fold (#2801) and the daObjBlockL_c promotion (#2887).
- Out-of-line destructor with stacked D1/D0 markers on one definition, following the dEnemyBase_c precedent; homeless D2 licensed deadstrip.
- Resource handles stay char[] with address-of casts (duplicate int spellings dropped as codegen-neutral); helpers stay mangled externs.


andrewboudreau:

```
FAIL src/actors/daObjBlockItemTag_c.cpp: isolate: requested functions are not emitted in ROM order
```

- differing indices — **exactly `[0, 1]`**, out of 10 symbols
- `[0]` ROM wants `_ZN19daObjBlockItemTag_cD1Ev`, compiler emitted `_ZN19daObjBlockItemTag_cD0Ev`
- `[1]` ROM wants `_ZN19daObjBlockItemTag_cD0Ev`, compiler emitted `_ZN19daObjBlockItemTag_cD1Ev`

### Cause

`include/daObjBlockItemTag_c.h` declares the destructor **out of line**:

```c
virtual ~daObjBlockItemTag_c();                         /* slots 16, 17 */
```

measured to bite:

   green.

---

andrewboudreau:

```
daObjBlockItemTag_c          ov002   is promoted but the queue says no (has 'no', want 'yes')
queue_audit: FAIL -- 1 row disagree.
```

The manifest records the class as `promoted`, but `notes/data/tu-promotion-queue.tsv` still carries `already_promoted=no`. The gate compares that single column against `config/tu_manifest.d` and nothing else, so a promotion has to flip the cell in the same change. Here that means 1 row: `daObjBlockItemTag_c`.

andrewboudreau:
Byte validation is green here. Pushed `a6eaca4d` to this branch fixing the four non-byte gates.

**1. `promoted`** — the manifest says `promoted` but `notes/data/tu-promotion-queue.tsv` still had `daObjBlockItemTag_c` at `already_promoted=no`. Flipped that one cell by hand; `queue_audit --write` rewrites dozens of unrelated rows.

**2. `references` (baseline well-formedness)** — this is the one worth reading. The branch grew `config/dead-reference-baseline.json` from main's **148 entries to 302**, banking 159 new ones. Most of them cite `tools/bin` and `tools/mwccarm` — gitignored build inputs that exist in a local checkout but are not in the tree, so they aren't real dead references. Banking them permanently weakens the gate for everyone. Three cited `.claude/skills/decomp-tu-build/SKILL.md`, which is in no tree here, and that is what tripped `test_the_baseline_is_wellformed`.

I restored main's baseline verbatim and re-ran the gate: **this promotion needs no baseline change at all**. If `--update` was run from a checkout that had those paths locally, that's the likely cause — worth re-running from a clean tree before banking anything.

**3. `references` (prose)** — `notes/runbook-type-reconstruction.md` used `src/_ZN19daObjBlockItemTag_cD0Ev.cpp` as its "Rung 2 — unmigrated function" exemplar, and this promotion folds that shard away. I went looking for a live replacement and there isn't one: zero files anywhere in `src/` still have that raw shape (hand-spelled mangled symbol, `int *` this). The migration has eliminated the form tree-wide, so the example is genuinely historical now. Named the symbol instead of the path — the convention already used elsewhere for folded-away citations — and said plainly that it predates the promotion. Nothing had to be banked.

**4. `check_src_tu_compiles`** — no CI job runs this one. Promoting the class un-declares its TU, so the leftover `src_tu/actors/daObjBlockItemTag_c.cpp` landed in `on_disk - declared` with no compiler pin and nothing to vouch for it. Deleted. Worth checking on every promotion for that reason; #2920 avoided it by moving the file with `git mv` instead.

Verified after the fixes: `queue_audit --check-promoted` clean, `check_dead_references` clean, `BaselineTests` 6/6, `check_src_tu_compiles` 260/260, `check_src_tu` 50 TUs all resolving, `port_refcheck` 408/0 stale, `check_decl_agreement` no new disagreements.

PR #2891, opened 2026-09-21
Promote ov022/daObjFl_Ukiyuka_c to a single translation unit (4/4 TEXT-VERIFIED)

lunavyqo:
Folds the 4 one-function ov022 floating-floor shards into src/actors/daObjFl_Ukiyuka_c.cpp over the contiguous linker run 0x02111f6c..0x021120b8, replacing the per-function delinks entries with one complete span. Builds on a prior commit on this branch renaming the coined FloatingFloorLllSmall alias to the ROM RTTI name daObjFl_Ukiyuka_c.

IDENTITY IS THE CARTRIDGE'S. ov022 holds _ZTS17daObjFl_Ukiyuka_c at 0x021140c0; no FloatingFloorLllSmall RTTI record exists anywhere. The 32-slot table at 0x0211412c carries this class's own _ZTI (0x021140b4) at its typeinfo word and overrides InitResources (slot 0), CleanupResources (slot 3) and the destructor pair (slots 16, 17). Two factories (FL_UKIYUKA_L, FL_UKIYUKA) build this one class. The coined _ZTV21FloatingFloorLllSmall alias is renamed at its unchanged address 0x0211412c via tools/class_rename.py (idempotent, 18 edits + 5 renames).

Gates, real output:
- tubuild verify ov022/daObjFl_Ukiyuka_c: 4/4 MATCH, objisolate clean, reloc-destinations clean. PARTIAL on ordinal pair (0,1) only -- out-of-line destructor order (pilot sec 3).
- dead references: zero hits on touched files.
- Header fan-out is just the promoted TU (only a byte-neutral assert-name fix).

Deliberate divergences: text-only promotion (vtable/RTTI externalized deadstrip-data, no data/BSS); out-of-line destructor with stacked D1/D0 markers; InitResources' missing return is faithful (documented in-TU, measured); ResourceDescriptor spelling matches the definition site in daObjUkiyuka_c.cpp.

Note: tools/tu_promote.py crashes with an unhandled CalledProcessError when legacy shards carry uncommitted edits (git rm refuses). Workaround used here: commit before promoting. Worth a guard upstream.


andrewboudreau:

```
FAIL src/actors/daObjFl_Ukiyuka_c.cpp: isolate: requested functions are not emitted in ROM order
```

- differing indices — **exactly `[0, 1]`**, out of 4 symbols
- `[0]` ROM wants `_ZN17daObjFl_Ukiyuka_cD1Ev`, compiler emitted `_ZN17daObjFl_Ukiyuka_cD0Ev`
- `[1]` ROM wants `_ZN17daObjFl_Ukiyuka_cD0Ev`, compiler emitted `_ZN17daObjFl_Ukiyuka_cD1Ev`

### Cause

`include/daObjFl_Ukiyuka_c.h` declares the destructor **out of line**:

```c
virtual ~daObjFl_Ukiyuka_c(); /* slots 16 (D1), 17 (D0) */
```

measured to bite:

   green.

---

andrewboudreau:

```
daObjFl_Ukiyuka_c            ov022   is promoted but the queue says no (has 'no', want 'yes')
queue_audit: FAIL -- 1 row disagree.
```

The manifest records the class as `promoted`, but `notes/data/tu-promotion-queue.tsv` still carries `already_promoted=no`. The gate compares that single column against `config/tu_manifest.d` and nothing else, so a promotion has to flip the cell in the same change. Here that means 1 row: `daObjFl_Ukiyuka_c`.

andrewboudreau:
Byte validation is green here — the only thing left was the `promoted` gate.

`queue_audit --check-promoted` compares `notes/data/tu-promotion-queue.tsv` column 6 (`already_promoted`) against `config/tu_manifest.d`. The manifest says `promoted`, the queue row still said `no`, so the gate read 1 row disagreeing.

Pushed `55e9f3ae` to this branch flipping that one cell. I edited it by hand rather than running `queue_audit --write`, which rewrites dozens of unrelated rows in the same file.

Also ran the gates that have no CI job, so nothing is hiding behind the green ticks: `check_src_tu_compiles` 261/261, `check_dead_references` clean, `port_refcheck` 408 checked / 0 stale, `check_decl_agreement` no new disagreements. No `src_tu` shadow left behind — that one is worth checking on every promotion, since promoting a class un-declares its TU and CI never looks.

PR #2893, opened 2026-09-21
Promote ov022 daObjFlMaruta_c+volcanoCannon to a single translation unit (16/16 TEXT-VERIFIED)

lunavyqo:
Depends on #2892 (the RollingLogLll rename) -- this branch contains that commit plus the promotion below. Merge #2892 first, then I will rebase this onto main.

Promotes the 16-function combined TU (daObjFlMaruta_c + daObj_volcanoCannon_c, run 0x02112498..0x02112950) to src/game/actors/d_a_obj_fl_maruta.cpp, replacing 16 per-function delinks entries with one complete span. The two classes cannot be separated: daObjFlMaruta_c's vtable (0x021143a4) owns slots 32/34 at 0x021126ac/0x02112710, interleaved between volcanoCannon's D0 and ChangeState (#2802's proof, re-verified here).

Gates, real output:
- tubuild verify: 16/16 MATCH, objisolate clean, reloc-destinations clean. PARTIAL on (0,1) and (5,6) -- both destructor D0/D1 groups, compiler order (pilot sec 3).
- dead references: zero hits on touched files (legacy enumeration dropped; history lives in manifest legacy_source).
- Manifest licenses 16 compiler-only records (both vtables + full RTTI chains as deadstrip-data, flMaruta D2 deadstrip) -- the exact licensing #2802's message left open.

Deliberate divergences: text-only route (no data/BSS claims); out-of-line destructors with stacked markers; per-class _ZTV decls at namespace scope above the factories (moved out of extern C per the promotion conventions -- byte-neutral, re-verified); factories keep hand-spelled fBase_c allocator form and &_ZTV[C][2] vptr spelling.


andrewboudreau:
### Why `PR validation` is red: a cross-class destructor hoist, not just D1/D0

The `stock control is stale` / `omitted the ROM-data measurement` headline is
downstream noise. The cause is:

```
FAIL src/game/actors/d_a_obj_fl_maruta.cpp: isolate: requested functions are not
emitted in ROM order
```

Diffed positionally (`ast.literal_eval` both lists, compare index by index):

- differing indices — **`[0..7]`** of 16; indices 8–15 already agree

| idx | ROM wants | compiler emitted |
|----:|---|---|
| 0 | `_ZN15daObjFlMaruta_cD1Ev` | `_ZN21daObj_volcanoCannon_cD1Ev` |
| 1 | `_ZN15daObjFlMaruta_cD0Ev` | `_ZN21daObj_volcanoCannon_cD0Ev` |
| 2 | `_ZN15daObjFlMaruta_c16CleanupResourcesEv` | `_ZN15daObjFlMaruta_cD0Ev` |
| 3 | `_ZN15daObjFlMaruta_c8BehaviorEv` | `_ZN15daObjFlMaruta_cD1Ev` |
| 4 | `_ZN15daObjFlMaruta_c13InitResourcesEv` | `_ZN15daObjFlMaruta_c16CleanupResourcesEv` |
| 5 | `daObjFlMaruta_c_classInit` | `_ZN15daObjFlMaruta_c8BehaviorEv` |
| 6 | `_ZN21daObj_volcanoCannon_cD1Ev` | `_ZN15daObjFlMaruta_c13InitResourcesEv` |
| 7 | `_ZN21daObj_volcanoCannon_cD0Ev` | `daObjFlMaruta_c_classInit` |

Two distinct effects, and they need separating before either is priced:

1. **`daObj_volcanoCannon_c`'s destructor pair is hoisted to the front of the
   object**, ahead of the whole `daObjFlMaruta_c` run, where the ROM puts it at
   ordinals 6/7 — after `daObjFlMaruta_c_classInit`.
2. **`daObjFlMaruta_c`'s own pair is additionally transposed** (`D0` before `D1`),
   the same single-lever failure that #2887, #2888 and #2891 hit on their own —
   the out-of-line destructor form emitting `D2, D0, D1` where the cartridge
   wants `D1` first.

Effect 2 alone is the well-understood case. Effect 1 is the one that makes this
PR harder than its three siblings: with two classes in one TU, the emission order
of the second class's variants is decided relative to the first class's codegen,
so fixing Maruta's destructor form does not by itself move volcanoCannon's pair
back down.

Worth confirming the TU boundary is right before spending source forms on it — an
explanation that needs two independent codegen coincidences is often a slicing
result instead. Either way this is matching work under `AGENTS.md`: it needs a
rebuild and byte proof, not just a green isolate.

---

andrewboudreau:

```
daObj_volcanoCannon_c+daObjFlMaruta_c ov022
    daObjFlMaruta_c              ov022   is promoted but the queue says no (has 'no', want 'yes')
queue_audit: FAIL -- 2 rows disagree.
```

The manifest records the class as `promoted`, but `notes/data/tu-promotion-queue.tsv` still carries `already_promoted=no`. The gate compares that single column against `config/tu_manifest.d` and nothing else, so a promotion has to flip the cell in the same change. Here that means 2 rows: `daObjFlMaruta_c` **and** `daObj_volcanoCannon_c+daObjFlMaruta_c` (note the queue spells the combined row with the two class names in the opposite order from the manifest id).

andrewboudreau:
Byte validation is green here. Pushed `a1a5ac30` to this branch fixing the two non-byte gates.

**1. `promoted`** — this is a two-class promotion, so two rows in `notes/data/tu-promotion-queue.tsv` were still at `already_promoted=no`: `daObj_volcanoCannon_c+daObjFlMaruta_c` and `daObjFlMaruta_c`. Flipped both cells by hand; `queue_audit --write` rewrites dozens of unrelated rows.

**2. `references`** — `notes/plan-cpp-conversion-queue.md` linked `func_ov022_02112710` to `../src/func_ov022_02112710.c`, the shard this promotion folds into `src/game/actors/d_a_obj_fl_maruta.cpp`. Repointed the link at the assembled TU, which is exactly what the neighbouring `func_ov027_02111680` entry in the same list already does for its own folded symbol. The trailing "(ROM ordinal 4 of `daObj_volcanoCannon_c`)" annotation is still accurate and is left alone.

`config/dead-reference-baseline.json` is untouched here and identical to main's — nothing needed banking, which is the right outcome for a promotion.

Verified after the fixes: `queue_audit --check-promoted` clean, `check_dead_references` clean with no new banked entries, `check_src_tu_compiles` 260/260 (no `src_tu` shadow left behind), `check_src_tu` 50 TUs all resolving, `port_refcheck` 408/0 stale, `check_decl_agreement` no new disagreements.

PR #2916, opened 2026-09-22
Update README.md

lunavyqo:
"Every arm-mode function in the game, drawn as..." text removed as reduandant since same info it contains is present in the treemap svg right under it


andrewboudreau:
Pushed one commit to this branch (`8f798e183`) — flagging it since I edited someone else's PR.

**The caption removal is right**, and I checked the reasoning rather than taking it on trust: `tools/treemap.py:333` draws

> Each rectangle is one function, sized by byte count. Green = matched, gray = unmatched.

*inside* the SVG, and that text is present in `docs/progress-treemap.svg` on main. So the paragraph above the image really is redundant. I also confirmed no tool writes that paragraph into `README.md`, so removing it is durable — a refresh won't put it back.

**But the commit carried a second change that wasn't in the description.** It reverted the generated row inside `<!-- tiers:start/end -->`:

```
-CONVERTED  ...  26.4%   3,000 / 11,385 functions
+CONVERTED  ...  26.3%   2,999 / 11,385 functions
```

That row is written by `tools/tiers.py`. Both the merge base (`3704c7b8f`) and current main carry **3,000**, and only this branch has 2,999 — the signature of a web-editor save from a stale rendering of the page, which writes the whole file back as it was when the editor opened.

Merging as-is would have stepped the published counter backwards until the next scheduled refresh rewrote it. **Nothing would have caught it:** `coverage` and `ratchet` only run when the paths they measure change, and a README-only PR doesn't trigger them — every check on this PR was green, including `PR validation`, because none of them look at that row.

I restored the row to main's value. The branch's only difference from main is now the three-line caption removal, which is what you described.

No action needed from you — just didn't want to silently rewrite your commit. If you edit `README.md` through the GitHub web editor again, it's worth re-checking the `progress:`/`tiers:` blocks afterward, since they change every few minutes.

PR #2951, opened 2026-09-22
Deslop ov033/daObjTtFuta_c: pounded lid comments

lunavyqo:
Keep 7/7 MATCH (tubuild verify: bytes, objisolate, relocs, emission order all clean). Real Sound::PlayBank3 call; member calls and member-anchored args; BSS file homes typed SharedFilePtr; match essays, shadow structs and ordinal banners gone; honest leftovers. Banked two SharedFilePtr decl rows by hand. Manifest untouched. Gates: decl --changed clean, tiers PASS, port 408, deadrefs clean.


andrewboudreau:
Heads-up: this now conflicts with `main` on `config/decl-agreement-baseline.json`. #2947 (`daObjTtWater_c`, same overlay) landed two adjacent `data_ov033_*` keys, so the two insertions collide textually.

Only the baseline conflicts — `src/actors/daObjTtFuta_c.cpp` merges clean.

The resolution is a plain sorted insertion: your two blocks

```
    "data_ov033_021124c0": { "src/actors/daObjTtFuta_c.cpp": ["return:SharedFilePtr"] },
    "data_ov033_021124c8": { "src/actors/daObjTtFuta_c.cpp": ["return:SharedFilePtr"] },
```

go immediately **before** `"data_ov033_021124e8"`, which is what `main` now has there.

I rebased it locally to confirm that is all it takes:

- baseline replays exactly — `(main − removed) | added`, 16383 entries, **0 lost, 0 invented**
- `src/actors/daObjTtFuta_c.cpp` blob is **identical** to this PR head (`b4ded4a64`)

I have not pushed it: that would be a force-push to your fork branch, which is yours to make. If you would rather I push the rebase, say so and I will.

Please **do not** resolve this by re-running `--update` — the full gate currently reports ~128 banked entries as "gone", and an `--update` would bank that loss set too and turn a +2 diff into a large one.

Worth knowing for the next one: the PR `declarations` job runs `--changed "$BASE"`, but the push-to-main job runs the **full** tree. A baseline edit for a file your branch does not otherwise touch is outside every `--changed` scope, so it is invisible here and only shows up once merged. That is exactly how #2936 turned `main` red earlier today (fix in #2950).

andrewboudreau:
`PR validation` is green on this head and the only non-advisory check outstanding is the merge itself — main has moved and `config/decl-agreement-baseline.json` now conflicts.

The conflict is pure textual adjacency. Your branch adds two blocks:

```
"data_ov033_021124c0": { "src/actors/daObjTtFuta_c.cpp": ["return:SharedFilePtr"] },
"data_ov033_021124c8": { "src/actors/daObjTtFuta_c.cpp": ["return:SharedFilePtr"] },
```

and main gained unrelated entries in the same region since you branched.

Please resolve it by replaying rather than blending. Flatten both sides to a set of `(symbol, file, reason)` triples, compute `(main − your_removed) | your_added`, and assert that nothing is lost and nothing is invented before writing the file back. Do **not** re-run `check_decl_agreement.py --update` on top of a merged file — `--update` appends, so a rebase plus a re-run double-banks entries, and running `--check` after `--update` is circular. Also note the file is not fully sorted, so insert your two blocks in place rather than re-sorting: re-sorting silently moves other classes' keys.

Once that's replayed the PR should go straight through.

PR #2962, opened 2026-09-22
Deslop ov002/daObjBlockItemTag_c: item-tag blocks comments

lunavyqo:
Keep 10/10 MATCH (tubuild verify: bytes, objisolate, relocs, emission order all clean). BSS homes typed with member Load/Release; match essays and ordinal banners gone; honest leftover block. Banked one SharedFilePtr decl row by hand. Manifest untouched. Gates: decl --changed clean, tiers PASS, port 408.


andrewboudreau:
Holding this one — it reds the `declarations` gate on **main**, even though the gate is green here on the branch. The change itself is the right idea; it just needs two more declarations moved with it.

**What I measured.** I built this PR's merge tree against current main and ran the *full-tree* `check_decl_agreement.py` (not `--changed`):

```
FAIL: 2 declaration(s) contradict the symbol's definition:

  data_ov002_0210da18: return
    src/actors/daObjHatenaBlock_c.cpp:145  declares  char []
    src/__sinit_ov002_02100560.c:7  plurality is  void *
    got char [], expected void *
  data_ov002_0210da30: return
    src/actors/daObjHatenaBlock_c.cpp:146  declares  char []
    src/_ZN16daKing_Donketu_c13InitResourcesEv.cpp:20  plurality is  SharedFilePtr
    got char [], expected SharedFilePtr
```

`src/actors/daObjHatenaBlock_c.cpp` is a file this PR never touches. Retyping `data_ov002_0210da18` and `data_ov002_0210da30` from `extern char data_...[]` to `extern SharedFilePtr data_...` moves the plurality for those two symbols, and the still-`char []` spelling in `daObjHatenaBlock_c.cpp` becomes the contradiction.

**Why the branch is green anyway.** `.github/workflows/decl-agreement.yml` runs `--changed "$BASE"` on a PR and the full tree only on push-to-main. `daObjHatenaBlock_c.cpp` is outside every `--changed` scope here: it isn't a changed file, it doesn't *define* either symbol, and neither symbol's `config/**/symbols.txt` row moved. So the disagreement is invisible until it lands.

I isolated it — this PR is the trigger on its own, not an interaction. #2968 and #2971 each pass the same full-tree check individually, and all three folded together fail on exactly these two lines and nothing else.

**The fix** is to carry `daObjHatenaBlock_c.cpp:145-146` along in this PR: give those two externs the same `SharedFilePtr` spelling, since that is what the `Release()` / `Model::LoadFile()` call sites here prove the objects actually are. That is matching work, not a text edit — changing a declaration can change instruction selection at the call site — so it needs a rebuild and byte proof for `daObjHatenaBlock_c.cpp`, and then the full-tree check should come back clean.

Worth checking `data_ov002_0210d9d8` at the same time; it isn't in the failure list, but it is the third symbol in the same group.

Everything else here is good — collapsing `((SharedFilePtr *)(data_...))->Release()` and `*(SharedFilePtr *)&data_...` down to a typed extern is exactly the direction, and the byte gate is green.

PR #2965, opened 2026-09-22
Promote ov077/daTgz_c and delete its 34 shard TUs

lunavyqo:
## Summary

Promotes the Spiny on the castle roof from the coined `Spiny` spelling to the cartridge class `daTgz_c`.

- One `.text` span, `0x02124b64..0x0212624c`, from `src/actors/daTgz_c.cpp`.
- Deletes the 34 one-function shards, including `src/d_a_tgz.c`. The factory is the last function in the span. `daJgm_c` ends at `0x02124b64` and `daPopoi_c` starts at `0x0212624c`.
- Drops the coined `_ZTV5Spiny` alias. `_ZTV7daTgz_c` stays at `0x02127984`. `_ZTS7daTgz_c` is at `0x02127948`.
- The destructor emits retail D1 then D0. D2 is not in the ROM.
- `func_ov077_021251d0` copies a flat `Matrix4x3`. `common.h` is included before the actor headers, so that copy stays a 12-word struct move. That is the cartridge instruction, not a compiler-invented assignment operator.

## Verification

- `tubuild.py verify ov077/daTgz_c`: 34/34 MATCH, ROM order, relocation destinations clean. Class and inherited RTTI/vtable are `deadstrip-data` at their existing homes.
- The linked ov077 image is 17376 bytes and matches retail, including this span.
- Contributor credit: 0 lost. All 34 symbols consolidated into `src/actors/daTgz_c.cpp`.
- The 106/106 figure is the build-box stock comparison. This note will be updated when that validator finishes.


andrewboudreau:
Reviewed the substance against main — the class-identity claim holds and the rename is done the right way. Recording the evidence so it doesn't have to be re-derived:

**`daTgz_c` is the cartridge name, `Spiny` was the coined one.** `config/arm9/overlays/ov077/symbols.txt` on main already carries `_ZTS7daTgz_c` at `0x02127948` and `_ZTI7daTgz_c` at `0x02127954` — 12 bytes for the type string, which fits `"7daTgz_c"` (8 + NUL, padded) and rules out `"5Spiny"` (6 + NUL would put `_ZTI` at `0x02127950`). Main was carrying *both* `_ZTV5Spiny` and `_ZTV7daTgz_c` at `0x02127984`; this PR collapses that to the single ROM spelling, which is the right call — the ROM RTTI name wins whenever one exists.

**The rename ledger is right.** `symbols/actor_renames.tsv` rewrites the seven ov077 rows in place so column 4 reads the new name in the present tense, rather than appending a second generation of rows. Appending alone would red the tree.

**`config_tu/arm9/overlays/ov077/delinks.txt` still names `src_tu/actors/Spiny.cpp`,** which this PR deletes — but that is the existing convention, not a regression: 16 of the 18 `src_tu/` rows in `config_tu/**/delinks.txt` on main already point at files earlier promotions removed. Flagging it only so it isn't mistaken for something this PR introduced.

**Gates, run on your head commit:** `check_tubuild_conflicts` ✅, `queue_audit --check-promoted` ✅, `check_src_tu` ✅ ("every reference resolves"), `check_dead_references` ✅ (no new dead references, no broken markdown links).

**The one thing blocking the merge** is `attribution.json`, which conflicts with main now that #2956 has landed and will keep conflicting while other promotions land — every promotion appends to the same `overrides` map, so the collision is pure textual adjacency, never semantic.

When you resolve it, please *replay* rather than blend: take main's post-merge `overrides` as the prefix, then append your branch's added keys in their existing order. Concretely — flatten all three versions to `{key: value}`, compute `added = branch − base`, assert `removed == 0` and `added ∩ main == ∅`, then write `main + added`. The file is `indent=2` with a trailing newline, and new entries go at the end of the map, so a correct resolution shows exactly one diff hunk: the old last line gaining a comma, plus your keys. A git auto-merge of that file is the thing to avoid.

Your ov033/`daObjTtWater_c` fix is byte-identical to what landed in #2957, so it will merge as a no-op.

PR #2968, opened 2026-09-22
Deslop ov002/daObjBlockL_c: big brick blocks comments

lunavyqo:
Keep 14/14 MATCH (tubuild verify: bytes, objisolate, relocs, emission order all clean). Mesh-collider setters typed; match essays, generator comments, conflict notes and ordinal banners gone; honest leftover block. Banked two decl rows by hand. Manifest untouched. Gates: decl --changed clean, tiers PASS, port 408.


andrewboudreau:
This will red the `conflicts` gate once that check posts — it isn't on the board yet, so the PR looks clear. Run against this PR's merge tree with current main:

```
ov002/daObjBlockL_c: STALE NOTE: extern declaration _ZN5Event6GetBitEj (from the legacy
file for _ZN13daObjBlockL_c13InitResourcesEv) is recorded as an open conflict once, but
src/actors/daObjBlockL_c.cpp does not carry a matching marker.
ov002/daObjBlockL_c: STALE NOTE: ... (from the legacy file for _ZN13daObjBlockL_c8BehaviorEv) ...
2 unreconciled conflict record(s)
```

`check_tubuild_conflicts` audits a **pair** — a `/* TUBUILD CONFLICT ... NOT applied: ... */` marker in the source and a `tubuild create warning: CONFLICT:` note in the manifest. Either half alone fails. This PR removes both markers and leaves both notes in `config/tu_manifest.d/ov002/daObjBlockL_c.json`.

**Removing them was correct.** The two "NOT applied" alternates are

```c
extern "C" int _ZN5Event6GetBitEj(unsigned int a);   // from Behavior
int _ZN5Event6GetBitEj(u32 bit);                     // from InitResources
```

and the surviving declaration at line 57 is `extern "C" int _ZN5Event6GetBitEj(unsigned int bit);`. The first differs only in the parameter name. The second differs only in spelling `unsigned int` as `u32` — one type, via `include/types.h:8` — and in dropping `extern "C"`, which the survivor keeps and which is what makes a mangled-name declaration link. Neither was ever a real second interface.

So close the notes rather than restoring the markers: in `config/tu_manifest.d/ov002/daObjBlockL_c.json`, change both `_ZN5Event6GetBitEj` notes from `tubuild create warning: CONFLICT:` to `tubuild create warning (RESOLVED): CONFLICT:`. That's the form the checker accepts, and it's what #2957 did for the identical `Event::GetBit` pair in ov033.

Everything else is green on the merge tree: `check_decl_agreement` full-tree reports no new disagreements, `check_dead_references` no new dead references, `queue_audit --check-promoted` agrees. The substance is good — retyping `func_020393a4`'s parameter to `dBgW_KcMbg *` matches the declaration `daObjPathLift_c.cpp` already carries, so it moves toward the tree's plurality rather than away from it.

